PR #2, opened 2016-06-06
Update Config Implementation

gabrielhartmann:
1. Remove old config implementation.  Shouldn't be used by anyone.
2. Add new config interfaces and implementations.
        ConfigStore and StateStore interfaces
        CuratorConfigStore and CuratorStateStore implementations
3. Unit tests for Storage implementations

PR #3, opened 2016-06-07
Adds a configuration storage library

nickbp:
At the base is the StorageAccess interface, with a default ZK implementation.

Above that layer are separate classes for each of the following categories:
- ConfigStorage: Managing/accessing active and historical configurations
- FrameworkIdStorage: Storing and retrieving the active framework id
- TaskStorage: Storing and retrieving the current task state

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,9 @@ allprojects {
 apply plugin: 'application'
 
 ext {
-  yamlVer = "2.5.4"
+  jacksonGuavaTypeVer = "2.7.4"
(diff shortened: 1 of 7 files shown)


nickbp:
This is effectively superceded by PR #2, just here as reference.

The main difference is this one breaks out the different types of data into separate access classes, all sharing a common underlying storage implementation. I'd found that in practice frameworks only wanted to access data in one of three provided categories.

nickbp:
Looks like PR#2 is merged so this one can go away now.

PR #4, opened 2016-06-08
Add a factory for creating MesosSchedulerDrivers.

nickbp:
Each framework should be constructing these in the same way. Can be expanded as things are normalized further.

### DIFF
--- a/src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
+++ b/src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
@@ -1,0 +1,84 @@
+package org.apache.mesos.scheduler;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.mesos.MesosSchedulerDriver;
+import org.apache.mesos.Scheduler;
+import org.apache.mesos.SchedulerDriver;
+import org.apache.mesos.Protos.Credential;
+import org.apache.mesos.Protos.FrameworkInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import com.google.protobuf.ByteString;
+
+/**
+ * Factory class for creating {@link MesosSchedulerDriver}s.
+ */
+public class SchedulerDriverFactory {
+
+  private static final Logger LOGGER =
+    LoggerFactory.getLogger(SchedulerDriverFactory.class);
+
+  /**
+   * Creates and returns a new {@link SchedulerDriver} without a credential secret.
+   *
+   * @param scheduler The Framework {@link Scheduler} implementation which should receive callbacks
+   * from the {@link SchedulerDriver}
+   * @param frameworkInfo The {@link FrameworkInfo} which describes the framework implementation.
+   * The 'principal' field MUST be populated and non-empty
+   * @param masterZkUrl The URL of the currently active Mesos Master, of the form "zk://host/mesos"
+   * @return A {@link SchedulerDriver} configured with the provided info
+   * @throws IllegalArgumentException if {@link FrameworkInfo}.principal is unset or empty
+   */
+  public SchedulerDriver create(
+      final Scheduler scheduler, final FrameworkInfo frameworkInfo, final String masterZkUrl) {
+    return create(scheduler, frameworkInfo, masterZkUrl, null /* credentialSecret */);
+  }
+
+  /**
+   * Creates and returns a new {@link SchedulerDriver} with the provided credential secret.
+   *
+   * @param scheduler The Framework {@link Scheduler} implementation which should receive callbacks
+   * from the {@link SchedulerDriver}
+   * @param frameworkInfo The {@link FrameworkInfo} which describes the framework implementation.
+   * The 'principal' field MUST be populated and non-empty
+   * @param masterZkUrl The URL of the currently active Mesos Master, of the form "zk://host/mesos"
+   * @param credentialSecret The secret to be included in the framework
+   *     {@link org.apache.mesos.Protos.Credential}, ignored if {@code null}/empty
+   * @return A {@link SchedulerDriver} configured with the provided info
+   * @throws IllegalArgumentException if {@link FrameworkInfo}.principal is unset or empty
+   */
+  public SchedulerDriver create(
+      final Scheduler scheduler,
+      final FrameworkInfo frameworkInfo,
+      final String masterZkUrl,
+      final byte[] credentialSecret) {
+    LOGGER.info("Creating MesosSchedulerDriver for "
+        + "scheduler = {}, frameworkInfo = {}, masterZkUrl = {}, credentialSecret = {}",
+        scheduler, frameworkInfo, masterZkUrl, credentialSecret);
+
+    if (!frameworkInfo.hasPrincipal() || StringUtils.isEmpty(frameworkInfo.getPrincipal())) {
+      throw new IllegalArgumentException(
+          "Unable to create MesosSchedulerDriver, FrameworkInfo lacks a principal: "
+              + frameworkInfo.toString());
+    }
+
+    Credential.Builder credentialBuilder = Credential.newBuilder()
+        .setPrincipal(frameworkInfo.getPrincipal());
+    if (credentialSecret != null && credentialSecret.length != 0) {
+      credentialBuilder.setSecretBytes(ByteString.copyFrom(credentialSecret));
+    }
+
+    return createInternal(scheduler, frameworkInfo, masterZkUrl, credentialBuilder.build());
+  }
+
+  /**
+   * Broken out into a separate function to allow testing with custom SchedulerDrivers.
+   */
+  protected SchedulerDriver createInternal(
+      final Scheduler scheduler,
+      final FrameworkInfo frameworkInfo,
+      final String masterZkUrl,
+      final Credential credential) {
+    return new MesosSchedulerDriver(scheduler, frameworkInfo, masterZkUrl, credential);
+  }
+}
--- a/src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
+++ b/src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
@@ -23,7 +23,7 @@ public class SchedulerDriverFactory {
    * some form of sidechannel auth. When this environment variable is present, we should always
    * provide a {@link Credential} with (only) the principal set.
    */
-  private static String SIDECHANNEL_AUTH_MAGIC_ENVNAME = "DCOS_SERVICE_ACCOUNT_CREDENTIAL";
+  private static String SIDECHANNEL_AUTH_ENV_NAME = "DCOS_SERVICE_ACCOUNT_CREDENTIAL";
 
   /**
    * Creates and returns a new {@link SchedulerDriver} without a credential secret.
@@ -32,14 +32,14 @@ public class SchedulerDriverFactory {
    *     from the {@link SchedulerDriver}
    * @param frameworkInfo The {@link FrameworkInfo} which describes the framework implementation.
    *     The 'principal' field MUST be populated and non-empty
-   * @param masterZkUrl The URL of the currently active Mesos Master, of the form "zk://host/mesos"
+   * @param masterUrl The URL of the currently active Mesos Master, of the form "zk://host/mesos"
    * @return A {@link SchedulerDriver} configured with the provided info
    * @throws IllegalArgumentException if {@link FrameworkInfo}.principal is unset or empty when
    *     authentication is needed
    */
   public SchedulerDriver create(
-      final Scheduler scheduler, final FrameworkInfo frameworkInfo, final String masterZkUrl) {
-    return create(scheduler, frameworkInfo, masterZkUrl, null /* credentialSecret */);
+      final Scheduler scheduler, final FrameworkInfo frameworkInfo, final String masterUrl) {
+    return create(scheduler, frameworkInfo, masterUrl, null /* credentialSecret */);
   }
 
   /**
@@ -49,7 +49,7 @@ public class SchedulerDriverFactory {
    *     from the {@link SchedulerDriver}
    * @param frameworkInfo The {@link FrameworkInfo} which describes the framework implementation.
    *     The 'principal' field MUST be populated and non-empty
-   * @param masterZkUrl The URL of the currently active Mesos Master, of the form "zk://host/mesos"
+   * @param masterUrl The URL of the currently active Mesos Master, of the form "zk://host/mesos"
    * @param credentialSecret The secret to be included in the framework
    *     {@link org.apache.mesos.Protos.Credential}, ignored if {@code null}/empty
    * @return A {@link SchedulerDriver} configured with the provided info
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,84 @@
...
+  public SchedulerDriver create(
+      final Scheduler scheduler,
+      final FrameworkInfo frameworkInfo,
+      final String masterZkUrl,
+      final byte[] credentialSecret) {
+    LOGGER.info("Creating MesosSchedulerDriver for "
+        + "scheduler = {}, frameworkInfo = {}, masterZkUrl = {}, credentialSecret = {}",
+        scheduler, frameworkInfo, masterZkUrl, credentialSecret);
+
+    if (!frameworkInfo.hasPrincipal() || StringUtils.isEmpty(frameworkInfo.getPrincipal())) {
```
gabrielhartmann: Requiring a principal is ok, as Mesos allows you to make one up.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,84 @@
...
+      throw new IllegalArgumentException(
+          "Unable to create MesosSchedulerDriver, FrameworkInfo lacks a principal: "
+              + frameworkInfo.toString());
+    }
+
+    Credential.Builder credentialBuilder = Credential.newBuilder()
```
gabrielhartmann: However, always generating credentials shouldn't be a requirement.  In an unauthenticated environment, I'm not sure if providing unneeded credentials will result in everything being ok, or in things not working.  Better to not generate credentials when we expect them not to be used.

nickbp:
Had another chat with security folks. Turns out the behavior should indeed be different than what's laid out here. Dropping assignee while I work on it.

nickbp:
...aand we're back. PTAL.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,126 @@
...
+
+  private static final Logger LOGGER =
+    LoggerFactory.getLogger(SchedulerDriverFactory.class);
+
+  /**
+   * If this environment variable is present in the scheduler environment, the master is using
+   * some form of sidechannel auth. When this environment variable is present, we should always
+   * provide a {@link Credential} with (only) the principal set.
+   */
+  private static String SIDECHANNEL_AUTH_MAGIC_ENVNAME = "DCOS_SERVICE_ACCOUNT_CREDENTIAL";
```
gabrielhartmann: Magic has bad connotations.  Please could you rename this variable?

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Creates and returns a new {@link SchedulerDriver} without a credential secret.
+   *
+   * @param scheduler The Framework {@link Scheduler} implementation which should receive callbacks
+   *     from the {@link SchedulerDriver}
+   * @param frameworkInfo The {@link FrameworkInfo} which describes the framework implementation.
+   *     The 'principal' field MUST be populated and non-empty
+   * @param masterZkUrl The URL of the currently active Mesos Master, of the form "zk://host/mesos"
```
gabrielhartmann: Could we please rename masterZkUrl to something less implementation specific.  Zookeeper is potentially a transient implementation detail.  Maybe `masterUrl`.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,126 @@
...
+    } else {
+      return new MesosSchedulerDriver(scheduler, frameworkInfo, masterZkUrl, credential);
+    }
+  }
+
+  /**
+   * Returns whether it appears that sidechannel auth should be used when creating the
+   * SchedulerDriver. Broken out into a separate function to customize behavior in tests.
+   */
+  protected boolean isSideChannelActive() {
```
gabrielhartmann: I hate to carp on about names, but "SideChannel" implies illegitimacy.  I think this is a totally legitimate authentication mechanism which is being supported...  Don't have a good name suggestion however.  Naming is hard.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,126 @@
...
+      LOGGER.info("Creating secret authenticated MesosSchedulerDriver for "
+          + "scheduler = {}, frameworkInfo = {}, masterZkUrl = {}, credentialSecret = {}",
+          scheduler, frameworkInfo, masterZkUrl, credentialSecret);
+    } else if (isSideChannelActive()) {
+      // Sidechannel auth is enabled. Provide a Credential with only the Principal set.
+      if (!frameworkInfo.hasPrincipal() || StringUtils.isEmpty(frameworkInfo.getPrincipal())) {
+        throw new IllegalArgumentException(
+            "Unable to create MesosSchedulerDriver for sidechannel auth, "
+                + "FrameworkInfo lacks required principal: " + frameworkInfo.toString());
+      }
```
gabrielhartmann: Lines 82-86 are a copy of 68-72.  Please, factor out into a function.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,126 @@
...
+            "Unable to create MesosSchedulerDriver for secret auth, "
+                + "FrameworkInfo lacks required principal: " + frameworkInfo.toString());
+      }
+      credential = Credential.newBuilder()
+          .setPrincipal(frameworkInfo.getPrincipal())
+          .setSecretBytes(ByteString.copyFrom(credentialSecret))
+          .build();
+      LOGGER.info("Creating secret authenticated MesosSchedulerDriver for "
+          + "scheduler = {}, frameworkInfo = {}, masterZkUrl = {}, credentialSecret = {}",
+          scheduler, frameworkInfo, masterZkUrl, credentialSecret);
```
gabrielhartmann: Please put the log message before the attempt to create the Credential object.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,126 @@
...
+        throw new IllegalArgumentException(
+            "Unable to create MesosSchedulerDriver for sidechannel auth, "
+                + "FrameworkInfo lacks required principal: " + frameworkInfo.toString());
+      }
+      credential = Credential.newBuilder()
+          .setPrincipal(frameworkInfo.getPrincipal())
+          .build();
+      LOGGER.info("Creating sidechannel authenticated MesosSchedulerDriver for "
+          + "scheduler = {}, frameworkInfo = {}, masterZkUrl = {}",
+          scheduler, frameworkInfo, masterZkUrl);
```
gabrielhartmann: Please put the log message before the attempt to create the Credential object.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,126 @@
...
+    } else {
+      // No auth. Provide no credential.
+      credential = null;
+      LOGGER.info("Creating unauthenticated MesosSchedulerDriver for "
+          + "scheduler = {}, frameworkInfo = {}, masterZkUrl = {}",
+          scheduler, frameworkInfo, masterZkUrl);
```
gabrielhartmann: Please put the log message before the attempt to create the Credential object.

gabrielhartmann:
Test coverage looks good, and mostly minor comments regarding some possible improvements.  @adam-mesos:  Logic look good to you?

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: Done

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: Done

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: Done

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: Done

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: This name is intended to encompass both Bouncer-based Framework Auth as well as potentially other auth in the future which would work via the same mechanism (eg Kerberos). Added a note to the comment explaining this.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: Done

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: Dropped "MAGIC" from the name

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,131 @@
...
+      final FrameworkInfo frameworkInfo,
+      final String masterUrl,
+      final byte[] credentialSecret) {
+    Credential credential;
+    if (credentialSecret != null && credentialSecret.length > 0) {
+      // User has manually provided a Secret. Provide a Credential with Principal + Secret.
+      // (note: we intentionally avoid logging the content of the credential secret, just in case)
+      LOGGER.info("Creating secret authenticated MesosSchedulerDriver for "
+          + "scheduler[{}], frameworkInfo[{}], masterUrl[{}], credentialSecret[{} bytes]",
+          scheduler, frameworkInfo, masterUrl, credentialSecret.length);
```
adam-mesos: I'd like to see the principal in this log message too, but I guess it's included in the frameworkInfo?

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Extracts the Principal name from the provided {@link FrameworkInfo}, or throws an
+   * {@link IllegalArgumentException} (mentioning the provided {@code authType}) if the Principal
+   * is unavailable.
+   */
+  private static String getPrincipal(final FrameworkInfo frameworkInfo, final String authType) {
+    if (!frameworkInfo.hasPrincipal() || StringUtils.isEmpty(frameworkInfo.getPrincipal())) {
+      throw new IllegalArgumentException(
+          "Unable to create MesosSchedulerDriver for " + authType + " auth, "
```
adam-mesos: `authType` seems irrelevant to this getter. Seems like better practice would be for the caller to catch the exception/error/empty-string and generate its own error message.
Similarly, this standalone static method shouldn't need to know about the MesosSchedulerDriver, or that its caller is trying to create one.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,131 @@
...
+    return create(scheduler, frameworkInfo, masterUrl, null /* credentialSecret */);
+  }
+
+  /**
+   * Creates and returns a new {@link SchedulerDriver} with the provided credential secret.
+   *
+   * @param scheduler The Framework {@link Scheduler} implementation which should receive callbacks
+   *     from the {@link SchedulerDriver}
+   * @param frameworkInfo The {@link FrameworkInfo} which describes the framework implementation.
+   *     The 'principal' field MUST be populated and non-empty
```
adam-mesos: You say MUST, and yet you don't assert or check/error.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Returns whether it appears that side channel auth should be used when creating the
+   * SchedulerDriver. Side channel auth may take the form of Bouncer-based framework authentication
+   * or potentially other methods in the future (eg Kerberos). Broken out into a separate function
+   * to allow easy customization in tests.
+   */
+  protected boolean isSideChannelActive() {
+    return System.getenv(SIDECHANNEL_AUTH_ENV_NAME) != null;
```
adam-mesos: I'll reiterate that my gut prefers an explicit "--authenticate" flag to decide whether to pass a Credential to the MesosSchedulerDriver, and then the "--secret_file" flag can determine whether the Credential will have a secret.
It seems weird to hardcode a magic "sidechannel" environment variable that would be different if trying to use the classic Kerberos authenticatee module, and could change names in the future (no promises).
But dcos-commons isn't my library, so I'll leave the final decision up to you guys.

adam-mesos:
Logic looks good, but I'm still wary of basing the decision to include a Credential on a magic environment variable name instead of an explicit flag.

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: Yes this just dumps the whole info which contains the principal

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: This is a private utility method called in two places elsewhere in the class. The authType field is used internally for logging

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: The check is in getPrincipal()

src/main/java/org/apache/mesos/scheduler/SchedulerDriverFactory.java
nickbp: I really don't want to punish users on auth-enabled clusters with manually having to check a box each time they install anything on their cluster. Meanwhile the envvar name is already going to be independently synchronized between other independent DCOS components, it's already semi-stable. This library only targets DCOS.

nickbp: And to add: due to how Universe works, each framework would also need to independently define that flag in their packaging, leaving a burden on the framework developer and leaving open a window for inconsistencies in how the flag is defined

Given both of the above, having the library deal with it automatically feels like the preferable option

gabrielhartmann: I agree with Nick.

PR #5, opened 2016-06-09
Missing Zookeeper gradle dependency

kaseyalusi:
Without these lines there were compiler errors about missing missing zookeeper.
`package org.apache.zookeeper does not exist`


gabrielhartmann:
@kaseyalusi: Thanks for the fix.  Looks good to me.  I tried it out and it builds for me too.

PR #8, opened 2016-06-10
Add fetchExecutorNames to StateStore interface

gabrielhartmann:
This allows the whole state store to be iterated over for retrieval of
all information


src/main/java/org/apache/mesos/state/StateStore.java
```diff
@@ -22,6 +22,7 @@
 
     void storeTasks(Collection<Protos.TaskInfo> tasks, String execName) throws StateStoreException;
     Collection<Protos.TaskInfo> fetchTasks(String execName) throws StateStoreException;
+    Collection<String> fetchExecutorNames() throws StateStoreException;
```
nickbp: Add some brief javadocs in this interface imo

PR #13, opened 2016-06-14
Fix TaskStatus updates for the wrong or non-existing TaskInfo

gabrielhartmann:


src/main/java/org/apache/mesos/state/CuratorStateStore.java
```diff
@@ -164,6 +164,22 @@ public void storeStatus(
             String execName) throws StateStoreException {
 
         try {
+            Protos.TaskInfo taskInfo;
+            try {
+                taskInfo = fetchTask(taskName, execName);
+            } catch (KeeperException.NoNodeException | StateStoreException ex) {
+                throw new StateStoreException(
+                    String.format("Failed to retrieve TaskInfo for TaskStatus: '%s'", status),
```
nickbp: Print taskName/execName instead of status?

PR #21, opened 2016-06-15
Test DefaultStageStrategy

gabrielhartmann:


src/test/java/org/apache/mesos/scheduler/plan/PhaseStrategyTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Test;
+import org.junit.Assert;
+
+import java.util.Arrays;
+import java.util.UUID;
+
+/**
+ * This class tests the DefaultStageStrategy.
+ */
+public class PhaseStrategyTest {
```
nickbp: DefaultStageStrategyTest?

PR #24, opened 2016-06-15
Update Block interface and update callers, add tests

nickbp:


src/main/java/org/apache/mesos/scheduler/plan/DefaultStageScheduler.java
```diff
@@ -14,47 +14,63 @@
...
+    @Inject
+    public DefaultStageScheduler(OfferAccepter offerAccepter) {
+        this(offerAccepter, new OfferEvaluator());
+    }
 
-  @Inject
-  public DefaultStageScheduler(OfferAccepter offerAccepter) {
-    this.offerAccepter = offerAccepter;
-  }
+    public DefaultStageScheduler(OfferAccepter offerAccepter, OfferEvaluator offerEvaluator) {
```
gabrielhartmann: Since an OfferEvaluator is constructed with an OfferRequirement, this implies that we need to reconstruct this Scheduler on every Offer cycle.  I'm not sure that's a good idea.  Why was this done?

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -26,7 +26,7 @@
 public class OfferEvaluator {
   private static final Log log = LogFactory.getLog(OfferEvaluator.class);
 
-  private OfferRequirement requirement;
+  private OfferRequirement requirement; //TODO(nick) remove state in favor of passing as an evaluate() parameter
```
gabrielhartmann: I would do this now.  I think that would resolve my concern below.

gabrielhartmann:
@nickbp:  I would just complete your thought regarding OfferEvaluator behavior now.  After that I'd be happy to merge.

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
nickbp: Done!

PR #38, opened 2016-06-17
Fix mount volume update

gabrielhartmann:
Updating a mount volume resource requirement cannot result in additional RESERVE or
UNRESERVE Operations.  However, it must be validated that the new
requirement is still met.

### DIFF
--- a/src/main/java/org/apache/mesos/offer/MesosResourcePool.java
+++ b/src/main/java/org/apache/mesos/offer/MesosResourcePool.java
@@ -113,8 +113,17 @@ public class MesosResourcePool {
 
   private MesosResource consumeReserved(ResourceRequirement resReq) {
     MesosResource mesRes = reservedPool.get(resReq.getResourceId());
+
     if (mesRes != null) {
-      reservedPool.remove(resReq.getResourceId());
+      if (mesRes.isAtomic()) {
+        if (sufficientValue(resReq.getValue(), mesRes.getValue())) {
+          reservedPool.remove(resReq.getResourceId());
+        } else {
+          return null;
+        }
+      } else {
+        reservedPool.remove(resReq.getResourceId());
+      }
     }
 
     return mesRes;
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/mesos/offer/MesosResourcePool.java
```diff
@@ -113,8 +113,16 @@ private void releaseAtomicResource(MesosResource mesRes) {
 
   private MesosResource consumeReserved(ResourceRequirement resReq) {
     MesosResource mesRes = reservedPool.get(resReq.getResourceId());
+
     if (mesRes != null) {
-      reservedPool.remove(resReq.getResourceId());
+      if (mesRes.isAtomic() &&
```
nickbp: how about:

if (isAtomic) {
  if (sufficientValue) {
    remove
  } else {
    return null
  }
} else {
  remove
}

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -151,7 +151,7 @@ public static FulfilledRequirement fulfillRequirement(
           logger.info("Expects Resource");
           // Compute any needed resource pool consumption / release operations
           // as well as any additional needed Mesos Operations
-          if (expectedValueChanged(resReq, mesRes)) {
+          if (expectedValueChanged(resReq, mesRes) && !mesRes.isAtomic()) {
```
nickbp: Add comment about why to skip when atomic?

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
nickbp: Couldn't hurt to add a quick comment to the 'return null' case explaining the 'why'

nickbp:
Just some nits, LGTM otherwise

PR #40, opened 2016-06-17
Nitpicks and one bug fix for CuratorStateStore

nickbp:
- Bug: Don't return the Framework ID as an executor name (test added)
- Nit: Have the underlying curator interface be a member rather than
  a superclass
- Nit: Cleanup to logging, use slf4j's built-in formatting
- Nit: Various tweaks/additions to comments

### DIFF
--- a/src/main/java/org/apache/mesos/state/CuratorStateStore.java
+++ b/src/main/java/org/apache/mesos/state/CuratorStateStore.java
@@ -13,39 +13,45 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * A CuratorStateStore stores TaskInfo and TaskStatus data in Zookeeper.
+ * CuratorStateStore is an implementation of {@link StateStore} which persists data in Zookeeper.
  *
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/mesos/state/CuratorStateStore.java
```diff
@@ -13,97 +13,107 @@
...
         this.rootPath = rootPath;
         this.fwkIdPath = rootPath + "/" + FWK_ID_PATH_NAME;
     }
 
     @Override
     public void storeFrameworkId(Protos.FrameworkID fwkId) throws StateStoreException {
         try {
-            logger.info(String.format("Storing FrameworkID in '%s'", fwkIdPath));
-            store(fwkIdPath, fwkId.toByteArray());
+            logger.info("Storing FrameworkID in '{}'", fwkIdPath);
```
gabrielhartmann: @nickbp:  I chose the wrong log level for these kinds of statements.  They should all be changed to debug in my opinion.

nickbp: Done

PR #46, opened 2016-06-21
Require valid names/ids in data passed to OfferEvaluator/StateStore

nickbp:
- TaskInfo requirements:
  - must have a name
  - if task id is non-empty, it must align with the name
  - if executor is provided, it must have a name (only at StateStore)
- ExecutorInfo requirements:
  - must have a name
  - if executor id is non-empty, it must align with the name
- TaskStatus requirements (only at StateStore):
  - must have a valid/parseable task id
  - must have a valid/parseable executor id

### DIFF
--- a/src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
+++ b/src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
@@ -17,7 +17,11 @@ public class ExecutorRequirement {
 
     public ExecutorRequirement(ExecutorInfo unverifiedExecutorInfo)
             throws InvalidRequirementException {
-        this.executorInfo = validateExecutorInfo(unverifiedExecutorInfo);
+        validateExecutorInfo(unverifiedExecutorInfo);
+        // ExecutorID is always overwritten with a new UUID, even if already present:
+        this.executorInfo = ExecutorInfo.newBuilder(unverifiedExecutorInfo)
+                .setExecutorId(ExecutorUtils.toExecutorId(unverifiedExecutorInfo.getName()))
+                .build();
         this.resourceRequirements =
                 RequirementUtils.getResourceRequirements(this.executorInfo.getResourcesList());
     }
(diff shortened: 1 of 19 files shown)


src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
```diff
@@ -2,56 +2,89 @@
...
+                        executorInfo, e));
+            }
+            if (!executorName.equals(executorInfo.getName())) {
+                throw new InvalidRequirementException(String.format(
+                        "When non-empty, ExecInfo.id must align with ExecInfo.name. Use "
+                        + "ExecutorUtils.toExecutorId(): %s", executorInfo));
+            }
+        }
+        // ExecutorID is always overwritten with a new UUID, even if already present:
+        return ExecutorInfo.newBuilder(executorInfo)
```
gabrielhartmann: I would separate validation from generation of the new ExecutorID

src/main/java/org/apache/mesos/offer/TaskRequirement.java
```diff
@@ -1,62 +1,81 @@
...
+                        + "TaskUtils.toTaskId(): %s", taskInfo));
+            }
+        }
+        if (taskInfo.hasExecutor()) {
+            throw new InvalidRequirementException(String.format(
+                    "TaskInfo must not contain ExecutorInfo. "
+                    + "Use ExecutorRequirement for any Executor requirements: %s", taskInfo));
+        }
+        // TaskID is always overwritten with a new UUID, even if already present:
+        return TaskInfo.newBuilder(taskInfo)
```
gabrielhartmann: Same here.  Validation and construction should be separate.

gabrielhartmann:
@nickbp: Other than those 2 comments, looks good to me.

src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
nickbp: Done

src/main/java/org/apache/mesos/offer/TaskRequirement.java
nickbp: Done

PR #48, opened 2016-06-23
Remove Executor Name from task paths

nickbp:
Consistently getting an executor id is a surprisingly difficult exercise. Switch to a flatter structure where we just require that distinct tasks have unique task names.

This also simplifies things a bit for the developer -- they now only need to think about task names; 'execName' is gone.

### DIFF
--- a/src/main/java/org/apache/mesos/state/CuratorStateStore.java
+++ b/src/main/java/org/apache/mesos/state/CuratorStateStore.java
@@ -148,6 +148,26 @@ public class CuratorStateStore implements StateStore {
                     "Failed to parse the Task Name from TaskStatus.task_id: '%s'", status), e);
         }
 
+        // Validate that a TaskInfo with the exact same UUID is currently present. We intentionally
+        // ignore TaskStatuses whose TaskID doesn't (exactly) match the current TaskInfo: We will
+        // occasionally get these for stale tasks that have since been changed (with new UUIDs).
+        Protos.TaskInfo taskInfo;
+        try {
+            taskInfo = fetchTask(taskName);
+        } catch (Exception e) {
+            throw new StateStoreException(String.format(
+                    "Unable to retrieve matching TaskInfo for the provided TaskStatus name %s. " +
+                    "Call storeTasks() before calling storeStatus()", taskName), e);
+        }
+        if (!taskInfo.getTaskId().getValue().equals(status.getTaskId().getValue())) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/apache/mesos/state/CuratorStateStore.java
```diff
@@ -150,48 +139,54 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
     }
 
     @Override
-    public Collection<Protos.TaskInfo> fetchTasks(String execName) throws StateStoreException {
-        Collection<Protos.TaskInfo> taskInfos = new ArrayList<>();
+    public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
```
gabrielhartmann: Spoke to you in person.  We must not store TaskStatus's with irrelevant TaskIDs.

src/main/java/org/apache/mesos/state/CuratorStateStore.java
```diff
@@ -200,82 +195,70 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
-            throw new StateStoreException(e);
+    public Collection<Protos.TaskInfo> fetchTasks() throws StateStoreException {
+        Collection<Protos.TaskInfo> taskInfos = new ArrayList<>();
+        for (String taskName : fetchTaskNames()) {
+            try {
+                byte[] bytes = curator.fetch(taskPathMapper.getTaskInfoPath(taskName));
+                taskInfos.add(Protos.TaskInfo.parseFrom(bytes));
+            } catch (KeeperException.NoNodeException e) {
+                // The task node exists, but it doesn't contain a TaskInfo node. This may occur if
+                // the only contents are a TaskStatus.
```
gabrielhartmann: This should never happen.  We should always have a TaskInfo.

src/main/java/org/apache/mesos/state/CuratorStateStore.java
nickbp: Done: storeStatus() now checks that an exactly-matching TaskInfo already exists

src/main/java/org/apache/mesos/state/CuratorStateStore.java
nickbp: Done: This special case was removed (with a comment added)

PR #54, opened 2016-06-28
Add registration and re-registration hooks to the CustomExecutor

gabrielhartmann:


src/test/java/org/apache/mesos/executor/CustomExecutorTest.java
```diff
@@ -4,25 +4,41 @@
...
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
+import static org.mockito.Mockito.times;
+
 public class CustomExecutorTest {
+    private Duration defaultSleep = Duration.ofMillis(100);
+    private Duration defaultTimeout = Duration.ofSeconds(1);
+    @Rule
+    public final ExpectedSystemExit exit = ExpectedSystemExit.none();
```
mohitsoni: That's neat. I didn't know it existed.

PR #55, opened 2016-06-29
Add ExecutorRequirement creation representing creating new

gabrielhartmann:
Executors and using existing Executors

### DIFF
--- a/src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
+++ b/src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
@@ -16,13 +16,48 @@ public class ExecutorRequirement {
     private ExecutorInfo executorInfo;
     private Collection<ResourceRequirement> resourceRequirements;
 
-    public ExecutorRequirement(ExecutorInfo unverifiedExecutorInfo)
+    /**
+     * This method generates one of two possible Executor requirements.  In the first case, if an empty Executor name is
+     * presented in the ExecutorInfo an ExecutorRequirement representing a need for a new Executor is returned.  In the
+     * second case, if an ExecutorInfo with a valid name is presented a requirement indicating use of an already running
+     * Executor is generated.
+     * @param executorInfo is the ExecutorInfo indicate what requirement should be generated.
+     * @return an ExecutorRequirement to be used to evaluate Offers by the OfferEvaluator
+     * @throws InvalidRequirementException when a malformed ExecutorInfo is presented indicating an invalid
+     * ExecutorRequirement.
+     */
+    public static ExecutorRequirement create(ExecutorInfo executorInfo)
+        throws InvalidRequirementException {
+        if (executorInfo.getExecutorId().getValue().isEmpty()) {
+            return createExecutorRequirement(executorInfo);
+        } else {
+            return getExistingExecutorRequirement(executorInfo);
+        }
+    }
+
+    private static ExecutorRequirement createExecutorRequirement(ExecutorInfo executorInfo)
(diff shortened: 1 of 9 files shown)


src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
```diff
@@ -16,13 +16,48 @@
     private ExecutorInfo executorInfo;
     private Collection<ResourceRequirement> resourceRequirements;
 
-    public ExecutorRequirement(ExecutorInfo unverifiedExecutorInfo)
+    /**
+     * This method generates one of two possible Executor requirements.  In the first case, if an empty Executor name is
```
nickbp: Empty ID, not name

src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
```diff
@@ -16,13 +16,48 @@
...
+    }
+
+    private ExecutorRequirement(ExecutorInfo executorInfo)
             throws InvalidRequirementException {
-        validateExecutorInfo(unverifiedExecutorInfo);
-        // ExecutorID is always overwritten with a new UUID, even if already present:
-        this.executorInfo = ExecutorInfo.newBuilder(unverifiedExecutorInfo)
-                .setExecutorId(ExecutorUtils.toExecutorId(unverifiedExecutorInfo.getName()))
-                .build();
+        validateExecutorInfo(executorInfo);
```
nickbp: Subjective: Validate call could be moved into new create() factory?

nickbp:
Would it be feasible to defer population of new executor IDs in the requirement until after the offer evaluator is run?

The offer evaluator could then use executor ID presence as an explicit signal to pair against a specific executor (and log when they don't match). This is currently being done implicitly via the resource label matching, but without an explicit check the task can still be launched despite the executor ID mismatch. Adding an explicit check with logging could make it easier to discover mistaken changes to the executor ID.

src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
gabrielhartmann: I like it in the constructor.  Someday someone may change how these are created.  Putting it in the constructor makes sure whatever path is taken to ultimate construction, validation still takes place.  I considered the approach you're describing.

src/main/java/org/apache/mesos/offer/ExecutorRequirement.java
gabrielhartmann: Thanks.

gabrielhartmann:
Your points about the ExecutorRequirements and population of the ExecutorID are good.  I had planned on doing as you describe later, but now is as good a time as any.  I'll do it now.

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -310,6 +331,10 @@ private TaskInfo getFulfilledTaskInfo(
               ExecutorInfo.newBuilder(execInfo)
                       .clearResources();
 
+      if (execInfo.getExecutorId().getValue().isEmpty()) {
+        execBuilder.setExecutorId(ExecutorUtils.toExecutorId(execInfo.getName()));
+      }
+
```
gabrielhartmann: This isn't the right place for this.  I'm going to move it up the ExecutorReqirement fulfillment section.

gabrielhartmann: Hmmm... maybe not.

PR #56, opened 2016-06-29
Add an accept function with customizable filters in OfferAccepter

verma7:
This is needed so that frameworks (e.g., dcos-cassandra-service) can customize the filter.refuse_seconds when they accept offers.


gabrielhartmann:
@verma7:  Looks good to me.

PR #61, opened 2016-07-08
StateStore now support properties

mohitsoni:


src/main/java/org/apache/mesos/state/CuratorStateStore.java
```diff
@@ -282,6 +286,51 @@ public void clearTask(String taskName) throws StateStoreException {
...
+            throw new StateStoreException(e);
+        }
+    }
+
+    @Override
+    public Collection<String> listPropertyKeys() throws StateStoreException {
+        Collection<String> keys = new ArrayList<>();
+        try {
+            final Collection<String> children = curator.getChildren(this.propertiesPath);
+            keys.addAll(children);
```
gabrielhartmann: Just return here.

src/main/java/org/apache/mesos/state/CuratorStateStore.java
```diff
@@ -282,6 +286,51 @@ public void clearTask(String taskName) throws StateStoreException {
...
+        try {
+            final String path = this.propertiesPath + "/" + key;
+            logger.debug("Storing property key: {} into path: {}", key, path);
+            curator.store(path, value);
+        } catch (Exception e) {
+            throw new StateStoreException(e);
+        }
+    }
+
+    private void validateKey(String key) {
```
gabrielhartmann: Move this function below all the public functions.

src/main/java/org/apache/mesos/state/StateStore.java
```diff
@@ -148,4 +148,27 @@
...
      */
     Protos.TaskStatus fetchStatus(String taskName) throws StateStoreException;
+
+
+    // Property storage
+
+    /**
+     * Stores an arbitrary key/value pair.
+     *
+     * @param key The key should be a String, and should not contain forward slash ('/')
```
gabrielhartmann: should not contain "a" forward slash

src/main/java/org/apache/mesos/state/StateStore.java
```diff
@@ -148,4 +148,27 @@
...
+     * @param value The value should be a byte array.
+     */
+    void storeProperty(String key, byte[] value);
+
+    /**
+     * Fetches the value byte array, stored against the Property 'key'.
+     *
+     * @param key The key should be a String, and should not contain forward slash ('/')
```
gabrielhartmann: should not contain "a" forward slash
and the sentence should in a period.

src/test/java/org/apache/mesos/state/CuratorStateStoreTest.java
```diff
@@ -404,6 +402,26 @@ public void testStoreInfoThenStatus() throws Exception {
         assertTrue(store.fetchStatuses().isEmpty());
     }
 
+    @Test
+    public void testProperties() {
+        store.storeProperty("hey", "DC/OS".getBytes(StandardCharsets.UTF_8));
```
gabrielhartmann: "DC/OS" should be a constant in the test function as it is used twice.

src/test/java/org/apache/mesos/state/CuratorStateStoreTest.java
```diff
@@ -404,6 +402,26 @@ public void testStoreInfoThenStatus() throws Exception {
...
+        final byte[] bytes = store.fetchProperty("hey");
+        assertEquals("DC/OS", new String(bytes, StandardCharsets.UTF_8));
+        final Collection<String> keys = store.listPropertyKeys();
+        assertTrue(keys.size() == 1);
+        assertEquals(keys.iterator().next(), "hey");
+    }
+
+    @Test(expected = StateStoreException.class)
+    public void testPropertiesSlashStore() {
+        store.storeProperty("hey/hi", "DC/OS".getBytes(StandardCharsets.UTF_8));
```
gabrielhartmann: Same for "hey/hi".  Change this to a variable named "BAD_KEY" or something.

PR #64, opened 2016-07-16
Fix offer evaluation loop bug.

gabrielhartmann:
If an OfferRequirement was not satisfied on the first Offer, all Offers were being rejected.  The test included here failed without the provided fix.  The rest of the change is refactoring of test code for clarity.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -88,7 +88,7 @@ ext {
 }
 
 group = "mesosphere"
-version = "0.4.21-SNAPSHOT"
+version = "0.4.22-SNAPSHOT"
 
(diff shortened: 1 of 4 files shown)


gabrielhartmann:
I tested this against both Kafka and Cassandra.  I need to unblock PRs over on those two projects.  @nickbp: Let me know if you see any changes that are needed and we can make them in a separate PR.

nickbp:
LGTM

gabrielhartmann:
Thanks @nickbp.  Sorry about the early merge.

PR #65, opened 2016-07-18
Added a clearProperty method

mohitsoni:


gabrielhartmann:
@mohitsoni:  Please write some separate tests.  Things to validate:
1. Clear good value
2. Clear non-existent value
3. Clear empty value
4. Exercise the Exception catch block.

PR #66, opened 2016-07-18
Bumped version up

mohitsoni:


mohitsoni:
@gabrielhartmann said merge it if tests pass.

PR #68, opened 2016-07-20
Added ConfigValidation interface

mohitsoni:


src/main/java/org/apache/mesos/config/ConfigurationValidation.java
```diff
@@ -0,0 +1,19 @@
+package org.apache.mesos.config;
+
+import java.util.List;
+
+/**
+ * The {@code ConfigurationValidation} interface should be implemented by any class
+ * which intend to validate a new {@code Configuration} w.r.t. an old {@code Configuration}.
+ */
+public interface ConfigurationValidation {
```
gabrielhartmann: ConfigurationValidator

src/main/java/org/apache/mesos/config/ConfigurationValidationError.java
```diff
@@ -0,0 +1,27 @@
+package org.apache.mesos.config;
+
+/**
+ * Signals a certain configurationField is incorrectly configured.
+ */
+public class ConfigurationValidationError {
+    private final String configurationField;
+    private final String message;
+
+    public ConfigurationValidationError(String configField, String message) {
```
gabrielhartmann: Should also contain `String configValue`?

mohitsoni: Good point!

src/main/java/org/apache/mesos/config/ConfigurationValidation.java
nickbp: Nit: ConfigurationValidator?

src/main/java/org/apache/mesos/config/ConfigurationValidationException.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.mesos.config;
```
nickbp: Looks like this exception isn't being used. Remove?

src/main/java/org/apache/mesos/config/ConfigurationValidation.java
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * Returns {@code List} of {@code ConfigurationValidationError} for the newly supplied
+     * {@code Configuration} object.
+     *
+     * @param oldConfig Currently persisted Configuration.
+     * @param newConfig Proposed new Configuration
+     * @return List of errors
```
nickbp: "List of errors, or an empty list if validation passed"

nickbp:
For 'validate the new config, regardless of the old config' (eg 'ram should be non-zero'), would that just be implemented as a ConfigValidator that ignores the oldConfig parameter?

src/main/java/org/apache/mesos/config/DefaultConfigurationValidator.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.List;
+
+/**
+ * Default Configuration validator, which validates a given new {@code Configuration} w.r.t.
+ * an old {@code Configuration}.
+ */
+public class DefaultConfigurationValidator {
+    private final Collection<ConfigurationValidation> validations = new ArrayList<>();
+
+    public DefaultConfigurationValidator(final Collection<ConfigurationValidation> validations) {
```
nickbp: Would be useful to have an alternate constructor of the form:

```
Constructor(ConfigurationValidator... validators) {
  this(Arrays.asList(validators));
}
```

Then callers could create a new validator like this:

```
new DefaultConfigurationValidator(
  new MyValidatorA(),
  new MyValidatorB(),
  ...);
```

src/main/java/org/apache/mesos/config/ConfigurationValidationException.java
mohitsoni: This is intended for SDK users.

src/main/java/org/apache/mesos/config/DefaultConfigurationValidator.java
mohitsoni: Good suggestion.

src/main/java/org/apache/mesos/config/ConfigurationValidationException.java
gabrielhartmann: I agree with @nickbp.  If it's not being used yet, we should remove it.

mohitsoni:
@nickbp Regarding new config only validation. See here: https://github.com/mesosphere/dcos-cassandra-service/blob/master/cassandra-scheduler/src/main/java/com/mesosphere/dcos/cassandra/scheduler/config/ConfigValidator.java#L32

gabrielhartmann:
@mohitsoni:  Not sure I understand what you're referring to in that link.  I could be wrong but I think @nickbp is talking about the case where there is no `oldConfig`.  Some validation should occur even on first startup to guard against things Universe validation can't catch.

mohitsoni:
@gabrielhartmann That was what I was trying to point out. That link demonstrates that we verify  `seeds < servers` always. Presence of an old config doesn't matter, in that example.

mohitsoni:
We can design(overdesign?) it, by having two separate validate methods to make the intention more explicit:
`validate(newConfig)`
`validate(oldConfig, newConfig)`

gabrielhartmann:
@mohitsoni:  I see, so on this line we're comparing `newConfig` to `newConfig`: https://github.com/mesosphere/dcos-cassandra-service/blob/master/cassandra-scheduler/src/main/java/com/mesosphere/dcos/cassandra/scheduler/config/ConfigValidator.java#L35

That's a little confusing from an interface perspective.  It would be nice to be able to specify Validations that take a single Configuration to make clear that it isn't doing a comparison to another configuration.

mohitsoni:
@gabrielhartmann Please see my earlier comment.

nickbp:
TBH I'm ok with using the same type/interface for both self-validation and cross-validation, was just thinking it'd be good to document the two possibilities.

gabrielhartmann:
Ok, just a thought.  Can be changed alter if needed.  Don't block on my comments.  LGTM.

mohitsoni:
@nickbp That was my un-documented intent. But, as I pointed out above, we can make things more explicit by having dedicated methods. It's a design choice.

nickbp:
LGTM

PR #70, opened 2016-07-21
Added Ken's QueuedSchedulerDriver

mohitsoni:


src/main/java/org/apache/mesos/testing/QueuedSchedulerDriver.java
```diff
@@ -281,7 +289,7 @@ public void printState() {
   @Override
   public Protos.Status launchTasks(Collection<Protos.OfferID> offerIds,
                                    Collection<Protos.TaskInfo> tasks) {
-    return launchTasks(offerIds, tasks);
+    return launchTasks(offerIds, tasks, null);
```
kow3ns: Can we make this Filters.DefaultInstance instead of null?

src/main/java/org/apache/mesos/testing/QueuedSchedulerDriver.java
```diff
@@ -294,7 +302,7 @@ public void printState() {
   @Override
   public Protos.Status launchTasks(Protos.OfferID offerId,
                                    Collection<Protos.TaskInfo> tasks) {
-    return launchTasks(offerId, tasks);
+    return launchTasks(offerId, tasks, null);
   }
```
kow3ns: Same as above

PR #71, opened 2016-07-21
BREAKING CHANGE: Put ZK data under a /dcos-service- prefix, while also implementing a stored schema version to simplify future migrations

nickbp:
1. BREAKING CHANGE of moving framework data under a '/dcos-service-' prefix:
The '/dcos-service-' prefix solves both of the following cases:
= Avoiding problems if someone creates a framework named 'mesos' (current problem)
= Avoiding deleting too much if someone mistakenly deletes the wrong node (problem for shared '/services/' dir)

2. Adds 'SchemaVersion':
Doesn't implement any migration, just starts off by storing a default value of "1", while updating CuratorConfigStore/CuratorStateStore to start storing/using the value.

### DIFF
--- a/src/main/java/org/apache/mesos/curator/CuratorConfigStore.java
+++ b/src/main/java/org/apache/mesos/curator/CuratorConfigStore.java
@@ -1,13 +1,15 @@
-package org.apache.mesos.config;
+package org.apache.mesos.curator;
 
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.UUID;
 
 import org.apache.curator.RetryPolicy;
 import org.apache.curator.retry.ExponentialBackoffRetry;
-import org.apache.mesos.state.CuratorSchemaVersionStore;
+import org.apache.mesos.config.ConfigStore;
+import org.apache.mesos.config.ConfigStoreException;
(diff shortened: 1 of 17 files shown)


src/main/java/org/apache/mesos/curator/CuratorConfigStore.java
```diff
@@ -41,30 +49,50 @@
...
-        this.configurationsPath = rootPath + "/" + CONFIGURATIONS_PATH_NAME;
+        String rootPath = CuratorUtils.toServiceRootPath(frameworkName);
+
+        // Check version up-front:
+        SchemaVersionStore versionStore = new CuratorSchemaVersionStore(curator, rootPath);
+        int currentVersion = versionStore.fetch();
+        if (!SchemaVersionStore.isSupported(
+                currentVersion, MIN_SUPPORTED_SCHEMA_VERSION, MAX_SUPPORTED_SCHEMA_VERSION)) {
+            throw new IllegalStateException(String.format(
+                    "Storage schema version %d is not supported by this software", currentVersion));
```
gabrielhartmann: Might be nice to include the MIN and MAX supported versions here in the exception, for logging/debugging purposes.

nickbp: Done

PR #73, opened 2016-07-28
Add DCOS Cluster Capabilities

gabrielhartmann:
In order to make decisions regarding whether or not to attempt to use a particular feature it is necessary to know whether a particular version of DC/OS supports the feature.  The code here exposes the first example of determining whether a capability (named VIPs) is supported.

### DIFF
--- a/src/main/java/org/apache/mesos/dcos/Capabilities.java
+++ b/src/main/java/org/apache/mesos/dcos/Capabilities.java
@@ -1,0 +1,19 @@
+package org.apache.mesos.dcos;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+
+/**
+ * This class represents a set of capabilities that may or may not be supported in a given version of DC/OS.
+ */
+public class Capabilities {
+    private DcosCluster dcosCluster;
+
+    public Capabilities(DcosCluster dcosCluster) {
+        this.dcosCluster = dcosCluster;
+    }
+
+    public boolean supportsNamedVips() throws IOException, URISyntaxException {
+        return dcosCluster.getDcosVersion().getVersion().startsWith("1.8");
+    }
+}
--- a/src/main/java/org/apache/mesos/dcos/DcosCluster.java
+++ b/src/main/java/org/apache/mesos/dcos/DcosCluster.java
@@ -13,10 +13,10 @@ import java.util.Optional;
  * Instances of this class represent DC/OS clusters.
  */
 public class DcosCluster {
-    public static final String INTERNAL_DCOS_URL = "http://master.mesos";
+    private static final String INTERNAL_DCOS_URL = "http://master.mesos";
     private static final String DCOS_VERSION_PATH = "/dcos-metadata/dcos-version.json";
 
-    private URI dcosUri;
+    private final URI dcosUri;
     private Optional<DcosVersion> dcosVersion = Optional.empty();
 
     DcosCluster(URI dcosUri) {
--- a/src/main/java/org/apache/mesos/dcos/DcosCluster.java
+++ b/src/main/java/org/apache/mesos/dcos/DcosCluster.java
@@ -1,0 +1,45 @@
+package org.apache.mesos.dcos;
+
+import org.apache.http.client.fluent.Content;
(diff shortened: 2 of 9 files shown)


src/main/java/org/apache/mesos/dcos/DcosCluster.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public DcosCluster(URI dcosUri) {
+        this.dcosUri = dcosUri;
+    }
+
+    public URI getDcosUri() {
+        return  dcosUri;
+    }
+
+    public DcosVersion getDcosVersion() throws IOException, URISyntaxException {
```
nickbp: Cache or fetch once on startup?

src/main/java/org/apache/mesos/dcos/DcosVersion.java
```diff
@@ -0,0 +1,56 @@
...
+public class DcosVersion {
+    public static final String BOOTSTRAP_ID_KEY = "bootstrap-id";
+    public static final String DCOS_IMAGE_COMMIT = "dcos-image-commit";
+    public static final String VERSION_KEY = "version";
+
+    private String bootstrapId;
+    private String dcosImageCommit;
+    private String version;
+
+    public DcosVersion(String bootstrapId, String dcosImageCommit, String version) {
```
nickbp: Constructors could be package-private

src/main/java/org/apache/mesos/dcos/DcosCluster.java
```diff
@@ -0,0 +1,33 @@
...
+import java.net.URISyntaxException;
+
+/**
+ * Instances of this class represent DC/OS clusters.
+ */
+public class DcosCluster {
+    private URI dcosUri;
+    private static final String DCOS_VERSION_PATH = "/dcos-metadata/dcos-version.json";
+
+    public DcosCluster(URI dcosUri) {
```
nickbp: What is the expected format for this URI? Where does a developer get it from?

src/main/java/org/apache/mesos/dcos/Capabilities.java
```diff
@@ -0,0 +1,19 @@
...
+import java.io.IOException;
+import java.net.URISyntaxException;
+
+/**
+ * This class represents a set of capabilities that may or may not be supported in a given version of DC/OS.
+ */
+public class Capabilities {
+    private DcosCluster dcosCluster;
+
+    public Capabilities(DcosCluster dcosCluster) {
```
nickbp: If DcosCluster URI can be autodetected (see comment there): Construct DcosCluster internally?
Else: Accept URI directly and construct DcosCluster from that?

src/main/java/org/apache/mesos/dcos/DcosCluster.java
nickbp: Assuming DcosCluster is constructed by Capabilities, this constructor could be package-private

src/main/java/org/apache/mesos/dcos/Capabilities.java
gabrielhartmann: This exposes the Capabilities of a DcosCluster, not the Capabilities of a URI.

src/main/java/org/apache/mesos/dcos/DcosCluster.java
gabrielhartmann: Done.

src/main/java/org/apache/mesos/dcos/DcosVersion.java
gabrielhartmann: Done.

src/main/java/org/apache/mesos/dcos/Capabilities.java
```diff
@@ -0,0 +1,19 @@
...
+        this.dcosCluster = dcosCluster;
+    }
+
+    public boolean supportsNamedVips() throws IOException, URISyntaxException {
+        return dcosCluster.getDcosVersion().getVersion().startsWith("1.8");
```
triclambert: I recommend we use TODOs in the code to identify future work.

PR #75, opened 2016-07-31
Implement a Go-based CLI module SDK

nickbp:
The current Python-based CLI modules are very simple (thin convenience wrappers around HTTP endpoints), but building Python binaries is very complicated/brittle. To resolve this, let's use a language that's trivial to cross-compile.
    
Includes:
- Helpers for:
  - Querying the service (with auth token/url automatically pulled from 'dcos config')
  - Pretty-printing data
- Implementations of all common scheduler HTTP endpoints (e.g. plan/config/state)
  - Easy for developers to add custom commands for implementation-specific endpoints.
- 'Example' CLI module which adds an `echo` command on top of the common commands, as a sample of what a framework developer would do.
- Sample `build.sh` for building Lin/Win/Mac binaries from any system which has go tools installed (see README)

### DIFF
--- a/cli/http.go
+++ b/cli/http.go
@@ -1,0 +1,128 @@
+package cli
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+	"path"
+)
+
+var (
+	dcosAuthToken string
+	dcosUrl       string
+	serviceName   string
+)
+
+func HTTPGet(urlPath string) *http.Response {
+	return HTTPQuery(CreateHTTPRequest("GET", urlPath), true)
+}
+func HTTPGetData(urlPath, payload, contentType string) *http.Response {
+	return HTTPQuery(CreateHTTPDataRequest("GET", urlPath, payload, contentType), true)
+}
+func HTTPGetJSON(urlPath, jsonPayload string) *http.Response {
+	return HTTPQuery(CreateHTTPJSONRequest("GET", urlPath, jsonPayload), true)
+}
+
+func HTTPDelete(urlPath string) *http.Response {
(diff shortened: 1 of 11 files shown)


nickbp:
Note: the huge line count is mainly due to vendored inclusion of a 3rd-party arg processing lib

https://blog.gopheracademy.com/advent-2015/vendor-folder/

cli/http.go
```diff
@@ -0,0 +1,91 @@
...
+		(response.StatusCode < 200 || response.StatusCode >= 300) {
+		log.Fatalf("HTTP %s Query for %s failed: %s", method, request.URL, response.Status)
+	}
+	if Verbose {
+		log.Printf("Response: %s (%d bytes)", response.Status, response.ContentLength)
+	}
+	return response
+}
+
+func CreateHTTPRequest(method string, urlPath string, jsonString string) *http.Request {
```
mgummelt: I'm not a big fan of having to rewrite the components of the dcos cli that we need, to get around the lack of a `dcos` python module.  I realize we have to do that if we're writing in go, but still, kinda sucks...

nickbp:
FYI: Dropped the huge vendor directory containing the CLI library, making this PR much smaller. With it present, I saw weird dependency issues while reimplementing existing CLI modules in other repos. I suspect it's that libraries aren't supposed to have vendored libraries of their own.

`./main.go:17: cannot use app (type *"github.com/mesosphere/dcos-commons/cli/vendor/gopkg.in/alecthomas/kingpin.v2".Application) as type *"gopkg.in/alecthomas/kingpin.v2".Application in argument to handleExampleSection`

PR #76, opened 2016-08-01
Test transition from TASK_LOST to TASK_RUNNING

mrbrowning:


src/test/java/org/apache/mesos/reconciliation/DefaultReconcilerTest.java
```diff
@@ -192,6 +192,32 @@ public void testForceCompleteReconciler() {
...
     }
 
+    @Test
+    public void testTaskLostToTaskRunningTransition() {
+        reconciler.start(Arrays.asList(TASK_STATUS_2));
+
+        assertFalse(reconciler.isReconciled());
+        assertEquals(1, reconciler.remaining().size());
+
+        reconciler.reconcile(mockDriver);
```
gabrielhartmann: This call is supposed to do nothing right?  We should validate it does nothing or remove the call.

PR #77, opened 2016-08-01
Automatically repairing scheduler implementation

dgrnbrg:
This PR implements a new scheduler called `RepairScheduler`. This helps frameworks handle failure & restarts for persistent-disk based applications.

This is implementation complete, but still needs tests that do not depend on the proprietary libraries they were originally developed against.

### DIFF
--- a/src/main/java/org/apache/mesos/config/RepairConfiguration.java
+++ b/src/main/java/org/apache/mesos/config/RepairConfiguration.java
@@ -1,0 +1,93 @@
+package org.apache.mesos.config;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Represents the repair configuration in JSON.
+ */
+public class RepairConfiguration {
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+        return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        RepairConfiguration that = (RepairConfiguration) o;
+
+        if (gracePeriodMins != that.gracePeriodMins) {
+            return false;
+        }
+        if (repairDelaySecs != that.repairDelaySecs) {
+            return false;
+        }
+        return enableReplacement == that.enableReplacement;
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = gracePeriodMins;
+        result = 31 * result + repairDelaySecs;
+        result = 31 * result + (enableReplacement ? 1 : 0);
+        return result;
+    }
+
+    @JsonProperty("repair_in_place_grace_period_mins")
+    private int gracePeriodMins;
+    @JsonProperty("min_delay_between_repairs_secs")
+    private int repairDelaySecs;
+    @JsonProperty("enable_replacement")
+    private boolean enableReplacement;
+
+    public  RepairConfiguration() {}
+
+    @JsonCreator
+    public RepairConfiguration(
+            @JsonProperty("repair_in_place_grace_period_mins") int gracePeriodMins,
+            @JsonProperty("min_delay_between_repairs_secs") int repairDelaySecs,
+            @JsonProperty("enableReplacement") boolean enableReplacement) {
+        this.gracePeriodMins = gracePeriodMins;
+        this.repairDelaySecs = repairDelaySecs;
+        this.enableReplacement = enableReplacement;
+    }
+
+    public boolean isReplacementEnabled() {
+        return enableReplacement;
+    }
+
+    @JsonProperty("enable_replacement")
+    public void setEnableReplacement(boolean enableReplacement) {
+        this.enableReplacement = enableReplacement;
+    }
+
+    public int getRepairDelaySecs() {
+        return repairDelaySecs;
+    }
+
+    @JsonProperty("min_delay_between_repairs_secs")
+    public void setRepairDelaySecs(int repairDelaySecs) {
+        this.repairDelaySecs = repairDelaySecs;
+    }
+
+    public int getGracePeriodMins() {
+        return gracePeriodMins;
+    }
+
+    @JsonProperty("repair_in_place_grace_period_mins")
+    public void setGracePeriodMins(int gracePeriodMins) {
+        this.gracePeriodMins = gracePeriodMins;
+    }
+
+    @Override
+    public String toString() {
+        return "RepairConfiguration{" +
+                "gracePeriodMins=" + gracePeriodMins +
+                ", repairDelaySecs=" + repairDelaySecs +
+                ", enableReplacement=" + enableReplacement +
+                '}';
+    }
+}
--- a/src/main/java/org/apache/mesos/scheduler/repair/RepairOfferRequirementProvider.java
+++ b/src/main/java/org/apache/mesos/scheduler/repair/RepairOfferRequirementProvider.java
@@ -1,0 +1,51 @@
+package org.apache.mesos.scheduler.repair;
+
+import org.apache.mesos.Protos.TaskInfo;
+import org.apache.mesos.config.ConfigStoreException;
+import org.apache.mesos.offer.InvalidRequirementException;
+import org.apache.mesos.offer.OfferRequirement;
+import org.apache.mesos.scheduler.plan.Block;
+import org.apache.mesos.scheduler.repair.monitor.FailureMonitor;
+
+import java.util.Optional;
+
+/**
+ * Implementions of this know how to relaunch tasks from a framework when those tasks crash or fail.
+ * <p>
+ * The method {@link RepairOfferRequirementProvider#getReplacementOfferRequirement(TaskInfo)} is called once a given
+ * task crashes. The implementation should look at the given {@link TaskInfo} in order to know which task it's
+ * replacing, and where that task's resources & data have been reserved.
+ * <p>
+ * The method {@link RepairOfferRequirementProvider#maybeGetNewOfferRequirement(String, Block)} is called once the
+ * framework's {@link FailureMonitor} decides that a task has permanently failed, and we should try to find a new
+ * location to run it. Typicaly, this means that if the task represented a database, the data on that node is
+ * permanently lost.
+ */
+public interface RepairOfferRequirementProvider {
+    /**
+     * Possibly returns an {@link OfferRequirement} that will replace a missing task.
+     * <p>
+     * Implementations should not try to relaunch the current {@link Block}'s task.
+     *
+     * @param targetConfigName The current configuration we're moving towards
+     * @param currentBlock     The current {@link Block} being executed
+     * @return {@link OfferRequirement} to relaunch a failed task, or none if all tasks are present
+     * @throws InvalidRequirementException
+     * @throws ConfigStoreException
+     */
+    Optional<OfferRequirement> maybeGetNewOfferRequirement(String targetConfigName, Block currentBlock)
+            throws InvalidRequirementException, ConfigStoreException;
+
+    /**
+     * Computes a new {@link OfferRequirement} that will relaunch-in-place the given {@link TaskInfo}.
+     * <p>
+     * This will only be called when the argument is in a terminal state, before the {@link FailureMonitor} decides that
+     * task is definitely gone forever. The implementation should ensure it's reusing the resources that were reserved
+     * by the given {@literal terminatedTask}.
+     *
+     * @param terminatedTask The task that has stopped and needs to be relaunched.
+     * @return {@link OfferRequirement} to relaunch the stopped task in place
+     * @throws InvalidRequirementException
+     */
+    OfferRequirement getReplacementOfferRequirement(TaskInfo terminatedTask) throws InvalidRequirementException;
+}
--- a/src/main/java/org/apache/mesos/scheduler/repair/RepairResource.java
+++ b/src/main/java/org/apache/mesos/scheduler/repair/RepairResource.java
@@ -1,0 +1,37 @@
+package org.apache.mesos.scheduler.repair;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Simple endpoint to expose repair state information.
+ * <p>
+ * Renders the stopped tasks and failed tasks as JSON, so that the operator can understand the status of the system.
+ */
+@Path("/v1/repair")
+public class RepairResource {
+    private static final Log log = LogFactory.getLog(RepairResource.class);
+
+    private final AtomicReference<RepairStatus> repairStatus;
+
+    public RepairResource(AtomicReference<RepairStatus> repairStatus) {
+        this.repairStatus = repairStatus;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public RepairStatus repair() {
+        try {
+            return repairStatus.get();
+        } catch (Exception ex) {
+            log.error("Failed to fetch data with exception:", ex);
+            throw ex;
+        }
+    }
+}
--- a/src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
+++ b/src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
@@ -1,0 +1,177 @@
+package org.apache.mesos.scheduler.repair;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.mesos.Protos.Offer;
+import org.apache.mesos.Protos.Offer.Operation;
+import org.apache.mesos.Protos.OfferID;
+import org.apache.mesos.Protos.TaskInfo;
+import org.apache.mesos.SchedulerDriver;
+import org.apache.mesos.offer.OfferAccepter;
+import org.apache.mesos.offer.OfferEvaluator;
+import org.apache.mesos.offer.OfferRecommendation;
+import org.apache.mesos.offer.OfferRequirement;
+import org.apache.mesos.scheduler.plan.Block;
+import org.apache.mesos.scheduler.repair.constrain.LaunchConstrainer;
+import org.apache.mesos.scheduler.repair.monitor.FailureMonitor;
+import org.apache.mesos.state.StateStore;
+
+import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+
+/**
+ * This class provides the core functionality of the automatic repair mechanisms.
+ */
+public class RepairScheduler {
+    private final Log log = LogFactory.getLog(getClass());
+
+    private final String targetConfigName;
+    private final StateStore stateStore;
+    private final OfferAccepter offerAccepter;
+    private final TaskFailureListener failureListener;
+    private final RepairOfferRequirementProvider offerReqProvider;
+    private final FailureMonitor failureMonitor;
+    private final LaunchConstrainer launchConstrainer;
+    private final AtomicReference<RepairStatus> repairStatusRef;
+
+    public RepairScheduler(
+            String targetConfigName,
+            StateStore stateStore,
+            TaskFailureListener failureListener,
+            RepairOfferRequirementProvider offerReqProvider,
+            OfferAccepter offerAccepter,
+            LaunchConstrainer launchConstrainer,
+            FailureMonitor failureMonitor,
+            AtomicReference<RepairStatus> repairStatusRef) {
+        this.targetConfigName = targetConfigName;
+        this.stateStore = stateStore;
+        this.offerReqProvider = offerReqProvider;
+        this.offerAccepter = offerAccepter;
+        this.failureMonitor = failureMonitor;
+        this.launchConstrainer = launchConstrainer;
+        this.repairStatusRef = repairStatusRef;
+        this.failureListener = failureListener;
+    }
+
+    /**
+     * This function runs the repair logic. It should be invoked periodically.
+     *
+     * This function is synchronized in order to avoid launchConstrainer race conditions.
+     * The situation we need to avoid is 2 threads getting approval from the launchConstrainer, when only one should
+     * have been allowed to proceed. If this issue was dealt with via another mechanism, we could make this function
+     * unsynchronized.
+     *
+     * @param driver The current SchedulerDriver
+     * @param offers A list of offers to use to launch tasks
+     * @param block The current block, or null if there isn't any
+     * @return IDs of accepted offers
+     * @throws Exception
+     */
+    public synchronized List<OfferID> resourceOffers(SchedulerDriver driver, List<Offer> offers, Block block)
+            throws Exception {
+        List<OfferID> acceptedOffers = new ArrayList<>();
+        List<TaskInfo> terminatedTasks = new ArrayList<>(getTerminatedTasks(block));
+
+        // Compute the stopped and failed pool for UI
+        RepairStatus oldStatus = repairStatusRef.get();
+        if (oldStatus == null) {
+            oldStatus = new RepairStatus(Collections.emptyList(), Collections.emptyList());
+        }
+        List<String> oldFailed = oldStatus.getFailed();
+        List<String> newFailed = new ArrayList<>(terminatedTasks.stream()
+                .filter(failureMonitor::hasFailed)
+                .map(TaskInfo::getName)
+                .collect(Collectors.toList()));
+        newFailed.addAll(oldFailed);
+        //TODO (ghartman) possibly write this list down in ZK, to ensure the UI doesn't change upon restart
+        newFailed = newFailed.stream().distinct().collect(Collectors.toList());
+        List<String> newStopped = terminatedTasks.stream()
+                .filter(it -> !failureMonitor.hasFailed(it))
+                .map(TaskInfo::getName)
+                .collect(Collectors.toList());
+
+        Optional<OfferRequirement> offerReq = Optional.empty();
+        boolean replacingFailed = false;
+
+        if (terminatedTasks.size() > 0) {
+            TaskInfo terminatedTask = terminatedTasks.get(new Random().nextInt(terminatedTasks.size()));
+            log.info("Found stopped task");
+            if (failureMonitor.hasFailed(terminatedTask)) {
+                log.info("Marking stopped task as failed");
+                failureListener.taskFailed(terminatedTask.getTaskId());
+            } else {
+                log.info("Relaunching stopped task in place");
+                offerReq = Optional.of(offerReqProvider.getReplacementOfferRequirement(terminatedTask));
+            }
+        } else {
+            log.info("Launching a failed task");
+            replacingFailed = true;
+            offerReq = offerReqProvider.maybeGetNewOfferRequirement(targetConfigName, block);
+        }
+
+        if (offerReq.isPresent() && launchConstrainer.canLaunch(offerReq.get())) {
+            log.info("Preparing to launch task");
+            OfferEvaluator offerEvaluator = new OfferEvaluator();
+            List<OfferRecommendation> recommendations = offerEvaluator.evaluate(offerReq.get(), offers);
+            if (!recommendations.isEmpty()) {
+                // we're replacing some task(s)
+                Set<String> relaunchedTasks = recommendations.stream()
+                        .filter(it -> it.getOperation().hasLaunch())
+                        .flatMap(it -> it.getOperation().getLaunch().getTaskInfosList().stream())
+                        .map(TaskInfo::getName)
+                        .collect(Collectors.toSet());
+                if (replacingFailed) {
+                    // we'll filter their names out of the failed pool
+                    newFailed = newFailed.stream()
+                            .filter(it -> !relaunchedTasks.contains(it))
+                            .collect(Collectors.toList());
+                } else {
+                    // we'll filter their names out of the stopped pool
+                    newStopped = newStopped.stream()
+                            .filter(it -> !relaunchedTasks.contains(it))
+                            .collect(Collectors.toList());
+                }
+            }
+            List<Operation> launchOperations = recommendations.stream()
+                    .map(OfferRecommendation::getOperation)
+                    .filter(Operation::hasLaunch)
+                    .collect(Collectors.toList());
+            if (launchOperations.size() > 1) {
+                throw new IllegalArgumentException("Repairs should only launch one task at a time, found "
+                        + launchOperations.size() + " tasks");
+            }
+            acceptedOffers = offerAccepter.accept(driver, recommendations);
+            if (launchOperations.size() == 1) {
+                // We could've launched nothing if the offer didn't fit
+                launchConstrainer.launchHappened(launchOperations.get(0));
+            }
+        }
+
+        repairStatusRef.set(new RepairStatus(newStopped, newFailed));
+
+        return acceptedOffers;
+    }
+
+    private Collection<TaskInfo> getTerminatedTasks(Block block) {
+        List<TaskInfo> filteredTerminatedTasks = new ArrayList<TaskInfo>();
+
+        try {
+            if (block == null) {
+                return stateStore.fetchTerminatedTasks();
+            }
+
+            String blockName = block.getName();
+            for (TaskInfo taskInfo : stateStore.fetchTerminatedTasks()) {
+                if (!taskInfo.getName().equals(blockName)) {
+                    filteredTerminatedTasks.add(taskInfo);
+                }
+            }
+        } catch (Exception ex) {
+            log.error("Stopped to fetch terminated tasks.");
+        }
+
+        return filteredTerminatedTasks;
+    }
+
+}
--- a/src/main/java/org/apache/mesos/scheduler/repair/TaskFailureListener.java
+++ b/src/main/java/org/apache/mesos/scheduler/repair/TaskFailureListener.java
@@ -1,0 +1,15 @@
+package org.apache.mesos.scheduler.repair;
+
+import org.apache.mesos.Protos.TaskID;
+
+/**
+ * Exposes simple task-oriented API for components that need to interact with the framework as a whole.
+ */
+public interface TaskFailureListener {
+    /**
+     * Notifies that the task identified by the given {@link TaskID} has failed.
+     *
+     * @param taskId The ID of the task to delete
+     */
+    void taskFailed(TaskID taskId);
+}
--- a/src/main/java/org/apache/mesos/scheduler/repair/constrain/TestingLaunchConstrainer.java
+++ b/src/main/java/org/apache/mesos/scheduler/repair/constrain/TestingLaunchConstrainer.java
@@ -1,0 +1,31 @@
+package org.apache.mesos.scheduler.repair.constrain;
+
+import org.apache.mesos.Protos.Offer.Operation;
+import org.apache.mesos.offer.OfferRequirement;
+
+/**
+ * {@link LaunchConstrainer} that makes it easy to enable/disable launches for testing.
+ * <p>
+ * Defaults to allowing all launches.
+ */
+public class TestingLaunchConstrainer implements LaunchConstrainer {
+    private boolean canLaunch;
+
+    public TestingLaunchConstrainer() {
+        this.canLaunch = false;
+    }
+
+    public void setCanLaunch(boolean canLaunch) {
+        this.canLaunch = canLaunch;
+    }
+
+    @Override
+    public void launchHappened(Operation launchOperation) {
+        // Does nothing when the launch happens
+    }
+
+    @Override
+    public boolean canLaunch(OfferRequirement offerRequirement) {
+        return this.canLaunch;
+    }
+}
--- a/src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
+++ b/src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
@@ -1,0 +1,38 @@
+package org.apache.mesos.scheduler.repair.constrain;
+
+import org.apache.mesos.Protos.Offer.Operation;
+import org.apache.mesos.offer.OfferRequirement;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Implements a {@link LaunchConstrainer} that requires a minimum number of seconds to elapse between launches, for
+ * rate-limiting purposes.
+ */
+public class TimedLaunchConstrainer implements LaunchConstrainer {
+    private AtomicLong lastLaunch;
+    private long minDelaySeconds;
+
+    /**
+     * Create a new constrainer with the given required minimum delay between recovery operations.
+     *
+     * @param minDelaySeconds Minimum number of seconds between each launch
+     */
+    public TimedLaunchConstrainer(long minDelaySeconds) {
+        this.minDelaySeconds = minDelaySeconds;
+        this.lastLaunch = new AtomicLong(0);
+    }
+
+    @Override
+    public void launchHappened(Operation launchOperation) {
+        if (!canLaunch(null)) {
+            throw new RuntimeException("It hasn't been long enough since the last launch");
+        }
+        lastLaunch.compareAndSet(lastLaunch.get(), System.currentTimeMillis());
+    }
+
+    @Override
+    public boolean canLaunch(OfferRequirement offerRequirement) {
+        return (lastLaunch.get() + minDelaySeconds * 1000) < System.currentTimeMillis();
+    }
+}
--- a/build.gradle
+++ b/build.gradle
@@ -96,7 +96,7 @@ ext {
 }
 
 group = "mesosphere"
-version = "0.4.29-SNAPSHOT"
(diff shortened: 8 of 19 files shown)


gabrielhartmann:
@dgrnbrg: This doesn't build.  `./gradlew clean build` yields:
```
$ ./gradlew clean build
:clean
:acme-scheduler-example:clean UP-TO-DATE
:compileJava
Note: /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/testing/QueuedSchedulerDriver.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
:compileGroovy UP-TO-DATE
:processResources UP-TO-DATE
:classes
:jar
:startScripts
:distTar
:distZip
:assemble
:checkstyleMain
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/RepairController.java:3:1: Redundant import from the same package - org.apache.mesos.scheduler.repair.RepairStatus.
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/RepairOfferRequirementProvider.java:15: Line is longer than 120 characters (found 121).
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/RepairOfferRequirementProvider.java:19: Line is longer than 120 characters (found 126).
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/RepairOfferRequirementProvider.java:20: Line is longer than 120 characters (found 121).
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/RepairOfferRequirementProvider.java:39: Line is longer than 120 characters (found 161).
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java:19: Missing a Javadoc comment.
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java:63: All TODOs should be named.  e.g. "TODO (johndoe): Refactor when v2 is released."
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java:110: Line is longer than 120 characters (found 129).
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/constrain/AllLaunchConstrainer.java:6:1: Redundant import from the same package - org.apache.mesos.scheduler.repair.constrain.LaunchConstrainer.
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/constrain/TestingLaunchConstrainer.java:6:1: Redundant import from the same package - org.apache.mesos.scheduler.repair.constrain.LaunchConstrainer.
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java:6:1: Redundant import from the same package - org.apache.mesos.scheduler.repair.constrain.LaunchConstrainer.
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/constrain/UnconstrainedLaunchConstrainer.java:6:1: Redundant import from the same package - org.apache.mesos.scheduler.repair.constrain.LaunchConstrainer.
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/monitor/FailureMonitor.java:15: Line is longer than 120 characters (found 148).
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/monitor/TestingFailureMonitor.java:4:1: Redundant import from the same package - org.apache.mesos.scheduler.repair.monitor.FailureMonitor.
[ant:checkstyle] /Users/gabriel/tmp/dcos-commons/src/main/java/org/apache/mesos/scheduler/repair/monitor/TimedFailureMonitor.java:67: Line is longer than 120 characters (found 182).
:checkstyleMain FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':checkstyleMain'.
> Checkstyle rule violations were found. See the report at: file:///Users/gabriel/tmp/dcos-commons/build/reports/checkstyle/main.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 9.225 secs
```

dgrnbrg:
I have been building and running tests via IntelliJ. I'll fix those checkstyle issues as soon as I've finished implementing my test plan this evening. In the meantime, why don't you disable checkstyle or check out the code/design/etc in the meantime?

gabrielhartmann:
I'll review in the mean time, but this can't be merged into master.  @dgrnbrg, can you address the conflicts?

src/main/java/org/apache/mesos/config/RepairConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public int hashCode() {
+        int result = gracePeriodMins;
+        result = 31 * result + repairDelaySecs;
+        result = 31 * result + (enableReplacement ? 1 : 0);
+        return result;
+    }
+
+    @JsonProperty("repairInPlaceGracePeriodMins")
```
gabrielhartmann: All JSON should be snake_case

src/main/java/org/apache/mesos/config/RepairConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+    public RepairConfiguration(
+            @JsonProperty("repairInPlaceGracePeriodMins") int gracePeriodMins,
+            @JsonProperty("minDelayBetweenRepairsSecs") int repairDelaySecs,
+            @JsonProperty("enableReplacement") boolean enableReplacement) {
+        this.gracePeriodMins = gracePeriodMins;
+        this.repairDelaySecs = repairDelaySecs;
+        this.enableReplacement = enableReplacement;
+    }
+
+    public boolean isEnableReplacement() {
```
gabrielhartmann: method should be named `isReplacmentEnabled`

src/main/java/org/apache/mesos/config/RepairConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+        return enableReplacement;
+    }
+
+    @JsonProperty("enableReplacement")
+    public void setEnableReplacement(boolean enableReplacement) {
+        this.enableReplacement = enableReplacement;
+    }
+
+    public int getRepairDelaySecs() {
```
gabrielhartmann: Any reason we're not using something like `Duration` instead of an int of unknown type?  It's in the function name, but that I prefer types that are bit more expressive about what they represent.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+        return repairDelaySecs;
+    }
+
+    @JsonProperty("minDelayBetweenRepairsSecs")
+    public void setRepairDelaySecs(int repairDelaySecs) {
+        this.repairDelaySecs = repairDelaySecs;
+    }
+
+    public int getGracePeriodMins() {
```
gabrielhartmann: Same thing regarding `Duration` here.

src/main/java/org/apache/mesos/scheduler/repair/CoreFrameworkState.java
```diff
@@ -0,0 +1,44 @@
...
+    List<TaskInfo> getTerminatedTaskInfos() throws Exception;
+
+    /**
+     * Remove the given task from this framework. This could cause the given task to be killed
+     * if it's still running, and it will remove the task from the framework's tracking.
+     * The absence of the task will be fixed by the {@link RepairScheduler}, if that task
+     * is still supposed to be running.
+     * @param taskId The ID of the task to delete
+     */
+    void deleteTask(TaskID taskId);
```
gabrielhartmann: As, stated previously.  Deletion of a Task is an implementation detail of how Kafka works, not how all users of Auto Recovery should work.  I don't think this method belongs here at all, and it's meaning is difficult to understand.  `This could cause the given task to be killed`...  It's unclear what happens when this method is called.  It might kill something, but it will definitely delete something from persistent storage?

src/main/java/org/apache/mesos/scheduler/repair/CoreFrameworkState.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.mesos.Protos.TaskInfo;
+import org.apache.mesos.state.StateStoreException;
+
+import java.util.List;
+
+/**
+ * Exposes simple task-oriented API for components that need to interact with the framework
+ * as a whole.
+ */
+public interface CoreFrameworkState {
```
gabrielhartmann: Why does this interface exist?  It looks almost identical to a subset of StateStore.  It adds only `deleteTask` which I don't think belongs here, or has unclear meaning at least.  It also adds `getTerminatedTasks`.  Why not just add that to the StateStore?

`getFrameworkId` and `getTaskInfos` are redundant with `StateStore` capabilities

src/main/java/org/apache/mesos/scheduler/repair/RepairController.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.ws.rs.core.MediaType;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Simple endpoint to expose repair state information.
+ * <p>
+ * Renders the stopped tasks and failed tasks as JSON, so that the operator can understand the status of the system.
+ */
+@Path("/v1/repair")
+public class RepairController {
```
gabrielhartmann: In dcos-commons these types of classes are called `RepairResource`.  The convention in Kafka of using a `Controller` suffix is outdated.

src/main/java/org/apache/mesos/scheduler/repair/RepairOfferRequirementProvider.java
```diff
@@ -0,0 +1,54 @@
...
+ * The method {@link RepairOfferRequirementProvider#maybeGetNewOfferRequirement(String, Block)} is called once the
+ * framework's {@link FailureMonitor} decides that a task has permanently failed, and we should try to find a new
+ * location to run it. Typicaly, this means that if the task represented a database, the data on that node is
+ * permanently lost.
+ */
+public interface RepairOfferRequirementProvider {
+    /**
+     * Possibly returns an {@link OfferRequirement} that will replace a missing task.
+     * <p>
+     * If no tasks are currently missing, then this should return null. Otherwise, it should return an {@link
```
gabrielhartmann: Comment is out of date.  Returns Optional now, not `null`.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
dgrnbrg: Given that JSON doesn't have a duration type, I didn't want to add custom serialization/deserialization code, since this object is a json data model. Do you require this changed? If so, do you want me to use java.time or joda.time Durations? And do you mind if I continue to document the duration in the method, since at the usage sites I need to convert to other APIs?

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+
+import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+
+/**
+ * This class provides the core functionality of the automatic repair mechanisms.
+ */
+public class RepairScheduler {
+    private final Log log = LogFactory.getLog(RepairScheduler.class);
```
gabrielhartmann: Should be `getClass()` instead of `RepairScheduler.class`

src/main/java/org/apache/mesos/scheduler/repair/CoreFrameworkState.java
dgrnbrg: I have clarified this.

src/main/java/org/apache/mesos/scheduler/repair/CoreFrameworkState.java
dgrnbrg: I created this interface between there's no interface for accessing and modifying live tasks, only the StateStore. Would you like me to split this into one new interface for removing tasks, and otherwise just extend statestore?

src/main/java/org/apache/mesos/config/RepairConfiguration.java
gabrielhartmann: If serialization/deserialization is barrier, I'd suggest separating the components which are serialized from those consumed for logic.  Components which need to understand the time elements should use Duration, elements which need to serialize/deserialize should use types appropriate to that need.  Documenting in method signatures would not I think be needed for those methods which return `Duration` types.

We are already using java.time Duration here, so probalby stick with that.
https://github.com/mesosphere/dcos-commons/blob/ad7dbe51a10a768b0e05b827fa30ce42b026ec4b/src/main/java/org/apache/mesos/executor/TimedExecutorTask.java#L9

src/main/java/org/apache/mesos/scheduler/repair/CoreFrameworkState.java
gabrielhartmann: I don't understand what you mean by live Tasks.  The StateStore references the current state of Tasks.  Everything in this method is a getter except for deleteTask which I don't think should be here.

dgrnbrg: The RepairScheduler needs to do something with stopped tasks. The way I implemented it is that stopped tasks are removed from the StateStore, which enables them to be replaced. Would you like me to use the `clearTask()` method here? It would have the same effect, except it may result in untracked orphaned tasks during partitions and potential races due to using the name rather than the ID. If you're okay with those shortcomings, I will change the implementation.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
dgrnbrg: Serdes is not a barrier--it's a style question. I would recommend using integral units, since by forcing the user to select units of a particular size, this helps to mentally anchor what appropriate choices for these values may be. 

I would strongly recommend against removing the units from the names of the properties, since those names are exposed to the YAML config file, which is totally untyped. Without the units in the name, you'd have to read the code or docs to understand the config file, rather than having it be self-documenting. Because of this, I don't think that Durations offer an advantage over `int`s.

I would, however, agree that Durations are a fine choice for programmatic APIs, where it's easy to construct whichever Duration you want. It's just that this API is interacted with via untyped YAML, so it's meant to be easier to use from there.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+        result = 31 * result + (enable_replacement ? 1 : 0);
+        return result;
+    }
+
+    @JsonProperty("repair_in_place_grace_period_mins")
+    private int gracePeriodMins;
+    @JsonProperty("min_delay_between_repairs_secs")
+    private int repairDelaySecs;
+    @JsonProperty("enable_replacement")
+    private boolean enable_replacement;
```
gabrielhartmann: variable should be camelCase

src/main/java/org/apache/mesos/scheduler/repair/RepairResource.java
```diff
@@ -0,0 +1,37 @@
...
+        this.repairStatus = repairStatus;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public RepairStatus repair() {
+        try {
+            return repairStatus.get();
+        } catch (Exception ex) {
+            log.error("Failed to fetch data with exception: " + ex);
```
gabrielhartmann: Please use:
`log.error("Failed to fetch data with exception: ", ex);`
which provides the full stack trace in the error message instead of just the Exception name.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+        if (oldStatus == null) {
+            oldStatus = new RepairStatus(Collections.emptyList(), Collections.emptyList());
+        }
+        List<String> oldStopped = oldStatus.getStopped();
+        List<String> newStopped = new ArrayList<>(terminatedTasks.stream()
+                .filter(failureMonitor::hasFailed)
+                .map(TaskInfo::getName)
+                .collect(Collectors.toList()));
+        newStopped.addAll(oldStopped);
+        //TODO (person) possibly write this list down in ZK, to ensure the UI doesn't change upon restart
```
gabrielhartmann: Please don't leave TODO's in code unassigned.  Feel free to assign to me.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+                log.info("Relaunching terminated task in place");
+                offerReq = offerReqProvider.getReplacementOfferRequirement(terminatedTask);
+            }
+        } else {
+            log.info("Launching a missing task");
+            replacingMissing = true;
+            offerReq = offerReqProvider.maybeGetNewOfferRequirement(targetConfigName, block).orElse(null);
+        }
+
+        if (offerReq != null && launchConstrainer.canLaunch(offerReq)) {
```
gabrielhartmann: `offerReq` should be Optional<OfferRequirement> or OfferRequirement.  Don't check `null`

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+        newStopped = newStopped.stream().distinct().collect(Collectors.toList());
+        List<String> newFailed = terminatedTasks.stream()
+                .filter(it -> !failureMonitor.hasFailed(it))
+                .map(TaskInfo::getName)
+                .collect(Collectors.toList());
+
+        OfferRequirement offerReq = null;
+        boolean replacingMissing = false;
+
+        if (terminatedTasks.size() > 0) {
```
gabrielhartmann: Always preferring merely stopped elements to elements which have been marked for destructive replacement is not necessarily always the right policy.  This would block replacement of permanently failed tasks due to a single flapping node which would in some cases cause unavailability.  The purpose of the system is to attempt to avoid unavailability.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,163 @@
...
+                            .filter(it -> !relaunchedTasks.contains(it))
+                            .collect(Collectors.toList());
+                }
+            }
+            List<Operation> launchOperations = recommendations.stream()
+                    .map(OfferRecommendation::getOperation)
+                    .filter(Operation::hasLaunch)
+                    .collect(Collectors.toList());
+            if (launchOperations.size() > 1) {
+                throw new RuntimeException("Tardigrade repairs should only launch one task at a time, found "
```
gabrielhartmann: Please remove the `Tardigrade` code name.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
dgrnbrg: One thing we discussed at the project kickoff was the challenges of composing scheduler logic together. That is one of the requirements that set us down the txnplan path. Once you decided that the simpler approach was necessary, we reverted back to this design, as we had discussed. I brought up my concern with this when I mentioned the need to thoughtfully created scheduler combinators, to express different sorts of logic.

If you have an alternative design you'd prefer for this logic, if it's simple I should be able to do it by the end of the week; otherwise, I can come up with a timeline for you.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,163 @@
...
+            log.info("Found terminated task");
+            if (failureMonitor.hasFailed(terminatedTask)) {
+                log.info("Purging terminated task");
+                failureListener.taskFailed(terminatedTask.getTaskId());
+            } else {
+                log.info("Relaunching terminated task in place");
+                offerReq = Optional.of(offerReqProvider.getReplacementOfferRequirement(terminatedTask));
+            }
+        } else {
+            log.info("Launching a missing task");
```
gabrielhartmann: The concept of a missing Task has leaked out of the Kafka framework.  It is an implementation detail of the Kafka framework.  The work that needs to be done here is replacing permanently failed Tasks and restarting transiently failed Tasks.  "Missing" doesn't enter into the equation.

dgrnbrg: The design of the system, as we've discussed several times, is that when a task is permanently stopped, we restart it. In other worsd, missing = stopped. I'll change the log message here for clarity.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,163 @@
...
+                            .filter(it -> !relaunchedTasks.contains(it))
+                            .collect(Collectors.toList());
+                }
+            }
+            List<Operation> launchOperations = recommendations.stream()
+                    .map(OfferRecommendation::getOperation)
+                    .filter(Operation::hasLaunch)
+                    .collect(Collectors.toList());
+            if (launchOperations.size() > 1) {
+                throw new RuntimeException("Repairs should only launch one task at a time, found "
```
gabrielhartmann: A `RuntimeException` seems harsh here.  What is the expectation?  Would you like the Scheduler to crash on this unexpected scenario?  I would prefer to raise an Exception which requires declaration in the method signature such that consumers are forced to decide what do when encountering a Recovery system failure.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,163 @@
...
+                        + launchOperations.size() + " tasks");
+            }
+            acceptedOffers = offerAccepter.accept(driver, recommendations);
+            if (launchOperations.size() == 1) {
+                // We could've launched nothing if the offer didn't fit; only make if it happened
```
gabrielhartmann: I don't understand the comment. `only make if it happened`

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,163 @@
...
+        newStopped = newStopped.stream().distinct().collect(Collectors.toList());
+        List<String> newFailed = terminatedTasks.stream()
+                .filter(it -> !failureMonitor.hasFailed(it))
+                .map(TaskInfo::getName)
+                .collect(Collectors.toList());
+
+        Optional<OfferRequirement> offerReq = Optional.empty();
+        boolean replacingStopped = false;
+
+        if (terminatedTasks.size() > 0) {
```
gabrielhartmann: Logging terminology in this `if / else` block is not consistent with the terms used in `RepairStatus`.  Use of the terms `stopped` and `failed` would be clearer.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
gabrielhartmann: Are you sure @JsonProperty annotation is unnecessary here?  I've been bitten in the past with being able to serialize, but not deserialize appropriately.  Do you have tests exercising serialization/deserialization?

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
dgrnbrg: I can switch to checked exceptions, but they haven't been popular for usability reasons for many years. I typically throw RuntimeExceptions when I would like to use an `assert`, but of course `assert`s can be disabled, and this is a critical check.

The point here is to help developers catch incorrect implementations of the provider at dev time. I don't think that requiring try/catch around callsites of this method will help, since this is meant to detect bad implementations early on in the development process.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
dgrnbrg: I deleted what comes after the semicolon; it was meant to emphasize we only make the invocation to `launchHappened` if the launch actually happened, rather than nearly happened (since we unconditionally accept recommendations.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
gabrielhartmann: Yet another reason I don't like YAML for config.  Oh well.  If it's too hard to have both config in YAML and have classes that reasonable signatures for internal consumption, we can stay with the current approach.  To be clear, my proposal was to essentially leave what you have here alone, and add something which provides a better interface for internal consumption, and translate between the two as necessary.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
gabrielhartmann: How would it be caught by developers during the development process?  Unit tests?  Regardless, in a production setting, I'd like developers to decide what they want to do when an error occurs in the Recovery system.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
dgrnbrg: Jackson automatically handles these. I just immitated your style for consistency. Here's a reference this: https://github.com/FasterXML/jackson-annotations#changing-property-auto-detection

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
dgrnbrg: Assuming that developers test what they write at any point, they'll see this issue if they incorrectly attempt to launch multiple tasks during a recovery operation. This particular error should never occur in production, unless the developer literally never tested any of the recovery logic.

If you still want this to be a checked exception, would you like me to make a new exception class, or use one of the existing ones?

src/main/java/org/apache/mesos/scheduler/repair/CoreFrameworkState.java
gabrielhartmann: Are we talking about `stopped` or `failed` tasks here.  I think you mean `failed`.  Working under that assumption, removing persistence of information about a Task is not categorically the right approach for permanent Task failure recovery.  A call against an Interface should be made.  In some cases the implementation of that interface may (as in Kafka's case) determine that removal of information from the StateStore is a valid approach.  In `dcos-commons` however, no assumption may be made regarding how a Service intends to handle the determination that a Task has permanently failed.  Removal of this persistent information will certainly be the wrong thing to do for some services.

src/main/java/org/apache/mesos/scheduler/repair/constrain/TestingLaunchConstrainer.java
```diff
@@ -0,0 +1,31 @@
+package org.apache.mesos.scheduler.repair.constrain;
```
gabrielhartmann: This file should be in the `test` not `main` src tree.

src/main/java/org/apache/mesos/scheduler/repair/constrain/TestingLaunchConstrainer.java
```diff
@@ -0,0 +1,31 @@
...
+    public TestingLaunchConstrainer() {
+        this.canLaunch = false;
+    }
+
+    public void setCanLaunch(boolean canLaunch) {
+        this.canLaunch = canLaunch;
+    }
+
+    @Override
+    public void launchHappened(Operation launchOperation) {
```
gabrielhartmann: Comment needed regarding empty implementation.

src/main/java/org/apache/mesos/scheduler/repair/CoreFrameworkState.java
dgrnbrg: This interface no longer exists. Please check out the changes to StateStore and TaskFailureListener instead, for the new approach I tried to use.

src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param minDelaySeconds Minimum number of seconds between each launch
+     */
+    public TimedLaunchConstrainer(long minDelaySeconds) {
+        this.minDelaySeconds = minDelaySeconds;
+        this.lastLaunch = new AtomicLong(0);
+    }
+
+    @Override
+    public void launchHappened(Operation launchOperation) {
+        if (!canLaunch(null)) {
```
gabrielhartmann: I don't get this call of `canLaunch` with `null`.  Can you clarify?

src/main/java/org/apache/mesos/scheduler/repair/constrain/TestingLaunchConstrainer.java
dgrnbrg: I'd be happy to move this; however, the 2 testing classes (for Constrainer and Monitor) are designed to make it easy to write tests for RepairScheduler integration. If these are moved to the `test` tree, the won't be included in the library, and so they couldn't be used for testing, which defeats the point. If you'd like, I can delete these helpers for testing consumers of this library.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
gabrielhartmann: It's very easy to write test scenarios that might only ever generate a single Launch Operation, but avoid scenarios that produce multiple.  I would like developers to consciously decide what they'd like to do recovery system failures.  Some relevant checked exception, new or existing would be good.

src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
dgrnbrg: Previously, you told me that `launchHappened` should take an `Operation` and not an `OfferRequirement`. I mentioned this broke some symmetries.

There's no way for me to create an `OfferRequirement` here, but this implementation's `canLaunch` doesn't interact with its argument.

I wanted to add this validation here, so I call `canLaunch(null)`.

src/main/java/org/apache/mesos/scheduler/repair/constrain/TestingLaunchConstrainer.java
gabrielhartmann: No, that makes sense.  I get it.  Keep it here.

src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
gabrielhartmann: Why is launchHappened being called, if it was already determined that canLaunch is false?  And again, I don't think a RuntimeException is the best idea here.  This seems like a strange validation.  What happens if we set `lastLaunch` again in the case where `canLaunch` returned false?

src/main/java/org/apache/mesos/config/RepairConfiguration.java
gabrielhartmann: Does that mean there aren't tests exercising serialization and deserialization?

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
```diff
@@ -0,0 +1,163 @@
...
+            TaskInfo terminatedTask = terminatedTasks.get(new Random().nextInt(terminatedTasks.size()));
+            log.info("Found failed task");
```
gabrielhartmann: Do you mean `stopped` task here?  Can you validate terminology in here for clarity.

src/main/java/org/apache/mesos/scheduler/repair/TaskFailureListener.java
```diff
@@ -0,0 +1,17 @@
...
+import org.apache.mesos.Protos.TaskID;
+
+/**
+ * Exposes simple task-oriented API for components that need to interact with the framework as a whole.
+ */
+public interface TaskFailureListener {
+    /**
+     * Remove the given task from this framework. This ensures that the given task is not running, possibly by killing
+     * it, and it will remove the task from the framework's state tracking. The absence of the task will be fixed by the
+     * {@link RepairScheduler}, if that task is still supposed to be running.
```
gabrielhartmann: Do these comments still make sense?  Isn't a call back here an indication that the Task has is already not running?  Why any mention of removal from the Framework's state tracking?

src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
dgrnbrg: So, I think you're thinking through the wrong case. The issue is if `canLaunch` returns **true**, and then `lastLaunch` is updated. In this case, we can observe a launch that shouldn't have happened. The issue is that we don't have any atomicity guarantee between these 2 methods. I try to avoid creating implicit threading demands on APIs, so that future concurrent code won't randomly break.

The issue here is that multiple threads can be making the `canLaunch` -> `launchHappened` paired calls. There are interleavings that would invalidate this logic, and I believe that when we can detect a race condition, we should throw something.

There are 3 solutions:

1. The solution I implemented, which detects the bug without preventing future concurrency.
1. We could put a lock around the repairscheduler, which would prevent the bug but also prevent future concurrency.
1. I can remove this check, and the behavior will remain the same, although then the bug will simply become a silent issue.

I think the solution I chose is best because it detects the issue w/o reducing the future design space.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
dgrnbrg: Exactly. We're using Jackson, one of the most popular and tested libraries in the java ecosystem.

Is there a specific risk you're concerned about?

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
gabrielhartmann: Offload to an implementation of an Interface.  The interface receives some indication of the transiently and permanently failed tasks and provides a `List<OfferRequirement>` to be fulfilled.  An implementation could be to always prefer transiently failed tasks over permanent, or vice versa.

An initial simple implementation in the RecoveryScheduler could be to attempt to satisfy every OfferRequirement in order until one is satisfied.  Later optimizations can be made to satisfy more OfferRequirements simultaneously.

dgrnbrg: I'm sorry, this is out of scope at this point. We've been collaborating on this design for two weeks. I showed you this API on July 21st in [this commit](https://github.com/mesosphere/dcos-kafka-service/pull/218/commits/6827267f4872d3601f74ca628ec9b9c04bd4a8e5), and you didn't comment on it then.

Changing this would require redesigning the control flow of the RepairScheduler and the API for the RepairOfferRequirementProvider, which will change the delivery timeline.

src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
gabrielhartmann: What is a launch "that shouldn't have happened"?  If a launch has occurred, then the timer should be reset.  If it "should not have happened" it still happened.

What are the possible logic invalidations?  Timer is reset on any launch.  `canLaunch` has logic based on that new time.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+    @JsonProperty("enableReplacement")
+    private boolean enableReplacement;
+
+    public  RepairConfiguration() {}
+
+    @JsonCreator
+    public RepairConfiguration(
+            @JsonProperty("repair_in_place_grace_period_mins") int gracePeriodMins,
+            @JsonProperty("min_delay_between_repairs_secs") int repairDelaySecs,
+            @JsonProperty("enableReplacement") boolean enableReplacement) {
```
gabrielhartmann: JSON should be snake case

src/main/java/org/apache/mesos/config/RepairConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+        this.gracePeriodMins = gracePeriodMins;
+        this.repairDelaySecs = repairDelaySecs;
+        this.enableReplacement = enableReplacement;
+    }
+
+    public boolean isReplacementEnabled() {
+        return enableReplacement;
+    }
+
+    @JsonProperty("enableReplacement")
```
gabrielhartmann: JSON should be snake case

src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
dgrnbrg: I need you to explicitly confirm that you want me to choose the third option: turning a potential bug that warns you that the constraints have been broken (and thus we're in a potentially catastrophic state) into a silent bug that will be very hard to discover.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+    public void setEnableReplacement(boolean enable_replacement) {
```
gabrielhartmann: java variables should be camelCase

src/main/java/org/apache/mesos/config/RepairConfiguration.java
gabrielhartmann: I'm concerned that serializing this configuration and storing it to ZK for example, then attempting to retrieve and deserialize won't work.  A test would be good.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
dgrnbrg: Did you know that all of this could be done automatically? [See here](https://fasterxml.github.io/jackson-databind/javadoc/2.2.0/com/fasterxml/jackson/databind/PropertyNamingStrategy.html#CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES)

I'll continue with the present style.

src/main/java/org/apache/mesos/config/RepairConfiguration.java
dgrnbrg: We do not need to test that Jackson works.

gabrielhartmann: It's not Jackson that may not work.  It's the improper use of annotations which may not work.

gabrielhartmann:
Please update the build.gradle version-SNAPSHOT to a number larger than what is currently in master.

src/main/java/org/apache/mesos/scheduler/repair/constrain/TimedLaunchConstrainer.java
gabrielhartmann: Before I choose one of the possible options, I'd like to understand more thoroughly the problem we're attempting to avoid.

Could you please provide a concrete example and/or a more detailed explanation of the failure scenario that is being addressed here?

dgrnbrg: Suppose that the repair scheduler gains concurrency, which as I recall is an area that's being considered in the future. At the moment, the following sequence could happen:

```
Thread 1                  Thread 2
canLaunch()
                          canLaunch()
launchHappened()
                          launchHappened() <---should not have happened
```

gabrielhartmann: In the example you provide, have two LAUNCH Operations already been issued to Mesos?

dgrnbrg: Yes they have. Because the safety constraint was violated, I think it's necessary to complain as loudly as possible, as a violated safety constraint could be catastrophic.

So, given this, what do you prefer?

gabrielhartmann: I like the idea of complaining loudly, but I'd prefer to avoid doing the wrong thing in the first place, so given the options you suggest, I'd go with 2.  That is, I'd disallow concurrency as long as it's correctness isn't enforced.  When it can be enforced, we should allow it.  Thanks for the explanation.

dgrnbrg: I've blocked concurrency in the most common case to try to further mitigate the issue, as well as a comment explaining why.

dgrnbrg:
This is now squashed, updated, and ready to go.

src/main/java/org/apache/mesos/scheduler/repair/RepairScheduler.java
gabrielhartmann: Got it.  Increase in scope is undesirable this close to the deadline.

dgrnbrg:
Thank you!

PR #79, opened 2016-08-04
Allow unverified TLS certs, reuse err handling for 'plan show'

nickbp:
New EE systems have custom-generated TLS certs

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -49,27 +49,31 @@ func HandleCommonArgs(app *kingpin.Application, defaultServiceName string, short
 // Standard Arguments
 
 func HandleCommonFlags(app *kingpin.Application, defaultServiceName string, shortDescription string) {
-	// -h (in addition to --help): standard help
-	app.HelpFlag.Short('h')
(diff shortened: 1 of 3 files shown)


gabrielhartmann:
@nickbp:  I'm not sure about this.  Isn't the "right" thing to do here to get good certs or something?  I'd like to hear what the plan is because the whole experience is a bit broken now in browsers and now in the CLI.  I understand the desire to unblock, but I'd like to understand more clearly what's going on first, ok?

nickbp:
Good point, I was feeling a little iffy about this as well. I think it'd make sense to do one of the following:
- fetch the CA cert from the cluster somehow? (but then what if THAT is forged...) this depends on what options we'd have available
- fallback: only allow the unverified certs if the user manually enables a flag saying that they consent to it
Once we hear back from them we should know which option is feasible. In the meantime I'll mark myself assignee for this PR

nickbp:
Added (off-by-default) flag to allow unvalidated certs, and mostly-complete support for automatically grabbing the cluster cert from the CLI (should just need to update the name of the config value once that's known)

PR #82, opened 2016-08-08
Fix Tardigrade

gabrielhartmann:
1. First implementation conflates presence/absence of Task metadata with transient/permanent
failure implementation details.
2. Stopped and Failed pools should be a view on StateStore and not
effected directly by the RepairScheduler.
3. Provide better non-name based failure and stopped pools
implementation, for appropriately abstracted recovery decision making.
4. Fix tests which were affected by 1).
5. Fix failedTaskCanBeRestarted test which should have failed.
Requirement that only 1 Operation be performed in the
DefaultRepairScheduler is a mistake, as it may entail producing a
totally new Task with (RESERVE etc. operations).
6. Standardize on "Recovery" not "Repair" as the term for recovering from transient and permanent faults.


src/main/java/org/apache/mesos/config/RecoveryConfiguration.java
```diff
@@ -43,10 +43,10 @@ public int hashCode() {
...
 
-    public  RepairConfiguration() {}
+    public RecoveryConfiguration() {}
 
     @JsonCreator
-    public RepairConfiguration(
+    public RecoveryConfiguration(
             @JsonProperty("repair_in_place_grace_period_mins") int gracePeriodMins,
             @JsonProperty("min_delay_between_repairs_secs") int repairDelaySecs,
             @JsonProperty("enableReplacement") boolean enableReplacement) {
```
nickbp: snake case

gabrielhartmann: nice catch.  Thanks.

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryScheduler.java
```diff
@@ -71,85 +69,36 @@ public RepairScheduler(
...
-                log.info("Relaunching stopped task in place");
-                offerReq = Optional.of(offerReqProvider.getReplacementOfferRequirement(terminatedTask));
-            }
-        } else {
-            log.info("Launching a failed task");
-            replacingFailed = true;
-            offerReq = offerReqProvider.maybeGetNewOfferRequirement(targetConfigName, block);
+        Optional<OfferRequirement> offerReq = Optional.empty();
+        if (candidateRepairs.size() > 0) {
+            offerReq = Optional.of(candidateRepairs.get(new Random().nextInt(candidateRepairs.size())));
```
gabrielhartmann: Will do.  The idea is that we're giving equal priority to both transiently and permanently failed tasks such that a flapping transient failure at the very least doesn't block recovery of permanently failed tasks.

src/main/java/org/apache/mesos/scheduler/recovery/RecoveryOfferRequirementProvider.java
```diff
@@ -0,0 +1,15 @@
+package org.apache.mesos.scheduler.recovery;
+
+import org.apache.mesos.Protos.TaskInfo;
+import org.apache.mesos.offer.OfferRequirement;
+
+import java.util.List;
+
+/**
+ * Implementations of this interface provide OfferRequirement objects when presented with Tasks which have failed
+ * permanently or transiently.
```
nickbp: What are the OfferRequirements intended to be used for? to re-launch the failed tasks? (cant hurt to list/specify what the expected responses are here)

nickbp: Looks like a lot of this could be copied from RepairOfferRequirementProvider

nickbp:
Just the above nits, otherwise LGTM

gabrielhartmann:
Thanks @nickbp.  Merging.

PR #83, opened 2016-08-09
Minor Improvements.

joerg84:
1. Added Null check (instead of size > 0)
2. Generalized supportNamedVips check.

### DIFF
--- a/src/main/java/org/apache/mesos/offer/OfferAccepter.java
+++ b/src/main/java/org/apache/mesos/offer/OfferAccepter.java
@@ -3,6 +3,7 @@ package org.apache.mesos.offer;
 import com.google.inject.Inject;
 import com.google.protobuf.TextFormat;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.mesos.Protos.Filters;
 import org.apache.mesos.Protos.Offer.Operation;
 import org.apache.mesos.Protos.OfferID;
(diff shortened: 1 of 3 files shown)


joerg84:
@nickbp Could you take a look, nothing urgent, TY!

src/main/java/org/apache/mesos/offer/OfferAccepter.java
```diff
@@ -45,7 +46,7 @@ public OfferAccepter(Set<OperationRecorder> recorders) {
   }
 
   public List<OfferID> accept(SchedulerDriver driver, List<OfferRecommendation> recommendations, Filters filters) {
-    if (recommendations.size() <= 0) {
+    if (CollectionUtils.isEmpty(recommendations)) {
```
nickbp: any reason to not just use recommendations.isEmpty()? (same with operations.isEmpty() below)

src/test/java/org/apache/mesos/dcos/CapabilitiesTest.java
```diff
@@ -30,10 +30,18 @@ public void testHasNamedVipsFails() throws URISyntaxException, IOException {
...
 
     @Test
-    public void testHasNamedVipsSucceeds() throws URISyntaxException, IOException {
+    public void testHasNamedVipsSucceeds_1_8() throws URISyntaxException, IOException {
         mockDcosCluster =  MockDcosCluster.create("1.8.0");
         dcosCluster = mockDcosCluster.getDcosCluster();
         Capabilities capabilities = new Capabilities(dcosCluster);
         Assert.assertTrue(capabilities.supportsNamedVips());
     }
+
```
nickbp: Couldn't hurt to add a 2.0 test to exercise that check

joerg84: Sure.

src/main/java/org/apache/mesos/offer/OfferAccepter.java
joerg84: CollectionUtils.isEmpty includes a `null` check. If you feel that is not needed here, happy to switch to recommendations.empty().

PR #87, opened 2016-08-11
Move serialized object members to snake_case

mrbrowning:
JSON properties should always be in snake case for consistency. Only the Acme scheduler example is affected here.


gabrielhartmann:
@mrbrowning: Broken CI build?  @mesosphere-ci: retest this please

mrbrowning:
@gabrielhartmann builds are passing now.

PR #88, opened 2016-08-11
Supply Reconciler with TaskStatusProvider in constructor and remove from arguments in start method

mrbrowning:
To ensure that task statuses are all coming from a single source of truth, this alters the `Reconciler` interface so that the `start()` method no longer accepts a list of `TaskStatus` protos. Instead, it is expected that the implementing class will accept a `TaskStatusProvider` in its constructor, which it will consult to fetch a collection of all task statuses currently known by the scheduler. To this end, the `StateStore` interface now extends the `TaskStatusProvider` interface with its `getTaskStatuses()` method. All default implementations of these interfaces now reflect the new semantics.


gabrielhartmann:
@mrbrowning:  Tests failing?  Also this PR will need to bump `build.gradle` version.

mrbrowning:
@gabrielhartmann Tests are passing.

gabrielhartmann:
@mrbrowning:  Unit test failing.

gabrielhartmann:
@mrbrowning:  I this for INFINITY-321 ?

gabrielhartmann:

PR #89, opened 2016-08-12
Fix version check for case of "1.8-dev"

nickbp:
Stays pretty strict for now, supporting only '-dev', and only on the second element.

Moves version element extraction upstream in case other utils need it someday.

### DIFF
--- a/src/main/java/org/apache/mesos/dcos/DcosVersion.java
+++ b/src/main/java/org/apache/mesos/dcos/DcosVersion.java
@@ -25,6 +25,7 @@ public class DcosVersion {
     private static final String BOOTSTRAP_ID_KEY = "bootstrap-id";
     private static final String DCOS_IMAGE_COMMIT = "dcos-image-commit";
     private static final String VERSION_KEY = "version";
+    private static final String DEV_VERSION_SUFFIX = "-dev";
 
     private final String bootstrapId;
     private final String dcosImageCommit;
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/mesos/dcos/DcosVersion.java
```diff
@@ -53,4 +54,26 @@ public String getDcosImageCommit() {
     public String getVersion() {
         return version;
     }
+
+    public int getVersionFirstElement() throws NumberFormatException {
```
gabrielhartmann: Can we encapsulate Version in a Class?  So we get something like:
```
public Version getVersion();
Version version = getVersion();
int firstElement = version.getFirstElement();
int secondElement = version.getSecondElement();
...
```

nickbp: Done

src/test/java/org/apache/mesos/dcos/CapabilitiesTest.java
```diff
@@ -12,7 +12,6 @@
  */
 public class CapabilitiesTest {
```
gabrielhartmann: @nickbp:  Thanks for all the tests.  Can you add one more that exercises the throwing of the NumberFormatException please?  So something that has version returned that looks like `1` for example and maybe something with `0.` as another example.

nickbp: Done

PR #92, opened 2016-08-15
Fix recovery throttling defect.

gabrielhartmann:
The RecoveryScheduler was throttling recovery operations regardless of
whether they were in reaction to a permanent or transient failure.  Only
destructive permanent recovery operations should be throttled.

### DIFF
--- a/src/main/java/org/apache/mesos/scheduler/recovery/RecoveryRequirement.java
+++ b/src/main/java/org/apache/mesos/scheduler/recovery/RecoveryRequirement.java
@@ -1,0 +1,24 @@
+package org.apache.mesos.scheduler.recovery;
+
+import org.apache.mesos.offer.OfferRequirement;
+
+/**
+ * Implementation of this interface allows RecoverySchedulers to make decisions about the scheduling of recovery of
+ * failed Tasks.
+ */
+public interface RecoveryRequirement {
+    /**
+     * Possible recovery types are NONE, TRANSIENT and PERMANENT.  NONE indicates that no recovery is under way.
+     * TRANSIENT indicates that non-destructive recovery is being attempted.  PERMANENT indicates that a destructive
+     * recovery operation is desired.
+     */
+    enum RecoveryType {
+        NONE,
+        TRANSIENT,
+        PERMANENT
+    }
+
+    RecoveryType getRecoveryType();
+
+    OfferRequirement getOfferRequirement();
+}
--- a/src/main/java/org/apache/mesos/scheduler/recovery/constrain/TimedLaunchConstrainer.java
+++ b/src/main/java/org/apache/mesos/scheduler/recovery/constrain/TimedLaunchConstrainer.java
@@ -1,8 +1,10 @@
 package org.apache.mesos.scheduler.recovery.constrain;
 
 import org.apache.mesos.Protos.Offer.Operation;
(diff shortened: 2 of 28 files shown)


src/main/java/org/apache/mesos/scheduler/recovery/constrain/TimedLaunchConstrainer.java
```diff
@@ -10,29 +10,33 @@
  * rate-limiting purposes.
  */
 public class TimedLaunchConstrainer implements LaunchConstrainer {
-    private AtomicLong lastLaunch;
+    private AtomicLong lastPermanentRecoveryLaunch;
```
nickbp: It'd make sense to mark this with the unit, ie "lastPermanentRecoveryLaunchMillis" or "lastPermanentRecoveryLaunchMs", to avoid mixup with minDelaySeconds.

For that matter it'd probably make sense to store minDelaySeconds as milliseconds, so that the units for comparison are consistent throughout.

src/main/java/org/apache/mesos/scheduler/recovery/RecoveryRequirement.java
```diff
@@ -0,0 +1,24 @@
...
+ * failed Tasks.
+ */
+public interface RecoveryRequirement {
+    /**
+     * Possible recovery types are None, Transient and Permanent.  None indicates that no recovery is under way.
+     * Transient indicates that non-destructive recovery is being attempted.  Permanent indicates that a destructive
+     * recovery operation is desired.
+     */
+    enum RecoveryType {
+        None,
```
nickbp: UNDERSCORE_CAPS for enums?

(or if the rest of commons is CamelCase then stick with consistency imo)

src/main/java/org/apache/mesos/scheduler/recovery/constrain/TimedLaunchConstrainer.java
```diff
@@ -10,29 +10,33 @@
...
+            lastPermanentRecoveryLaunch.compareAndSet(lastPermanentRecoveryLaunch.get(), System.currentTimeMillis());
         }
-        lastLaunch.compareAndSet(lastLaunch.get(), System.currentTimeMillis());
     }
 
     @Override
-    public boolean canLaunch(OfferRequirement offerRequirement) {
-        return (lastLaunch.get() + minDelaySeconds * 1000) < System.currentTimeMillis();
+    public boolean canLaunch(RecoveryRequirement recoveryRequirement) {
+        if (recoveryRequirement.getRecoveryType().equals(RecoveryRequirement.RecoveryType.Permanent)) {
```
nickbp: Could add static utility methods to RecoveryRequirement of this form:
- public static boolean isPermanent(RecoveryRequirement)
- public static booleas isTemporary(RecoveryRequirement)

Similar to (but the opposite of) Block.getStatus(): https://github.com/mesosphere/dcos-commons/blob/3581e7b5697d9789c1162200398637186df82423/src/main/java/org/apache/mesos/scheduler/plan/Block.java

gabrielhartmann:

gabrielhartmann:
@nickbp: Comments addressed.  Standardized enums everywhere as they were inconsistent.

gabrielhartmann:
@nickbp:  Standardized enum names.  Comments addressed.

PR #96, opened 2016-08-20
Health checks

gabrielhartmann:


src/main/java/org/apache/mesos/executor/CustomExecutor.java
```diff
@@ -7,18 +7,21 @@
...
+import java.util.Optional;
+import java.util.concurrent.*;
 
 /**
  * An {@code Executor} implementation that supports execution of long-running tasks and supporting short-lived tasks.
  */
 public class CustomExecutor implements Executor {
     private static final Logger LOGGER = LoggerFactory.getLogger(
             CustomExecutor.class);
+    private static final int HEALTH_CHECK_THREAD_POOL_SIZE = 50;
```
nickbp: seems like a lot?

gabrielhartmann:
@nickbp @mohitsoni: ping!

src/main/java/org/apache/mesos/executor/CustomExecutor.java
gabrielhartmann: Yeah, I didn't want to run out of threads... what do you think 5?  There's really only 1 check running at a time normally.  Although it depends on the interval/timeout/delay settings.

gabrielhartmann:
@nickbp @mohitsoni:  Any other comments?

src/main/java/org/apache/mesos/executor/CustomExecutor.java
```diff
@@ -111,6 +115,45 @@ public void launchTask(final ExecutorDriver driver, final Protos.TaskInfo task)
...
+                            HealthCheckHandler.create(
+                                    taskInfo,
+                                    scheduledExecutorService,
+                                    new HealthCheckStats(taskInfo.getName())),
+                            executorTask);
+            LOGGER.info("Submitting health check monitor.");
+            Future<Optional<HealthCheckStats>> futureOptionalHealthCheckStats =
+                    executorService.submit(healthCheckMonitor);
+
+            executorService.submit(new Runnable() {
```
nickbp: could this logging wrapper be placed inside of HealthCheckMonitor? it's not strictly operating within the monitor process itself but it's pretty closely tied to its output

src/main/java/org/apache/mesos/executor/HealthCheckHandler.java
```diff
@@ -0,0 +1,177 @@
...
+                TimeUnit.MILLISECONDS);
+    }
+
+    private void validateTask(Protos.TaskInfo taskInfo) throws HealthCheckValidationException {
+        // Validate TaskInfo
+        if (!taskInfo.hasHealthCheck()) {
+            throw new HealthCheckValidationException("The following task does not contain a HealthCheck: " + taskInfo);
+        }
+
+        // Validate HealthCheck
```
nickbp: maybe add a comment mentioning that this healthcheck configuration is only viewed by us in practice.
per conversation, its never visible to marathon and mesos doesn't support healthchecks on custom executors

src/main/java/org/apache/mesos/executor/CustomExecutor.java
```diff
@@ -111,6 +115,45 @@ public void launchTask(final ExecutorDriver driver, final Protos.TaskInfo task)
         }
     }
 
+    private void executeHealthCheck(Protos.TaskInfo taskInfo, ExecutorTask executorTask) {
```
nickbp: maybe call this 'scheduleHealthCheck'? looks like it's scheduling a regular retriggering of the check, whereas 'execute' implies it's only being run once

src/main/java/org/apache/mesos/executor/HealthCheckHandler.java
```diff
@@ -0,0 +1,177 @@
...
+            Protos.CommandInfo commandInfo = healthCheck.getCommand();
+
+            try {
+                final Map<String, String> envMap = TaskUtils.fromEnvironmentToMap(commandInfo.getEnvironment());
+                List<String> command = new ArrayList<>();
+                command.add("/bin/bash");
+                command.add("-c");
+                command.add(commandInfo.getValue());
+
+                String[] cmdArray = new String[command.size()];
```
mohitsoni: This is not required, and can be replaced with:
String[] cmdArray = command.toArray(cmdArray);

Not doing so, will cause us to perform String[] allocation, which is immediately discarded and is made available for GC.

src/main/java/org/apache/mesos/executor/HealthCheckHandler.java
```diff
@@ -0,0 +1,177 @@
...
+                cmdArray = command.toArray(cmdArray);
+                ProcessBuilder processBuilder = new ProcessBuilder(cmdArray);
+                processBuilder.inheritIO();
+                processBuilder.environment().putAll(envMap);
+
+                logger.info("Starting health check process: " + command);
+                Process process = processBuilder.start();
+                synchronized (process) {
```
mohitsoni: Why put a synchronized here ? How is the object referenced by `process` shared across more than 1 thread ? Or is there any other non-obvious reason ?

src/main/java/org/apache/mesos/executor/CustomExecutor.java
gabrielhartmann: I moved the "exited without statistics" one.  Actually the thread here doesn't have much to do other than wait on the get() call.  Since for testing it's handing to be able to get() and receive statistics I don't want to remove that.  So since it's getting the stats it may as well log them here.  The Interrupted|ExecutionException case isn't avoidable due to the get().  Anyway, did what I could.

src/main/java/org/apache/mesos/executor/CustomExecutor.java
gabrielhartmann: done

src/main/java/org/apache/mesos/executor/CustomExecutor.java
nickbp: makes sense

src/main/java/org/apache/mesos/executor/HealthCheckHandler.java
gabrielhartmann: done

src/main/java/org/apache/mesos/executor/HealthCheckHandler.java
gabrielhartmann: Doesn't compile. `Error:(134, 53) java: variable cmdArray might not have been initialized`

src/main/java/org/apache/mesos/executor/HealthCheckHandler.java
gabrielhartmann: You're not allowed to call wait on an Object for which you do not hold the Monitor.  If you do you get a `NotHoldingtheObjectMonitorException` or something.

nickbp:
LGTM

nickbp:
retest this please

gabrielhartmann:
@mesosphere-ci: please retest this

gabrielhartmann:
retest this please

gabrielhartmann:
https://jenkins.mesosphere.com/service/jenkins/job/infinity-dcos-commons-pr/402/console

src/main/java/org/apache/mesos/executor/HealthCheckHandler.java
mohitsoni: My bad! I was thinking something else.

PR #99, opened 2016-08-25
Reduce Curator/Zookeeper client creation and logging level

gabrielhartmann:


src/main/java/org/apache/mesos/curator/CuratorConfigStore.java
```diff
@@ -183,6 +182,10 @@ public UUID getTargetConfig() throws ConfigStoreException {
         }
     }
 
+    public void close() {
```
nickbp: Will anyone ever want to call this instead of just destroying the instance?

Maybe make them package-private if they're only for testing

gabrielhartmann: Tried making them package private.  I'll try again.  Tests didn't have access.  You're right though.  Tests were failing because so many curatorpersister's were created that the simultaneous connection limit was hit for ZK.

gabrielhartmann:
@nickbp: Made those methods package private.  Just the CuratorPersister method must remain public as it's in a different package.  I think that makes sense as consumers of the Persister should make the decision about whether to keep it closed or not.  Actually adding it to the interface.

PR #101, opened 2016-08-26
Support DC/OS clusters which have auth disabled

nickbp:
DC/OS may be configured with auth disabled, in which case
dcos_acs_token isn't needed nor provided.


cli/http.go
```diff
@@ -182,10 +182,9 @@ func CreateHTTPRawRequest(method, urlPath, urlQuery, payload, contentType string
 		log.Fatalf("Unable to parse DC/OS Cluster URL '%s': %s", dcosUrl, err)
 	}
 	if len(dcosAuthToken) == 0 {
-		dcosAuthToken = RequiredCLIConfigValue(
-			"core.dcos_acs_token",
-			"DC/OS Authentication Token",
-			"Run 'dcos auth login' to log in to the cluster.")
+		// if the token wasnt manually provided by the user, try to fetch it from the main CLI.
+		// this value is optional: clusters can be configured to not require any auth
+		dcosAuthToken = OptionalCLIConfigValue("core.dcos_acs_token")
```
gabrielhartmann: @nickbp: Can we log a warning here?  Concerned that it will fail now on clusters with authentication enabled in an unclear way.

nickbp: Mistakenly missing this setting (forgot to log in) results in a 401 error from the server. When this happens we emit an error telling the user to log in.

Did a quick cleanup of moving that check alongside the other status code checks.

PR #104, opened 2016-08-29
DefaultScheduler

gabrielhartmann:
The private repository here: https://github.com/mesosphere/sdk-reference-framework, consumes the changes made here and works as described below.  This PR changes code coverage from 79.4% to 78.3%.  https://github.com/mesosphere/dcos-commons/pull/109 pays this debt.

Use https://github.com/mesosphere/dcos-commons/pull/104/files?w=0 to view non-whitespace changes.  

# Goal:
Provide a general implementation of a Scheduler which can deploy, scale, update, and recover Tasks.  One should be able to specify that they would like to deploy a service with 2 Tasks of type Foo, 3 of type Bar.  Then one should be able to change this specification later and have the appropriate scaling (up) operation complete.  So if the `ServiceSpecification` is later changed to indicate that 3 Task of type Foo are desired, then an additional Foo Task should be launched.

If the specification of a Task type changes, then a rolling configuration update should occur to meet this new service specification.  e.g. If Foo Tasks used to consume 1.0 CPU, but now should consume 2.0 CPU, then a rolling restart of all Foo Tasks should occur to meet this requirement.  

Finally, crashed Tasks should be restarted or replaced depending on how long a Task has not been running.  This is the generalized normal behavior of `dcos-kafka-service`, `dcos-cassandra-service`, and `dcos-hdfs-service` today.

The overall flow of control is as follows.  A user should provide a `ServiceSpecification` which should be provided to the DefaultScheduler which should do the rest of the work.  Given a `ServiceSpecification` a `PlanSpecification` should be generated to deploy the Service.  The `PlanSpecification` is then further instantiated into an actual implementation of the `Plan` interface.  This can then be executed using the normal `PlanManager` and `PlanScheduler` interfaces.

# Broad categorization of changes:
* DefaultScheduler with attendant (`ServiceSpecification`, `TaskSpecification`, `ResourceSpecification`, `PlanSpecification`, `PhaseSpecification`, `BlockSpecification`)
* Change of ConfigStore and StateStore interfaces to return Optional objects.
* Rename Stage to Plan

# Notable omissions (because it's already 3000+ lines):
* Handling of Persistent volumes is not yet covered by the DefaultScheduler.  This work should entail a minor extension of the ResourceSpecification interface.
* A web server for serving the API endpoints
* Scaling down of TaskTypes
* Configuration management


src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -137,18 +138,20 @@ public void clearFrameworkId() throws StateStoreException {
     }
 
     @Override
-    public Protos.FrameworkID fetchFrameworkId() throws StateStoreException {
+    public Optional<Protos.FrameworkID> fetchFrameworkId() throws StateStoreException {
```
mohitsoni: :sparkles: `Optional` :sparkles:

build.gradle
```diff
@@ -1,173 +1,175 @@
...
+    compile "org.json:json:${jsonVer}"
+    compile "com.google.inject:guice:${guiceVer}"
+    compile "com.google.inject.extensions:guice-assistedinject:${guiceVer}"
+    compile "org.glassfish.jersey.containers:jersey-container-jetty-http:2.23.2"
+    testCompile "com.github.stefanbirkner:system-rules:${systemRulesVer}"
+    testCompile "io.dropwizard:dropwizard-testing:${dropwizardVer}"
+    testCompile "org.mockito:mockito-all:${mockitoVer}"
+    testCompile "org.powermock:powermock-mockito-release-full:${powerMockVer}"
+    testCompile "org.mock-server:mockserver-netty:${mockServerVer}"
+    testCompile "org.springframework.integration:spring-integration-http:${springVer}"
```
mohitsoni: Unrelated to this PR, but we should remove this dependency on `Spring`. It brings in many transitive dependencies. It's only being use in a Test class: https://github.com/mesosphere/dcos-commons/blob/master/src/test/java/org/apache/mesos/dcos/MockDcosCluster.java#L76

It can be replaced with https://jersey.java.net/apidocs/2.7/jersey/javax/ws/rs/core/MediaType.html#APPLICATION_JSON_TYPE (already a dependency on the project).

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -1,33 +1,40 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * This class encapsulates common methods for manipulating Resources.
  */
 public class ResourceUtils {
+  private static final Logger LOGGER = LoggerFactory.getLogger(TaskUtils.class);
```
mohitsoni: `s/TaskUtils/ResourceUtils/g` OR `s/TaskUtils/getClass()/g`

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -211,6 +218,45 @@ public static ExecutorInfo clearResourceIds(ExecutorInfo executorInfo) {
...
+          ResourceSpecification newResourceSpecification) {
+
+    Protos.Value oldValue = oldResourceSpecification.getValue();
+    Protos.Value newValue = newResourceSpecification.getValue();
+    if (!ValueUtils.equal(oldValue, newValue)) {
+      LOGGER.info("Values are different.");
+      return true;
+    }
+
+    if (!oldResourceSpecification.getRole().equals(newResourceSpecification.getRole())) {
```
mohitsoni: We should replace this with `Objects.equal()` call, as that's null safe like above `!ValueUtils.equal(oldValue, newValue)`

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -211,6 +218,45 @@ public static ExecutorInfo clearResourceIds(ExecutorInfo executorInfo) {
...
+      LOGGER.info("Roles are different.");
+      return true;
+    }
+
+    if (!oldResourceSpecification.getPrincipal().equals(newResourceSpecification.getPrincipal())) {
```
mohitsoni: We should replace this with `Objects.equal()` call, as that's null safe like above `!ValueUtils.equal(oldValue, newValue)`

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,309 @@
...
+
+    @Override
+    public void resourceOffers(SchedulerDriver driver, List<Protos.Offer> offers) {
+        executor.execute(new Runnable() {
+            @Override
+            public void run() {
+                logOffers(offers);
+
+                // Task Reconciliation
+                reconciler.reconcile(driver);
```
mohitsoni: In this implementation, we don't reconcile right after we are registered OR re-registered. Instead we wait for the first offer cycle. This is sub-optimal. Is there any specific reason why it's done this way?

In busy clusters the scheduler will waste time until first offer cycle sitting idle.

src/main/java/org/apache/mesos/scheduler/DefaultTaskKiller.java
```diff
@@ -0,0 +1,95 @@
...
+            for (Protos.TaskInfo taskInfo : tasksToRestart) {
+                if (taskInfo != null) {
+                    logger.info("Restarting task: " + taskInfo.getTaskId().getValue());
+                    driver.killTask(taskInfo.getTaskId());
+                } else {
+                    logger.warn("Asked to restart null task.");
+                }
+            }
+
+            tasksToRestart = new ArrayList<>();
```
mohitsoni: Just curious: Is it possible for Master to drops `killTask` messages ? If not, ignore it.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,309 @@
...
+ * when possible.  Changes to the ServiceSpecification will result in rolling configuration updates, or the creation of
+ * new Tasks where applicable.
+ */
+public class DefaultScheduler implements Scheduler {
+    private static final Integer DELAY_BETWEEN_DESTRUCTIVE_RECOVERIES_SEC = 10 * 60;
+    private static final Integer PERMANENT_FAILURE_DELAY_SEC = 20 * 60;
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private final ServiceSpecification serviceSpecification;
+    private final String zkConnectionString;
+    private final ExecutorService executor = Executors.newFixedThreadPool(1);
```
mohitsoni: Is there any reason why you have used a fixedSizeThreadPool of size 1.
I'm trying to reason whether it's going to bite us back when it's managing a larger cluster, viz 10, 100, 1000, 10000 nodes.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
gabrielhartmann: This just hands the Reconciler a fresh SchedulerDriver whenever one's available.  I added a call to initiate reconciliation with the driver provided at registration time as well.

src/main/java/org/apache/mesos/scheduler/DefaultTaskKiller.java
gabrielhartmann: It's possible I'm sure.  Components will need to retry if a kill doesn't happen.  This will happen if you use the DefaultPlanScheduler and DefaultBlock implementation.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
gabrielhartmann: Yes this is intentional.  Parallel processing of Offers isn't supported by having multiple threads processing different potentially overlapping sets of Offers on the Blocks of a Plan.  It  will be supported by starting more than Block at once.

How would having multiple threads processing multiple batches of Offers work?  What if they are optimistic Offers?  What if multiple Blocks try to consume the same Offer in different batches?  What if a single Block accepts two different Offers on different threads?  Parallel processing of Offers is a separate effort.

mohitsoni: I was more concerned about handling status updates vs handling offers in the same thread.

I think handling offer batches one at a time is better and keeps things simple.

gabrielhartmann: It's possible to have separate thread pools for status updates and Offer handling.  I thought it might be simpler to have a single thread since then the State of the system isn't changing while Offers are being processed.  By handling Offers synchronously on all frameworks today, that is the same behavior we get.  So this represents no change in behavior.

build.gradle
gabrielhartmann: Please do this in a separate PR.  This PR is large, so I'd prefer keeping unrelated comments elsewhere.

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -211,6 +218,49 @@ public static ExecutorInfo clearResourceIds(ExecutorInfo executorInfo) {
...
+
+    if (areDifferent(oldResourceSpecification.getPrincipal(), newResourceSpecification.getPrincipal())) {
+      LOGGER.info("Principals are different.");
+      return true;
+    }
+
+    return false;
+  }
+
+  private static boolean areDifferent(String string1, String string2) {
```
mohitsoni: This is unnecessary. Can be replaced with `!Objects.equals(string1, string2)`

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -1,223 +1,282 @@
...
+    private static final String TASK_NAME_DELIM = "__";
+
+    private TaskUtils() {
+        // do not instantiate
+    }
+
+    /**
+     * Converts the unique {@link TaskID} into a Framework defined task name.
+     *
+     * For example: "instance-0_aoeu5678" => "instance-0"
```
mohitsoni: s/instance-0_aoeu5678/instance-0__aoeu5678/g

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -1,223 +1,282 @@
...
+                            + "TaskIDs should be generated with TaskUtils.toTaskId().", taskId, TASK_NAME_DELIM));
+        }
+
+        return taskId.getValue().substring(0, underScoreIndex);
+    }
+
+    /**
+     * Converts the Framework defined task name into a unique {@link TaskID}.
+     *
+     * For example: "instance-0" => "instance-0_aoeu5678"
```
mohitsoni: s/instance-0_aoeu5678/instance-0__aoeu5678/g

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,311 @@
...
+
+    public DefaultScheduler(ServiceSpecification serviceSpecification, String zkConnectionString) {
+        this.serviceSpecification = serviceSpecification;
+        this.zkConnectionString = zkConnectionString;
+        recoveryStatusRef = new AtomicReference<>(new RecoveryStatus(Collections.emptyList(), Collections.emptyList()));
+    }
+
+    void awaitTermination() throws InterruptedException {
+        executor.shutdown();
+        executor.awaitTermination(10, TimeUnit.SECONDS);
```
mohitsoni: Can we maybe make this a CONSTANT or better configuration ?

gabrielhartmann: I'll make it a constant. Configuration management is going to be a different PR.

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -137,18 +138,20 @@ public void clearFrameworkId() throws StateStoreException {
...
             byte[] bytes = curator.fetch(fwkIdPath);
             if (bytes.length > 0) {
-                return Protos.FrameworkID.parseFrom(bytes);
+                return Optional.of(Protos.FrameworkID.parseFrom(bytes));
             } else {
                 throw new StateStoreException(String.format(
                         "Failed to retrieve FrameworkID in '%s'", fwkIdPath));
             }
+        } catch (KeeperException.NoNodeException e) {
+            logger.warn("No FrameworkId found.");
```
nickbp: Log the path that failed to fetch: should ease debugging in the field

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -262,19 +269,21 @@ public void clearTask(String taskName) throws StateStoreException {
...
             byte[] bytes = curator.fetch(path);
             if (bytes.length > 0) {
-                return Protos.TaskInfo.parseFrom(bytes);
+                return Optional.of(Protos.TaskInfo.parseFrom(bytes));
             } else {
                 throw new StateStoreException(String.format(
                         "Failed to retrieve TaskInfo for TaskName: %s", taskName));
             }
+        } catch (KeeperException.NoNodeException e) {
+            logger.warn("No TaskInfo found for the requested name: " + taskName);
```
nickbp: Log the `path` that failed to fetch: should ease debugging in the field

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -183,21 +186,25 @@ public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
...
+            optionalTaskInfo = fetchTask(taskName);
         } catch (Exception e) {
             throw new StateStoreException(String.format(
-                    "Unable to retrieve matching TaskInfo for the provided TaskStatus name %s. " +
-                            "Call storeTasks() before calling storeStatus()", taskName), e);
+                    "Unable to retrieve matching TaskInfo for the provided TaskStatus name %s.", taskName), e);
+        }
+
+        if (!optionalTaskInfo.isPresent()) {
+            throw new StateStoreException("The following TaskStatus is not present in the StateStore: " + taskName);
```
nickbp: TaskInfo, not TaskStatus?

In the error, could add a blurb explaining that 'TaskInfo must be present before writing TaskStatus'

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -373,6 +382,32 @@ public void clearProperty(final String key) throws StateStoreException {
         }
     }
 
+    @Override
+    public List<Protos.Resource> getExpectedResources() {
```
nickbp: Given that this is effectively a utility method on top of fetchTasks(), maybe it would fit better in a separate utility that's upstream of/separate from the StateStore itself?

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -211,6 +218,49 @@ public static ExecutorInfo clearResourceIds(ExecutorInfo executorInfo) {
...
   }
 
+  public static boolean areDifferent(
+          ResourceSpecification oldResourceSpecification,
+          ResourceSpecification newResourceSpecification) {
+
+    Protos.Value oldValue = oldResourceSpecification.getValue();
+    Protos.Value newValue = newResourceSpecification.getValue();
+    if (!ValueUtils.equal(oldValue, newValue)) {
+      LOGGER.info("Values are different.");
```
nickbp: Would it make sense to debug() for these (or will people often want to see these)?

nickbp: Also, these could print the things being compared: the two values, the two roles, the two principals

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -211,6 +218,49 @@ public static ExecutorInfo clearResourceIds(ExecutorInfo executorInfo) {
...
+    Protos.Resource.Builder builder = Protos.Resource.newBuilder(resource);
+    switch (resource.getType()) {
+      case SCALAR:
+        return builder.setScalar(resourceSpecification.getValue().getScalar()).build();
+      case RANGES:
+        return builder.setRanges(resourceSpecification.getValue().getRanges()).build();
+      case SET:
+        return builder.setSet(resourceSpecification.getValue().getSet()).build();
+      default:
+        LOGGER.error("Encountered unexpected Value type: " + resource.getType());
```
nickbp: Throw IllegalArgumentException here? this feels like a good time to fail early

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -1,223 +1,282 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
 
 /**
  * Various utility methods for manipulating data in {@link TaskInfo}s.
  */
 public class TaskUtils {
-  private static final String TARGET_CONFIGURATION_KEY = "target_configuration";
```
nickbp: It looks like a lot of the diff is 2sp -> 4sp indent, but the other files surrounding this one are 2sp?

Maybe reindent this back to 2sp to at least help with the patch size a bit.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
nickbp: Add a comment explaining this in the code imo

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
nickbp: Ideally this should be configurable by tests, and maybe go with milliseconds so that tests don't need to wait for long.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,311 @@
...
+            if (acceptedOfferId.equals(offer.getId())) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @SuppressWarnings({"DM_EXIT"})
+    private void hardExit(SchedulerErrorCode errorCode) {
+        System.exit(errorCode.ordinal());
```
nickbp: Have this log something to stderr containing the reason for exiting

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    @Override
+    public void resourceOffers(SchedulerDriver driver, List<Protos.Offer> offers) {
+        executor.execute(new Runnable() {
+            @Override
+            public void run() {
+                logOffers(offers);
+
+                // Task Reconciliation
```
nickbp: Given that this is the core logic to any scheduler (and that SDK users will likely look at this early on in their work), I think it'd make sense for to add a few lines on each step which describes what exactly the steps are doing.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,311 @@
...
+
+    @Override
+    public void disconnected(SchedulerDriver driver) {
+        logger.error("Disconnected from Master.");
+        hardExit(SchedulerErrorCode.DISCONNECTED);
+    }
+
+    @Override
+    public void slaveLost(SchedulerDriver driver, Protos.SlaveID slaveId) {
+        logger.warn("Slave lost: " + slaveId);
```
nickbp: Could add either a TODO describing what should be done by us in an ideal implementation, or a comment saying why we don't need to do anything

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    @Override
+    public void executorLost(SchedulerDriver driver, Protos.ExecutorID executorId, Protos.SlaveID slaveId, int status) {
+        logger.warn(String.format("Lost Executor: %s on Agent: %s", executorId, slaveId));
```
nickbp: Could add either a TODO describing what should be done by us in an ideal implementation, or a comment saying why we don't need to do anything

src/main/java/org/apache/mesos/scheduler/DefaultTaskKiller.java
```diff
@@ -0,0 +1,95 @@
...
+    public void killTask(String taskName, boolean destructive) {
+        Optional<Protos.TaskInfo> taskInfoOptional = stateStore.fetchTask(taskName);
+        if (!taskInfoOptional.isPresent()) {
+            logger.warn("Attempted to kill unknown task: " + taskName);
+            return;
+        }
+
+        Protos.TaskInfo taskInfo = taskInfoOptional.get();
+        if (!stateStore.fetchStatus(taskName).getState().equals(Protos.TaskState.TASK_RUNNING)) {
+            logger.warn("Task is not running, no need to kill: " + taskInfo);
```
nickbp: Print the TaskStatus (possibly in addition to the TaskInfo)

src/main/java/org/apache/mesos/scheduler/DefaultTaskKiller.java
```diff
@@ -0,0 +1,95 @@
...
+            return;
+        }
+
+        if (destructive) {
+            synchronized (rescheduleLock) {
+                logger.info("Scheduling task to be killed destructively: " + taskInfo);
+                tasksToReschedule.add(taskInfo);
```
nickbp: Throw IllegalArgumentException (or just `warn`) here if taskInfo is null. Then the null checks in processTasksTo[Restart|Reschedule] below can be removed.

src/main/java/org/apache/mesos/scheduler/DefaultTaskKiller.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    public void killTask(String taskName, boolean destructive) {
+        Optional<Protos.TaskInfo> taskInfoOptional = stateStore.fetchTask(taskName);
+        if (!taskInfoOptional.isPresent()) {
+            logger.warn("Attempted to kill unknown task: " + taskName);
+            return;
+        }
+
+        Protos.TaskInfo taskInfo = taskInfoOptional.get();
```
nickbp: Throw IllegalArgumentException (or just `warn`) here if taskInfo is null. Then the null checks in processTasksTo[Restart|Reschedule] below can be removed.

src/main/java/org/apache/mesos/scheduler/SchedulerUtils.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.mesos.scheduler;
+
+/**
+ * This class provides utilities common to the construction and operation of Mesos Schedulers.
+ */
+public class SchedulerUtils {
```
nickbp: This name could be more specific, or is there an expectation that more functionality will be added here later?

src/main/java/org/apache/mesos/scheduler/TaskKiller.java
```diff
@@ -0,0 +1,25 @@
+package org.apache.mesos.scheduler;
+
+import org.apache.mesos.SchedulerDriver;
+
+/**
+ * This interface should be implemented to allow components to request the killing of Mesos Tasks.
```
nickbp: Could provide some examples of reasons for wanting to kill a task

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
```diff
@@ -0,0 +1,161 @@
...
+    @Override
+    public void forceComplete() {
+        logger.warn("Forcing completion of block: " + getName());
+        setStatus(Status.COMPLETE);
+    }
+
+    @Override
+    public void update(Protos.TaskStatus status) {
+        if (!tasks.containsKey(status.getTaskId())) {
+            logger.warn(getName() + " ignoring irrelevant TaskStatus: " + status);
```
nickbp: Will this mismatch happen often? If not, this could be a throw. Otherwise it could be an info().

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
```diff
@@ -0,0 +1,161 @@
...
+        return "Block: " + name;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public String toString() {
```
nickbp: Could use ToStringBuilder, like this: https://github.com/mesosphere/dcos-commons/blob/71de6c54646600ed510a0fcc94495b5df4960f5a/src/main/java/org/apache/mesos/offer/OfferRequirement.java#L115

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
```diff
@@ -182,6 +181,10 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return ReflectionToStringBuilder.toString(this);
+        return "DefaultPhase{" +
```
nickbp: Any reason for making this manual? It looks like `byId` is no longer being printed. If that's intentional then maybe mention that in a comment.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlan.java
```diff
@@ -0,0 +1,84 @@
...
+        for (Phase phase : getPhases()) {
+            if (!phase.isComplete()) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    @Override
+    public String toString() {
```
nickbp: ReflectionToStringBuilder imo

src/main/java/org/apache/mesos/specification/DefaultResourceSpecification.java
```diff
@@ -0,0 +1,50 @@
...
+        return principal;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public String toString() {
```
nickbp: ReflectionToStringBuilder imo

src/main/java/org/apache/mesos/specification/DefaultTaskSpecification.java
```diff
@@ -0,0 +1,73 @@
...
+                            resource.getName(),
+                            ValueUtils.getValue(resource),
+                            resource.getRole(),
+                            resource.getReservation().getPrincipal()));
+        }
+        return resourceSpecifications;
+    }
+
+    @Override
+    public String toString() {
```
nickbp: ReflectionToStringBuilder imo

src/main/java/org/apache/mesos/specification/DefaultTaskTypeSpecification.java
```diff
@@ -0,0 +1,56 @@
...
+        return command;
+    }
+
+    @Override
+    public Collection<ResourceSpecification> getResources() {
+        return resources;
+    }
+
+    @Override
+    public String toString() {
```
nickbp: ReflectionToStringBuilder imo

src/main/java/org/apache/mesos/specification/Named.java
```diff
@@ -0,0 +1,8 @@
+package org.apache.mesos.specification;
+
+/**
+ * A named object provides a name.
```
nickbp: This feels so abstract that I think we could just leave it out. Will anyone need to have a collection of different types of `Named`s?

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanManager.java
```diff
@@ -9,75 +9,73 @@
...
     this.strategyFactory = strategyFactory;
-    setStage(stage);
+    setPlan(plan);
   }
 
-  @Override
-  public void setStage(final Stage stage) {
-    LOGGER.info("Setting stage : state = {}", stage);
-    this.stage = stage;
+  public void setPlan(final Plan plan) {
```
nickbp: Bring back `@Override`

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanManager.java
```diff
@@ -9,75 +9,73 @@
...
+    LOGGER.info("Setting plan : state = {}", plan);
+    this.plan = plan;
   }
 
-  @Override
-  public Stage getStage() {
-    return stage;
+  public Plan getPlan() {
```
nickbp: Bring back `@Override`

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanScheduler.java
```diff
@@ -0,0 +1,113 @@
...
+        logger.info("Processing resource offers for block: {}", block.getName());
+        Optional<OfferRequirement> offerRequirementOptional = block.start();
+        if (!offerRequirementOptional.isPresent()) {
+            logger.info("No OfferRequirement for block: {}", block.getName());
+            block.updateOfferStatus(Optional.empty());
+            return acceptedOffers;
+        }
+
+        OfferRequirement offerRequirement = offerRequirementOptional.get();
+        // It is harmless to attempt to kill tasks which have never been launched.
```
nickbp: What does killTasks() accomplish? Delete data from StateStore? (maybe add a comment describing the answer)

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanScheduler.java
```diff
@@ -0,0 +1,113 @@
...
+            block.updateOfferStatus(Optional.empty());
+            return acceptedOffers;
+        }
+
+        acceptedOffers = offerAccepter.accept(driver, recommendations);
+        // Notify block of offer outcome:
+        if (acceptedOffers.size() > 0) {
+            block.updateOfferStatus(getOperations(recommendations));
+        } else {
+            block.updateOfferStatus(Optional.empty());
```
nickbp: Does it make sense to pass an empty operation list to the block? (Maybe leave a comment explaining the use case here)

src/main/java/org/apache/mesos/scheduler/plan/api/CurrentlyActiveInfo.java
```diff
@@ -60,7 +63,7 @@ public CurrentlyActivePhaseInfo getPhaseStatus() {
       return phaseStatus;
     }
 
-    @JsonProperty("stage")
+    @JsonProperty("plan")
```
nickbp: Any docs/samples need updating for this?

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryRequirementProvider.java
```diff
@@ -0,0 +1,53 @@
...
+                            RecoveryRequirement.RecoveryType.TRANSIENT));
+        }
+
+        return transientRecoveryRequirements;
+    }
+
+    @Override
+    public List<RecoveryRequirement> getPermanentRecoveryOfferRequirements(List<Protos.TaskInfo> failedTasks)
+            throws InvalidRequirementException {
+        List<RecoveryRequirement> transientRecoveryRequirements = new ArrayList<>();
```
nickbp: permanentRecoveryRequirements

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryScheduler.java
```diff
@@ -127,7 +127,7 @@ public DefaultRecoveryScheduler(
         return filteredTerminatedTasks;
     }
 
-    private void updateRecoveryPools(Block block) {
+    private void updateRecoveryPools(Optional<Block> block) {
```
nickbp: Subjective: How about switching this parameter to `List<TaskInfo> terminatedTasks` and then having the caller do something like `updateRecoveryPolls(getTerminatedTasks(block))`?

src/main/java/org/apache/mesos/scheduler/recovery/DefaultTaskFailureListener.java
```diff
@@ -0,0 +1,38 @@
...
+        try {
+            Optional<Protos.TaskInfo> optionalTaskInfo = stateStore.fetchTask(TaskUtils.toTaskName(taskId));
+            if (optionalTaskInfo.isPresent()) {
+                Protos.TaskInfo taskInfo = FailureUtils.markFailed(optionalTaskInfo.get());
+                stateStore.storeTasks(Arrays.asList(taskInfo));
+            } else {
+                logger.error("TaskInfo for TaskID was not present in the StateStore: " + taskId);
+            }
+        } catch (TaskException e) {
+            logger.error("Failed to fetch Task for taskId: " + taskId + " with exception:", e);
```
nickbp: maybe say 'fetch/update' since technically either fetchTask or storeTask could've failed

src/main/java/org/apache/mesos/scheduler/recovery/FailureUtils.java
```diff
@@ -0,0 +1,45 @@
...
+        if (!labeledAsFailed(taskInfo)) {
+            taskInfo = ResourceUtils.clearResourceIds(taskInfo);
+            return Protos.TaskInfo.newBuilder(taskInfo)
+                    .setLabels(Protos.Labels.newBuilder(taskInfo.getLabels())
+                            .addLabels(Protos.Label.newBuilder()
+                                    .setKey(PERMANENTLY_FAILED_KEY)
+                                    .setValue(String.valueOf(true))))
+                    .build();
+        } else {
+            return  taskInfo;
```
nickbp: double space

src/main/java/org/apache/mesos/scheduler/recovery/FailureUtils.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public class FailureUtils {
+    private static final String PERMANENTLY_FAILED_KEY = "permanently-failed";
+
+    /**
+     * Check if a Task has been marked as permanently failed.
+     * @param taskInfo The Task to check for failure.
+     * @return True if the Task has been marked, false otherwise.
+     */
+    public static boolean labeledAsFailed(Protos.TaskInfo taskInfo) {
```
nickbp: isLabeledAsFailed

src/main/java/org/apache/mesos/scheduler/recovery/FailureUtils.java
```diff
@@ -0,0 +1,45 @@
...
+            if (label.getKey().equals(PERMANENTLY_FAILED_KEY) && Boolean.valueOf(label.getValue())) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Mark a task as permanently failed.  This new marekd Task should be persistently stored.
```
nickbp: - "marked"
- Maybe mention that this may return a copy of the provided TaskInfo

src/main/java/org/apache/mesos/specification/DefaultPhaseSpecification.java
```diff
@@ -0,0 +1,37 @@
...
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public List<TaskSpecification> getTaskSpecifications() {
+        return taskSpecifications;
+    }
+
+    private TaskSpecification getTaskSpecification(int id, TaskTypeSpecification taskTypeSpecification) {
```
nickbp: `private static TaskSpecification create(...)` ?

src/main/java/org/apache/mesos/specification/DefaultPhaseSpecification.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+
+/**
+ * This class provides a default implementation of the PhaseSpecification interface.
+ */
+public class DefaultPhaseSpecification implements PhaseSpecification {
+    private final String name;
+    private final List<TaskSpecification> taskSpecifications;
+
+    public DefaultPhaseSpecification(String name, TaskTypeSpecification taskTypeSpecification) {
```
nickbp: Given that the only caller of this constructor (in `DefaultPlanSpecificationFactory`) is getting the `name` param by just calling `taskTypeSpecification.getName()`, maybe remove that param and move the `getName()` call here inside the constructor?

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -137,18 +138,20 @@ public void clearFrameworkId() throws StateStoreException {
...
             if (bytes.length > 0) {
-                return Protos.FrameworkID.parseFrom(bytes);
+                return Optional.of(Protos.FrameworkID.parseFrom(bytes));
             } else {
                 throw new StateStoreException(String.format(
                         "Failed to retrieve FrameworkID in '%s'", fwkIdPath));
             }
+        } catch (KeeperException.NoNodeException e) {
+            logger.warn("No FrameworkId found.");
+            return Optional.empty();
```
mrbrowning: I'm unclear on the what the move to `Optional` buys us here -- in general, I think it's best used when semantics would otherwise dictate returning `null`, rather than as an alternative to exceptional control flow -- but even then, we swallow the one exception here and not the other, which seems to strictly increase the complexity foisted on clients.

src/main/java/org/apache/mesos/specification/DefaultPlanSpecificationFactory.java
```diff
@@ -0,0 +1,21 @@
...
+ * This class provides a default implementation of the PlanSpecifiactionFactory interface.
+ */
+public class DefaultPlanSpecificationFactory implements PlanSpecificationFactory {
+    @Override
+    public PlanSpecification getPlanSpecification(ServiceSpecification serviceSpecification) {
+        List<PhaseSpecification> phaseSpecifications = new ArrayList<>();
+
+        for (TaskTypeSpecification taskTypeSpecification : serviceSpecification.getTaskSpecifications()) {
+            phaseSpecifications.add(
+                    new DefaultPhaseSpecification(taskTypeSpecification.getName(), taskTypeSpecification));
```
nickbp: See note above -- DefaultPhaseSpecification could accept just the taskTypeSpecification and call getName() internally.

src/main/java/org/apache/mesos/specification/DefaultServiceSpecification.java
```diff
@@ -0,0 +1,46 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+
+        DefaultServiceSpecification that = (DefaultServiceSpecification) o;
+
+        if (!name.equals(that.name)) return false;
+        return taskTypeSpecifications.equals(that.taskTypeSpecifications);
+
+    }
+
+    @Override
```
nickbp: Is the default implementation known to be bad? If so add a comment explaining why

src/main/java/org/apache/mesos/specification/DefaultServiceSpecification.java
```diff
@@ -0,0 +1,46 @@
...
+    public List<TaskTypeSpecification> getTaskSpecifications() {
+        return taskTypeSpecifications;
+    }
+
+    @Override
+    @SuppressWarnings("PMD.IfStmtsMustUseBraces")
+    public boolean equals(Object o) {
```
nickbp: Is the default implementation known to be bad? If so add a comment explaining why

src/main/java/org/apache/mesos/specification/Named.java
nickbp: To illustrate, among interfaces edited in this PR, each of `Phase`, `TaskSpecification`, and `ServiceSpecification` could all also extend `Named` as-is but I don't think we'd want those to all be jumbled into a single collection together.

src/main/java/org/apache/mesos/specification/TaskTypeSpecification.java
```diff
@@ -0,0 +1,8 @@
+package org.apache.mesos.specification;
+
+/**
+ * A TaskTypeSpecification describes the count of a particular TaskSpecification which should be launched.
```
nickbp: - Why does the TypeSpecification extension just add a count? This name feels odd
- Any reason to not just merge this into TaskSpecification? What is the count of?

src/main/java/org/apache/mesos/state/StateStore.java
```diff
@@ -207,4 +207,9 @@
      */
     void clearProperty(final String key) throws StateStoreException;
 
+    /**
+     * Provides the list of Resources which are expected to be reserved to a particular role.
+     * @return
+     */
+    List<Protos.Resource> getExpectedResources();
```
nickbp: As mentioned before it feels like this would be better placed as a helper method upstream of StateStore.

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
gabrielhartmann: This makes it more likely that implementers get the ResourceCleaner for free.  While not strictly necessary, I would say this is a point we can have an opinion about what a well formed StateStore should look like.  It should make it possible to not leak resources.  There are a lot of interfaces already.  I think in practice every StateStore implementation should want to provide this capability, so let's enforce it.

src/main/java/org/apache/mesos/offer/TaskUtils.java
gabrielhartmann: concat `?w=0` to see only non-whitespace changes.  I'm moving everything to 4 spaces as I encounter it.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
gabrielhartmann: This is logged by all reasonably implemented Executors including the CmdExecutor used by Marathon.

src/main/java/org/apache/mesos/scheduler/SchedulerUtils.java
gabrielhartmann: I expect more to migrate here.

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -1,223 +1,282 @@
...
+            builder.setData(ByteString.copyFrom(data));
+        }
+
+        final Protos.TaskStatus taskStatus = builder.build();
+        driver.sendStatusUpdate(taskStatus);
+    }
+
+    public static boolean areDifferent(TaskSpecification oldTaskSpecification, TaskSpecification newTaskSpecification) {
+        if (!oldTaskSpecification.getName().equals(newTaskSpecification.getName())) {
+            LOGGER.info("Task names are different.");
```
nickbp: Print the names being compared?

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -1,223 +1,282 @@
...
+            return true;
+        }
+
+        if (!oldTaskSpecification.getCommand().equals(newTaskSpecification.getCommand())) {
+            LOGGER.info("Task commands are different.");
```
nickbp: Print the commands being compared?

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -1,223 +1,282 @@
...
+            return true;
+        }
+
+        Map<String, ResourceSpecification> oldResourceMap = getResourceSpecMap(oldTaskSpecification.getResources());
+        Map<String, ResourceSpecification> newResourceMap = getResourceSpecMap(newTaskSpecification.getResources());
+
+        if (oldResourceMap.size() != newResourceMap.size()) {
+            LOGGER.info("Resource lengths are different.");
```
nickbp: Print the maps being compared?

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -1,223 +1,282 @@
...
+            return true;
+        }
+
+        for (Map.Entry<String, ResourceSpecification> newEntry : newResourceMap.entrySet()) {
+            String resourceName = newEntry.getKey();
+            LOGGER.info("Checking resource difference for: " + resourceName);
+            ResourceSpecification oldResourceSpec = oldResourceMap.get(resourceName);
+            if (oldResourceSpec == null) {
+                LOGGER.info("Resource not found.");
```
nickbp: Print resourceName

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
gabrielhartmann: I would expect this to occur on stale TaskStatus messages and so should be pretty rare, but it's a normal thing to occur so it's not worthy of an exception.  I can change to Info.  I put it as Warn because it's something worthy of investigation.  Definitely if you're seeing it a lot.  Let me know what you think.

src/main/java/org/apache/mesos/specification/Named.java
gabrielhartmann: @nickbp:  You suggested this!  Removing...

src/main/java/org/apache/mesos/scheduler/plan/api/CurrentlyActiveInfo.java
gabrielhartmann: I looked around in Kafka and Cassandra's docs and see no mention.  The endpoint was always `v1/plan` and the word `stage` never showed up in that output.

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
gabrielhartmann: It distinguishes between two cases.  The thrown exception indicates that something bad and unexpected happened.  The Optional.empty() indicates that something hasn't been set before which is an expected scenario.  Clients should know how to handle the the expected empty case as well as unexpected communication errors.

This also helps the fact that our logs were filled with stack traces in expected situations which was confusing users.

src/main/java/org/apache/mesos/specification/DefaultServiceSpecification.java
gabrielhartmann: Spoke away from keyboard.

src/main/java/org/apache/mesos/specification/TaskTypeSpecification.java
gabrielhartmann: TaskTypSpecification:
count: 3
name: foo

generates --> 

TaskSpecification
name: foo-0

TaskSpecificaiton
name: foo-1

etc.

nickbp: Got it. Maybe call it TaskCountSpecification?

src/main/java/org/apache/mesos/scheduler/DefaultTaskKiller.java
gabrielhartmann: Done

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
gabrielhartmann: Done

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlan.java
gabrielhartmann: Done.

src/main/java/org/apache/mesos/specification/DefaultResourceSpecification.java
gabrielhartmann: Done.

src/main/java/org/apache/mesos/specification/DefaultTaskSpecification.java
gabrielhartmann: Done.

src/main/java/org/apache/mesos/specification/DefaultTaskTypeSpecification.java
gabrielhartmann: Done.

src/main/java/org/apache/mesos/specification/DefaultServiceSpecification.java
gabrielhartmann: Spoke away from keyboard.

src/main/java/org/apache/mesos/specification/TaskTypeSpecification.java
gabrielhartmann: It isn't just a count.  It's also the name and the command and the resources and in the future volume specifications etc.

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
mrbrowning: Ah, ok, that makes sense. I misread the `NoNodeException` as an unexpected error itself. Is `curator` a part of this repo? I like the idea of standardizing these semantics across the board, so it might be appropriate for `CuratorPersister::fetch` to return an `Optional` itself if its returning an empty value is a normal part of proceedings here.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    @Override
+    public void frameworkMessage(
+            SchedulerDriver driver,
+            Protos.ExecutorID executorId,
+            Protos.SlaveID slaveId,
+            byte[] data) {
+        logger.error("Received a Framework Message, but don't know how to process it");
+        hardExit(SchedulerErrorCode.FRAMEWORK_MESSAGE);
```
mrbrowning: Why the hard exit? I guess if we don't expect the executor to send a message it would be a very exceptional case to receive this, but.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    @Override
+    public void disconnected(SchedulerDriver driver) {
+        logger.error("Disconnected from Master.");
+        hardExit(SchedulerErrorCode.DISCONNECTED);
```
mrbrowning: This may be my ignorance of practical handling of master failure, but why can't we wait for it to come back? Is this handled by Marathon?

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
```diff
@@ -0,0 +1,161 @@
...
+    private Status status;
+    private Map<Protos.TaskID, Status> tasks = new HashMap<>();
+
+    public DefaultBlock(String name, OfferRequirement offerRequirement, Status status) {
+        this(name, Optional.of(offerRequirement), status);
+    }
+
+    private DefaultBlock(String name, Optional<OfferRequirement> offerRequirementOptional, Status status) {
+        this.name = name;
+        this.offerRequirementOptional = offerRequirementOptional;
```
mrbrowning: IMO it's the compiler's job to keep track of types, no need to encode them in variable names.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
gabrielhartmann: To avoid a scenario where it's possible that more than 1 Scheduler is doing anything.  Because the DefaultScheduler cannot do anything until it is registered and only one Scheduler may be registered at a time, we guarantee that no work is accidentally done by more than 1 Scheduler simultaneously by crashing on disconnect.

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
gabrielhartmann: What would you call it?  I started out calling things like this `offerRequirement` although that is still the type.  However when I did the `offerRequirement.get()` and assign it to some new variable, I'd also like to call that one `offerRequirement`.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
gabrielhartmann: Ok.

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
mrbrowning: For myself I'd call it `offerRequirement` and then the result from the `.get()` could just be `requirement` -- the heuristic I employ is that the importance of a variable name being descriptive scales in proportion to the length of its scope, thus e.g. the use of single-letter variable names in lambda expressions. As always, a matter of taste.

gabrielhartmann:
@nickbp:  Anything left here?

nickbp:
LGTM

PR #109, opened 2016-08-31
Remove the protobuf helper methods.

gabrielhartmann:
They introducted confusion and were never fully used, as they couldn't
keep up with an actual protobuf implmemenation.  They added little value
and were dragging test coverage numbers down.

Test coverage goes from 79.4% to 81.4%.  This is me paying my test coverage regression debt from #104.


src/main/java/org/apache/mesos/offer/CreateOfferRecommendation.java
```diff
@@ -17,9 +16,10 @@
 
     public CreateOfferRecommendation(Offer offer, Resource resource) {
         this.offer = offer;
-        this.operation = new OperationBuilder()
+        this.operation = Operation.newBuilder()
                 .setType(Operation.Type.CREATE)
-                .setCreate(Arrays.asList(resource))
+                .setCreate(Operation.Create.newBuilder()
+                        .addAllVolumes(Arrays.asList(Resource.newBuilder(resource).build())))
```
mrbrowning: Can't we still just pass `resource` instead of `Resource.newBuilder(resource).build()`? It's not going to get mutated.

gabrielhartmann: Yep.

src/test/java/org/apache/mesos/testutils/OfferRequirementTestUtils.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * This class provides utility methods for tests concerned with OfferRequirements.
+ */
+public class OfferRequirementTestUtils {
+    public static OfferRequirement getOfferRequirement(Protos.Resource resource) throws InvalidRequirementException {
+        return getOfferRequirement(Arrays.asList(resource));
```
mrbrowning: What does this abstraction buy us? I understand the desire to keep things dry, but sometimes adding an extra layer of indirection without much logic can hamper readability, and I like the idea of tests being super-explicit since they often function as a de facto spec.

src/test/java/org/apache/mesos/offer/MesosResourcePoolTest.java
```diff
@@ -1,41 +1,28 @@
...
-  private static final String testRole = "test-role";
-  private static final String testPrincipal = "test-principal";
-  private static final String testContainerPath = "test-containter-path";
-  private static final String testResourceId = "test-resource-id";
+public class MesosResourcePoolTest {
 
   @Test
   public void testEmptyUnreservedAtomicPool() {
-    Offer offer = getOffer(ResourceBuilder.cpus(1.0));
+    Offer offer = OfferTestUtils.getOffer(ResourceTestUtils.getUnreservedCpu(1.0));
```
gabrielhartmann: This library is constantly attempting to reserver resources.  It's hard to say what's the default for us.  I like being explicit that something isn't reserved yet, with the implication that it's about to be probably.

src/test/java/org/apache/mesos/testutils/OfferRequirementTestUtils.java
gabrielhartmann: Yes, to keep things DRY.  It was being reimplemented a few times in a few different test classes.

PR #110, opened 2016-09-01
[KAF-227] suppress/revive support

mgummelt:
To support suppress/revive 

Also adds ObjectPropertyDeserializer to support reading serialized java objects from zk

Adds SchedulerState (factored out from KafkaStateStore)


mgummelt:
@gabrielhartmann Needs review

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -354,6 +355,35 @@ public void storeProperty(final String key, final byte[] value) throws StateStor
         }
     }
 
+    public void storePropertyAsObj(final String key, final Object obj) throws StateStoreException {
```
gabrielhartmann: Should these be part of the `StateStore` interface?

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -138,7 +138,7 @@ public static Resource getUnreservedScalar(String name, double value) {
 
     public static Resource getExpectedScalar(
             String name,
-            double value,
+           double value,
```
gabrielhartmann: Accidental deletion?

src/main/java/org/apache/mesos/reconciliation/Reconciler.java
```diff
@@ -12,8 +12,7 @@
 public interface Reconciler {
 
     /**
-     * Starts reconciliation against the provided tasks, which should represent what the Scheduler
-     * currently knows about task status.
+     * Starts reconciliation.
```
gabrielhartmann: `Starts Task reconciliation`

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryScheduler.java
```diff
@@ -85,7 +98,8 @@ public DefaultRecoveryScheduler(
             recoveryRequirement = Optional.of(recoveryCandidates.get(new Random().nextInt(recoveryCandidates.size())));
         }
 
-        if (recoveryRequirement.isPresent() && launchConstrainer.canLaunch(recoveryRequirement.get())) {
+        boolean b = recoveryRequirement.isPresent() && launchConstrainer.canLaunch(recoveryRequirement.get());
```
gabrielhartmann: Why does the `b` variable exist?  Shouldn't this all be part of the if statement below?

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryScheduler.java
```diff
@@ -127,11 +148,13 @@ public DefaultRecoveryScheduler(
...
     }
 
-    private void updateRecoveryPools(Collection<TaskInfo> terminatedTasks) {
+    private void updateRecoveryStatus(Collection<TaskInfo> terminatedTasks) {
         List<TaskInfo> failed = new ArrayList<>(terminatedTasks.stream()
                 .filter(failureMonitor::hasFailed)
                 .collect(Collectors.toList()));
         failed = failed.stream().distinct().collect(Collectors.toList());
+
+        // Why? They're already stored as failed.
```
gabrielhartmann: File a JIRA or a TODO, please.

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.mesos.reconciliation.TaskStatusProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+/**
+ * Read/write interface for the state of a Scheduler.
+ */
+public class SchedulerState extends CuratorStateStore implements TaskStatusProvider {
```
gabrielhartmann: Prefer composition over inheritance.

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private static final String SUPPRESSED_KEY = "suppressed";
+
+    public SchedulerState(String frameworkName, String mesosZkURI) {
+        super(frameworkName, mesosZkURI);
+    }
+
+    public FrameworkID getFrameworkId() {
+        try {
+            return fetchFrameworkId().orElse(null);
```
gabrielhartmann: We just went through a series of changes to avoid returning nulls.  Let's not transform [the underlying `StateStore` interface](https://github.com/mesosphere/dcos-commons/blob/6a5e665c56dcb9eb478b89619453f4f5e0fdbad2/src/main/java/org/apache/mesos/state/StateStore.java#L53) here.

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,154 @@
...
+        try {
+            log.info(String.format("Storing framework id: %s", fwkId));
+            storeFrameworkId(fwkId);
+        } catch (StateStoreException ex) {
+            log.error("Failed to set FrameworkID: " + fwkId, ex);
+            throw ex;
+        }
+    }
+
+    public void recordTasks(List<TaskInfo> taskInfos) throws StateStoreException {
```
gabrielhartmann: This should be transactional.  Maybe not best as part of this PR, though.

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,154 @@
...
+            taskStatuses.add(taskStatus);
+        }
+
+        storeTasks(taskInfos);
+        for (TaskStatus taskStatus : taskStatuses) {
+            recordTaskStatus(taskStatus);
+        }
+    }
+
+    public void updateStatus(TaskStatus taskStatus) throws StateStoreException {
```
gabrielhartmann: What's the value of wrapping this?

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,154 @@
...
+        log.info(String.format("Recording updated TaskStatus to state store: %s", taskStatus));
+        recordTaskStatus(taskStatus);
+    }
+
+    public List<TaskInfo> getTerminatedTaskInfos() throws Exception {
+        return new ArrayList(fetchTerminatedTasks());
+    }
+
+    public int getRunningBrokersCount() throws StateStoreException {
```
gabrielhartmann: Not relevant to dcos-commons?

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public Set<TaskStatus> getTaskStatuses() throws StateStoreException {
+        Set<TaskStatus> taskStatuses = new HashSet<TaskStatus>();
+        taskStatuses.addAll(fetchStatuses());
+        return taskStatuses;
+    }
+
+    public List<TaskInfo> getTaskInfos() throws StateStoreException {
```
gabrielhartmann: Why wrap this?

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,154 @@
...
+        List<TaskInfo> taskInfos = new ArrayList<TaskInfo>();
+        taskInfos.addAll(fetchTasks());
+        return taskInfos;
+    }
+
+    public List<Resource> getExpectedResources() {
```
gabrielhartmann: [Already implemented here](https://github.com/mesosphere/dcos-commons/blob/6a5e665c56dcb9eb478b89619453f4f5e0fdbad2/src/main/java/org/apache/mesos/offer/ResourceCleaner.java#L103).  Should be factored out of both of these.

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,154 @@
...
+                }
+            }
+        } catch (Exception ex) {
+            log.error("Failed to retrieve all Task information", ex);
+            return resources;
+        }
+        return resources;
+    }
+
+    public void recordTaskInfo(TaskInfo taskInfo) throws StateStoreException {
```
gabrielhartmann: Not much value in wrapping this.

src/main/java/org/apache/mesos/offer/ResourceUtils.java
mgummelt: whoops, yea

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
mgummelt: I'll create an AbstractStateStore with these methods, so all implementations can share the functionality.

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryScheduler.java
mgummelt: Whoops, this is from my debugging.  Good catch.  Removed.

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryScheduler.java
mgummelt: whoops, removed.

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: changed

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: fixed

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: whoops, removed.

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: I just copy+pasted this from the previous `FrameworkState` class in kafka.  I can remove it and we can just use `Collection` if you like.

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
gabrielhartmann: Please use composition not inheritance.

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: removed

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: Well, since I'm moving from inheritance to composition, we have to wrap this in order to expose it, unless you want me to expose the underlying `StateStore`

mgummelt:
@gabrielhartmann ready for re-review

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public SchedulerState(AbstractStateStore stateStore) {
+        this.stateStore = stateStore;
+    }
+
+    public AbstractStateStore getStateStore() {
+        return stateStore;
+    }
+
+    public Optional<FrameworkID> getFrameworkId() {
```
gabrielhartmann: Is the added value here the removal of the `throws StateStoreException` from the method signature?  I don't understand why this is needed or wanted.  [The CuratorStateStore in particular already performs the appropriate Exception handling and logging](https://github.com/mesosphere/dcos-commons/blob/master/src/main/java/org/apache/mesos/curator/CuratorStateStore.java#L140-L156).

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            return stateStore.fetchFrameworkId();
+        } catch (StateStoreException ex) {
+            log.warn("Failed to get FrameworkID. "
+                    + "This is expected when the service is starting for the first time.", ex);
+        }
+        return Optional.empty();
+    }
+
+    public void setFrameworkId(FrameworkID fwkId) throws StateStoreException {
```
gabrielhartmann: Similar comment here, as to above.  These should be straightforward pass throughs to the `CuratorStateStore`.

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,135 @@
...
+            recordTaskStatus(taskStatus);
+        }
+    }
+
+    public void updateStatus(TaskStatus taskStatus) throws StateStoreException {
+        log.info(String.format("Recording updated TaskStatus to state store: %s", taskStatus));
+        recordTaskStatus(taskStatus);
+    }
+
+    public List<TaskInfo> getTerminatedTaskInfos() throws Exception {
```
gabrielhartmann: This only converts a `Collection` to a `List`

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,135 @@
...
+        return new ArrayList(stateStore.fetchTerminatedTasks());
+    }
+
+    @Override
+    public Set<TaskStatus> getTaskStatuses() throws StateStoreException {
+        Set<TaskStatus> taskStatuses = new HashSet<TaskStatus>();
```
gabrielhartmann: `Collection` to `Set` transformation, not necessary.

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,135 @@
...
+        taskStatuses.addAll(stateStore.fetchStatuses());
+        return taskStatuses;
+    }
+
+    public List<TaskInfo> getTaskInfos() throws StateStoreException {
```
gabrielhartmann: `Collection` to `List` transformation.

src/main/java/org/apache/mesos/state/AbstractStateStore.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+
+/**
+ * Abstract StateStore implementation.
+ */
+public abstract class AbstractStateStore implements StateStore {
```
nickbp: These feel like they could be static utility methods, rather than putting them in statestore proper

src/main/java/org/apache/mesos/state/AbstractStateStore.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final Logger log = LoggerFactory.getLogger(SchedulerState.class);
+
+    public void storePropertyAsObj(final String key, final Object obj) throws StateStoreException {
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        ObjectOutputStream oos = null;
+        try {
+            oos = new ObjectOutputStream(bos);
```
nickbp: Will there be compatibility issues if we change the JDK version in the future?

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.mesos.reconciliation.TaskStatusProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+/**
+ * Read/write interface for the state of a Scheduler.
+ */
+public class SchedulerState implements TaskStatusProvider {
```
gabrielhartmann: I haven't gone through ever method, but I don't see this class adding a lot value, but it's adding some complexity, and doesn't really implement a clear interface, but selectively wraps a lot of what `StateStore` is doing.

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,135 @@
...
+    private static final Logger log = LoggerFactory.getLogger(SchedulerState.class);
+    private static final String SUPPRESSED_KEY = "suppressed";
+
+    protected final AbstractStateStore stateStore;
+
+    public SchedulerState(String frameworkName, String mesosZkURI) {
```
nickbp: It feels odd pulling in one of the concrete state store types here. Maybe just omit this constructor in favor of the following one?

src/main/java/org/apache/mesos/state/AbstractStateStore.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.*;
+
+/**
+ * Abstract StateStore implementation.
+ */
+public abstract class AbstractStateStore implements StateStore {
+    private static final Logger log = LoggerFactory.getLogger(SchedulerState.class);
+
+    public void storePropertyAsObj(final String key, final Object obj) throws StateStoreException {
```
gabrielhartmann: These could both be static methods of StateStore and avoid the definition of an abstract class, of which there are none (to my knowledge) in dcos-commons.

src/main/java/org/apache/mesos/state/SchedulerState.java
nickbp: It looks like a lot of functions in this class are just performing datatype conversions against the underlying StateStore. Are those really needed or could they be omitted here?

src/main/java/org/apache/mesos/state/AbstractStateStore.java
gabrielhartmann: If this is just to transform boolean (isSuppressed) to byte array and back.  We could defer this complexity with something like:

```
boolean vIn = true;
byte vOut = (byte)(vIn?1:0);
```

src/main/java/org/apache/mesos/state/StateStoreUtils.java
```diff
@@ -1,13 +1,16 @@
 package org.apache.mesos.state;
 
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Utilities for implementations and users of {@link StateStore}.
  */
 public class StateStoreUtils {
```
mgummelt: This seems to create an inconsistency in how scheduler state is stored:

If I want to set the frameworkID:

```
frameworkState.setFrameworkId(id)
```

But if I want to set the suppressed state:

```
StateStoreUtils.setSuppressed(frameworkState.getStateStore(), false);
```

These are both just two different properties of the scheduler, but now they have significantly different interfaces.  I'd rather just write:

```
frameworkState.setFrameworkId(id)
frameworkState.setSuppressed(false)
```

Can we instead make a concrete super class: `class SchedulerState` that implements `setSuppressed`, `getSuppressed`, `setFrameworkId`, etc?

mgummelt:
@gabrielhartmann ready for re-review

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final Logger log = LoggerFactory.getLogger(SchedulerState.class);
+
+    private static final String SUPPRESSED_KEY = "suppressed";
+    private final StateStore stateStore;
+
+    public SchedulerState(StateStore stateStore) {
+        this.stateStore = stateStore;
+    }
+
+    public Optional<FrameworkID> getFrameworkId() {
```
nickbp: This and setFrameworkId() feel unrelated to the rest of the class. Could the caller just invoke the stateStore functions directly?

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.mesos.Protos.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+/**
+ * Read/write interface for the state of a Scheduler.
+ */
+public class SchedulerState  {
```
nickbp: Double space

src/main/java/org/apache/mesos/state/SchedulerState.java
```diff
@@ -0,0 +1,65 @@
...
+        if (suppressed.length != 1) {
+            log.error("Unexpected SUPPRESSED byte array length: " + suppressed.length);
+            return false;
+        }
+
+        return (suppressed[0] == 1) ? true : false;
+    }
+
+    public void setSuppressed(boolean isSuppressed) {
+        final byte[] suppressed = new byte[]{(byte) (isSuppressed ? 1 : 0)};
```
nickbp: Could just store a 'true' or 'false' ascii string. Then it'd be easy to diagnose when accessing the data directly (eg in exhibitor)

src/test/java/org/apache/mesos/scheduler/recovery/RecoverySchedulerTest.java
```diff
@@ -41,7 +41,7 @@
  * as failed
  */
 public class RecoverySchedulerTest {
-    private DefaultRecoveryScheduler recoveryScheduler;
+    private DefaultRecoveryScheduler repairScheduler;
```
nickbp: Any reason for the rename? the type is RecoveryScheduler..

src/test/java/org/apache/mesos/state/SchedulerStateTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Test
+    public void testGetFrameworkId() {
+        schedulerState.setFrameworkId(TestConstants.frameworkId);
+        assertEquals(schedulerState.getFrameworkId().get(), TestConstants.frameworkId);
+    }
+
+    @Test
+    public void testIsSuppressed() {
+        schedulerState.setSuppressed(true);
```
nickbp: Maybe add a call before setSuppressed() has been called, to test the behavior when `suppressed` is unset

src/test/java/org/apache/mesos/scheduler/recovery/RecoverySchedulerTest.java
gabrielhartmann: This is just a mistake from before.  It used to be called a `DefaultRepairScheduler` but the variable name didn't keep up.

gabrielhartmann: Also, I'm confused.  I thought [I updated these already](https://github.com/mesosphere/dcos-commons/blob/master/src/test/java/org/apache/mesos/scheduler/recovery/RecoverySchedulerTest.java#L44).  @mgummelt: What's going on with the diff here?  Regardless, let's leave changes irrelevant to the content of this PR to other PRs.

gabrielhartmann:
@mgummelt:  In general, I agree with @nickbp 's comments.

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: I will, sure.  

But how I really think this should be designed (possibly a future change) is:
- StateStore should be a lower level interface that just deals with key/value storage.
- SchedulerState should have all the scheduler specific state (frameworkId, suppressed, task state, etc.)

As it is, some scheduler state is exposed in StateStore, and some is exposed in SchedulerState (or equivalently, KafkaState and CassandraState), and the distinction seems arbitrary.

Thoughts?

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: The 0 and 1 are visible in exhibitor, and given that they're more canonical binary representations of a boolean, I'd rather keep it as is.

src/test/java/org/apache/mesos/scheduler/recovery/RecoverySchedulerTest.java
mgummelt: Yea, not sure what happened here.  Will fix.

src/main/java/org/apache/mesos/state/SchedulerState.java
mgummelt: what do you mean?

mgummelt: fixed

mgummelt:
@nickbp @gabrielhartmann ready for re-review

PR #114, opened 2016-09-07
Add VolumeSpecifications to TaskSpecifications

gabrielhartmann:
This allows persistent volumes to be specified as part of a ServiceSpecification.


src/main/java/org/apache/mesos/specification/DefaultTaskSpecification.java
```diff
@@ -3,36 +3,49 @@
...
+                getVolumes(taskInfo));
     }
 
     public static DefaultTaskSpecification create(
             String name,
             Protos.CommandInfo commandInfo,
-            Collection<ResourceSpecification> resourceSpecifications) {
-        return new DefaultTaskSpecification(name, commandInfo, resourceSpecifications);
+            Collection<ResourceSpecification> resourceSpecifications,
+            Optional<Collection<VolumeSpecification>> volumeSpecifications) {
```
nickbp: what's the difference between an empty list of volumes and an unset option?

PR #115, opened 2016-09-07
Switch to WORKSPACE, get run_tests.py to actually work

nickbp:
Bootstraps/runs both shakedown and dcos-tests, the latter notably without a docker image for the CLI.

### DIFF
--- a/tools/run_tests.py
+++ b/tools/run_tests.py
@@ -82,7 +82,7 @@ class CITester(object):
     def setup_cli(self):
         try:
             self.__github_updater.update('pending', 'Setting up CLI')
-            self.__configure_cli_sandbox() # adds to os.environ
+            self.__configure_cli_sandbox()  # adds to os.environ
             cli_filepath = self.__download_cli_to_sandbox()
             self.__configure_cli(self.__dcos_url)
             dcos_login.DCOSLogin(self.__dcos_url).login()
@@ -165,7 +165,7 @@ SSH_KEY_FILE="" PYTHONPATH=$(pwd) py.test {jenkins_args}-vv -s -m "{pytest_types
 
     def delete_sandbox(self):
         if not self.__sandbox_path:
-            return # no-op
+            return  # no-op
         logger.info('Deleting CLI sandbox: {}'.format(self.__sandbox_path))
         shutil.rmtree(self.__sandbox_path)
 
--- a/tools/run_tests.py
+++ b/tools/run_tests.py
@@ -105,6 +105,8 @@ class CITester(object):
             requirements_cmd = 'pip install -r {}'.format(requirements_file)
         else:
             requirements_cmd = ''
+        # TODO(nick): remove this inlined script with external templating
+        #             (or find a way of entering the virtualenv that doesn't involve a shell script)
         script_file.write('''
 #!/bin/bash
 set -e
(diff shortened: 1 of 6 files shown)


tools/run_tests.py
```diff
@@ -19,28 +19,28 @@
...
         self.__dcos_url = dcos_url
         self.__sandbox_path = ''
-        self.__sandbox_env = {}
         self.__github_updater = github_update.GithubStatusUpdater('test:{}'.format(github_label))
 
 
-    def __get_cli_sandbox(self):
-        self.__sandbox_path = tempfile.mkdtemp(prefix='ci-test')
-        custom_env = os.environ.copy()
+    def __configure_cli_sandbox(self):
```
mrbrowning: It's probably not in the scope of this diff, but why do we lean so heavily on name-mangled class members in this repo? It's non-idiomatic in a way that kind of obscures readability.

tools/run_tests.py
```diff
@@ -67,50 +67,72 @@ def __download_cli_to_sandbox(self):
...
         for cmd in cmds:
-            subprocess.check_call(cmd.split(' '), env=self.__sandbox_env)
+            subprocess.check_call(cmd.split(' '))
 
 
     def setup_cli(self):
         try:
             self.__github_updater.update('pending', 'Setting up CLI')
-            self.__sandbox_env = self.__get_cli_sandbox()
+            self.__configure_cli_sandbox() # adds to os.environ
```
mrbrowning: Small thing, but if we add linting for py scripts in this repo in the future, PEP8 specifies that end-of-line comments begin two spaces after the last non-comment, non-whitespace character in the line.

tools/run_tests.py
```diff
@@ -67,50 +67,72 @@ def __download_cli_to_sandbox(self):
...
+#!/bin/bash
+set -e
+echo "VIRTUALENV CREATE/UPDATE: {venv_path}"
+virtualenv -p $(which python3) --always-copy {venv_path}
+echo "VIRTUALENV ACTIVATE: {venv_path}"
+source {venv_path}/bin/activate
+echo "REQUIREMENTS INSTALL: {reqs_file}"
+{reqs_cmd}
+echo "SHAKEDOWN RUN: {test_dirs}"
+shakedown --dcos-url "{dcos_url}" --ssh-key-file "" --stdout=all --stdout-inline {test_dirs}
```
mrbrowning: It might be worth separating this out into its own file, and then using a templating engine to fill in the specifics. This would aid maintainability.

tools/run_tests.py
nickbp: I've just been using it to indicate 'no really this is private' -- I haven't really kept up with python idioms since 2.4 or so.

tools/run_tests.py
nickbp: I was thinking of doing something like that but ended up leaving it inlined, as I'm expecting the templates and their params to change a lot in the short term. It felt like having them inlined made it easier to keep the template in line with the input params.

tools/run_tests.py
nickbp: Fixed here and in another instance below.

tools/run_tests.py
nickbp: Talked offline, going to have a bulk rename in a separate PR

tools/run_tests.py
nickbp: Added TODOs to remove the inlined scripts (either with external templating, or with a better method to invoke the virtualenv)

mrbrowning:
Looks good to me!

PR #118, opened 2016-09-08
API server

gabrielhartmann:
Provides a Jetty server which can serve Resources, in particular the resources provided by default by the DefaultScheduler.

Configuration is not yet handled by the DefaultScheduler, and so is omitted.


build.gradle
```diff
@@ -157,7 +160,9 @@ dependencies {
     compile "org.json:json:${jsonVer}"
     compile "com.google.inject:guice:${guiceVer}"
     compile "com.google.inject.extensions:guice-assistedinject:${guiceVer}"
-    compile "org.glassfish.jersey.containers:jersey-container-jetty-http:2.23.2"
+    compile "org.glassfish.jersey.containers:jersey-container-jetty-http:${jerseyContainerVer}"
+    compile "org.eclipse.jetty:jetty-servlet:${jettyServletVer}"
+    compile "com.owlike:genson:${gensonVer}"
```
mohitsoni: Why use this vs `Gson` or `Jackson` ? What advantages does Genson offers ? 

My recommendation is to go with projects that are more stable and have good communities behind them.

src/main/java/org/apache/mesos/api/JettyApiServer.java
```diff
@@ -0,0 +1,55 @@
...
+
+import javax.ws.rs.core.UriBuilder;
+import java.net.URI;
+import java.util.Arrays;
+import java.util.Collection;
+
+/**
+ * A JettyApiServer takes a list of POJO JAX-RS Resources and serves them on the indicated port of 0.0.0.0.
+ */
+public class JettyApiServer {
```
mohitsoni: Why create this wrapper ? Bootstrapping and configuring jetty is fairly trivial. By creating this wrapper layer, we are taking away all the native config options our developers can use.

src/main/java/org/apache/mesos/api/JettyApiServer.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public JettyApiServer(int port, Collection<Object> resources) {
+        ResourceConfig resourceConfig = new ResourceConfig();
+        for (Object resource : resources) {
+            resourceConfig.register(resource);
+        }
+
+        URI baseUri = UriBuilder
+                .fromUri("http://0.0.0.0/")
```
mohitsoni: Why only `http` ? Why limit binding to all interfaces ?

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -124,6 +129,31 @@ private void initializeRecoveryScheduler() {
                 recoveryStatusRef);
     }
 
+    private void initializeApi() {
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                JettyApiServer apiServer = new JettyApiServer(
```
mohitsoni: This is limiting. What if the framework/service author wants to add more resources, other than {Plan, Recovery, State}Resource ?

src/test/java/org/apache/mesos/testutils/NetworkTestUtils.java
```diff
@@ -0,0 +1,16 @@
+package org.apache.mesos.testutils;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+/**
+ * This class provides utilities useful for all tests which may have Network related needs.
+ */
+public class NetworkTestUtils {
+    public static int getRandomPort() throws IOException {
```
mohitsoni: Why throw IOException that's thrown by `serverSocket.close()` ? Just silence it, you already got your port.

src/test/java/org/apache/mesos/testutils/NetworkTestUtils.java
```diff
@@ -0,0 +1,16 @@
+package org.apache.mesos.testutils;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+/**
+ * This class provides utilities useful for all tests which may have Network related needs.
+ */
+public class NetworkTestUtils {
```
mohitsoni: I'm a bit skeptical about having a `NetworkTestUtils` for a single use-case. Not completely against it, but keeping our ongoing focus on code-quality and simplicity in mind, just wanted to keep code-base real tight and only generalize something when it's required at more than 2 places.

build.gradle
gabrielhartmann: The advantage was that it worked, when others did not.  I can hunt around for something that works from someone else.

src/main/java/org/apache/mesos/api/JettyApiServer.java
gabrielhartmann: Only `http` is necessary to expose the endpoints through admin router.  You would prefer `https` as well?  Investigation seems to indicate this would be non-trivial, and I don't understand the benefit.  These endpoints are by design publicly available through the admin router.  Can you clarify?

You would prefer to have the ip address configurable?

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
gabrielhartmann: This is the DefaultScheduler, and so provides the default APIs.  Further, more flexible implementations of the Scheduler interface are possible.  This isn't one of them.  As you say setting up a Jetty Server is easy.  If service authors wish to server other elements they can do so in that way.

src/test/java/org/apache/mesos/testutils/NetworkTestUtils.java
gabrielhartmann: Why unnecessarily silence any exceptions in Test code?

src/test/java/org/apache/mesos/testutils/NetworkTestUtils.java
gabrielhartmann: Why more than 2 places and not 1?  How would you solve the multiple uses of this functionality without factoring it out?

build.gradle
mohitsoni: I'm surprised Gson and Jackson doesn't work with Jersey. Do you remember the exact issue ? Let's work together to identify the problem, and solve it. I'll be surprised if Gson and Jackson doesn't work for this widely practiced use-case.

gabrielhartmann: I'll figure it out.  Hard to test, as `dcos-commons` unit tests work due to, I believe, test testCompile elements bringing in needed dependencies.

src/main/java/org/apache/mesos/api/JettyApiServer.java
mohitsoni: Think Protobuf helpers and py.test wrappers, and why we dislike them, and  why we ultimately removed Protobuf helpers ? And, want to get rid of py.text wrappers.

Config options like, but not limited to:

- Thread pool size
- Timeout
- Buffer size
- logging, etc

I don't want the users of this library to be limited from not being able to configure these operationally important things in there own code-base that they control and own.

gabrielhartmann: This is to serve the default APIs for Plans, Recovery, State (and eventually Config).  That's why it's called an API server.  How do you propose serving those Resources?  Are you saying we shouldn't serve them at all and we should leave that to Service authors?

src/main/java/org/apache/mesos/api/JettyApiServer.java
mohitsoni: I want to be very careful in what features we limit and why ? Maybe for this specific one we can for the most part. But, why limit this feature under a layer for no good reason.

gabrielhartmann: Huh.  I have the opposite view.  I want to be careful about what is exposed.  Why expose users to a decision when not necessary?  If they don't have to make a decision or answer a question in order to get the outcome they want, all the better.

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
mohitsoni: DefaultScheduler should schedule, and not bootstrap API interfaces. These are two separate concerns, and shouldn't be coupled to each other.

I would rather provide a DefaultSchedulerResourceRegistry or something similar, which provides a `List` of resources that default scheduler provides, and let the framework/service author use those in-addition to service/framework specific resources while bootstrapping Jetty and Jersey.

gabrielhartmann:
@mohitsoni:  Ping

src/main/java/org/apache/mesos/api/JettyApiServer.java
mohitsoni: @gabrielhartmann I understand the intent, I am just trying to understand it's implications.

If a service author would like to expose resources other than the default provided ones like a FooResource and a BarResource, how are they going to do that ? Will they need to spin up another HTTP server within the same process listening on a different port ?

gabrielhartmann: @mohitsoni:  Ultimately, it's up to the service author.  If this simple Jetty server is sufficient for their needs, they can use it.  If it's not sufficient for their needs, they can use something else.  Providing a general purpose HTTP server is far outside the scope of this PR, or any PR really, as there are many general purpose HTTP servers available, including Jetty.

The author's choices would boil down to whether they want to use Default implementations of interfaces (which are not yet fully defined).  They could not use the DefaultService implementation.  They could not use the DefaultScheduler implementation.  If either of these layers are customized then the author would be free to use their own HTTP server.

The Default implementations need something, this is at that something.  The SDK is not making an opinionated statement on how HTTP resources should be served.  It is providing something, rather than nothing.

mohitsoni: True. I believe if a service author provides there own `Service` implementation, then they can still use a single Jetty instance to serve up `DefaultScheduler` provided resources and any other custom resources. 
I think that's acceptable.

src/main/java/org/apache/mesos/api/JettyApiServer.java
mohitsoni: We can iterate over in future. Let's see if it even comes up :)

gabrielhartmann:
Great.  Thanks @mohitsoni

PR #119, opened 2016-09-09
Add first draft of Stateful Service README

gabrielhartmann:


gabrielhartmann:
@NimaVaziri:  Any comments on this?

NimaVaziri:
As discussed previously, since the task command is not as simple as its resource specifications, it would be useful to either link to how it's constructed or at the very least mention the minimally required fields of the CommandInfo and why they're needed.

gabrielhartmann:
@NimaVaziri:  [Comment addressed here.](https://github.com/mesosphere/dcos-commons/pull/119/files#diff-4b68ea8c4b30924e190a021d01748b8cR108)

NimaVaziri:
nice, LGTM!

PR #121, opened 2016-09-12
Provide the ability to customize Commands on a per Task basis.

gabrielhartmann:


src/main/java/org/apache/mesos/specification/TaskTypeSpecification.java
```diff
@@ -1,8 +1,16 @@
...
+import org.apache.mesos.Protos;
+
+import java.util.Collection;
+
 /**
- * A TaskTypeSpecification describes the count of a particular TaskSpecification which should be launched.
+ * A TaskTypeSpecification describes a Type of Task and how many should be launched.
  */
-public interface TaskTypeSpecification extends TaskSpecification {
+public interface TaskTypeSpecification {
```
nickbp: Why does the id need to be passed into just the command function? Would name/resources/volumes need it too?

Feels like someone could have any id/type handling inside of their implementation, though that implies the developer owns the creation of TaskTypeSpecification objects.

gabrielhartmann: Name is [taken care of here](https://github.com/mesosphere/dcos-commons/blob/master/src/main/java/org/apache/mesos/specification/DefaultPhaseSpecification.java#L33) or through other user implementations, although I don't know why any service would ever want to be opinionated on the matter of `Task` name.

I don't see any reason `Resources` would need ID specifiable resources.  It's sort of the defining characteristic of a `TaskType` that all instances of it consume the same `Resources`.  For Volumes it's possible that a per `Task` instance volume container-path might be desirable, but since each is running in their own container (Executor) I don't see it as a necessity.  So far of the services we've encountered the only necessary change is on a Command basis, and none of the services have any special preference regarding `Task` name.

Make sense?

gabrielhartmann: Oh, I didn't address the final point.  The developer definitely owns the creation of `TaskTypeSpecification` objects.  Without ID input, it's not possible, as far as I can see to turn `TaskTypeSpecification`s into `TaskSpecification`s.  Open to ideas.

gabrielhartmann:
@nickbp:  I'm of two minds about the change to exposing both name and type name.  It seems to expose some complexity that wasn't absolutely necessary.  Users were getting Task naming for free, although they still do if they use default implementations.  In any case, this is more consistent.  I think in the end it's better.

src/main/java/org/apache/mesos/specification/TaskTypeSpecification.java
nickbp: I'm picturing someone accomplishing per-task handling via the construction of their task specification. Something like this:

```
public CustomTask implements Task {
  /** My custom constructor which does any per-task customization */
  public CustomTask(int index) {
    this.command = MyCustomLogic.getCommandFor(index);
  }
  [...]
  public String getCommand() {
    return this.command;
  }
}
```

Then the developer is free to make any per-instance customizations via that construction. Would this make sense, or is there something else going on here?

gabrielhartmann: This looks like you don't define a command at the `TaskType` layer.  Is that the intention?

nickbp:
Talked offline about the index handling, LGTM! Resyncing to master then should be good to go.

PR #126, opened 2016-09-13
fix stacktrace

mgummelt:
cc @nickbp


nickbp:
Sorry about that, got a similar fix pending here: https://github.com/mesosphere/dcos-commons/pull/125

mgummelt:
np, already merged this though :)

PR #128, opened 2016-09-14
add S3_URL and ARTIFACT_DIR to ci_upload.py

mgummelt:
cc @nickbp 

to support using ci_upload.py for production upload

I'd also like it if we can start trying to maintain relatively pep8-compliant python code (line breaks at 80, etc.)

### DIFF
--- a/tools/ci_upload.py
+++ b/tools/ci_upload.py
@@ -1,4 +1,11 @@
 #!/usr/bin/python
+#
+# Env:
+#   S3_BUCKET (default: infinity-artifacts)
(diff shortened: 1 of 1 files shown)


nickbp:
Maybe take a look at https://github.com/mesosphere/dcos-commons/blob/master/tools/release_builder.py if you're looking at doing releases. It'll also create a universe PR for you with a comment showing the diff vs what was previously there.

See eg https://github.com/mesosphere/universe/pull/653

mgummelt:
I don't want it to create a universe PR for me, but if I can make that optional, then I might use it.  I'll have to take a look.

Issue #134, opened 2016-09-15
Upload library to Maven Central Repository


activatedgeek:
I need to build upon this library but it looks like it is not available on the Maven Central Repository. I think it would be great to have this library released as a versioned JAR artifact rather than cloning and building manually.

nickbp:
We've been uploading artifacts to a maven repository at `downloads.mesosphere.com`. You should be able to configure your project to use that, with something like these examples (gradle configs):
- [dcos-cassandra-service](https://github.com/mesosphere/dcos-cassandra-service/blob/master/build.gradle#L41)
- [dcos-kafka-service](https://github.com/mesosphere/dcos-kafka-service/blob/master/build.gradle#L52)

activatedgeek:
Oh perfect. Sorry about the confusion. I think it is worth mentioning in the README. :) Thanks!

PR #136, opened 2016-09-15
Added usage via Maven

activatedgeek:
Updated README with a gradle config to be included in dependent projects. Addresses #134.


nickbp:
ok to test

PR #138, opened 2016-09-16
Fix failure to ignore TaskStatuses when the DefaultBlock is Complete

gabrielhartmann:


gabrielhartmann:
@mrbrowning:  Actually let me add a unit test.

PR #143, opened 2016-09-19
Remove Optional portion of updateOfferStatus

gabrielhartmann:
The Optional portion adds no value.  An empty Collection communicates
the same meaning.


gabrielhartmann:
@nickbp:  You correctly requested a similar change regarding Optionals and Collections in a previous PR.

PR #149, opened 2016-09-21
Plan admin tasks are POST requests

nickbp:
Compare to https://github.com/mesosphere/dcos-commons/blob/6a5e665c56dcb9eb478b89619453f4f5e0fdbad2/src/main/java/org/apache/mesos/scheduler/plan/api/PlanResource.java#L45


nickbp:
ping

PR #152, opened 2016-09-22
Nits for DefaultService

nickbp:
See added note about 'user' field in class javadoc. Is that accurate?

### DIFF
--- a/src/main/java/org/apache/mesos/specification/DefaultService.java
+++ b/src/main/java/org/apache/mesos/specification/DefaultService.java
@@ -32,7 +32,7 @@ public class DefaultService implements Service {
     private ServiceSpecification serviceSpecification;
 
     /**
-     * Creates and registers the service with Mesos, while starting a Jetty HTTP API service on the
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/mesos/specification/DefaultService.java
```diff
@@ -15,59 +15,72 @@
...
     private static final String MASTER_URI = "zk://master.mesos:2181/mesos";
     private static final String USER = "root";
-    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultService.class);
     private final int apiPort;
     private StateStore stateStore;
     private ServiceSpecification serviceSpecification;
 
+    /**
+     * Creates and registers the service with Mesos, while starting a Jetty HTTP API service on the
```
gabrielhartmann: I think you accidentally copy/pasted comments.  The constructor doesn't `Creates and registers etc.`

nickbp: Fixed

PR #154, opened 2016-09-23
Fix issues caused by destructively killing task with TaskKiller

xtenzo:
This fixes some issues found while using TaskKiller.killTask with destructive set to true.
1. "permanently-failed" attribute was not taken into account  by recovery scheduler.
2. Resource cleaner could use offers previously accepted by recovery  scheduler.

### DIFF
--- a/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
+++ b/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
@@ -86,19 +86,19 @@ public class DefaultScheduler implements Scheduler {
         return plan;
     }
 
-    private void initialize() throws InterruptedException {
+    private void initialize(SchedulerDriver driver) throws InterruptedException {
         logger.info("Initializing.");
-        initializeGlobals();
+        initializeGlobals(driver);
         initializeRecoveryScheduler();
         initializeDeploymentPlan();
         initializeResources();
     }
 
-    private void initializeGlobals() {
+    private void initializeGlobals(SchedulerDriver driver) {
         logger.info("Initializing globals");
         stateStore = new CuratorStateStore(serviceSpecification.getName(), zkConnectionString);
         taskFailureListener = new DefaultTaskFailureListener(stateStore);
-        taskKiller = new DefaultTaskKiller(stateStore, taskFailureListener);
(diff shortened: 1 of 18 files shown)


gabrielhartmann:
@xtenzo:  Thanks a lot for the PR.  Looks good.  Could you please write a unit test which fails without this change, and which passes with this change?

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -279,7 +279,8 @@ public void run() {
                 // UNRESERVE and DESTROY (in the case of persistent volumes) Operations.
                 ResourceCleanerScheduler cleanerScheduler = getCleanerScheduler();
                 if (cleanerScheduler != null) {
-                    acceptedOffers.addAll(getCleanerScheduler().resourceOffers(driver, offers));
+                    unacceptedOffers = filterAcceptedOffers(unacceptedOffers, acceptedOffers);
```
gabrielhartmann: Why do we need to filter Offers here?  The `ResourceCleanerScheduler` only `UNRESERVE`s and `DESTROY`s unexpected resources and volumes.  I think I've considered doing this in the past but couldn't convince myself that it mattered whether the Offers had been accepted or not.

xtenzo: This fixes an error `W0926 08:37:13.355495 26116096 sched.cpp:1348] Attempting to accept an unknown offer 30b17bfc-685c-49a8-8fe5-e1c3fa6cf188-O7` when cleaner tries to unreserve resources and destroy volume.

This is what happens without the fix:
Scheduler receives an offer with both unreserved and reserved (remaining from the failed task) resources. Recovery scheduler accepts this offer to reserve resources and launch a new task instead of the failed one. It does not use reserved resources, because the task is marked as permanently failed. Next, resource cleaner detects unused reserved resources in the same offer, and accepts it with `UNRESERVE` and `DESTROY` operations, however it fails with the above message. Then cleaner successfully repeats operations on the next `resourceOffers`.

nickbp:

nickbp:

gabrielhartmann:
@xtenzo:  Thanks a lot for the PR and explanation.

PR #155, opened 2016-09-23
Reindent Offer<Evaluator/Requirement>[Test], fix lint warnings

nickbp:
Intentionally no changes to functionality.


gabrielhartmann:
@nickbp:  Looks good to me.  How did you find all this?  Is it something that could be checked in such that `gradle` would catch deviations on build?

nickbp:
Indent was just found by being in these files for placement constraints. Figured it'd make sense to frontload the reindent operation in a separate PR while I was in there..

PR #156, opened 2016-09-23
Executortask

gabrielhartmann:
1. Allow [`ProcessTasks`](https://github.com/mesosphere/dcos-commons/pull/156/files#diff-b86b0d7e63b3fef82fed54c8cf3f66f7R46) to be created with ProcessBuilder objects.
2. Remove Onregistration and Onre-registration stuff that was never used.


src/main/java/org/apache/mesos/executor/CustomExecutor.java
```diff
@@ -162,7 +128,8 @@ public void killTask(ExecutorDriver driver, Protos.TaskID taskId) {
                 return;
             }
             LOGGER.info("Stopping task as part of killTask: {}", taskId);
-            launchedTasks.get(taskId).stop();
+            final LaunchedTask launchedTask = launchedTasks.get(taskId);
+            launchedTask.getExecutorTask().stop(launchedTask.getFuture());
```
nickbp: this API feels a little weird: passing launchedTask's future back to a member of launchedTask

src/main/java/org/apache/mesos/executor/CustomExecutor.java
```diff
@@ -86,23 +63,12 @@ public void launchTask(final ExecutorDriver driver, final Protos.TaskInfo task)
...
-                throw new ExecutorTaskException("Unable to determine task type: " + taskEnv);
-            }
-
-            final String taskType = taskEnv.get(DcosTaskConstants.TASK_TYPE);
-            final ExecutorTask taskToExecute = executorTaskFactory.createTask(taskType, task, driver);
-            executorService.submit(taskToExecute);
-            scheduleHealthCheck(task, taskToExecute);
-            launchedTasks.put(task.getTaskId(), taskToExecute);
+            Future<?> future = executorService.submit(taskToExecute);
+            LaunchedTask launchedTask = new LaunchedTask(taskToExecute, future);
```
nickbp: Could pass the executor service into LaunchedTask and have it create the future internally.

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -295,6 +297,27 @@ public static boolean areDifferent(TaskSpecification oldTaskSpecification, TaskS
         return false;
     }
 
+    public static String getTaskType(TaskInfo taskInfo) throws InvalidProtocolBufferException {
+        final Protos.CommandInfo commandInfo = getCommand(taskInfo);
+        final Map<String, String> envMap = TaskUtils.fromEnvironmentToMap(commandInfo.getEnvironment());
+        return envMap.get(DcosTaskConstants.TASK_TYPE);
```
nickbp: Check for null type due to unset in map?

src/main/java/org/apache/mesos/executor/CustomExecutor.java
gabrielhartmann: I generally don't like having construction of an object imply starting or execution, butI could do that.  It seems a little weird to create something called `LaunchedTask` which at creation launches itself?

gabrielhartmann:
@nickbp: PR looks screwy.  I made changes addressing 2 of 3 comments, but it's not reflected above...

src/main/java/org/apache/mesos/executor/CustomExecutor.java
nickbp: Good point, carry on then

PR #157, opened 2016-09-23
[WIP] Observer pattern

mgummelt:


src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
```diff
@@ -12,7 +15,7 @@
  * their ordering. It is an immutable class that can be constructed either
  * directly, or using a fluent style builder.
  */
-public class DefaultPhase implements Phase {
+public class DefaultPhase extends ChainedObserver implements Phase, Observer {
```
gabrielhartmann: Do you need to write `implements Observer` if you already extend the `ChainedObserver`?

gabrielhartmann:
@mgummelt: LGTM if tests pass.

PR #160, opened 2016-09-26
Reliably, destructively, replace Tasks

gabrielhartmann:
* Tasks which were in the state `TASK_LOST`, weren't being handled properly by destructive killing.  They weren't getting marked as permanently failed because they were being detected as not needing to be killed because they weren't `TASK_RUNNING`.
* There was also a conflation between Tasks being in states which required recovery and being in a terminal state.  These have been clarified.
* Integration tested by having Kafka consume this change, terminated a AWS instance, had a Broker go to TASK_LOST.  `dcos kafka broker replace <id>` worked as expected.  It did not work as expected before this change.
* It also appears that `dcos kafka broker replace <id>` while suppressed was broken.  So the TaskKiller now takes a SchedulerDriver so it can work anytime, even without an Offer cycle.

Note: test coverage goes from 79.3% to 79.7% in this PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -101,7 +101,7 @@ ext {
 }
 
 group = "mesosphere"
-version = "0.7.6-SNAPSHOT"
+version = "0.7.7-SNAPSHOT"
 
 task sourceJar(type: Jar) {
     from sourceSets.main.allJava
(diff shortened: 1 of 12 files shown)


gabrielhartmann:

PR #164, opened 2016-09-27
[IGNOREME] Original Placement Constraints PR

nickbp:
New concepts:
- PlacementRule: A static rule which accepts/declines Offers for the
  purpose of launching a single task.
- PlacementRuleGenerator: A dynamic rule generator which creates a
  PlacementRule for the next task to be launched, based on current
  task deployment.

Note that PlacementRules may internally be composites of other PlacementRules, via AndRule/OrRule/NotRule (which contrary to their names are actually performing set logic on returned Offer Resources).

The provided PlacementRuleGenerator implementations support things like 'avoid (or colocate with) any tasks with this type', and 'don't launch more than N tasks of this type on agents with this attribute'.

Summary of changes to existing dcos-commons:
- Store a string representation of the offer attributes, if any, as
  a label in the TaskInfo. The string representation matches the spec
  provided by Mesos for attribute strings.
- Replace existing avoidAgents/colocateAgents logic with a rules-based
  implementation.
- Replace "taskType + '-' + index" in DefaultTaskSpecification with a
  a new 'toTaskName()' utility function in TaskUtils.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -101,7 +101,7 @@ ext {
 }
 
 group = "mesosphere"
-version = "0.7.7-SNAPSHOT"
+version = "0.7.8-SNAPSHOT"
 
 task sourceJar(type: Jar) {
     from sourceSets.main.allJava
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 // Config shared by the dcos-commons library and the examples:
 
 plugins {
+    id 'com.github.johnrengelman.shadow' version '1.2.3'
     id 'com.github.ksoichiro.console.reporter' version '0.4.0'
 }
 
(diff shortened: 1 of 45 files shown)


mrbrowning:
@nickbp is there a way to break this into smaller logical chunks, with the latter chunks being dependent on the earlier ones? I worry that trying to swallow this all at once might lead to us missing things we might want to change but that we'll be committed to down the line.

nickbp:
Sure, the bulk of the change is the example implementations, I could rip out the extra ones, leaving only what's needed to replace the previous hand-coded avoidAgent/colocateAgent. Other fixes like logging improvements could also be broken out. That should bring it down quite a bit.

mrbrowning:
I would appreciate that mightily, thank you!

nickbp:
Supplanted by https://github.com/mesosphere/dcos-commons/pull/186

nickbp:
Keeping this around for reference for a bit, will close it once the feature is merged via other PRs.

PR #170, opened 2016-09-29
Convert TaskTypeSpecification to TaskSet for more customization flexibility

gabrielhartmann:


src/test/java/org/apache/mesos/specification/CustomTaskSetTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public void testCustomCommand() {
+        TaskSet taskSet = CustomTaskSet.create("custom", 3);
+        Assert.assertEquals("custom 0", taskSet.getTaskSpecifications().get(0).getCommand().getValue());
+        Assert.assertEquals("custom 2", taskSet.getTaskSpecifications().get(2).getCommand().getValue());
+
+        Assert.assertEquals("custom_0", taskSet.getTaskSpecifications().get(0).getName());
+        Assert.assertEquals("custom_2", taskSet.getTaskSpecifications().get(2).getName());
+    }
+
+    public static class CustomTaskSet extends DefaultTaskSet {
```
gabrielhartmann: @nickbp @mrbrowning:  Here's an example of providing your own custom TaskSpecifications.  What do you think?  This change also fixes the weirdness around `DefaultPhaseSpecification` always ending up being composed of `DefaultTaskSpecification`s

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
```diff
@@ -0,0 +1,52 @@
...
+
+        List<TaskSpecification> taskSpecifications = new ArrayList<>();
+        for (int i = 0; i < count; i++) {
+            taskSpecifications.add(new DefaultTaskSpecification(name + "-" + i, command, resources, volumes));
+        }
+
+        return new DefaultTaskSet(name, taskSpecifications);
+    }
+
+    public DefaultTaskSet(String name, List<TaskSpecification> taskSpecifications) {
```
nickbp: Mark this private and add a create() method with this signature (to keep the parallel with the create() method added above)?

Alternate option: Could mark protected instead of private if needed by eg the CustomTaskSet example.

src/main/java/org/apache/mesos/specification/ServiceSpecification.java
```diff
@@ -14,8 +14,8 @@
 
 
     /**
-     * Gets the list of TaskTypeSpecifications defining the types of Tasks of which the Service is made.
+     * Gets the list of TaskSets defining the types of Tasks of which the Service is made.
```
nickbp: Could do `{@link TaskSet}s`

src/main/java/org/apache/mesos/specification/TaskSet.java
```diff
@@ -0,0 +1,11 @@
+package org.apache.mesos.specification;
+
+import java.util.List;
+
+/**
+ * A TaskSet groups a set of Tasks that should be deployed using the same deployment strategy.
```
nickbp: Maybe specify how in practice a TaskSet corresponds to a Phase(correct term?) in Plans, where each TaskSpecification in turn corresponds to a Block.

src/test/java/org/apache/mesos/specification/CustomTaskSetTest.java
nickbp: :+1:

src/main/java/org/apache/mesos/specification/TaskSet.java
mrbrowning: The single-deployment-strategy aspect of a `TaskSet` just falls out of the fact that `PhaseSpecification` contains a `TaskSet`, right? That might be helpful to document so that developers can make that explicit connection.

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
gabrielhartmann: Why not both?

nickbp: Both is good too :)

src/main/java/org/apache/mesos/specification/TaskSet.java
gabrielhartmann: @nickbp and @mrbrowning:  Started looking at `PhaseSpecification` and it turns out to be identical to `TaskSet`.  So it's gone.

PR #171, opened 2016-09-30
Ensure that TaskInfo has command in data field when executor has command set

mrbrowning:
Mesos will not accept a `TaskInfo` that both has its own `CommandInfo` as well as an `ExecutorInfo` with a `CommandInfo`. We end up in violation of this when setting dynamic port requirements on both our tasks and our executors, since those ports are stored as in environment variables on `TaskInfo.CommandInfo` and `TaskInfo.ExecutorInfo.CommandInfo`, respectively. To work around this case generally, we already have the capability of supplying a serialized `CommandInfo` in the `TaskInfo`'s `data` field, so this change causes the `OfferEvaluator` to check for a `TaskInfo` that has both `CommandInfo` fields, and, if so, moves the task's `CommandInfo` to its `data` field.

### DIFF
--- a/src/main/java/org/apache/mesos/offer/ResourceUtils.java
+++ b/src/main/java/org/apache/mesos/offer/ResourceUtils.java
@@ -203,6 +203,17 @@ public class ResourceUtils {
         return resBuilder.build();
     }
 
+    public static TaskInfo.Builder serializeCommandInfo(Protos.TaskInfo.Builder builder) {
+        Protos.CommandInfo command = builder.getCommand();
+
+        if (builder.hasExecutor()) {
+            builder.clearCommand();
+            builder.setData(command.toByteString());
+        }
+
+        return builder;
+    }
+
     public static Resource setDynamicPortName(Resource resource, String name) {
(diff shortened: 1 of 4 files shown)


mrbrowning:

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -203,6 +203,17 @@ public static Resource getExpectedRanges(
         return resBuilder.build();
     }
 
+    public static TaskInfo.Builder serializeCommandInfo(Protos.TaskInfo.Builder builder) {
+        Protos.CommandInfo command = builder.getCommand();
+
+        if (builder.hasExecutor() && builder.getExecutor().hasCommand()) {
```
gabrielhartmann: The second portion of this if clause i.e. `builder.getExecutor().hasCommand()` is not necessary.

gabrielhartmann:
@mrbrowning: One minor comment.  Otherwise LGTM.  When you're happy with the test, go ahead and merge.

PR #173, opened 2016-10-01
Implement a StateStore write-through cache.

nickbp:
Only as consistent as the underlying StateStore implementation.


src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -386,7 +388,8 @@ public void clearProperty(final String key) throws StateStoreException {
         }
     }
 
-    void close() {
+    @VisibleForTesting
```
gabrielhartmann: Neat.

gabrielhartmann:
@nickbp:  It looks like what is missing here is the ability to transactionally store both a TaskInfo and a TaskStatus simultaneously.  We really want this feature so we can guarantee that when a TaskInfo is present in the StateStore, that it also has a TaskStatus (initially TASK_STAGING).

nickbp:
Yes this is just covering caching of the StateStore interface as-is and allow applications to avoid DIYing their own caching. In addition to this there are additional changes to be made in CuratorStateStore to make that more transactional/atomic, plus the earlier discussion about how to (re)initialize TaskStatus on TaskInfo changes and what form that should take

src/main/java/org/apache/mesos/state/StateStoreCache.java
```diff
@@ -0,0 +1,301 @@
...
+        } finally {
+            RLOCK.unlock();
+        }
+    }
+
+    @Override
+    public void storeTasks(Collection<TaskInfo> tasks) throws StateStoreException {
+        RWLOCK.lock();
+        try {
+            store.storeTasks(tasks);
```
mohitsoni: Is it possible that storeTasks method persists tasks collection fully or partially and throws an exception ? If yes, then that particular case will make the cache inconsistent.

src/main/java/org/apache/mesos/state/StateStoreCache.java
```diff
@@ -0,0 +1,301 @@
...
+                        taskName, nameToId, idToTask.keySet()));
+            }
+            return Optional.of(task);
+        } finally {
+            RLOCK.unlock();
+        }
+    }
+
+    @Override
+    public Set<TaskStatus> getTaskStatuses() throws Exception {
```
mohitsoni: Can we de-dup `getTaskStatuses` vs `fetchStatuses` ? Or there's a good reason why both should exist ?

src/main/java/org/apache/mesos/state/StateStoreCache.java
nickbp: Yes. This will only be as consistent as the underlying state store.

mohitsoni: Okay.

src/main/java/org/apache/mesos/state/StateStoreCache.java
nickbp: I agree, don't know why there need to be both. getTaskStatuses() is from TaskStatusProvider which the base StateStore interface extends.

We could either:
- Update TaskStatusProvider interface (Set<TaskInfo>) to align with StateStore interface (List<TaskInfo>)
- Update StateStore interface to align with TaskStatusProvider interface

src/main/java/org/apache/mesos/state/StateStoreCache.java
nickbp: Filed INFINITY-570 for fixing CuratorStateStore

src/main/java/org/apache/mesos/state/StateStoreCache.java
mohitsoni: If there's no good reason for `TaskStatusProvider` to be around, maybe we can remove `TaskStatusProvider` in favor of just `StateStore` ? As `StateStore` provides the required functionality.

src/main/java/org/apache/mesos/state/StateStoreCache.java
mohitsoni: Thanks!

src/main/java/org/apache/mesos/state/StateStoreCache.java
nickbp: Done: Removed `TaskStatusProvider` interface and switched `DefaultReconciler` (its sole user) to use `StateStore` directly.

mohitsoni:
Regarding @gabrielhartmann's comment on transactionally storing TaskInfo and TaskStatus:
@nickbp Are you going to update the CuratorStateStore to support that in this PR or you have planned additional work for it ?

nickbp:
I was planning on just having a cache that works against the current interface, figuring that interface changes to StateStore itself would be done separately.

nickbp:
Ping, any other comments?

PR #174, opened 2016-10-01
Proto construction nitpick

nickbp:


gabrielhartmann:
@nickbp:  Style violation.

nickbp:
green!

PR #180, opened 2016-10-03
Explicitly add TASK_LOST as non-terminal

nickbp:
No effective change to logic (I assume this is correct?), just makes the case more explicit


nickbp:
retest this please

nickbp:
retest this please

PR #182, opened 2016-10-03
Serialize and deserialize TaskInfos with custom Executors and data fields

gabrielhartmann:
TaskInfos may not simultaneously have CommandInfo and ExecutorInfo fields.  That doesn't make sense and there is [a ticket with a shepherd to fix it on the Mesos side](https://issues.apache.org/jira/browse/MESOS-6294).

In the mean time we had been serializing the CommandInfo when a custom Executor is used and putting it in the data field.  However, this overwrites anything that was in the data field.  So one more level of encapsulation is needed.  I am reusing the ExecutorInfo protobuf for this as it's the smallest protobuf with both CommandInfo and data fields.

So, when a TaskInfo with both CommandInfo and ExecutorInfo is encountered, the necessary information is serialized and encapsulated in an ExecutorInfo (in the data field) on the Scheduler side.  The deserialization operation is performed on our CustomExecutor's launchTask side.

### DIFF
--- a/src/main/java/org/apache/mesos/offer/TaskUtils.java
+++ b/src/main/java/org/apache/mesos/offer/TaskUtils.java
@@ -21,6 +21,7 @@ public class TaskUtils {
     private static final Logger LOGGER = LoggerFactory.getLogger(TaskUtils.class);
     private static final String TARGET_CONFIGURATION_KEY = "target_configuration";
     private static final String TASK_NAME_DELIM = "__";
+    private static final String COMMAND_DATA_PACKAGE_EXECUTOR = "command_data_package_executor";
 
     private TaskUtils() {
         // do not instantiate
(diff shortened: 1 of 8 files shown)


src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -309,12 +310,49 @@ public static String getTaskType(TaskInfo taskInfo) throws InvalidProtocolBuffer
         }
     }
 
-    public static CommandInfo getCommand(TaskInfo taskInfo) throws InvalidProtocolBufferException {
-        return Protos.CommandInfo.parseFrom(taskInfo.getData());
+    public static TaskInfo serializeTaskInfo(TaskInfo taskInfo) {
```
mrbrowning: Maybe slightly more explanatory names for this method pair? It's not the `TaskInfo` being serialized, but the attached `CommandInfo`.

gabrielhartmann:
@mrbrowning:  That [test was modified to pass](https://github.com/mesosphere/dcos-commons/pull/182/files#diff-60c23c4e7c4d059d13dbb384f7ecb7d3R166).

src/main/java/org/apache/mesos/offer/TaskUtils.java
nickbp: How about 'withSerializedExecutorInfo'?

gabrielhartmann: @nickbp and @mrbrowning:  I went with pack/unpack and added some comments.  It's not a serialization of ExecutorInfo.  The fact that there's an ExecutorInfo involved at all is an abomination.

PR #184, opened 2016-10-03
Use Optionals in OfferEvaluator/MesosResourcePool

nickbp:
- Adds logging of OfferEvaluator decisions
- Cleans up MesosResourcePool internals a bit, mainly the init logic, while adding Javadocs

### DIFF
--- a/src/main/java/org/apache/mesos/offer/MesosResourcePool.java
+++ b/src/main/java/org/apache/mesos/offer/MesosResourcePool.java
@@ -11,42 +11,65 @@ import org.slf4j.LoggerFactory;
 import java.util.*;
 
 /**
- * A representation of the pool of resources available in a single Offer.
- **/
+ * A representation of the pool of resources available in a single {@link Offer}. Tracks the
+ * consumption of the {@link Offer}'s resources, as they are matched with
+ * {@link ResourceRequirement}s.
+ */
 public class MesosResourcePool {
     private static final Logger logger = LoggerFactory.getLogger(MesosResourcePool.class);
 
-    private Offer offer;
-    private Collection<MesosResource> mesosResources;
-    private Map<String, List<MesosResource>> unreservedAtomicPool;
-    private Map<String, Value> unreservedMergedPool;
-    private Map<String, MesosResource> reservedPool;
+    private final Offer offer;
+    private final Map<String, List<MesosResource>> unreservedAtomicPool;
+    private final Map<String, Value> unreservedMergedPool;
+    private final Map<String, MesosResource> reservedPool;
 
+    /**
+     * Creates a new pool of resources based on what's available in the provided {@link Offer}.
+     */
     public MesosResourcePool(Offer offer) {
         this.offer = offer;
-        this.mesosResources = getMesosResourcesInternal();
-        this.unreservedAtomicPool = getUnreservedAtomicPool(offer);
-        this.unreservedMergedPool = getUnreservedMergedPool(offer);
-        this.reservedPool = getReservedPool(offer);
+        final Collection<MesosResource> mesosResources = getMesosResources(offer);
+        this.unreservedAtomicPool = getUnreservedAtomicPool(mesosResources);
+        this.unreservedMergedPool = getUnreservedMergedPool(mesosResources);
+        this.reservedPool = getReservedPool(mesosResources);
     }
 
+    /**
+     * Returns the underlying offer which this resource pool represents.
+     */
     public Offer getOffer() {
         return offer;
     }
 
+    /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/mesos/offer/MesosResourcePool.java
```diff
@@ -63,14 +86,19 @@ public MesosResource consume(ResourceRequirement resReq) {
...
+     * Consumes and returns a {@link MesosResource} which meets the provided
+     * {@link DynamicPortRequirement}, or does nothing and returns an empty {@link Optional} if no
+     * available resources meet the requirement.
+     */
+    public Optional<MesosResource> consume(DynamicPortRequirement dynamicPortRequirement) {
         Value availableValue = unreservedMergedPool.get(dynamicPortRequirement.getName());
 
         if (availableValue == null) {
-            return null;
+            return Optional.empty();
```
mrbrowning: @gabrielhartmann you mentioned only using `Optional` at interaction boundaries, how do you feel about this? I'm in favor it, for ease of internal development.

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
```diff
@@ -219,7 +251,8 @@ private boolean sufficientValue(Value desired, Value available) {
         return mesRsrcs;
     }
 
-    private Map<String, MesosResource> getReservedPool(Offer offer) {
+    private static Map<String, MesosResource> getReservedPool(
+            Collection<MesosResource> mesosResources) {
```
mrbrowning: Since this method is private, it'll only ever be able to operate on `this.mesosResources` anyway. Is the parameter here for testability? If so, great, otherwise, it could be dropped.

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
```diff
@@ -266,8 +301,32 @@ private boolean sufficientValue(Value desired, Value available) {
...
+        return getUnreservedResources(getAtomicResources(mesosResources));
+    }
+
+    private static Collection<MesosResource> getUnreservedMergedResources(
+            Collection<MesosResource> mesosResources) {
+        return getUnreservedResources(getMergedResources(mesosResources));
+    }
+
+    private static Collection<MesosResource> getUnreservedResources(
+            Collection<MesosResource> mesResources) {
```
mrbrowning: Keep parameter name consistent with the earlier methods as `mesosResources`.

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -74,67 +56,73 @@ private boolean hasExpectedExecutorId(Offer offer, Protos.ExecutorID executorID)
...
+        if (execReq.isPresent()) {
+            execInfo = Optional.of(execReq.get().getExecutorInfo());
+
+            if (execReq.get().desiresResources()
+                    || execInfo.get().getExecutorId().getValue().isEmpty()) {
                 fulfilledExecutorRequirement = FulfilledRequirement.fulfillRequirement(
-                        execReq.getResourceRequirements(),
-                        execReq.getDynamicPortRequirements(),
+                        execReq.get().getResourceRequirements(),
+                        execReq.get().getDynamicPortRequirements(),
```
mrbrowning: `execReq.get()` appears often enough here that it might be worth extracting into its own variable.

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
nickbp: The mesosResources member was removed as it was effectively only being used for initialization within the constructor

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
nickbp: Done

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
nickbp: Done here and throughout

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
nickbp: I'd leaned towards doing it for the public consume() methods as they make the intended behavior a little clearer.

Ended up carrying it over to the private methods where relevant since their return values were effectively passed up through the consume() methods.

gabrielhartmann: fine with me.

PR #187, opened 2016-10-04
Enabling Multi-Plan support via PlanCoordinator

mohitsoni:
Done:

- [x] Default implementation of `PlanCoordinator`
- [x] Implementation of `DefaultRecoveryPlanManager`
- [x] Implementation of `RandomRecoveryStrategy`
- [x] Addition of `PlansResource`
- [x] Deprecation of `PlanResource`
- [x] Refactoring to support all the above
- [x] Unit tests
- [x] Documentation


mohitsoni:
retest this please

mohitsoni:
retest this please

mohitsoni:
retest this please

build.gradle
```diff
@@ -102,7 +102,7 @@ ext {
 }
 
 group = "mesosphere"
-version = "0.7.10-SNAPSHOT"
+version = "0.7.11-SNAPSHOT"
```
mgummelt: Please either keep it at 0.7.10-SNAPSHOT, or do a release of 0.7.10, then bump to 0.7.11-SNAPSHOT

https://docs.google.com/document/d/1S-zNC80CeW3ETHJGpgt22ALiEQtz4otLKYqVWj1nA4I/edit

mohitsoni: Good point. I'll bump this to 0.8.0, as this changes PlanManager interface, and introduces new ones.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinator.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public Collection<Protos.OfferID> processOffers(final SchedulerDriver driver,
+                                                    final List<Protos.Offer> offersToProcess) {
+        final Set<Protos.OfferID> dirtiedOffers = new HashSet<>();
+        final List<Block> dirtiedAssets = new ArrayList<>();
+        final List<Protos.Offer> offers = new ArrayList<>(offersToProcess);
+        for (final PlanManager planManager : planManagers) {
+            try {
+                LOGGER.info("Current plan {} interrupted.", planManager.isInterrupted() ? "is" : "is not");
```
nickbp: Can plans (or plan managers) have a name? Would be useful for debugging. Could then log the name of the plan here and below.

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
```diff
@@ -18,7 +18,7 @@
 
     Optional<Phase> getCurrentPhase();
 
-    Optional<Block> getCurrentBlock();
+    Optional<Block> getCurrentBlock(List<Block> dirtiedAssets);
```
nickbp: Needs javadoc explaining what dirtiedAssets is

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
```diff
@@ -18,7 +18,7 @@
```
nickbp: Is setPlan needed? Feels like it'd be fixed after the PlanManager is created.

src/main/java/org/apache/mesos/scheduler/plan/RandomRecoveryStrategy.java
```diff
@@ -0,0 +1,26 @@
+package org.apache.mesos.scheduler.plan;
+
+import org.apache.commons.collections.CollectionUtils;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Random;
+
+/**
+ * RandomRecoveryStrategy.
```
nickbp: Javadoc

src/main/java/org/apache/mesos/scheduler/plan/api/PlansResource.java
```diff
@@ -0,0 +1,133 @@
...
+        this.planManagers.putAll(planManagers);
+    }
+
+    /**
+     * Returns the status of the currently active Phase/Block.
+     */
+    @GET
+    @Path("/{planName}/status")
+    public Response getStatus(@PathParam("planName") String planName) {
+        if (planManagers.containsKey(planName)) {
```
nickbp: Here and below, could just do 'planManagers.get(planName)' up-front and then check 'manager != null' to determine if planName is valid.

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryBlock.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.mesos.Protos;
+import org.apache.mesos.offer.OfferRequirement;
+import org.apache.mesos.scheduler.plan.DefaultBlock;
+import org.apache.mesos.scheduler.plan.Status;
+import org.apache.mesos.scheduler.recovery.constrain.LaunchConstrainer;
+
+import java.util.Collection;
+
+/**
+ * DefaultRecoveryBlock.
```
nickbp: Javadoc

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
```diff
@@ -0,0 +1,321 @@
...
+import org.apache.mesos.specification.DefaultTaskSpecification;
+import org.apache.mesos.specification.InvalidTaskSpecificationException;
+import org.apache.mesos.state.StateStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+/**
+ * An implementation of {@code PlanManager} that performs task recovery using dynamically generated {@code Plan}.
```
nickbp: Could mention that the dynamic generation is based on finding tasks with failed status.

src/main/java/org/apache/mesos/scheduler/recovery/monitor/TestingFailureMonitor.java
```diff
@@ -25,6 +26,7 @@ public void setFailedList(TaskInfo... failed) {
         failedList = new ArrayList<>();
         for (TaskInfo info : failed) {
             failedList.add(info);
+            FailureUtils.markFailed(info);
```
nickbp: markFailed doesn't update in-place

build.gradle
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -81,65 +79,73 @@ void awaitTermination() throws InterruptedException {
...
         Collection<Object> resources = new ArrayList<>();
-        resources.add(new PlanResource(planManager));
-        resources.add(new RecoveryResource(recoveryStatusRef));
+        resources.add(new PlanResource(deployPlanManager));
+        resources.add(new PlansResource(ImmutableMap.of(
+                "deploy", deployPlanManager,
+                "recovery", recoveryPlanManager)));
         resources.add(new StateResource(stateStore));
         resourcesQueue.put(resources);
+        logger.info("Done initializing resources.");
```
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
mohitsoni: There are three major themes:
1. External entity generates a Plan, and instantiates PlanManager with it (DeployManager). And, this plan doesn't change ever for the life-cycle of PlanManager. `setPlan` is not needed.
2. PlanManager generates it's own Plan (SimpleRecoveryPlanManager). `setPlan` is not needed.
3. An external entity generates plan at various points, and provides PlanManager with a new one whenever a new one is generates. (No such implemenation yet). `setPlan` is needed. 

Because we don't have use-case 3 right now, I would say we don't need `setPlan`. Thoughts ?

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/plan/api/PlansResource.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryBlock.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/recovery/monitor/TestingFailureMonitor.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -81,63 +79,67 @@ void awaitTermination() throws InterruptedException {
     }
 
     Plan getPlan() {
```
gabrielhartmann: `getDeployPlan`

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -81,63 +79,67 @@ void awaitTermination() throws InterruptedException {
...
                 taskFailureListener,
                 recoveryRequirementProvider,
-                offerAccepter,
                 constrainer,
-                new TimedFailureMonitor(Duration.ofSeconds(PERMANENT_FAILURE_DELAY_SEC)),
-                recoveryStatusRef);
+                new TimedFailureMonitor(Duration.ofSeconds(PERMANENT_FAILURE_DELAY_SEC)));
     }
 
     private void initializeDeploymentPlan() {
```
gabrielhartmann: This initializes both the `DeploymentPlan` and the `DeploymentPlanManager`.  Maybe they should be separated or the name should be changed.  It also makes `initialize()` confusing because we intialize the recovery manager, but just the deployment plan.  I think breaking it up a bit makes the initialize method clearer, but up to you.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinator.java
```diff
@@ -0,0 +1,83 @@
...
+                final Optional<Block> currentBlock = planManager.getCurrentBlock(dirtiedAssets);
+                if (currentBlock.isPresent()) {
+                    final Block blockToSchedule = currentBlock.get();
+                    LOGGER.info("Current block to schedule: {}", blockToSchedule.getName());
+                    List<Protos.OfferID> usedOffers = planScheduler.resourceOffers(driver, offers, blockToSchedule);
+                    dirtiedOffers.addAll(usedOffers);
+
+                    // If an offer was used, the block was also scheduled, so let's mark it dirty.
+                    if (CollectionUtils.isNotEmpty(usedOffers)) {
```
gabrielhartmann: This is the correct logic for now, but ideally PlanManagers should be more explicit about what Assets they've dirtied.  The PlanCoordinator shouldn't be determining this by implication.  [INFINITY-596](https://mesosphere.atlassian.net/browse/INFINITY-596)

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanManager.java
```diff
@@ -61,7 +60,7 @@ public Plan getPlan() {
     }
 
     @Override
-    public Optional<Block> getCurrentBlock() {
+    public Optional<Block> getCurrentBlock(List<Block> dirtiedAssets) {
```
gabrielhartmann: I would expect that in the `DefaultPlanManager` the default behavior is to treat any dirtied Asset as indicating that it should not be worked on by another `PlanManager`.  So I would expect us to return `Optional.empty()` here if the Block we'd ordinarily like to return is in the `dirtiedAssets` list.

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
gabrielhartmann: I agree with Mohit.

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
```diff
@@ -0,0 +1,325 @@
...
+    @VisibleForTesting
+    protected synchronized void refreshPlan(List<Block> dirtiedAssets) {
+        Map<String, Protos.TaskInfo> recoveryCandidates = getRecoveryCandidates(getTerminatedTasks(dirtiedAssets));
+
+        final DefaultPhase.Builder newRecoveryPhaseBuilder = DefaultPhase.builder();
+        newRecoveryPhaseBuilder.setId(phaseId);
+
+        final OfferRequirementProvider offerRequirementProvider = new DefaultOfferRequirementProvider();
+
+        // If recovery candidates are not part of plan, generate a recovery block for them and mark it as pending.
```
gabrielhartmann: `... are not part of a different PlanManager's plan, ...`

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
mohitsoni: Good that you brought this one up. I was debating between:
1. `getDeployPlan`
2. `getPlans` (which would query PlanManagers, collect and return Plans)

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinator.java
```diff
@@ -0,0 +1,83 @@
...
+                    dirtiedOffers);
+            offers.clear();
+            offers.addAll(unacceptedOffers);
+        }
+
+        return dirtiedOffers;
+    }
+
+    @VisibleForTesting
+    protected static List<Protos.Offer> filterAcceptedOffers(List<Protos.Offer> offers,
```
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinator.java
mohitsoni: Okay.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinator.java
```diff
@@ -0,0 +1,83 @@
...
+    public DefaultPlanCoordinator(List<PlanManager> planManagers, PlanScheduler planScheduler) {
+        if (CollectionUtils.isEmpty(planManagers)) {
+            throw new IllegalArgumentException("Atleast one plan manager is required");
+        }
+        this.planManagers.addAll(planManagers);
+        this.planScheduler = planScheduler;
+    }
+
+    @Override
+    public Collection<Protos.OfferID> processOffers(final SchedulerDriver driver,
```
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -229,59 +231,36 @@ public void reregistered(SchedulerDriver driver, Protos.MasterInfo masterInfo) {
...
+        executor.execute(() -> {
+            logOffers(offers);
+
+            // Task Reconciliation:
+            // Task Reconciliation must complete before any Tasks may be launched.  It ensures that a Scheduler and
+            // Mesos have agreed upon the state of all Tasks of interest to the scheduler.
+            // http://mesos.apache.org/documentation/latest/reconciliation/
+            reconciler.reconcile(driver);
+            if (!reconciler.isReconciled()) {
+                logger.info("Accepting no offers: Reconciler is still in progress");
```
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
mohitsoni: Makes sense. Done!

src/main/java/org/apache/mesos/scheduler/plan/RandomRecoveryStrategy.java
mohitsoni: Done!

mohitsoni:
retest this please

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
gabrielhartmann: I think `getPlans` is better as long as one can distinguish between plans by name as requested by @nickbp

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
```diff
@@ -0,0 +1,325 @@
...
+        LOGGER.info("Ignoring status: {}", TextFormat.shortDebugString(status));
+    }
+
+    @Override
+    public boolean hasDecisionPoint(Block block) {
+        return phaseStrategy != null && phaseStrategy.hasDecisionPoint(block);
+    }
+
+    @Override
+    public Status getStatus() {
```
gabrielhartmann: Is this implementation copy/pasted from elsewhere?  Let's factor it out if it is.  It's pretty much _the_ implementation for determining Plan status.

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
```diff
@@ -0,0 +1,325 @@
...
+            LOGGER.info("(status={}) At least one phase has status '{}' and one has status '{}'",
+                    result, Status.COMPLETE, Status.PENDING);
+        } else {
+            result = null;
+            LOGGER.error("(status={}) Unexpected state. Plan: {}", result, plan);
+        }
+        return result;
+    }
+
+    public boolean allHaveStatus(Status status, Plan plan) {
```
gabrielhartmann: Same as above.  I think this has already been implemented elsewhere.

src/test/java/org/apache/mesos/scheduler/DefaultSchedulerTest.java
```diff
@@ -429,8 +429,8 @@ public void testOfferAcceptedOnce() throws Exception {
         defaultScheduler.resourceOffers(mockSchedulerDriver, Arrays.asList(offer));
         defaultScheduler.awaitTermination();
 
-        // Verify scheduler accepted resources only once
-        verify(mockSchedulerDriver, times(1)).acceptOffers(any(), any(), any());
+        // Verify scheduler accepted resources twice, once for scheduling and once for recovery.
+        verify(mockSchedulerDriver, times(2)).acceptOffers(any(), any(), any());
```
gabrielhartmann: Now that we're accepting twice, you may want to change the name of this test.  `testLaunchAndRecovery`?  Almost anything else is better.

src/test/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinatorTest.java
```diff
@@ -0,0 +1,296 @@
...
+                        ResourceTestUtils.getUnreservedDisk(disk))).get(0))
+                .setId(Protos.OfferID.newBuilder().setValue("other-offer"))
+                .build());
+        return offers;
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testNoPlanManager() {
+        final DefaultPlanCoordinator coordinator = new DefaultPlanCoordinator(Arrays.asList(), planScheduler);
+        coordinator.processOffers(schedulerDriver, getOffers(1, 1, 1));
```
gabrielhartmann: Is this line necessary?  This should fail on construction.

src/test/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinatorTest.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    @Test
+    public void testOnePlanManagerPendingSufficientOffer() throws Exception {
+        when(stateStore.fetchTask(anyString())).thenReturn(Optional.empty());
+        final Plan plan = new DefaultPlanFactory(stateStore).getPlan(serviceSpecification);
+        final DefaultPlanCoordinator coordinator = new DefaultPlanCoordinator(
+                Arrays.asList(new DefaultPlanManager(plan, new DefaultStrategyFactory())),
+                planScheduler);
+        Assert.assertTrue(coordinator.processOffers(schedulerDriver, getOffers(2, 2000, 10000)).size() == 1);
```
gabrielhartmann: Can we rename these constants to stuff like `SUFFICENT_CPU` and `INSUFFICIENT_MEM`.

src/test/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinatorTest.java
```diff
@@ -0,0 +1,296 @@
...
+    @Test
+    public void testTwoPlanManagersPendingPlansDisjointAssets() throws Exception {
+        when(stateStore.fetchTask(anyString())).thenReturn(Optional.empty());
+        final Plan planA = new DefaultPlanFactory(stateStore).getPlan(serviceSpecification);
+        final Plan planB = new DefaultPlanFactory(stateStore).getPlan(serviceSpecificationB);
+        final DefaultPlanManager planManagerA = new DefaultPlanManager(planA, new DefaultStrategyFactory());
+        final DefaultPlanManager planManagerB = new DefaultPlanManager(planB, new DefaultStrategyFactory());
+        final DefaultPlanCoordinator coordinator = new DefaultPlanCoordinator(
+                Arrays.asList(planManagerA, planManagerB), planScheduler);
+        Assert.assertEquals(2, coordinator.processOffers(schedulerDriver, getOffers(2, 2000, 10000)).size());
```
gabrielhartmann: I prefer this style of assertion, but please just choose one or the other.  You were using `assertTrue` on an equality expression before.

src/test/java/org/apache/mesos/scheduler/plan/RandomRecoveryStrategyTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Test
+    @edu.umd.cs.findbugs.annotations.SuppressWarnings(value="RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT")
+    public void testGetCurrentBlockSingleBlock() {
+        Phase phase = mock(Phase.class);
+        final Block mock = mock(Block.class);
+        final List<? extends Block> blocks = Arrays.asList(mock);
+        Mockito.doReturn(blocks).when(phase).getBlocks();
+        final RandomRecoveryStrategy randomRecoveryStrategy = new RandomRecoveryStrategy(phase);
+        Assert.assertEquals(true, randomRecoveryStrategy.getCurrentBlock().isPresent());
```
gabrielhartmann: assertTrue

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
mohitsoni: See [INFINITY-597](https://mesosphere.atlassian.net/browse/INFINITY-597)

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanManager.java
```diff
@@ -61,10 +61,34 @@ public Plan getPlan() {
...
```
gabrielhartmann: Should take a `Collection<Block>`

src/test/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManagerTest.java
```diff
@@ -0,0 +1,472 @@
...
+        when(mockDeployManager.isInterrupted()).thenReturn(false);
+        final DefaultPlanScheduler planScheduler = new DefaultPlanScheduler(offerAccepter,
+                new OfferEvaluator(stateStore),
+                new DefaultTaskKiller(stateStore, taskFailureListener, schedulerDriver));
+        planCoordinator = new DefaultPlanCoordinator(Arrays.asList(mockDeployManager, recoveryManager),
+                planScheduler);
+    }
+
+    @After
+    public void teardownTest() {
```
gabrielhartmann: Why is this method necessary as they're re-mocked before every test.

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
mohitsoni: This is slightly special:
`} else if (plan == null || CollectionUtils.isEmpty(plan.getPhases())) {` (SimpleRecoveryPlanManager)

vs

`} else if (plan.getPhases().isEmpty()) {` (DefaultPlanManager)

But, I think `SimpleRecoveryPlanManager` version would also work for `DefaultPlanManager`.

Let me move this to `PlanManager` interface as the default impl for `getStatus()` method.

gabrielhartmann:
@mohitsoni:  I'm done with my review.  Looks very good to me.  You've already addressed the couple things that I thought were substantive.  The rest is mostly style.  Would like at least one more sign off from someone else on the team, then feel free to merge.

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
```diff
@@ -8,7 +8,7 @@
 import java.util.UUID;
 
 /**
- * Interface for PlanManagers.
+ * PlanManager is the management interface for {@code Plan}s.
```
nickbp: Could do `{@link Plan}s`

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
```diff
@@ -18,7 +18,14 @@
 
     Optional<Phase> getCurrentPhase();
 
-    Optional<Block> getCurrentBlock();
+    /**
+     * Determines the next block that should be considered for scheduling. Blocks that are being selected by other
+     * PlanManager are provided as {@code dirtiedAssets} as a hint to this PlanManager to assist with scheduling.
+     *
+     * @param dirtiedAssets Other blocks/assets that are already selected for scheduling.
+     * @return An Optional Block that can be scheduled.
```
nickbp: Could specify the situation where the return value may be empty, something like `... that can be scheduled, or an empty Optional when [situation]`

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
```diff
@@ -0,0 +1,325 @@
...
+import java.util.*;
+
+/**
+ * {@link SimpleRecoveryPlanManager} enables monitoring and management of recovery plan.
+ *
+ * This is an implementation of {@code PlanManager} that performs task recovery using dynamically generated
+ * {@code Plan}. {@link SimpleRecoveryPlanManager} tracks currently failed (permanent) and stopped (transient) tasks,
+ * generates a new {@link DefaultRecoveryBlock} for them and adds them to the recovery Plan, if not already added.
+ */
+public class SimpleRecoveryPlanManager implements PlanManager {
```
nickbp: So far the default/sample implementations have been named `DefaultFoo`, would it make sense to continue that pattern here?

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
```diff
@@ -0,0 +1,325 @@
...
+                            newRecoveryPhaseBuilder.addBlock(block);
+                        }
+                    }
+                }
+            }
+        }
+
+        final DefaultPhase recoveryPhase = newRecoveryPhaseBuilder.build();
+        plan = DefaultPlan.fromArgs(recoveryPhase);
+        phaseStrategy = new RandomRecoveryStrategy(recoveryPhase);
```
nickbp: Could break out the recovery strategy into a `protected` member function. If someone wants to use a custom strategy they can then extend the class and override that function

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
```diff
@@ -0,0 +1,325 @@
...
+            FailureMonitor failureMonitor) {
+        this.stateStore = stateStore;
+        this.offerReqProvider = offerReqProvider;
+        this.failureMonitor = failureMonitor;
+        this.launchConstrainer = launchConstrainer;
+        this.failureListener = failureListener;
+    }
+
+    @VisibleForTesting
+    protected synchronized void refreshPlan(List<Block> dirtiedAssets) {
```
nickbp: Is this expected to be accessed by multiple threads?

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanManager.java
mohitsoni: Glad that you brought this up. I wanted to also use Collection, because then we can also pass in Set. 

Another thing that I want to discuss while we are at it is, we need to think a bit more on this interface.

`Collection<Block>` vs `Collection<String>`. Because, current we assume that two blocks trying to act on the same task should be named same. This is currently not enforced, and can lead to bugs.

Moving `Collection<String>` is slightly better, because then we are stating that it contains the name of the assets being worked upon.

Although, I'm also not fully convinced yet with above. Thoughts?

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
mohitsoni: It is not. This is called from `getCurrentBlock()` method which is called from `PlanCoordinator` (intended). But because `getCurrentBlock()` is a public method and can be called from anywhere else, I did some defensive programming to ensure that this is thread safe.

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
mohitsoni: Good point.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanManager.java
gabrielhartmann: I think you're right.  I think moving to `Collection<String>` makes sense.  It's the most explicit regarding actual behavior.

src/main/java/org/apache/mesos/scheduler/plan/PlanManager.java
mohitsoni: Done!

src/test/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManagerTest.java
mohitsoni: Removed!

src/test/java/org/apache/mesos/scheduler/plan/RandomRecoveryStrategyTest.java
mohitsoni: Done!

src/test/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinatorTest.java
mohitsoni: Done!

src/test/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinatorTest.java
mohitsoni: Done!

src/test/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinatorTest.java
```diff
@@ -0,0 +1,296 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testNoPlanManager() {
+        final DefaultPlanCoordinator coordinator = new DefaultPlanCoordinator(Arrays.asList(), planScheduler);
+        coordinator.processOffers(schedulerDriver, getOffers(1, 1, 1));
+    }
+
+    @Test
+    public void testOnePlanManagerPendingSufficientOffer() throws Exception {
+        when(stateStore.fetchTask(anyString())).thenReturn(Optional.empty());
```
mohitsoni: Done!

src/test/java/org/apache/mesos/scheduler/DefaultSchedulerTest.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/recovery/SimpleRecoveryPlanManager.java
```diff
@@ -0,0 +1,325 @@
...
+
+    public boolean allHaveStatus(Status status, Plan plan) {
+        final List<? extends Phase> phases = plan.getPhases();
+        return phaseStrategy != null &&
+                phases
+                        .stream()
+                        .allMatch(phase -> phaseStrategy.getStatus() == status);
+    }
+
+    public boolean anyHaveStatus(Status status, Plan plan) {
```
mohitsoni: Done!

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanManager.java
mohitsoni: Okay. Let me refactor this.

mohitsoni: Done!

src/test/java/org/apache/mesos/scheduler/DefaultSchedulerTest.java
mohitsoni: @xtenzo Thanks for bringing this up. This should be fixed via: https://github.com/mesosphere/dcos-commons/pull/209

The root cause behind this was basically, two separate acceptOffer calls on the same offer in one Offer cycle. One from PlanManager and one from Resource cleaner.

PR #189, opened 2016-10-05
Support atomically storing multiple values in CuratorPersister

nickbp:
(continued from PR #173)

This allows atomic operations such as:
- Create and/or update multiple TaskInfos at once (used here)
- Create/update a TaskInfo+TaskStatus at once (in the future)

Curator transactions require foreknowledge of whether nodes need to be created or merely modified, as there is no in-transaction statement for 'create or update'. In addition, any parent nodes need to be created manually.

The logic here is a workaround for these deficiencies. The idea is to compose a transaction based on the current state of the ZK tree which creates nodes that don't exist and updates nodes that do exist, then the composed transaction is executed.

This composition/execution cycle is attempted a maximum of three times, to handle potential situations where the tree is modified out-of-band by another party between the composition step and the execution step. In practice this shouldn't come up but it's nice to have.


nickbp:
Note: A lot of the the displayed is a result of moving `CuratorPersister.java` from `storage` to `curator` package. The only effective change in that file should be the addition of a new `storeMany()` function.

src/main/java/org/apache/mesos/storage/Persister.java
```diff
@@ -1,14 +1,43 @@
...
 public interface Persister {
+    /**
+     * Atomically writes many values to storage at once.
+     */
+    void storeMany(Map<String, byte[]> pathBytesMap) throws Exception;
+
+    /**
+     * Writes a single value to storage at the specified path.
+     */
     void store(String path, byte[] bytes) throws Exception;
```
nickbp: Renamed interface to 'get/set'

src/main/java/org/apache/mesos/storage/Persister.java
```diff
@@ -1,14 +1,43 @@
...
+     * Retrieves the previously stored data at the specified path, or throws an exception if the
+     * data is missing.
+     */
     byte[] fetch(String path) throws Exception;
+
+    /**
+     * Deletes the data at the specified path, or throws an exception if the data didn't already
+     * exist.
+     */
     void clear(String path) throws Exception;
```
nickbp: Renamed interface to 'delete'

src/main/java/org/apache/mesos/curator/CuratorPersister.java
```diff
@@ -0,0 +1,141 @@
...
+    private final CuratorFramework client;
+
+    public CuratorPersister(String connectionString, RetryPolicy retryPolicy) {
+        this.connectionString = connectionString;
+        this.retryPolicy = retryPolicy;
+        this.client = startClient();
+    }
+
+    @Override
+    public void storeMany(Map<String, byte[]> pathBytesMap) throws Exception {
```
nickbp: Good point. Done.

src/main/java/org/apache/mesos/curator/CuratorPersister.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+        for (int i = 0; i < ATOMIC_WRITE_ATTEMPTS; ++i) {
+            // Phase 1: Determine which nodes already exist. This determination can be rendered
+            //          invalid by an out-of-band change to the data.
+            Set<String> pathsWhichExist = new HashSet<>();
+            List<String> parentPathsToCreate = new ArrayList<>();
+            for (String path : pathBytesMap.keySet()) {
+                if (client.checkExists().forPath(path) != null) {
+                    pathsWhichExist.add(path);
+                    continue;
```
nickbp: Done via decompose/refactoring

PR #191, opened 2016-10-06
ConnectionResource

mgummelt:
Creates an endpoint `/v1/state/tasks/connection/{taskName}`, which returns

```
{
  "dns": "task-name.framework-name.mesos"
  "ports": "2000-3000,8080"
}
```

### DIFF
--- a/src/main/java/org/apache/mesos/state/api/StateResource.java
+++ b/src/main/java/org/apache/mesos/state/api/StateResource.java
@@ -5,6 +5,7 @@ import com.googlecode.protobuf.format.JsonFormat;
 import org.apache.mesos.Protos;
 import org.apache.mesos.state.StateStore;
 import org.json.JSONArray;
+import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -13,7 +14,9 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 
 /**
@@ -26,6 +29,7 @@ public class StateResource {
 
     private final StateStore stateStore;
     private final PropertyDeserializer propertyDeserializer;
+    private final String frameworkName;
 
     /**
      * Creates a new StateResource which cannot deserialize Properties. Callers will receive a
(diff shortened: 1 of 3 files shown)


mgummelt:
cc @nickbp

mgummelt:
Actually, to consolidate, what do you think about creating a `/v1/tasks` endpoint instead, that serves up an array of task objects, where each object looks something like this:

```
{
  "status": "TASK_RUNNING"
  "info": {...}
  "name": "task-0'
  "dns": "task-0.framework.mesos"
  ...
}
```

Users can then project with `/v1/tasks?_fields=dns,status`, and filter with `/v1/tasks?name=<name>`?

nickbp:
From examples in cassandra/kafka docs, would we want to have support for ports and ip-based addrs in this upstreamed version? Also note the custom `zookeeper` entry for kafka (would probably want a hook to allow for custom entries like that):

```
$ dcos kafka connection
{
    "address": [
        "10.0.0.211:9843",
        "10.0.0.217:10056",
        "10.0.0.214:9689"
    ],
    "dns": [
        "broker-0.kafka.mesos:9843",
        "broker-1.kafka.mesos:10056",
        "broker-2.kafka.mesos:9689"
    ],
    "vip": "broker.kafka.l4lb.thisdcos.directory:9092",
    "zookeeper": "master.mesos:2181/dcos-service-kafka"
}
```

```
$ dcos cassandra connection
{
    "address": [
        "10.0.2.136:9042",
        "10.0.2.138:9042",
        "10.0.2.137:9042"
    ],
    "dns": [
         "node-0.cassandra.mesos:9042",
         "node-1.cassandra.mesos:9042",
         "node-2.cassandra.mesos:9042"
    ]
}
```

mgummelt:
@nickbp I moved the endpoint for consistency, and I included a "ports" field.  dcos-commons isn't storing IPs of hosts right now, so I can't include IPs.  We should add it later, but it will be a bigger change.

src/main/java/org/apache/mesos/state/api/StateResource.java
```diff
@@ -125,10 +131,63 @@ public Response getTaskStatus(@PathParam("taskName") String taskName) {
...
+     *
+     * {
+     *     "dns": "task-name.framework-name.mesos"
+     *     "ports": "8080,2000-3000"
+     * }
+     *
+     * @param taskName Name of the task
+     * @return 200 or 404
+     */
+    @Path("/tasks/connection/{taskName}")
```
nickbp: This endpoint doesn't quite map to the VIP case or the Kafka zk case, where a single endpoint goes to all tasks. Would it make sense to also have a separate 'summary' endpoint, somewhat like the current '/connection' per-framework one, that returns a list of all tasks along with these 'global' endpoints?

nickbp: This endpoint could also provide an interface/hook for frameworks to provide their own custom entries (like the kafka zk case) in a standardized way

mgummelt: dcos-commons isn't assigning VIPs right now, so we can't include them.  

We should add a way for schedulers to define their own resources, but let's do that as the need arises.

PR #186, opened 2016-10-04
Implement Placement Constraints

nickbp:
This pass includes most of the dcos-commons integration, replacing
the previous hand-coded colocation/avoidance constraints. A following
commit will expand on this by adding additional implementations of
PlacementRules and PlacementRuleGenerators, but the dcos-commons
plumbing should all be included here.

The most notable change is the additional data being serialized into
the TaskInfo, namely the Task type name (from TaskSet), and the
Offer Attributes, if any, that were present in the accepted offer
upon task launch. Strictly speaking these values will only be used
by PlacementRuleGenerators that aren't included in this PR (e.g.
one which limits the number of instances of a given type running
against a given attribute) but I regardless wanted to include all
the plumbing up-front in this PR.

PR #192, opened 2016-10-07
Add placement constraint Rules and RuleGenerators

nickbp:
Adds placement constraints (and associated tests) to support checks against:
- Offer Attributes
- Offer Roles
- Task Types
- Distribution of existing tasks against Offer Attributes

(continued from PR #186)


src/main/java/org/apache/mesos/offer/constrain/RoleRule.java
```diff
@@ -0,0 +1,41 @@
+package org.apache.mesos.offer.constrain;
+
+import org.apache.mesos.Protos.Offer;
+import org.apache.mesos.Protos.Resource;
+
+/**
+ * Requires that the Offer's resources be assigned to a provided role.
+ */
+public class RoleRule implements PlacementRule {
```
gabrielhartmann: This RoleRule is a bit dangerous in my view.  I don't know of a use case for it.  Pre-reserved out of band resources are not currently supported by the rest of the Offer Evaluation stack.  If anyone actually tried to use this today, nothing would work.  I'm tempted to say we leave it out since it can't be used yet (or maybe ever) depending on whether we ever fully support static partitioning.

src/main/java/org/apache/mesos/offer/constrain/TaskTypeGenerator.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.List;
+import java.util.Set;
+
+import org.apache.mesos.Protos.TaskInfo;
+import org.apache.mesos.offer.TaskException;
+import org.apache.mesos.offer.TaskUtils;
+
+
+/**
+ * This rule ensures that the given Offer is colocated with the specified task 'type', whose
```
gabrielhartmann: I think this comment is inaccurate.  Doesn't this class deal with both colocation and avoidance?

src/main/java/org/apache/mesos/offer/constrain/TaskTypeGenerator.java
```diff
@@ -0,0 +1,153 @@
...
+        this.typeConverter = typeConverter;
+        this.behaviorType = behaviorType;
+    }
+
+    @Override
+    public PlacementRule generate(Collection<TaskInfo> tasks) {
+        Set<String> agentsWithMatchingType = new HashSet<>();
+        for (TaskInfo task : tasks) {
+            if (typeToFind.equals(typeConverter.getTaskType(task))) {
+                // Matching task type found. Colocate with it on this agent.
```
gabrielhartmann: Comment should read `Colocate with or avoid this agent.`

src/main/java/org/apache/mesos/offer/constrain/RoleRule.java
nickbp: Removed

src/main/java/org/apache/mesos/offer/constrain/TaskTypeGenerator.java
nickbp: Done

src/main/java/org/apache/mesos/offer/constrain/TaskTypeGenerator.java
nickbp: Done

PR #193, opened 2016-10-07
[INFINITY-588] Restart task endpoint

mgummelt:
Moved /v1/state/tasks to /v1/tasks
Added POST /v1/tasks/restart/<task-name>?replace=<replace>

### DIFF
--- a/src/main/java/org/apache/mesos/scheduler/api/TaskResource.java
+++ b/src/main/java/org/apache/mesos/scheduler/api/TaskResource.java
@@ -1,0 +1,168 @@
+package org.apache.mesos.scheduler.api;
+
+import com.googlecode.protobuf.format.JsonFormat;
+import org.apache.mesos.Protos;
+import org.apache.mesos.scheduler.TaskKiller;
+import org.apache.mesos.state.StateStore;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * A read-only API for accessing task and frameworkId state from persistent storage.
+ */
+@Path("/v1/tasks")
+public class TaskResource {
+
+    private static final Logger logger = LoggerFactory.getLogger(TaskResource.class);
+
+    private final StateStore stateStore;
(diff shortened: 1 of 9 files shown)


mgummelt:
cc @gabrielhartmann

src/main/java/org/apache/mesos/scheduler/api/TaskResource.java
```diff
@@ -0,0 +1,189 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(TaskResource.class);
+
+    private final StateStore stateStore;
+    private final PropertyDeserializer propertyDeserializer;
+    private final TaskKiller taskKiller;
+    private final String frameworkName;
+
+    /**
+     * Creates a new TaskResource which cannot deserialize Properties. Callers will receive a
+     * "204 NO_CONTENT" HTTP response when attempting to view the content of a property.
```
gabrielhartmann: Looks like all the property handling is still over in the `StateResource`.  Was the intent to replace the `StateResource` with the `TaskResource`?

src/main/java/org/apache/mesos/state/api/StateResource.java
```diff
@@ -75,122 +69,6 @@ public Response getFrameworkId() {
...
-            }
-        }
-        return String.join(",", ports);
-    }
-
-
-    /**
-     * Produces a listing of the names of all stored tasks.
-     */
     @Path("/properties")
```
gabrielhartmann: The `TaskResource` constructors make reference to properties, but don't appear to actually serve property query responses.  Needs to be resolved one way or the other.  Was the intention of the `StateResource` to now solely present properties?

src/main/java/org/apache/mesos/scheduler/api/TaskResource.java
```diff
@@ -0,0 +1,189 @@
...
+        this(stateStore, null, taskKiller, frameworkName);
+    }
+
+    /**
+     * Creates a new TaskResource which can deserialize Properties. Callers will be able to view
+     * the content of individual Properties.
+     *
+     * @param stateStore the source of data to be returned to callers
+     * @param propertyDeserializer a deserializer which can turn any Property in the provided
+     *                             {@code stateStore} to valid JSON
```
mgummelt: same format as `StateResource`

src/main/java/org/apache/mesos/state/api/StateResource.java
mgummelt: I removed the reference to properties.  There were some vestiges of copy+pasting this from `StateStore`

PR #195, opened 2016-10-07
Add PlacementUtils for avoid and collocate agents

gabrielhartmann:


src/main/java/org/apache/mesos/offer/constrain/PlacementUtils.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Given a list of agents to avoid and collocate with, the appropriate rule is generated.
+     * @param avoidAgents Agents which should not have Tasks placed on them.
+     * @param collocateAgents Agents which should have Tasks placed on them.
+     * @return The appropriate placement rule.
+     */
+    public static Optional<PlacementRuleGenerator> getAgentPlacementRule(
+            List<String> avoidAgents,
+            List<String> collocateAgents) {
```
nickbp: colocate

src/test/java/org/apache/mesos/offer/OfferEvaluatorTest.java
```diff
@@ -912,26 +911,9 @@ public void testLaunchNotOnFirstOffer() throws InvalidRequirementException {
     }
 
     private static OfferRequirement getOfferRequirement(
-            Protos.Resource resource, List<String> avoidAgents, List<String> colocateAgents)
+            Protos.Resource resource, List<String> avoidAgents, List<String> collocateAgents)
```
nickbp: colocate!!

PR #198, opened 2016-10-10
Task config routing

nickbp:
Defines developer-facing APIs for defining task config, and handles scheduler-side forwarding of config to tasks by task type:
- any `TASKCFG_` envvars
- any config templates provided in task specifications

TODO in separate PRs as this one is already pretty big:
- Template handling at executor
- Detecting config changes and triggering task updates as needed


src/main/java/org/apache/mesos/specification/TaskSpecification.java
```diff
@@ -38,8 +38,17 @@
     Collection<VolumeSpecification> getVolumes();
 
     /**
-     * Returns the placement constraint for this task type. This can be anything from "colocate with
-     * another type", "ensure instances are only launched on nodes with attribute X", etc. This may
+     * Returns the Configuration Files to be written for this task, or an empty list if no custom
+     * configurations should be written.
+     *
+     * Note: There is a limit of 100KB (102,400B) across all template data for a given task. If you
```
NimaVaziri: I don't see anywhere that we're actually enforcing this. Is this threshold set by `TaskInfo` by default?

nickbp: Good catch, implemented the limit.

docs/tutorial.md
```diff
@@ -550,6 +550,34 @@ the scheduler starts up, it detects that it is already running 3
...
+While the above describes configuration of the scheduler itself via
+environment variables, there's also a need for configuration of the
+underlying service tasks themselves in a flexible way.
+
+To simplify the common task of getting user-facing configuration to
+service tasks, the developer may follow the following convention in
+naming the environment variables for these configuration options:
+`TASKCFG_<TASK_TYPE>_<NAME>`
+
+For example, an option for tasks of type `index-mgr` could be named
```
NimaVaziri: Perhaps instead of saying it env vars _could_ be we can say it _should_ be since it's not really a suggestion, otherwise if the developer has a lower-cased task name and copies it case-sensitively into the environment variable, the parsing will break.

NimaVaziri:
Besides the comment, LGTM - I've tested this in combination with the custom executor and can confirm that the correct env vars as well as the (config template, config path) pairs are present on the executor side.

docs/tutorial.md
nickbp: Done

docs/tutorial.md
```diff
@@ -550,6 +550,34 @@ the scheduler starts up, it detects that it is already running 3
...
+environment variables, there's also a need for configuration of the
+underlying service tasks themselves in a flexible way.
+
+To simplify the common task of getting user-facing configuration to
+service tasks, the developer may follow the following convention in
+naming the environment variables for these configuration options:
+`TASKCFG_<TASK_TYPE>_<NAME>`
+
+For example, an option for tasks of type `index-mgr` should be named
+`TASKCFG_INDEX_MGR_FOO`, while an option for tasks of type `data`
```
nickbp: Gave more detail on the conversion and updated the examples to better demonstrate the behavior.

Updated the implementation to convert all non-alphanumerics in task types to underscores (to fit envvar format)

src/main/java/org/apache/mesos/config/ConfigNamespace.java
```diff
@@ -0,0 +1,113 @@
...
+     * Returns an immutable copy of the full configuration within this namespace, with any prefixes
+     * omitted from keys.
+     */
+    public ImmutableMap<String, ConfigValue> getAll() {
+        return new ImmutableMap.Builder<String, ConfigValue>().putAll(config).build();
+    }
+
+    /**
+     * Appends the internal state as strings onto the provided list of environment variables. Any
+     * variables already present in the provided list are NOT overwritten, allowing the developer
```
gabrielhartmann: So, the behavior changes based on when this is called?  If you call this first you override, if you call it later you don't?

src/main/java/org/apache/mesos/config/ConfigNamespace.java
```diff
@@ -0,0 +1,113 @@
...
+     * to manually override these values.
+     */
+    public CommandInfo addMissingToEnvironment(CommandInfo command) {
```
gabrielhartmann: I don't understand this method name.

src/main/java/org/apache/mesos/config/TaskConfigRouter.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+import org.apache.mesos.specification.TaskSet;
+
+/**
+ * Default implementation of a {@link Configuration} which provides built-in serialization to
```
gabrielhartmann: This doesn't implement any interface.  Should it?

src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -12,22 +13,30 @@
  * A default implementation of the OfferRequirementProvider interface.
  */
 public class DefaultOfferRequirementProvider implements OfferRequirementProvider {
-    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private static final Logger logger = LoggerFactory.getLogger(DefaultOfferRequirementProvider.class);
```
gabrielhartmann: If the logger is static final, please change to `LOGGER`.

src/main/java/org/apache/mesos/offer/TaskUtils.java
```diff
@@ -18,6 +21,8 @@
  */
 public class TaskUtils {
     private static final Logger LOGGER = LoggerFactory.getLogger(TaskUtils.class);
+    private static final int CONFIG_TEMPLATE_LIMIT_BYTES = 100 * 1024; // 100KB
```
gabrielhartmann: I don't know why this concerns me, but this is a very small not configurable number.  I don't think we should make it configurable, at least not yet.  Maybe bump this to 500KB or something?  I have no data to back up this concern.  Feel free to ignore.

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlockFactory.java
```diff
@@ -22,9 +21,9 @@
     private final StateStore stateStore;
     private final OfferRequirementProvider offerRequirementProvider;
 
-    public DefaultBlockFactory(StateStore stateStore) {
+    public DefaultBlockFactory(StateStore stateStore, OfferRequirementProvider offerRequirementProvider) {
```
gabrielhartmann: Can we make this an additional constructor instead of a replacement?

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanFactory.java
```diff
@@ -14,8 +15,8 @@
 public class DefaultPlanFactory {
     private final DefaultPhaseFactory phaseFactory;
 
-    public DefaultPlanFactory(StateStore stateStore) {
-        this.phaseFactory = new DefaultPhaseFactory(new DefaultBlockFactory(stateStore));
+    public DefaultPlanFactory(StateStore stateStore, OfferRequirementProvider offerRequirementProvider) {
```
gabrielhartmann: Can we make this an additional constructor instead of a replacement?

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlockFactory.java
nickbp: Done here and elsewhere

src/main/java/org/apache/mesos/offer/TaskUtils.java
nickbp: This number is an arbitrary limit just to reduce the likelihood of TaskInfo growing beyond some reasonable level. Bumped to 512KB.

src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -12,22 +13,30 @@
...
+
+    private final TaskConfigRouter configRouter;
```
nickbp: Ended up changing the constructor structure to allow passing in a custom TaskConfigRouter (now an interface) or to use a DefaultConfigRouter.

src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
nickbp: Done

src/main/java/org/apache/mesos/config/TaskConfigRouter.java
nickbp: Yes definitely. Done.

src/main/java/org/apache/mesos/config/ConfigNamespace.java
nickbp: Changed to new name to reflect new behavior: `updateEnvironment()`

PR #202, opened 2016-10-11
Move task config mgmt out of 'features' section

nickbp:
Now moved adjacent to the scheduler reconfiguration tutorial

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -681,6 +681,13 @@ The DC/OS Stateful Services SDK includes deployment strategies supporting
 these different requirements.
 
 ## Configuration management
+For any framework, at least two components must be configured: the
+scheduler and the service.  Scheduler configuration includes things
+like node count, deployment strategies, and security parameters.
+Service configuration includes resource settings such memory, cpu,
+ports, etc., as well as any configuration passed on to the underlying
+service.
+
 The DC/OS Stateful Services SDK includes logic for reading
 configuration from an external source (default: environment
 variables), detecting changes to the configuration, and redeploying
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -376,6 +376,45 @@ the scheduler starts up, it detects that it is already running 3
 `data` tasks, and starts 2 more in order to reach the goal state of 5.
 We should observe two more tasks starting, `data-3` and `data-4`.
 
+## Task Configuration management
+
+While the above describes configuration of the scheduler itself via
+environment variables, there's also a need for configuration of the
+underlying service tasks themselves in a flexible way.
+
+The DC/OS Stateful Services SDK includes logic for reading
+configuration from an external source (default: environment
+variables), detecting changes to the configuration, and redeploying
+any affected tasks.  It also supports marking certain configuration
+parameters as "unmodifiable", so that the user can't change them after
+install time.  For example, the disk size of permanent volumes cannot
+be modified because volume size is static.
+
+(Note: As of this writing, the following is still a work in progress,
+but should be available in the next week or two.)
+
+To simplify the common task of getting user-facing configuration to
+service tasks, the developer may follow the following convention in
+naming the environment variables for these configuration options:
+`TASKCFG_<TASK_TYPE>_<NAME>`
+
+For example, an option for tasks of type `index-mgr` could be named
+`TASKCFG_INDEX_MGR_FOO`, while an option for tasks of type `data`
+could be named `TASKCFG_DATA_FOO`. These configuration options will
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -649,47 +688,6 @@ the cluster one at at time.  Others may permit parallel deployment.
 The DC/OS Stateful Services SDK includes deployment strategies supporting
 these different requirements.
 
-## Configuration management
```
mgummelt: "Configuration management" is still a feature, and we should include it here.  But only at a high level.  Only the "how to" part should be moved.

docs/tutorial.md
```diff
@@ -376,6 +376,37 @@ the scheduler starts up, it detects that it is already running 3
...
+environment variables, there's also a need for configuration of the
+underlying service tasks themselves in a flexible way.
+
+(Note: As of this writing, the following is still a work in progress,
+but should be available in the next week or two.)
+
+To simplify the common task of getting user-facing configuration to
+service tasks, the developer may follow the following convention in
+naming the environment variables for these configuration options:
+`TASKCFG_<TASK_TYPE>_<NAME>`
```
mgummelt: What is a TASK_TYPE?  Is this the name of the task?

docs/tutorial.md
```diff
@@ -376,6 +376,37 @@ the scheduler starts up, it detects that it is already running 3
 `data` tasks, and starts 2 more in order to reach the goal state of 5.
 We should observe two more tasks starting, `data-3` and `data-4`.
 
+## Task Configuration management
+
+While the above describes configuration of the scheduler itself via
+environment variables, there's also a need for configuration of the
+underlying service tasks themselves in a flexible way.
```
mgummelt: s/service tasks/tasks

docs/tutorial.md
```diff
@@ -376,6 +376,37 @@ the scheduler starts up, it detects that it is already running 3
 `data` tasks, and starts 2 more in order to reach the goal state of 5.
 We should observe two more tasks starting, `data-3` and `data-4`.
 
+## Task Configuration management
```
mgummelt: I've been using the term "executor" instead of task.  See the architecture section. I don't know which is better, but we should be consistent.

mgummelt: Can you integrate this into the reference framework?  This tutorial is for the reference framework.

docs/tutorial.md
```diff
@@ -376,6 +376,37 @@ the scheduler starts up, it detects that it is already running 3
...
+
+(Note: As of this writing, the following is still a work in progress,
+but should be available in the next week or two.)
+
+To simplify the common task of getting user-facing configuration to
+service tasks, the developer may follow the following convention in
+naming the environment variables for these configuration options:
+`TASKCFG_<TASK_TYPE>_<NAME>`
+
+For example, an option for tasks of type `index-mgr` could be named
```
mgummelt: s/index-mgr/meta-data, for consistency with this tutorial

docs/tutorial.md
```diff
@@ -376,6 +376,37 @@ the scheduler starts up, it detects that it is already running 3
...
+removed. A special prefix of `TASKCFG_ALL_<NAME>` may be used for
+any options that should be passed to *every* task type.
+
+A common need for service developers is an easy way to write
+configuration files before launching tasks. To fulfill this need,
+the developer may provide configuration file template(s) in their
+TaskSet(s). These templates follow the [mustache](https://mustache.github.io/)
+templating format, similar to what's used in DC/OS packaging. The
+templates will be automatically rendered against the task's
+environment (which is customized as described above), and then
```
mgummelt: what does it mean to be rendered against a task's environment?

docs/tutorial.md
nickbp: It isn't merged yet (hence the disclaimer at the top)

mgummelt: Can you mark this PR as a WIP, then?

docs/tutorial.md
nickbp: Template contains parameter `{{FOO}}` which is populated with the content of environment variable `FOO`

docs/tutorial.md
mgummelt: The whole PR.  There's no point in having it in the tutorial if the feature doesn't exist yet.

nickbp: Wait. This PR is for moving the content. Would you rather it be removed, or...

mgummelt: Oh yea.  I think we should just remove the content for now, keep it as a PR, and block it on the feature.

PR #203, opened 2016-10-11
Add placement rule implementation for simple hostname checks

nickbp:
Just does exact string matching. Largely a copy/paste of AgentRule,
with some rewording there as well ('require' instead of 'colocate')


gabrielhartmann:
cc @verma7

Issue #176, opened 2016-10-03
Misleading error when dcos binary is missing from PATH


nickbp:
From https://github.com/mesosphere/dcos-cassandra-service/issues/269

```
$ ./dcos cassandra connection
2016/09/30 15:35:48 Unable to retrieve configuration value core.dcos_url (DC/OS Cluster URL) from CLI. Run 'dcos config set core.dcos_url http://your-cluster.com' to configure.:
Error: exec: "dcos": executable file not found in $PATH
Output:
CLI configuration value core.dcos_url (DC/OS Cluster URL) is missing/unset. Run 'dcos config set core.dcos_url http://your-cluster.com' to configure.
```

NastyNash:
I was able to fix this by adding my current path to the $PATH. 
export PATH=$PATH:<current dir>

nickbp:
Better error is now available as of PR #204

PR #204, opened 2016-10-11
Provide better error when 'dcos' isn't in PATH

nickbp:
The CLI modules need to call into 'dcos' to retrieve config settings.
To keep this straightforward, the modules assume that 'dcos' is
present in the PATH. This handling detects when that isn't the case,
and provides available options for solving it.

Fixes issue #176


mohitsoni:
LGTM. We can merge once the build pass.

PR #211, opened 2016-10-14
Fix filtering terminated tasks

xtenzo:
Original code does not filter dirtied assets correctly. Assuming there are tasks A, B and C, that need recovery, and dirtied assets are A and C, the original code returns tasks B, C, A, B, instead of expected task B only.

This issue appeared during cluster update (such as increasing cpus/mem requirements, or changing command) with custom scheduler based on DefaultScheduler. Cluster update seems broken in this version of dcos-commons, when DefaultScheduler is used, and more fixes could be required.


gabrielhartmann:
@xtenzo: Could you provide a description in this PR describing what problem you're solving please?

xtenzo:
@gabrielhartmann Sorry for the missing description. I've added the details in the description.

PR #212, opened 2016-10-14
Adds support for the custom executor

NimaVaziri:
In order to support the custom executor, it made sense to make it a subproject and so the main project had to be restructured. Essentially the main project moved to `core` and the custom executor is at `executor`. The new structure is:
```
├── LICENSE
├── README.md
├── build
│   ├── classes
│   ├── libs
│   ├── reports
│   ├── test-results
│   └── tmp
├── build.gradle
├── build.sh
├── cli
│   ├── README.md
│   ├── _example
│   ├── commands.go
│   ├── http.go
│   ├── response.go
│   └── runcli.go
├── core
│   ├── build
│   ├── build.gradle
│   ├── gradle
│   ├── gradlew
│   ├── gradlew.bat
│   └── src
├── docs
│   ├── architecture.png
│   ├── automatic-repair.md
│   └── tutorial.md
├── executor
│   ├── build
│   ├── build.gradle
│   ├── gradle
│   └── src
├── gradle
│   ├── checkstyle
│   ├── findbugs
│   ├── quality.gradle
│   ├── spock.gradle
│   └── wrapper
├── gradle.properties
├── gradlew
├── gradlew.bat
...
...
```


NimaVaziri:
Logging issue has been fixed, `slf4j-simple:1.6.2` was a dependency for `log4j-over-slf4j:1.7.10`

dcos-commons/src/main/java/org/apache/mesos/executor/CustomExecutorTaskFactory.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.mesos.executor;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import org.apache.mesos.ExecutorDriver;
+import org.apache.mesos.Protos;
+
+/**
+ * Implements a factory for a generic custom executor.
+ */
+public class CustomExecutorTaskFactory implements ExecutorTaskFactory {
```
gabrielhartmann: `Custom` --> `Default`

dcos-commons/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -155,4 +157,23 @@ private void validateVolumes(Protos.TaskInfo taskInfo, TaskSpecification taskSpe
...
+
+    private Protos.ExecutorInfo getExecutorInfo(TaskSpecification taskSpecification) {
+
+        Protos.ExecutorInfo.Builder executorInfoBuilder = Protos.ExecutorInfo.newBuilder()
+                .setName(taskSpecification.getName())
+                .setExecutorId(Protos.ExecutorID.newBuilder().setValue("").build()); // Set later by ExecutorRequirement
+
+        Protos.CommandInfo.Builder commandInfoOrBuilder = Protos.CommandInfo.newBuilder()
+                .setValue("./custom-executor/bin/custom-executor")
+                .addUris(TaskUtils.uri("https://s3-us-west-2.amazonaws.com/infinity-artifacts/custom-executor.zip"))
```
gabrielhartmann: We should move the customer-executor.zip to a prod S3 bucket so we can have have `http://downloads.mesosphere.com/...`

dcos-commons/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -155,4 +157,23 @@ private void validateVolumes(Protos.TaskInfo taskInfo, TaskSpecification taskSpe
...
         return volumes;
     }
+
+    private Protos.ExecutorInfo getExecutorInfo(TaskSpecification taskSpecification) {
+
+        Protos.ExecutorInfo.Builder executorInfoBuilder = Protos.ExecutorInfo.newBuilder()
+                .setName(taskSpecification.getName())
+                .setExecutorId(Protos.ExecutorID.newBuilder().setValue("").build()); // Set later by ExecutorRequirement
+
+        Protos.CommandInfo.Builder commandInfoOrBuilder = Protos.CommandInfo.newBuilder()
```
gabrielhartmann: `commandInfoBuilder`  drop the `Or`

dcos-commons/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -155,4 +157,23 @@ private void validateVolumes(Protos.TaskInfo taskInfo, TaskSpecification taskSpe
...
+                .setValue("./custom-executor/bin/custom-executor")
+                .addUris(TaskUtils.uri("https://s3-us-west-2.amazonaws.com/infinity-artifacts/custom-executor.zip"))
+                .setEnvironment(taskSpecification.getCommand().getEnvironment());
+
+        for (Protos.CommandInfo.URI uri : taskSpecification.getCommand().getUrisList()) {
+            commandInfoOrBuilder.addUris(uri);
```
gabrielhartmann: There should be an `addAll` instead of having to loop.

dcos-commons/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -35,14 +35,14 @@
  * new Tasks where applicable.
  */
 public class DefaultScheduler implements Scheduler {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultScheduler.class);
```
gabrielhartmann: Why make the logger static?

dcos-commons/src/test/java/org/apache/mesos/scheduler/DefaultSchedulerTest.java
```diff
@@ -245,7 +245,7 @@ public String getName() {
         register();
 
         Plan plan = defaultScheduler.getPlan();
-        Assert.assertTrue(inExpectedState(plan, Arrays.asList(Status.PENDING, Status.COMPLETE, Status.PENDING)));
+        Assert.assertTrue(inExpectedState(plan, Arrays.asList(Status.PENDING, Status.PENDING, Status.PENDING)));
```
gabrielhartmann: Pretty sure adding the `CustomExecutor` should not have changed any test behavior here.  This looks like a bug was introduced.

dcos-commons/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
NimaVaziri: Consistency - it's static everywhere else.

dcos-commons/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
nickbp: Shouldn't this URI instead be provided via `resource.json` in packaging? Otherwise this will break in airgapped clusters.

dcos-commons/src/test/java/org/apache/mesos/scheduler/DefaultSchedulerTest.java
NimaVaziri: Got bit again by https://issues.apache.org/jira/browse/MESOS-6294.

dcos-commons/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
gabrielhartmann: Yes, it should be.

NimaVaziri:
Addressed comments, though not too sure about how we want to restructure the project.

core/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -155,4 +156,35 @@ private void validateVolumes(Protos.TaskInfo taskInfo, TaskSpecification taskSpe
...
+
+        Protos.CommandInfo.URI executorURI;
+        Protos.ExecutorInfo.Builder executorInfoBuilder = Protos.ExecutorInfo.newBuilder()
+                .setName(taskSpecification.getName())
+                .setExecutorId(Protos.ExecutorID.newBuilder().setValue("").build()); // Set later by ExecutorRequirement
+
+        try {
+            executorURI = TaskUtils.uri(System.getenv(EXECUTOR_URI));
+        } catch (NullPointerException e) {
+            throw new NullPointerException(String.format("Failed at setting the URI for the executor: %s", e));
```
gabrielhartmann: `NullPointerException` seems an odd choice.

core/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -155,4 +156,35 @@ private void validateVolumes(Protos.TaskInfo taskInfo, TaskSpecification taskSpe
...
+        }
+
+        Protos.CommandInfo.Builder commandInfoBuilder = Protos.CommandInfo.newBuilder()
+                .setValue("./executor/bin/executor")
+                .addUris(executorURI)
+                .setEnvironment(taskSpecification.getCommand().getEnvironment())
```
gabrielhartmann: Shouldn't you be getting the environment from the `ExecutorInfo`'s command and not the `TaskInfo`'s here?

core/src/main/java/org/apache/mesos/scheduler/plan/DefaultBlockFactory.java
```diff
@@ -50,6 +50,10 @@ public Block getBlock(String taskType, TaskSpecification taskSpecification) thro
             } catch (InvalidTaskSpecificationException e) {
                 logger.error("Failed to generate TaskSpecification for existing Task with exception: ", e);
                 throw new InvalidRequirementException(e);
+            } catch (InvalidProtocolBufferException e) {
+                logger.error("Failed to unpack taskInfo: {}", e);
+                throw new InvalidRequirementException(
+                        "Failed to generate correct TaskSpecification for existing Task with exception: ", e);
```
gabrielhartmann: This error message is identical to the catch clause above.  It should be updated to match the logged error.

core/src/test/java/org/apache/mesos/offer/DefaultOfferRequirementProviderTest.java
```diff
@@ -26,10 +27,13 @@
     private static final double MEM = 1024.0;
     private static final PlacementRuleGenerator ALLOW_ALL = new PassthroughGenerator(new PassthroughRule("test"));
     private DefaultOfferRequirementProvider defaultOfferRequirementProvider;
+    private EnvironmentVariables environmentVariables;
```
gabrielhartmann: Is this used somewhere or does setting env vars on it simulate a real environment for tests?

executor/src/main/java/org/apache/mesos/executor/Main.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.mesos.ExecutorDriver;
+import org.apache.mesos.Protos;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * The main entry point for the custom executor program.
```
gabrielhartmann: Remove the word `program` please.

executor/src/main/java/org/apache/mesos/executor/Main.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public class Main {
+    private static final Logger LOGGER = LoggerFactory.getLogger(Main.class);
+    private ExecutorService executorService = null;
+
+    public Main() {
```
gabrielhartmann: Is this method needed?

core/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
NimaVaziri: I don't have a `TaskInfo` or `ExecutorInfo` here though so not exactly sure what you mean. And based on my tests I didn't see the env vars being set by the CustomExecutor if I didn't explicitly set them here as part of the `ExecutorInfo`.

core/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -35,14 +35,14 @@
  * new Tasks where applicable.
  */
 public class DefaultScheduler implements Scheduler {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultScheduler.class);
```
NimaVaziri: Consistency - it's static elsewhere as well.

core/src/test/java/org/apache/mesos/offer/DefaultOfferRequirementProviderTest.java
NimaVaziri: It gets set to simulate an environment before each test on L35.

core/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
nickbp: This feels like taking advantage of a side effect, how about a more explicit check like...:
```
String executorStr = System.getenv(EXECUTOR_URI);
if (executorStr == null) {
  throw new IllegalStateException("Missing environment variable: " + EXECUTOR_URI);
}
executorUri = TaskUtils.uri(executorStr);
```

nickbp:
:bike: :house: How about `sdk` instead of `core`?

mohitsoni:
I think SDK better suits as a collective name for these library modules. Because we are breaking down the code-base into multiple modules, I think we should call out each module with a name that conveys it's place in whole SDK.

core/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
NimaVaziri: Sure, that works.

PR #213, opened 2016-10-17
Add ContainerInfo for TaskSpecification

activatedgeek:
The existing `DefaultTaskSpecification` does not provide appropriate entry points for running containers as tasks. This PR proposes an updated `TaskSpecification` interface.

cc @gabrielhartmann

### DIFF
--- a/sdk/core/src/main/java/org/apache/mesos/specification/DefaultService.java
+++ b/sdk/core/src/main/java/org/apache/mesos/specification/DefaultService.java
@@ -82,7 +82,7 @@ public class DefaultService implements Service {
     }
 
     private static void registerFramework(Scheduler sched, Protos.FrameworkInfo frameworkInfo, String masterUri) {
-        LOGGER.info(String.format("Registering framework: %s", frameworkInfo));
+        LOGGER.info("Registering framework: {}", frameworkInfo);
         SchedulerDriver driver = new SchedulerDriverFactory().create(sched, frameworkInfo, masterUri);
         driver.run();
     }
--- a/sdk/core/src/main/java/org/apache/mesos/specification/DefaultService.java
+++ b/sdk/core/src/main/java/org/apache/mesos/specification/DefaultService.java
@@ -82,7 +82,7 @@ public class DefaultService implements Service {
     }
 
     private static void registerFramework(Scheduler sched, Protos.FrameworkInfo frameworkInfo, String masterUri) {
-        LOGGER.info("Registering framework: {}", frameworkInfo);
+        LOGGER.info(String.format("Registering framework: %s", frameworkInfo));
         SchedulerDriver driver = new SchedulerDriverFactory().create(sched, frameworkInfo, masterUri);
         driver.run();
     }
--- a/examples/reference/src/main/java/com/mesosphere/sdk/reference/scheduler/Main.java
+++ b/examples/reference/src/main/java/com/mesosphere/sdk/reference/scheduler/Main.java
@@ -54,6 +54,7 @@ public class Main {
                 return Arrays.asList(
                         DefaultTaskSet.create(TASK_METADATA_COUNT,
                                 TASK_METADATA_NAME,
+                                TASK_METADATA_NAME,
                                 getCommand(TASK_METADATA_NAME),
                                 getResources(TASK_METADATA_CPU, TASK_METADATA_MEM_MB),
                                 getVolumes(TASK_METADATA_DISK_MB, TASK_METADATA_NAME),
@@ -63,6 +64,7 @@ public class Main {
                                 Optional.of(TaskTypeGenerator.createAvoid(TASK_METADATA_NAME)),
                                 Optional.of(getHealthCheck(TASK_METADATA_NAME))),
                         DefaultTaskSet.create(TASK_DATA_COUNT,
+                                TASK_DATA_NAME,
                                 TASK_DATA_NAME,
                                 getCommand(TASK_DATA_NAME),
                                 getResources(TASK_DATA_CPU, TASK_DATA_MEM_MB),
--- a/sdk/core/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
+++ b/sdk/core/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
@@ -233,8 +233,11 @@ public class DefaultOfferRequirementProvider implements OfferRequirementProvider
 
         Protos.CommandInfo.Builder commandInfoBuilder = Protos.CommandInfo.newBuilder()
                 .setValue("./executor/bin/executor")
-                .addUris(executorURI)
-                .addAllUris(taskSpecification.getCommand().getUrisList());
+                .addUris(executorURI);
+
+        if (taskSpecification.getCommand().isPresent()) {
+            commandInfoBuilder.addAllUris(taskSpecification.getCommand().get().getUrisList());
+        }
 
         // some version of the JRE is required to kickstart the executor
         setJREVersion(commandInfoBuilder, taskSpecification);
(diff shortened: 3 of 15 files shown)


activatedgeek:
I would request somebody to point me to the final `TaskInfo` construction so that the `getContainer` interface method can be utilized to add `ContainerInfo` to the protobuf.

gabrielhartmann:
@activatedgeek:  Thanks for the PR.  Looks reasonable to me.  Could you add a few unit tests validating that this generates expected protobufs.  In particular I'm interested in validating that `getExistingOfferRequirement` works as expected after launching with a `ContainerInfo`.  Should be fine, but would be nice to validate with a test or two.

mohitsoni:
What happens when ContainerInfo and CommandInfo are set simultaneously ?

gabrielhartmann:
Also, @activatedgeek and @mohitsoni.  I this situation we'd probably want to make CommandInfo optional.  I think with entry points all combinations are possible, and I don't see how we could validate TaskSpecifications actually.  Maybe trust the developer?  A revolutionary thought.

activatedgeek:
@mohitsoni, From what I remember after seeing `mesos.proto`, I think `ContainerInfo` uses `CommandInfo` if it is available for the `[command]` part in `docker run [options] image [command]`). Please correct me if I am wrong.

@gabrielhartmann I did think about making both the `ContainerInfo` and `CommandInfo` optional which in fact does keep us in line with the protobuf definition anyways (where both are optional and "trust the developer" kind of thing). I believe it should be done and an exception can be thrown in case both are `empty()`. What do you think?

gabrielhartmann:
@activatedgeek:  I think we're getting somewhere.  I was trying to think what validation we could perform.  Both empty throwing the exception you specify sounds good to me.  Can't validate the developer hasn't done something incorrect if both are specified.

Making the change discussed above and writing some unit tests should be sufficient to get this across the line.

/cc @mohitsoni

activatedgeek:
@gabrielhartmann I rebased from master and updated my code. Requesting for review.

* To ensure that the developer doesn't end up missing both `CommandInfo` and `ContainerInfo`, only appropriate constructors have been provided. More precisely, constructor arguments of `CommandInfo` and `ContainerInfo` are not optional in the `DefaultTaskSpecification`.

* The static `DefaultTaskSpecification.create` method throws appropriate exceptions and uses appropriate constructors based on which of the two protobufs are available.

* Updated tests for `CustomTaskSet` (covering all possible constructors)

gabrielhartmann:

src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -29,8 +29,9 @@ public DefaultOfferRequirementProvider(TaskConfigRouter taskConfigRouter) {
     @Override
     public OfferRequirement getNewOfferRequirement(String taskType, TaskSpecification taskSpecification)
             throws InvalidRequirementException {
+
         Protos.CommandInfo updatedCommand = taskConfigRouter.getConfig(taskType)
-                .updateEnvironment(taskSpecification.getCommand());
+                .updateEnvironment(taskSpecification.getCommand().get());
```
gabrielhartmann: Doesn't this need an `isPresent()` check before use?  If it's optional now we'll we want to skip adding it when it's not present correct?

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
```diff
@@ -16,6 +16,7 @@
     private final String name;
     private final List<TaskSpecification> taskSpecifications;
 
+    // TODO: this should be deprecated
```
gabrielhartmann: Why should this be deprecated?  Creation with just a CommandInfo is valid.

gabrielhartmann:
@activatedgeek:  Looks good.  A couple comments.  Also, this needs OfferRequirement tests.  I'd like to see something that rejects creation of an OfferRequirement if input is bad, and validates expected content in expected cases.

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
```diff
@@ -58,6 +59,58 @@ public static DefaultTaskSet create(
...
+            int count,
+            String name,
+            Protos.ContainerInfo container,
+            Collection<ResourceSpecification> resources,
+            Collection<VolumeSpecification> volumes,
+            Collection<ConfigFileSpecification> configs,
+            Optional<PlacementRuleGenerator> placementOptional,
+            Optional<Protos.HealthCheck> healthCheck) {
+
+        List<TaskSpecification> taskSpecifications = new ArrayList<>();
```
gabrielhartmann: A lot of code duplication in this loop and the loop below.

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
activatedgeek: Indeed that is valid but it misses out on various other optional elements and I thought it might be unnecessary to strip down the `TaskInfo` so much. But I guess, let's keep it.

src/main/java/org/apache/mesos/specification/DefaultTaskSpecification.java
```diff
@@ -45,26 +46,102 @@ public static DefaultTaskSpecification create(Protos.TaskInfo taskInfo, Optional
...
+                    placement,
+                    healthCheck);
+        }
+    }
+
+    /**
+     * Create task with only {@link Protos.CommandInfo}.
+     */
+    protected DefaultTaskSpecification(
+            String name,
```
activatedgeek: I'll fix this by making all the constructors now go via a private constructor. That removes the bad duplication. Thanks!

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
activatedgeek: Here again, a private static common function should fix bad duplication.

activatedgeek: Same as below, fixed via a `private static` sink function.

src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
activatedgeek: Oh! That's a big miss. I also want to understand that why is there a special update method for the `CommandInfo`. Why not simply set it into the new `Protos.TaskInfo.Builder`?

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
```diff
@@ -16,22 +16,23 @@
...
             String name,
             Protos.CommandInfo command,
             Collection<ResourceSpecification> resources,
             Collection<VolumeSpecification> volumes) {
+
         return create(
                 count,
                 name,
-                command,
+                Optional.empty(),
```
gabrielhartmann: Please put comments back in for what the empty elements represent.

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
```diff
@@ -44,19 +45,16 @@ public static DefaultTaskSet create(
...
-                    command,
-                    resources,
-                    volumes,
-                    configs,
-                    placementOptional,
-                    healthCheck));
-        }
-
-        return create(name, taskSpecifications);
+        return create(
```
gabrielhartmann: same here.

src/main/java/org/apache/mesos/specification/DefaultTaskSet.java
```diff
@@ -69,19 +67,16 @@ public static DefaultTaskSet create(
...
-                    container,
-                    resources,
-                    volumes,
-                    configs,
-                    placementOptional,
-                    healthCheck));
-        }
-
-        return create(name, taskSpecifications);
+        return create(
```
gabrielhartmann: and here

gabrielhartmann:
@activatedgeek:  mostly copy/paste from Slack:

The test is fine but more is needed.  It looks like all the tests in `DefaultOfferRequirementProviderTest` are focused around testing already launched Tasks.  Ignoring that oddness, it would be nice to have a couple tests that use the `getNewOfferRequirement` method of the `DefaultOfferRequirementProvider` with the new kinds of `TaskSpecification` you’ve enabled.

And then check that the OfferRequirement has been constructed as expected.  One with just a ContainerInfo, one with just a CommandInfo, one with both, and one with neither should be sufficient.  Then there should be a couple tests that test the `getExistingOfferRequirement` method to make sure one can recover from crash and detect changes to the TaskSpec.

e.g. One might start with a valid TaskSpecification, and then change it so that it's no invalid.  That should cause a failure in generating an an `OfferRequirement`.

activatedgeek:
@gabrielhartmann Right, that falls in line. Thanks!

sdk/core/src/main/java/org/apache/mesos/specification/DefaultService.java
```diff
@@ -82,7 +82,7 @@ public void run() {
     }
 
     private static void registerFramework(Scheduler sched, Protos.FrameworkInfo frameworkInfo, String masterUri) {
-        LOGGER.info("Registering framework: {}", frameworkInfo);
+        LOGGER.info(String.format("Registering framework: %s", frameworkInfo));
         SchedulerDriver driver = new SchedulerDriverFactory().create(sched, frameworkInfo, masterUri);
```
activatedgeek: Oh I believe that happened after I had committed. I kept my changes during the rebase. I'll revert. Thanks!

gabrielhartmann:

PR #217, opened 2016-10-18
[INFINITY-502] Add suppress/revive support to DefaultScheduler

mgummelt:
Also makes significant changes to `DefaultRecoveryPlanManager`

### DIFF
--- a/src/main/java/org/apache/mesos/curator/CuratorStateStore.java
+++ b/src/main/java/org/apache/mesos/curator/CuratorStateStore.java
@@ -6,10 +6,7 @@ import org.apache.mesos.Protos;
 import org.apache.mesos.dcos.DcosConstants;
 import org.apache.mesos.offer.TaskException;
 import org.apache.mesos.offer.TaskUtils;
-import org.apache.mesos.state.SchemaVersionStore;
-import org.apache.mesos.state.StateStore;
-import org.apache.mesos.state.StateStoreException;
-import org.apache.mesos.state.StateStoreUtils;
+import org.apache.mesos.state.*;
 import org.apache.mesos.storage.Persister;
 import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
@@ -17,6 +14,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.VisibleForTesting;
 
+import java.io.IOException;
 import java.util.*;
 
 /**
@@ -56,6 +54,7 @@ public class CuratorStateStore implements StateStore {
     private static final String FWK_ID_PATH_NAME = "FrameworkID";
     private static final String PROPERTIES_PATH_NAME = "Properties";
     private static final String TASKS_ROOT_NAME = "Tasks";
+    private static final String SUPPRESSED_KEY = "suppressed";
 
     private final Persister curator;
     private final TaskPathMapper taskPathMapper;
@@ -413,5 +412,34 @@ public class CuratorStateStore implements StateStore {
         private String getTasksRootPath() {
             return tasksRootPath;
         }
+    }
+
+    public boolean isSuppressed() throws StateStoreException {
+        byte[] bytes = StateStoreUtils.fetchPropertyOrEmptyArray(this, SUPPRESSED_KEY);
+        Serializer serializer = new JsonSerializer();
+
+        boolean suppressed;
+        try {
+            suppressed = serializer.deserialize(bytes, Boolean.class);
+        } catch (IOException e){
+            logger.error("Error converting property " + SUPPRESSED_KEY + " to boolean.", e);
+            throw new StateStoreException(e);
+        }
+
+        return suppressed;
+    }
+
+    public void setSuppressed(final boolean isSuppressed) {
+        byte[] bytes;
+        Serializer serializer = new JsonSerializer();
+
+        try {
+            bytes = serializer.serialize(isSuppressed);
+        } catch (IOException e) {
+            logger.error("Error serializing property " + SUPPRESSED_KEY + ": " + isSuppressed + ".", e);
+            throw new StateStoreException(e);
+        }
+
+        storeProperty(SUPPRESSED_KEY, bytes);
     }
 }
--- a/src/main/java/org/apache/mesos/scheduler/plan/Block.java
+++ b/src/main/java/org/apache/mesos/scheduler/plan/Block.java
@@ -11,6 +11,8 @@ import java.util.UUID;
  * Defines the interface for a Block of a {@link Phase}. The block is the base unit of a set of
  * tasks to perform, such as launching a Task, updating a Task, or reconciling Mesos state with
  * Framework state. A Block may be in one of four states: PENDING, IN_PROGRESS, COMPLETE, or ERROR.
+ *
+ * A {@Block} is an {@Observable}, and will notify its observers when its state changes.
  * <p>
  * See {@Stage} docs for more background.
  */
--- a/src/main/java/org/apache/mesos/scheduler/plan/DefaultInstallStrategy.java
+++ b/src/main/java/org/apache/mesos/scheduler/plan/DefaultInstallStrategy.java
@@ -1,5 +1,6 @@
 package org.apache.mesos.scheduler.plan;
 
+import java.util.Collection;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -18,12 +19,12 @@ public class DefaultInstallStrategy implements PhaseStrategy {
     }
 
     @Override
-    public Optional<Block> getCurrentBlock() {
+    public Optional<Block> getCurrentBlock(Collection<String> dirtiedAssets) {
         if (interrupted.get()) {
             return Optional.empty();
         } else {
             for (Block block : phase.getBlocks()) {
-                if (!block.isComplete()) {
+                if (!block.isComplete() && !dirtiedAssets.contains(block.getName())) {
                     return Optional.of(block);
                 }
             }
--- a/src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
+++ b/src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
@@ -12,6 +12,8 @@ import java.util.*;
  * Default Phase implementation tracks Blocks both by their UUID and by
  * their ordering. It is an immutable class that can be constructed either
  * directly, or using a fluent style builder.
+ *
+ * A {@DefaultPhase} is an {@Observable} and will forward updates from its blocks.
  */
 public class DefaultPhase extends ChainedObserver implements Phase {
 
--- a/src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseStrategy.java
+++ b/src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseStrategy.java
@@ -1,5 +1,6 @@
 package org.apache.mesos.scheduler.plan;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
@@ -10,21 +11,20 @@ import org.slf4j.LoggerFactory;
 /**
  * Provides the default strategy for executing a phase where there is a pause after the first block.
  */
-public class DefaultStageStrategy implements PhaseStrategy {
-  private static final Logger logger = LoggerFactory.getLogger(DefaultStageStrategy.class);
+public class DefaultPhaseStrategy implements PhaseStrategy {
+  private static final Logger logger = LoggerFactory.getLogger(DefaultPhaseStrategy.class);
 
   protected final Phase phase;
-
   private boolean[] shouldStart;
   private int currPos = 0;
 
-  public DefaultStageStrategy(Phase phase) {
+  public DefaultPhaseStrategy(Phase phase) {
     this.phase = phase;
-    this.shouldStart = initializeDecisionMap(phase.getBlocks());
+    this.shouldStart = getShouldStart(phase.getBlocks());
     advancePosition();
   }
 
-  private static boolean[] initializeDecisionMap(List<? extends Block> blocks) {
+  private static boolean[] getShouldStart(List<? extends Block> blocks) {
     int blockCount = blocks.size();
     boolean[] shouldStart = new boolean[blockCount];
 
@@ -63,15 +63,19 @@ public class DefaultStageStrategy implements PhaseStrategy {
   }
 
   @Override
-  public Optional<Block> getCurrentBlock() {
+  public Optional<Block> getCurrentBlock(Collection<String> dirtiedAssets) {
     //TODO(nick) the returned Block is not guaranteed to stay in a consistent state.
     //alternative: instead provide a thread-safe wrapper for viewing/changing current block state
     synchronized (this) {
       advancePosition();
       if (shouldStart[currPos]) {
         Block block = phase.getBlocks().get(currPos);
-        logger.info(String.format("Returning block '%s' at position: %d", block.getName(), currPos));
-        return Optional.of(block);
+        if (!dirtiedAssets.contains(block.getName())) {
+          logger.info(String.format("Returning block '%s' at position: %d", block.getName(), currPos));
+          return Optional.of(block);
+        } else {
+          return Optional.empty();
+        }
       } else {
         // WAITING for input, don't continue to next block
         return Optional.empty();
--- a/src/main/java/org/apache/mesos/scheduler/plan/DefaultPlan.java
+++ b/src/main/java/org/apache/mesos/scheduler/plan/DefaultPlan.java
@@ -10,6 +10,8 @@ import java.util.List;
 /**
  * An ordered list of {@link Phase}s, composed into a {@link Plan}. It may
  * optionally contain a List of errors associated with the phase.
+ *
+ * A {@DefaultPlan} is an {@Observable} and will forward updates from its {@Phase}s.
  */
 public class DefaultPlan extends ChainedObserver implements Plan {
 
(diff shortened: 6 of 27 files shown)


src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -25,95 +23,140 @@
...
-public class DefaultRecoveryPlanManager implements PlanManager {
+public class DefaultRecoveryPlanManager extends ChainedObserver implements PlanManager {
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultRecoveryPlanManager.class);
 
     protected volatile Plan plan;
     protected volatile PhaseStrategy phaseStrategy;
     protected volatile Map<UUID, PhaseStrategy> phaseStrategies;
 
     private final StateStore stateStore;
-    private final TaskFailureListener failureListener;
```
mgummelt: Is this safe to remove?  It seemed superfluous

mgummelt:
cc @mohitsoni for review

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
```diff
@@ -414,4 +413,33 @@ private String getTasksRootPath() {
...
+            suppressed = serializer.deserialize(bytes, Boolean.class);
+        } catch (IOException e){
+            logger.error("Error converting property " + SUPPRESSED_KEY + " to boolean.", e);
+            throw new StateStoreException(e);
+        }
+
+        return suppressed;
+    }
+
+    public void setSuppressed(final boolean isSuppressed) {
```
mohitsoni: Is this used to persist `suppressed` flag, so that when a scheduler restarts, it restarts suppressed ? If yes, how  is that implemented ?

src/main/java/org/apache/mesos/scheduler/plan/DefaultInstallStrategy.java
```diff
@@ -18,12 +19,12 @@ public DefaultInstallStrategy(Phase phase) {
...
 
     @Override
-    public Optional<Block> getCurrentBlock() {
+    public Optional<Block> getCurrentBlock(Collection<String> dirtiedAssets) {
         if (interrupted.get()) {
             return Optional.empty();
         } else {
             for (Block block : phase.getBlocks()) {
-                if (!block.isComplete()) {
+                if (!block.isComplete() && !dirtiedAssets.contains(block.getName())) {
```
mohitsoni: `getCurrentBlock()` is a public interface method. `dirtiedAssets` can be `null`. Maybe validate ? `... !CollectionUtils.isNotEmpty(dirtiedAssets) ...`

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
```diff
@@ -12,6 +12,8 @@
  * Default Phase implementation tracks Blocks both by their UUID and by
  * their ordering. It is an immutable class that can be constructed either
  * directly, or using a fluent style builder.
+ *
+ * A {@DefaultPhase} is an {@Observable} and will forward updates from its {@b}.
```
mohitsoni: `from it's {@b}` ? `Block` ?

src/main/java/org/apache/mesos/scheduler/plan/Block.java
```diff
@@ -11,6 +11,8 @@
  * Defines the interface for a Block of a {@link Phase}. The block is the base unit of a set of
  * tasks to perform, such as launching a Task, updating a Task, or reconciling Mesos state with
  * Framework state. A Block may be in one of four states: PENDING, IN_PROGRESS, COMPLETE, or ERROR.
+ *
+ * A {@Block} is an {@Observable}, and will notify its observers when its state changes.
```
mohitsoni: +1 for keeping docs up-to-date.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseStrategy.java
```diff
@@ -63,15 +63,19 @@ private void advancePosition() {
...
+  public Optional<Block> getCurrentBlock(Collection<String> dirtiedAssets) {
     //TODO(nick) the returned Block is not guaranteed to stay in a consistent state.
     //alternative: instead provide a thread-safe wrapper for viewing/changing current block state
     synchronized (this) {
       advancePosition();
       if (shouldStart[currPos]) {
         Block block = phase.getBlocks().get(currPos);
-        logger.info(String.format("Returning block '%s' at position: %d", block.getName(), currPos));
-        return Optional.of(block);
+        if (!dirtiedAssets.contains(block.getName())) {
```
mohitsoni: Good catch!

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlan.java
```diff
@@ -10,6 +10,8 @@
 /**
  * An ordered list of {@link Phase}s, composed into a {@link Plan}. It may
  * optionally contain a List of errors associated with the phase.
+ *
+ * A {@DefaultPlan} is an {@Observable} and will forward updates from its {@Phase}s.
```
mohitsoni: 👍

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanCoordinator.java
```diff
@@ -12,8 +13,10 @@
 
 /**
  * Default implementation of PlanCoordinator.
+ *
+ * A {@DefaultPlanCoordinator} is an {@Observable} and will forward updates from its plans.
```
mohitsoni: 👍

src/main/java/org/apache/mesos/scheduler/plan/PlanCoordinator.java
```diff
@@ -2,18 +2,22 @@
...
 import org.apache.mesos.SchedulerDriver;
+import org.apache.mesos.scheduler.Observable;
 
 import java.util.Collection;
 import java.util.List;
 
 /**
- * PlanCoordinator job is to perform co-ordination of offers among configured {@code PlanManager}s.
+ * PlanCoordinator's job is to coordinate offers among configured {@code PlanManager}s.
+ * A {@code PlanCoordinator} updates its observers
```
mohitsoni: `A {@code PlanCoordinator} being an {@link Observable} updates its observers.`

src/main/java/org/apache/mesos/scheduler/plan/PlanCoordinator.java
```diff
@@ -2,18 +2,22 @@
...
  */
-public interface PlanCoordinator {
+public interface PlanCoordinator extends Observable {
     /**
      * Provides offers to each {@code PlanManager} for processing. Keeps tracks of dirtied offers and assets.
      *
      * @return List of OfferID's that were used for scheduling operations.
      */
     Collection<Protos.OfferID> processOffers(final SchedulerDriver driver, final List<Protos.Offer> offers);
+
```
mohitsoni: Can you please add docs here as well ?

src/main/java/org/apache/mesos/scheduler/plan/RandomRecoveryStrategy.java
```diff
@@ -19,12 +16,16 @@ public RandomRecoveryStrategy(Phase phase) {
     }
 
     @Override
-    public Optional<Block> getCurrentBlock() {
+    public Optional<Block> getCurrentBlock(Collection<String> dirtiedAssets) {
         final List<? extends Block> blocks = filterOnlyPendingBlocks(getPhase().getBlocks());
-        if (isInterrupted() || CollectionUtils.isEmpty(blocks)) {
+        if (isInterrupted() || CollectionUtils.isEmpty(blocks) || blocks.size() <= dirtiedAssets.size()) {
```
mohitsoni: `null` check `dirtiedAssets` ?

src/main/java/org/apache/mesos/scheduler/plan/RandomRecoveryStrategy.java
```diff
@@ -19,12 +16,16 @@ public RandomRecoveryStrategy(Phase phase) {
...
             return Optional.empty();
         } else {
-            return Optional.of(blocks.get(new Random().nextInt(blocks.size())));
+            List<Block> cleanBlocks = blocks.stream()
+                    .filter(block -> !dirtiedAssets.contains(block.getName()))
```
mohitsoni: `null` check dirtiedAssets` ?

src/main/java/org/apache/mesos/scheduler/plan/ReconciliationStrategy.java
```diff
@@ -23,9 +24,14 @@ private ReconciliationStrategy(final ReconciliationPhase phase) {
     }
 
     @Override
-    public Optional<Block> getCurrentBlock() {
+    public Optional<Block> getCurrentBlock(Collection<String> dirtiedAssets) {
         // ReconciliationPhases only have a single Block
-        return Optional.of(phase.getBlock(0));
+        for (Block block : phase.getBlocks()) {
```
mohitsoni: `null` check `dirtiedAssets` ? Also, if `dirtiedAssets` is empty or null, just `return Optional.of(phase.getBlock(0));` ?

src/main/java/org/apache/mesos/scheduler/plan/Status.java
```diff
@@ -4,9 +4,9 @@
  * Status of the block. Useful for reporting
  */
 public enum Status {
-  ERROR,
-  WAITING,
-  PENDING,
-  IN_PROGRESS,
-  COMPLETE
+  ERROR, // execution experienced an error
```
mohitsoni: 👍  for updating documentation.

mohitsoni: Can you convert these to JavaDoc style comments `/** */` while you are at it ? That way these will become part of our JavaDocs.

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -25,93 +23,68 @@
...
-            TaskFailureListener failureListener,
             RecoveryRequirementProvider offerReqProvider,
             LaunchConstrainer launchConstrainer,
             FailureMonitor failureMonitor) {
         this.stateStore = stateStore;
         this.offerReqProvider = offerReqProvider;
         this.failureMonitor = failureMonitor;
         this.launchConstrainer = launchConstrainer;
-        this.failureListener = failureListener;
+        updatePlan(new ArrayList<>());
```
mohitsoni: Nit: s/`new ArrayList<>()`/`Collections.EMPTY_LIST`/g (feel free to ignore this one)

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -203,9 +174,53 @@ public void forceComplete(UUID phaseId, UUID blockId) {
...
+                    Collection<Block> blocks = new ArrayList<>();
+
+                    // 2. remove any COMPLETED blocks for this task.
+                    for (Phase phase : this.plan.getPhases()) {
+                        blocks.addAll(phase.getBlocks().stream()
+                                .filter(block -> !(block.isComplete() && block.getName().equals(taskName)))
+                                .collect(Collectors.toList()));
+                    }
+
+                    // 3. Add new PENDING block.
```
mohitsoni: 1. If a block was already pending, and for whatever reason (reconcile, etc.) we get another `TaskStatus` needing recovery, the code will add another `PENDING` block for it. I think we should make is explicit to only add a block in the phase if it's not already present.
2. Also, when a recovery Block is in `IN_PROGRESS` state, and the it gets another `TaskStatus`.

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -203,9 +174,53 @@ public void forceComplete(UUID phaseId, UUID blockId) {
...
+                    blocks.add(newBlock.get());
+
+                    updatePlan(blocks);
+                }
+            }
+        } catch (TaskException ex) {
+            LOGGER.error("Error updating status for task (id=%s)".format(status.getTaskId().toString()), ex);
```
mohitsoni: Style nit:
`LOGGER.error("Error updating status for task (id={}). Reason: {}", status.getTaskId(), ex);`

src/test/java/org/apache/mesos/scheduler/DefaultSchedulerTest.java
```diff
@@ -554,4 +515,59 @@ private int countBlocks(Plan plan) {
 
         return i;
     }
+
+    private Protos.TaskID installBlock(int phaseIndex, int blockIndex, Protos.Offer offer) {
```
mohitsoni: 👍  for creating helpers.

src/main/java/org/apache/mesos/curator/CuratorStateStore.java
mgummelt: No, this is only used to serve the `/v1/state/properties/suppressed` endpoint

src/main/java/org/apache/mesos/scheduler/plan/DefaultInstallStrategy.java
mgummelt: Validate that it's non-null, or that it's non-empty?  We don't do null checks anywhere else in the code (and rightly so, it would be too verbose).  And it's valid for `dirtiedAssets` to be empty.

src/main/java/org/apache/mesos/scheduler/plan/PlanCoordinator.java
mgummelt: fixed

src/main/java/org/apache/mesos/scheduler/plan/PlanCoordinator.java
mgummelt: fixed

src/main/java/org/apache/mesos/scheduler/plan/RandomRecoveryStrategy.java
mgummelt: addressed above

src/main/java/org/apache/mesos/scheduler/plan/ReconciliationStrategy.java
mgummelt: addressed above, and the existing code has that same effect.

src/main/java/org/apache/mesos/scheduler/plan/Status.java
mgummelt: fixed

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
mgummelt: fixed

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
gabrielhartmann: @mgummelt:  It's used here to mark Tasks as permanently failed correct? https://github.com/mesosphere/dcos-commons/blob/0df897784b8e68e42a31c09c9af79ea25ed3fc01/src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java#L275

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
gabrielhartmann: @mohitsoni:  This PR is for suppress revive right?  I think if we want to make improvements to the recovery manager's behavior we should probably do that in a different PR.  Do you want to make a JIRA to track the improvements you'd like?

/cc @mgummelt

mgummelt: His comments are related to the correctness of my changes to this class, which I needed to make to support the observer pattern.  I'm addressing them now.

mgummelt: Fixed, and added regression test.

mgummelt:
cc @mohitsoni please re-review

src/main/java/org/apache/mesos/scheduler/plan/DefaultInstallStrategy.java
mohitsoni: Yeah, it will work fine when `dirtiedAssets` is empty. 
I think not-doing null check in methods which are not part of public interfaces is okay. But, I think we should exercise caution by validating arguments of methods which are part of public interfaces like `PhaseStrategy`. If we are not doing it at other places in code, maybe we should start doing it.

src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
mohitsoni: @gabrielhartmann @mgummelt refactored this class to make use of `update()` method for receiving task updates, and regenerating the plan. Instead of the previous model, where the plan was re-generated only with each offer cycle. These comments are to address bug(s) introduced during that refactoring.

I think @mgummelt's approach is better is terms of consistency with PlanManager design, and it also allows for a single way for all PlanManagers to receives TaskStatuses via Observer pattern.

mohitsoni:
👍

PR #220, opened 2016-10-19
Plan Refactor

gabrielhartmann:
* `Plan`, `Phase`, `Block` hierarchy is now just a special case of hierarchy made possible by `Element<C>`
* `TaskSpecification` contains `type`
* `Element` contains `Strategy`
* `Strategy` is what determines what `Elements` can be worked on at anytime.  So parallel strategies are now possible and written
* `hasDecisionPoint` is gone as it doesn't make sense.  Superseded by the concept of a Strategy being interrupted.


nickbp:
retest this please

nickbp:
Given that TaskSpecification now has a `type`, any reason to keep TaskSets around anymore? Could represent all tasks as a flat list of TaskSpecifications, which in turn may have different `type` values.

src/main/java/org/apache/mesos/scheduler/plan/Block.java
```diff
@@ -63,38 +32,14 @@ static Status getStatus(Block block) {
...
-    /**
-     * Returns a user-visible message describing the current status of the Block.
-     */
-    String getMessage();
-
-    /**
-     * Returns a user-visible name describing the purpose of the Block.
-     */
-    String getName();
+    class InvalidException extends Exception {
```
nickbp: Feels like this exception could be in BlockFactory

src/main/java/org/apache/mesos/scheduler/plan/BlockFactory.java
```diff
@@ -11,6 +12,6 @@
  * detected for a Task.
```
nickbp: Will this be provided all defined/available TaskSpecifications at once, or just some subset? Could specify this in the javadoc for reference.

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
```diff
@@ -17,24 +20,23 @@
...
-
-    private DefaultBlock(String name, Optional<OfferRequirement> offerRequirementOptional, Status status) {
+    public DefaultBlock(
+            String name,
+            Optional<OfferRequirement> offerRequirement,
+            Status status,
+            List<String> errors) {
         this.name = name;
-        this.offerRequirementOptional = offerRequirementOptional;
+        this.offerRequirementOptional = offerRequirement;
```
nickbp: May as well drop the 'optional' on the internal value's name as well (dont care which way this goes, as long as the way is consistent for the file)

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
```diff
@@ -17,24 +20,23 @@
...
         this.status = status;
+        this.errors = errors;
     }
 
-    private boolean isStatus(Status status) {
-        return status.equals(this.status);
-    }
-
-    private void setTaskIds(Collection <Protos.Offer.Operation> operations) {
+    private synchronized void setTaskIds(Collection <Protos.Offer.Operation> operations) {
```
nickbp: Maybe specify in the class' javadoc that it's expected to be called from multiple threads, and specify which threads those would be. I'm assuming it's something like:
- Main execution thread
- Web service/Resource thread
- others?

src/main/java/org/apache/mesos/scheduler/plan/BlockFactory.java
```diff
@@ -11,6 +12,6 @@
...
  */
 public interface BlockFactory {
-    Block getBlock(String taskType, TaskSpecification taskSpecification)
-            throws InvalidRequirementException;
+    Block getBlock(TaskSpecification taskSpecification) throws Block.InvalidException;
+    List<Block> getBlocks(List<TaskSpecification> taskSpecifications) throws Block.InvalidException;
```
nickbp: Any reason for the switch? Looks like DefaultBlockFactory isn't really doing anything to take advantage of this at the moment.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
```diff
@@ -15,161 +19,82 @@
...
-            UUID id, String name, Collection<? extends Block> blocks) {
-        return new DefaultPhase(id, name, blocks);
+        return PlanUtils.getStatus(getChildren());
     }
 
-    public static Builder builder() {
-        return new Builder();
+    @Override
+    public void setStatus(Status status) {
+        logger.warn("Setting status does not effect status. Status is derived from Strategy and child statuses.");
```
nickbp: Would it make sense to just `throw new UnsupportedOperationException()` or similar here? Or is this expected to be called in practice?

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseBuilder.java
```diff
@@ -0,0 +1,26 @@
+package org.apache.mesos.scheduler.plan;
+
+import org.apache.mesos.scheduler.plan.strategy.DependencyStrategy;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Created by gabriel on 10/19/16.
```
nickbp: Fill real javadocs here and below

src/main/java/org/apache/mesos/scheduler/plan/ElementBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+    public ElementBuilder(String name) {
+        this.name = name;
+    }
+
+    public ElementBuilder add(C element) throws DependencyStrategyHelper.InvalidDependencyException {
+        dependencyStrategyHelper.addElement(element);
+        return this;
+    }
+
+    public ElementBuilder add(Collection<C> elements) throws DependencyStrategyHelper.InvalidDependencyException {
```
nickbp: addAll?

src/main/java/org/apache/mesos/scheduler/plan/ElementBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.mesos.scheduler.plan.strategy.DependencyStrategyHelper;
+
+import java.util.Collection;
+
+/**
+ * Created by gabriel on 10/19/16.
+ *
+ * @param <C> is the type of {@link Element}s to which the Strategy applies.
+ */
+public abstract class ElementBuilder<C extends Element> {
```
nickbp: Missing an abstract `build()`?

src/main/java/org/apache/mesos/scheduler/plan/Element.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.mesos.Protos.TaskStatus;
+import org.apache.mesos.scheduler.Observable;
+import org.apache.mesos.scheduler.plan.strategy.Strategy;
+
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * An Element is a generic component in a tree of elements.  It is the basic unit for the description of Plans used to
+ * launch Mesos Tasks. {@link Plan}s, @{link Phase}s and {@link Block}s are all implementations of Elements.  Elements
```
nickbp: `{@link`

src/main/java/org/apache/mesos/scheduler/plan/Element.java
```diff
@@ -0,0 +1,112 @@
...
+ * have Strategies (see: {@link Strategy} associated with them which describe when child elements should be presented as
```
nickbp: missing closing `)`

src/main/java/org/apache/mesos/scheduler/plan/Element.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    void setStatus(Status status);
+
+    /**
+     * Gets the children of an Element.
+     */
+    List<C> getChildren();
+
+    /**
+     * Gets the Strategy of an Element.
```
nickbp: Is this strategy for the _children_ of the element? (given that it's templated against C)

src/main/java/org/apache/mesos/scheduler/plan/api/BlockInfo.java
```diff
@@ -19,44 +17,33 @@
     private final Status status;
     private final String name;
     private final String message;
-    private final Boolean hasDecisionPoint;
 
     @JsonCreator
     public static BlockInfo create(
```
nickbp: Would it make sense to squash the `*Info`s into a single `ElementInfo`?

src/main/java/org/apache/mesos/scheduler/plan/strategy/DependencyStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.Collection;
+
+/**
+ * A DependencyStrategy implements the strategy designated by the {@link DependencyStrategyHelper} with which it was
+ * constructed.
+ *
+ * @param <C> is the type of {@link Element}s to which the Strategy applies.
+ */
+public class DependencyStrategy<C extends Element> extends InterruptableStrategy<C> {
```
nickbp: Given how small this is, any reason to have it distinct from DependencyStrategyHelper? Feels like Helper's logic could be merged into here.

tools/setup_permissions.sh
```diff
@@ -0,0 +1,68 @@
+SERVICE_ACCOUNT_NAME=service-acct
```
nickbp: unrelated? (also `*.pem` in `.gitignore`)

src/main/java/org/apache/mesos/scheduler/plan/Block.java
```diff
@@ -14,43 +13,13 @@
...
-            return Status.PENDING;
-        } else if (block.isInProgress()) {
-            return Status.IN_PROGRESS;
-        } else if (block.isComplete()) {
-            return Status.COMPLETE;
-        }
-        return Status.ERROR;
-    }
-
+public interface Block extends Element {
```
nickbp: Sorta surprised this works -- doesn't it complain that Element isn't being passed a template type?

src/main/java/org/apache/mesos/scheduler/plan/Element.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * An Element is a generic component in a tree of elements.  It is the basic unit for the description of Plans used to
+ * launch Mesos Tasks. {@link Plan}s, @{link Phase}s and {@link Block}s are all implementations of Elements.  Elements
+ * have Strategies (see: {@link Strategy} associated with them which describe when child elements should be presented as
+ * candidates for execution.
+ *
+ * @param <C> is the type of child {@link Element}s.
+ */
+public interface Element<C extends Element> extends Observable {
```
nickbp: Would it make sense to have a base `Element` class without anything to do with child elements (`Block` would be one of these), and then a `ParentElement extends Element` class that has the child element templating and `getChildren()`/`getStrategy()`?

src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -43,29 +43,45 @@
     private static final Integer PERMANENT_FAILURE_DELAY_SEC = 20 * 60;
     private static final Integer AWAIT_TERMINATION_TIMEOUT_MS = 10000;
     private final Logger logger = LoggerFactory.getLogger(getClass());
-    private final ServiceSpecification serviceSpecification;
     private final String zkConnectionString;
     private final ExecutorService executor = Executors.newFixedThreadPool(1);
     private final BlockingQueue<Collection<Object>> resourcesQueue;
+    private final String name;
```
mohitsoni: Style nit: What does this `name` represent ?  Service name ? Can we please rename this ?

src/main/java/org/apache/mesos/scheduler/plan/Block.java
mohitsoni: API Style nit: Can we also rename this to something more explicit ? Maybe `InvalidElementException` or `InvalidBlockException` ?

mohitsoni: 👍

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
```diff
@@ -17,24 +20,23 @@
     private final String name;
     private final Optional<OfferRequirement> offerRequirementOptional;
     private final UUID id = UUID.randomUUID();
+    private final List<String> errors;
+    private final Strategy strategy = new ParallelStrategy();
```
mohitsoni: Are these assumptions correct:
- Every Element now has a strategy
- A block by default has a a `ParallelStrategy` because we want to launch all the tasks (1 by default) that it composes at once.

src/main/java/org/apache/mesos/scheduler/plan/BlockFactory.java
gabrielhartmann: See here: https://github.com/mesosphere/dcos-commons/pull/220/files#diff-b30db19793859d5a98b8fddd400047eeR145

Are you asking about the method addition or the Exception?

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlockFactory.java
```diff
@@ -28,32 +28,46 @@ public DefaultBlockFactory(StateStore stateStore) {
...
-            return new DefaultBlock(
-                    taskSpecification.getName(),
-                    offerRequirementProvider.getNewOfferRequirement(taskType, taskSpecification),
-                    Status.PENDING);
-        } else {
-            try {
+
+        try {
+            if (!taskInfoOptional.isPresent()) {
+                logger.info("Generating new block for: " + taskSpecification.getName());
```
mohitsoni: Redundant log line...

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
```diff
@@ -15,161 +19,82 @@
...
-                ImmutableMap.builder();
-
-        for (int i = 0; i < this.blocks.size(); ++i) {
-            builder.put(this.blocks.get(i).getId(), i);
-            this.blocks.get(i).subscribe(this);
-        }
-        byId = builder.build();
+    @Override
+    public void update(Protos.TaskStatus status) {
+        PlanUtils.update(status, getChildren());
```
mohitsoni: Can we please add logging either here OR inside `PlanUtils` ?

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
```diff
@@ -15,161 +19,82 @@
...
     }
 
     @Override
-    public List<? extends Block> getBlocks() {
-        return blocks;
+    public void restart() {
+        PlanUtils.restart(getChildren());
```
mohitsoni: Can we please add logging either here OR inside `PlanUtils` ?

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
```diff
@@ -15,161 +19,82 @@
...
     }
 
     @Override
-    public Block getBlock(UUID id) {
-        if (id == null) {
-            return null;
-        }
-        return byId.containsKey(id) ? blocks.get(byId.get(id)) : null;
+    public void forceComplete() {
+        PlanUtils.forceComplete(getChildren());
```
mohitsoni: Can we please add logging either here OR inside `PlanUtils` ?

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseBuilder.java
mohitsoni: 👍

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseBuilder.java
```diff
@@ -0,0 +1,26 @@
...
+        super(name);
+    }
+
+    public DefaultPhase build() {
+        Set<Block> blocks = dependencyStrategyHelper.getDependencies().keySet();
+
+        return new DefaultPhase(
+                name,
+                new ArrayList<>(blocks),
+                new DependencyStrategy<>(dependencyStrategyHelper),
```
mohitsoni: Very subtle design nit: DefaultPhase will know about dependencyStrategyHelper which HAS `Set<Block> blocks`. Hence, doing `new ArrayList<>(blocks)` feels weird.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
gabrielhartmann: We can do that.  We'll just have to be careful not to call this, especially through HTTP endpoints.  Or I'm thinking we could set the same status on all child elements...  more consistent with the `forceComplete` and `restart` behavior.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanBuilder.java
```diff
@@ -0,0 +1,26 @@
+package org.apache.mesos.scheduler.plan;
+
+import org.apache.mesos.scheduler.plan.strategy.DependencyStrategy;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Created by gabriel on 10/19/16.
```
mohitsoni: Real doc please...

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseBuilder.java
gabrielhartmann: Will do, working on it.

src/main/java/org/apache/mesos/scheduler/plan/ElementBuilder.java
gabrielhartmann: done

src/main/java/org/apache/mesos/scheduler/plan/ElementBuilder.java
gabrielhartmann: done

src/main/java/org/apache/mesos/scheduler/plan/Element.java
gabrielhartmann: done

src/main/java/org/apache/mesos/scheduler/plan/Element.java
gabrielhartmann: done

src/main/java/org/apache/mesos/scheduler/plan/Element.java
gabrielhartmann: Yes, it is.

src/main/java/org/apache/mesos/scheduler/plan/api/BlockInfo.java
gabrielhartmann: Yes.  We spoke about this.  This PR, is too big.  Will follow up.  https://mesosphere.atlassian.net/browse/INFINITY-624

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
mohitsoni: Got it.

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseFactory.java
```diff
@@ -1,38 +1,68 @@
...
+    }
 
-        return blocks;
+    private List<Block> getBlocks(TaskSet taskSet) {
+        return taskSet.getTaskSpecifications().stream()
+                .map(taskSpec -> {
+                    try {
+                        return blockFactory.getBlock(taskSpec);
+                    } catch (Block.InvalidBlockException e) {
+                        return new DefaultBlock(
```
mohitsoni: Just curious about implication of this. If there was a failure in getting a block from a taskSpec it should just fail, right ? Or the implication here is that if there's any block that's in `ERROR` state, the plan will be in error state (status propagates upwards) and will not make any progress, but it will be used to tell the operator via API that there's an error ?

src/main/java/org/apache/mesos/scheduler/plan/DefaultPlanManager.java
```diff
@@ -38,235 +19,14 @@ public Plan getPlan() {
...
-        }
-
-        for (Phase phase : plan.getPhases()) {
-            if (phaseId.equals(phase.getId())) {
-                return phase;
-            }
-        }
+    public Collection<? extends Block> getCandidates(Collection<String> dirtyAssets) {
+        Collection<Phase> candidatePhases = plan.getStrategy().getCandidates(plan, dirtyAssets);
+        Collection<Block> candidateBlocks = candidatePhases.stream()
```
mohitsoni: Doesn't this break the execution order amongst phases ? This will work just fine for a single phase and multiple parallel phases. But, will break the execution order when serial phase execution is desired ? OR this relies on the underlying `PhaseStrategy` to do the right thing ?

src/main/java/org/apache/mesos/scheduler/plan/Element.java
mohitsoni: @nickbp Sounds interesting. Can you give an example ?

nickbp: I had been thinking there could be two different types for leaf nodes vs branch nodes. We'd talked about this offline and it ended up being really messy to have the separate types.

src/main/java/org/apache/mesos/scheduler/plan/BlockFactory.java
gabrielhartmann: I don't understand this question.

nickbp: I mean, which TaskSpecifications are supposed to be passed in a single call to `BlockFactory.getBlocks()`, in practice?

All TaskSpecifications across the service? Or just the TaskSpecifications from a single TaskSet? Or something else?

(I think this is something that a developer writing their own BlockFactory implementation would be wondering)

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlock.java
gabrielhartmann: Added an explanation regarding why it is synchronized.

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlockFactory.java
gabrielhartmann: Redundant in what sense?  There are four log lines.

1. Trying
2. New
3. Existing
4. Failed

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
gabrielhartmann: done

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
gabrielhartmann: done

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhase.java
gabrielhartmann: done

src/main/java/org/apache/mesos/scheduler/plan/DefaultPhaseFactory.java
gabrielhartmann: Yes, the second one.  The failure to construct a Block won't crash the Scheduler, but it will provide an ERROR state which will be handled appropriately.  It'll roll up to the Phase and Plan as a whole, where strategies should do the appropriate thing depending on their internal logic.  As implemented today anything in ERROR is not considered a candidate by any strategy.

gabrielhartmann:
@nickbp and @mohitsoni thanks for all the helpful review comments.  I think I've addressed them all except for one I didn't understand.  Please let me know if there's anything else you'd like.

src/main/java/org/apache/mesos/scheduler/plan/DefaultBlockFactory.java
mohitsoni: 36: logger.info("Generating block for: " + taskSpecification.getName());
41: logger.info("Generating new block for: " + taskSpecification.getName());

Oh, is `new` basically a cue that this method generated a branch new block v/s an existing block ?

nickbp:
The only thing I have left is with the new getBlocks() function in BlockFactory.java. Feels like the function doesn't add any benefit (unless there's some future benefit that it's looking ahead towards), and brings up questions about the contexts in which it'd be called instead of singular getBlock(). Added a comment but it apparently got buried by Github.

gabrielhartmann:
Cool, thanks @nickbp.  Addressed the BlockFactory comment.  Anything left @mohitsoni?

nickbp:
:+1: for me, Mohit may have other comments

src/main/java/org/apache/mesos/specification/TaskSpecification.java
```diff
@@ -16,6 +16,8 @@
      */
     String getName();
 
+    String getType();
```
mohitsoni: Can you please add docs ?

src/main/java/org/apache/mesos/scheduler/plan/strategy/Strategy.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * A strategy indicates which {@link Element}s are ready to be processed.  It may be interrupted such that it will
+ * indicate that no elements are available for processing.  To resume normal operation {@link #proceed()} may be called.
+ *
+ * @param <C> is the type of {@link Element}s to which the Strategy applies.
+ */
+public interface Strategy<C extends Element> {
+    Collection<C> getCandidates(Element<C> parentElement, Collection<String> dirtyAssets);
+
+    void interrupt();
```
mohitsoni: Can you please add JavaDocs to all the methods, as this is the top-level interface.

src/main/java/org/apache/mesos/scheduler/plan/strategy/CanaryStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public Collection<C> getCandidates(Element<C> parentElement, Collection<String> dirtyAssets) {
+        if (!initialized.get()) {
+            children = parentElement.getChildren().stream()
+                    .filter(element -> element.isPending())
+                    .collect(Collectors.toList());
+
+            for (int i = 0; i < 2; i++) {
+                children.get(i).getStrategy().interrupt();
```
mohitsoni: For a single node deployment, this will throw Index out of bounds ?

gabrielhartmann: Good catch, thanks.  Fixed.

PR #221, opened 2016-10-19
Adds config management support on executor side

NimaVaziri:
This gets the config data on the executor side and writes it to disk. Most of the logic for this specific functionality is at the end of `TaskUtils.java`

- [x] Live test
- [ ] Add tests

Rebased on #198 and #212 

@gabrielhartmann @nickbp


nickbp:
needs rebase: #198 is merged

PR #225, opened 2016-10-20
make github update optional

mgummelt:
cc @nickbp 

explicitly adding GITHUB_DISABLE shouldn't be necessary.  The update script should be smart enough to know when it can and can't update.


nickbp:
this duplicates existing checks at L169

mgummelt:
It's not duplicate.  This handles the case where the it is a CI job, but it doesn't come from a GITHUB PR (e.g. tagged release).

But I updated the code to be more consistent with the existing checks.

nickbp:
Oh got it. Makes sense

PR #227, opened 2016-10-21
Now detecting DC/OS version using on-disk file

mohitsoni:
Each DC/OS node has a file at location: `/opt/mesosphere/active/dcos-metadata/etc/dcos-version.json`. Use this file to determine DC/OS version, and do so in best-effort fashion.

### DIFF
--- a/src/main/java/org/apache/mesos/dcos/DcosCluster.java
+++ b/src/main/java/org/apache/mesos/dcos/DcosCluster.java
@@ -1,19 +1,28 @@
 package org.apache.mesos.dcos;
 
-import org.apache.http.client.fluent.Content;
-import org.apache.http.client.fluent.Request;
+import org.apache.commons.io.Charsets;
 import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Optional;
 
 /**
  * Instances of this class represent DC/OS clusters.
  */
 public class DcosCluster {
-    private static final String DCOS_VERSION_PATH = "/dcos-metadata/dcos-version.json";
+    private static final Logger LOGGER = LoggerFactory.getLogger(DcosCluster.class);
+
+    /**
+     * Location of file on DC/OS cluster that can be used to extract version information.
+     */
+    public static final String DCOS_VERSION_PATH = "/opt/mesosphere/active/dcos-metadata/etc/dcos-version.json";
 
     private final URI dcosUri;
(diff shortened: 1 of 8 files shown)


mohitsoni:
@nickbp It doesn't fall back to HTTP. In fact the HTTP endpoint in context, basically just serves the contents of this file from disk. This file is basically present on all DC/OS nodes.

Are you saying that this file did not exist on all hosts pre 1.8 ?

Also, if this file is not found then the library gives up all hopes on determining DC/OS version and any version based capability detection logic return false.

mohitsoni:
@mesosphere-ci retest this please.

nickbp:
I didn't know either way, just wanted to check that the new system would work for all 1.7+. From offline discussion it sounds like we're good.

src/main/java/org/apache/mesos/dcos/DcosCluster.java
```diff
@@ -1,19 +1,28 @@
...
  * Instances of this class represent DC/OS clusters.
  */
 public class DcosCluster {
-    private static final String DCOS_VERSION_PATH = "/dcos-metadata/dcos-version.json";
+    private static final Logger LOGGER = LoggerFactory.getLogger(DcosCluster.class);
+
+    /**
+     * Location of file on DC/OS cluster that can be used to extract version information.
+     */
+    public static final String DCOS_VERSION_PATH = "/opt/mesosphere/active/dcos-metadata/etc/dcos-version.json";
```
triclambert: Should we take another dependency on /opt/mesosphere?

mohitsoni: @triclambert There are only two ways to determine DC/OS version:
1. Read `/opt/mesosphere/active/dcos-metadata/etc/dcos-version.json` on the host.
2. HTTP GET `<admin-router-url>/dcos-metadata/dcos-version.json` which serves the same file in step 1.

Both approaches are similar in terms on dependency on DC/OS. 

This PR implements this dependency carefully, and doesn't bleed failure, which still makes this safe for use in non-DC/OS environments.

Thoughts ?

gabrielhartmann: @mohitsoni I think what @triclambert is saying is that in general we're not supposed to access anything `/opt/mesosphere` according to the Foundations team.  Of course we violate this for at least `/opt/mesosphere/libmesos.so`.  This change was originally intended to avoid problems reaching the endpoint in strict mode clusters right?

Maybe we should still attempt to hit the HTTP endpoint and degrade gracefully, instead of looking for a local file...

mohitsoni: Got it. I'm going to update the PR, and call out of review again.

mohitsoni:
Closing this in favor of https://github.com/mesosphere/dcos-commons/pull/238. Once the certificate is installed in the JRE, the existing version detection logic works just fine.

PR #228, opened 2016-10-21
Include shell scripts in tools release tgz

nickbp:
It'll technically includes `release.sh` itself but whatever


nickbp:
Going to merge. Already successfully tested locally. The CLI build failure would be unrelated and is likely due to today's DNS issues

PR #230, opened 2016-10-21
Add support for --strict flag

mohitsoni:


mgummelt:
@mrbrowning @mohitsoni We use env vars everywhere else for program configuration.  Please stay consistent, and don't use `--args`

PR #233, opened 2016-10-24
Fix multiple Task failure bug. [INFINITY-628]

gabrielhartmann:
@mohitsoni:  Tested on sdk-reference-framework with 1s sleep and plan behaves as expected, and framework continuously recovers.


mohitsoni:
This is looking much better now.

PR #238, opened 2016-10-25
DCOS Certificate Installer

mohitsoni:
This PR implements a DC/OS certificate installer, which installs the `.ssl/ca.crt` file injected inside the sandbox.


mohitsoni:
retest this please

nickbp:
Add unit tests? Can do something like PR #226 to mock out the execution of the command itself.

mohitsoni:

mohitsoni:
retest this please

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
```diff
@@ -0,0 +1,83 @@
...
+
+            final Path jreDefaultKeystorePath = Paths.get(pathToJRE, "/lib/security/cacerts");
+            final String jreDefaultKeystoreAbsolutePath = jreDefaultKeystorePath.toAbsolutePath().toString();
+
+            final Path keytoolPath = Paths.get(pathToJRE, "/bin/keytool");
+            final String keytoolAbsolutePath = keytoolPath.toAbsolutePath().toString();
+
+            final String certAbsolutePath = sandboxCertPath.toAbsolutePath().toString();
+
+            String command = keytoolAbsolutePath + " -importcert -noprompt";
```
nickbp: Maybe just do a single String.format() call for building the command?

something like `String command = String.format("%s -importcert -noprompt -alias dcoscert -keystore %s -file %s -storepass %s", paths, here)`

mohitsoni:
retest this please

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            final Path sandboxCertPath = Paths.get(CERT_PATH);
+            // Check if cert is present, there's nothing to do.
+            if (!Files.exists(sandboxCertPath)) {
+                LOGGER.info("Cert file not found at path: {}", sandboxCertPath.toAbsolutePath());
+                return false;
+            }
+
+            final Path jreDefaultKeystorePath = Paths.get(pathToJRE, "/lib/security/cacerts");
```
nickbp: Could have an explicit sanity check that jreDefaultKeystorePath is a valid path to a regular file with write permission (or read permission?)

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
```diff
@@ -0,0 +1,83 @@
...
+            final String jreDefaultKeystoreAbsolutePath = jreDefaultKeystorePath.toAbsolutePath().toString();
+
+            final Path keytoolPath = Paths.get(pathToJRE, "/bin/keytool");
```
nickbp: Maybe include an explicit sanity check that keytoolPath exists and has execute permissions?

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
```diff
@@ -0,0 +1,83 @@
...
+            command += " -storepass " + DEFAULT_JRE_KEYSTORE_PASS;
+
+            LOGGER.info("Installing DC/OS cert using command: {}", command);
+
+            final ProcessBuilder processBuilder = new ProcessBuilder("/bin/sh", "-c", command).inheritIO();
+            final int exitCode = processRunner.run(processBuilder, 10);
+            LOGGER.info("Certificate install process completed with exit code: {}", exitCode);
+            return exitCode == 0;
+        } catch (Throwable t) {
+            LOGGER.error("Error installing cert", t);
```
nickbp: could include pathToJRE in log message

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
```diff
@@ -0,0 +1,83 @@
...
+        return installCertificate(pathToJRE, new ProcessRunner());
+    }
+
+    @VisibleForTesting
+    protected static boolean installCertificate(String pathToJRE, ProcessRunner processRunner) {
+        try {
+            final Path jrePath = Paths.get(pathToJRE);
+            // Check if JRE is present, there's nothing to do.
+            if (!Files.exists(jrePath)) {
+                LOGGER.info("JRE not found at path: {}", jrePath.toAbsolutePath());
```
nickbp: error()

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
```diff
@@ -0,0 +1,83 @@
...
+                return false;
+            }
+
+            final Path sandboxCertPath = Paths.get(CERT_PATH);
+            // Check if cert is present, there's nothing to do.
+            if (!Files.exists(sandboxCertPath)) {
+                LOGGER.info("Cert file not found at path: {}", sandboxCertPath.toAbsolutePath());
```
nickbp: error()

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
mohitsoni: This file location is standard for the JRE. Also, the file should be writable by default if fetcher does the right thing, which it should.

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
mohitsoni: Same as above.

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
```diff
@@ -0,0 +1,83 @@
...
+            command += " -alias dcoscert";
+            command += " -keystore " + jreDefaultKeystoreAbsolutePath;
+            command += " -file " + certAbsolutePath;
+            command += " -storepass " + DEFAULT_JRE_KEYSTORE_PASS;
+
+            LOGGER.info("Installing DC/OS cert using command: {}", command);
+
+            final ProcessBuilder processBuilder = new ProcessBuilder("/bin/sh", "-c", command).inheritIO();
+            final int exitCode = processRunner.run(processBuilder, 10);
+            LOGGER.info("Certificate install process completed with exit code: {}", exitCode);
```
nickbp: could check the code value and log this as a warning when the value is non-zero

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
nickbp: Could there be problems with file access if the process is running as `nobody`, or is the mesos fetcher supposed to chmod the output of the tarball?

mohitsoni: `keytool` resides in the same `/bin` directory as `java` executable, and shares the same permissions. And, because the scheduler is run using `java` executable, the fact that we reach this line of code, means that we'll not have trouble running it.

sdk/core/src/main/java/org/apache/mesos/dcos/DCOSCertInstaller.java
```diff
@@ -0,0 +1,83 @@
...
+                    DEFAULT_JRE_KEYSTORE_PASS);
+
+            LOGGER.info("Installing DC/OS cert using command: {}", cmd);
+
+            final ProcessBuilder processBuilder = new ProcessBuilder("/bin/sh", "-c", cmd).inheritIO();
+            final int exitCode = processRunner.run(processBuilder, 10);
+            LOGGER.info("Certificate install process completed with exit code: {}", exitCode);
+            return exitCode == 0;
+        } catch (Throwable t) {
+            LOGGER.error("Error installing cert inside JRE: {}. Reason: {}", pathToJRE, t);
```
nickbp: As a workaround to ensure that `LOGGER.error(String, Throwable)` is called, this could be written as: `LOGGER.error(String.format("Error installing cert inside JRE: %s", pathToJRE), t);`

mohitsoni:
Done!

mohitsoni:
retest this please

PR #239, opened 2016-10-25
more beautiful architecture diagram

sascala:
@mrbrowning I'm creating a PR just so you can check that the lovelier diagram is also accurate.


mrbrowning:
Looks good!

PR #241, opened 2016-10-26
Config tardigrade

gabrielhartmann:
Adding new construction options such that Tardigrade can be configured in the DefaultScheduler.


sdk/core/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -76,14 +81,60 @@ public static DefaultScheduler create(String frameworkName, PlanManager deployme
...
+            String zkConnectionString,
+            boolean destructiveRecoveryEnabled,
+            Integer permanentFailureTimeout,
+            Integer destructiveRecoveryDelay) {
         this.frameworkName = frameworkName;
         this.deployPlanManager = deploymentPlanManager;
         this.zkConnectionString = zkConnectionString;
         this.resourcesQueue = new ArrayBlockingQueue<>(1);
+        this.destructiveRecoveryEnabled = destructiveRecoveryEnabled;
+        this.permanentFailureTimeout = permanentFailureTimeout;
```
nickbp: permanentFailureTimeoutSeconds

gabrielhartmann: Fixed both names to include `Sec`.

PR #242, opened 2016-10-26
Env pass resource requirement

mbdcos:
Made modifications in ResourceRequirements. Enabled passing any resource value as an environment variable to the application. EnvName is sync'ed to the reservation labels. 
Modifications in OfferEvaluation as well.
specific problems: dynamic ports and VIP setup.
IF port value is 0, it is processed differently in the Offer evaluation. A port value is assigned and ResourceRequirement is updated.
VIP setup is done by providing a label. Since port value can be empty, VIP setup (adding to TaskInfo Discovery label) is delayed till and offer is found - fulfilledRequirement.
While returning from Offer evaluate I still use Resource, not ResourceRequirement. This does unnecessary conversion. Intentionally done like that for the moment because Resource object are passed in general in the cod. This helps to make sure Resource labels are properly populated, so conversion can happen with no information change.
ResourceRequirement add to the reservation labels with a predefined key (to save the bookkeeping information). If label key (i.e. ENV_KEY or VIP_KEY) exits, it overwrites them.


src/main/java/org/apache/mesos/offer/MesosResourcePool.java
```diff
@@ -89,38 +92,23 @@ public Offer getOffer() {
...
-                                        .setBegin(range.getBegin())
-                                        // Use getBegin again, since we just want the one port.
-                                        .setEnd(range.getBegin())))
-                        .build());
-
-            return consumeUnreservedMerged(new ResourceRequirement(resource));
-        }
+            resReq.addPort((int) range.getBegin());
+            return consumeUnreservedMerged(resReq);
+        }//if
```
gabrielhartmann: Please remove this comment.

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
```diff
@@ -89,38 +92,23 @@ public Offer getOffer() {
...
-                        .setRanges(Value.Ranges.newBuilder()
-                                .addRange(Value.Range.newBuilder()
-                                        .setBegin(range.getBegin())
-                                        // Use getBegin again, since we just want the one port.
-                                        .setEnd(range.getBegin())))
-                        .build());
-
-            return consumeUnreservedMerged(new ResourceRequirement(resource));
-        }
+            resReq.addPort((int) range.getBegin());
```
gabrielhartmann: This shouldn't be part of the ResourceRequirement.  It should be part of the FulfilledRequirement.

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
```diff
@@ -223,6 +211,7 @@ private void releaseAtomicResource(MesosResource mesosResource) {
 
         if (sufficientValue(desiredValue, availableValue)) {
             unreservedMergedPool.put(resourceRequirement.getName(), ValueUtils.subtract(availableValue, desiredValue));
+            //// !!!! HERE We OVERWRITE THE RESOURCE AND ITS RESERVATION  /// TODO
```
gabrielhartmann: Please remove this comment.  I believe this code is consuming an unreserved resource in any case.  It should not have a reservation.

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -63,6 +63,20 @@ public OfferEvaluator(StateStore stateStore) {
         return Collections.emptyList();
     }
 
+    public List<OfferRecommendation> evaluate(OfferRequirement offerRequirement, Offer offer)
```
gabrielhartmann: I believe this function was removed because nothing called it.  Is it called today?

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -317,16 +305,17 @@ private FulfilledRequirement(
         public List<OfferRecommendation> getCreateRecommendations() {
             return createRecommendations;
         }
+
```
gabrielhartmann: This relies on modified `ResourceRequirement`s which shouldn't occur.

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -379,8 +367,7 @@ private static Resource getFulfilledResource(ResourceRequirement resReq, MesosRe
         return Optional.of(builder.build());
     }
```
gabrielhartmann: Please remove commented code.

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -379,8 +367,7 @@ private static Resource getFulfilledResource(ResourceRequirement resReq, MesosRe
...
 
-    private static Optional<ReservationInfo> getFulfilledReservationInfo(
-            ResourceRequirement resReq, MesosResource mesRes) {
+    private static Optional<ReservationInfo> getFulfilledReservationInfo(ResourceRequirement resReq) {
         if (!resReq.reservesResource()) {
```
gabrielhartmann: Please remove extra newlines.

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -426,15 +414,26 @@ private static TaskInfo getFulfilledTaskInfo(
                     .newBuilder(execInfo.get())
                     .clearResources()
```
gabrielhartmann: Please remove commented code.

src/main/java/org/apache/mesos/offer/RequirementUtils.java
```diff
@@ -1,55 +1,51 @@
...
-    Collection<String> resourceIds = new ArrayList<String>();
+    public static Collection<String> getResourceIds(Collection<ResourceRequirement> resourceRequirements) {
+        Collection<String> resourceIds = new ArrayList<String>();
 
-    for (ResourceRequirement resReq : resourceRequirements) {
-      if (resReq.expectsResource()) {
-        resourceIds.add(resReq.getResourceId());
-      }
+        for (ResourceRequirement resReq : resourceRequirements) {
+            if (resReq.expectsResource()) {
```
gabrielhartmann: Please remove commented code.

src/main/java/org/apache/mesos/offer/ResourceRequirement.java
```diff
@@ -15,12 +20,103 @@
  * or an already determined value.
  */
 public class ResourceRequirement {
+    /* keys for log */
+    public static final String ENV_KEY = "Environment_Name";
+    public static final String VIP_KEY = "VIP_KEY";
+    public static final String VIP_VALUE = "VIP_VALUE";
+
+
```
gabrielhartmann: Please remove extra newline.

src/main/java/org/apache/mesos/offer/ResourceRequirement.java
```diff
@@ -15,12 +20,103 @@
...
 
+    private Optional<String> envName = Optional.empty();
+    private Optional<Label> vipLabel = Optional.empty();
+
     public ResourceRequirement(Resource resource) {
         this.mesosResource = new MesosResource(resource);
         this.diskInfo = getDiskInfo();
+        ResourceUtils.setEnvName(this);
+        ResourceUtils.setVIPLabel(this);
+
```
gabrielhartmann: Please remove extra newline.

src/main/java/org/apache/mesos/offer/ResourceRequirement.java
```diff
@@ -15,12 +20,103 @@
  * or an already determined value.
  */
 public class ResourceRequirement {
```
gabrielhartmann: VIP and Port information is not relevant to every `ResourceRequirement`.  A sub-class (or some abstraction) should be used to add the information only where relevant.

src/main/java/org/apache/mesos/offer/ResourceRequirement.java
```diff
@@ -15,12 +20,103 @@
...
+        if (hasVIPLabel()) {
+            return vipLabel.get();
+        }
+        return null;
+    }
+
+
+    public void setEnvName(String envName) {
+        this.envName = Optional.of(envName);
+        if (!this.envName.isPresent()) {
```
gabrielhartmann: How could this condition ever occur?  If `envName` is null?  You can't create an `Optional.of(null)`.  It will throw.

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -203,66 +204,155 @@ public static Resource getExpectedRanges(
...
+                                                   List<Resource> resourceList) {
+        TaskInfo.Builder taskBuilder = builder;
+
+        for (Resource resource : resourceList) {
+            ResourceRequirement resReq = new ResourceRequirement(resource);
+                taskBuilder = setVIPDiscovery(taskBuilder, execName, resReq);
+        }
+        return taskBuilder;
+    }
+
```
gabrielhartmann: please rename to `setVipDiscovery`.  A different argument list is sufficient to distinguish from the above.

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -203,66 +204,155 @@ public static Resource getExpectedRanges(
...
+    public static TaskInfo.Builder setVIPDiscovery2(TaskInfo.Builder builder, String execName,
+                                                    List<ResourceRequirement> resReqList) {
+        TaskInfo.Builder taskBuilder = builder;
+
+        for (ResourceRequirement resReq : resReqList) {
+            taskBuilder = setVIPDiscovery(taskBuilder, execName, resReq);
+        }
+        return taskBuilder;
+    }
+
```
gabrielhartmann: Please format this as a `// TODO: ` and file a JIRA capturing the work you think needs to be done here.

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -203,66 +204,155 @@ public static Resource getExpectedRanges(
...
+    /* what if we have multiple ports, or multiple VIPs, we need to work on here !!!! */ //FIX Later
```
gabrielhartmann: I don't understand this variable name.  `execName`?  Is it from Executor?  There are no executors in this code.

src/main/java/org/apache/mesos/offer/ResourceUtils.java
```diff
@@ -203,66 +204,155 @@ public static Resource getExpectedRanges(
...
+                    .build();
+            builderArg.setDiscovery(discoveryInfo);
+            LOGGER.info(String.format("Discovery VIP is set for port resource  %s ",
+                    resReq.getValue()));
+        } catch (Exception e) {
+            LOGGER.error(String.format("Discovery VIP is NOT set for port resource  %s ",
+                    resReq.getValue()));
+            return builderArg;
+        }
+        return builderArg;
```
gabrielhartmann: newline needed here

src/test/java/org/apache/mesos/offer/DynamicPortRequirementTest.java
```diff
@@ -5,19 +5,26 @@
...
+                TestConstants.ROLE, TestConstants.PRINCIPAL,
+                Optional.of(TestConstants.PORT_NAME));
+
+        dynamicPortRequirement1.setVIPLabel( Protos.Label.newBuilder().setKey("some_key").setValue("some_value").build());
+
+        Assert.assertNotNull(dynamicPortRequirement1);
+        Boolean flag=dynamicPortRequirement1.getResource().getReservation().getLabels().getLabelsList().contains(
+                Protos.Label.newBuilder().setKey(ResourceRequirement.VIP_KEY).setValue("some_key").build());
+
+        Assert.assertEquals(true, flag);
```
gabrielhartmann: `Assert.assertTrue`

src/main/java/org/apache/mesos/offer/ResourceRequirement.java
mbdcos: can be replaced with ofNullable to be safe. If this is called with null, then there is a bug - I prefer it throw  in that case

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
mbdcos: did not understand the comment

src/main/java/org/apache/mesos/offer/OfferEvaluator.java
mbdcos: ResourceRequirement is immutable? there is non-final object there already. I did not see any comment saying this should be immutable :)

src/main/java/org/apache/mesos/offer/MesosResourcePool.java
mbdcos: public Optional<MesosResource> consume(ResourceRequirement resourceRequirement) 


    public Optional<MesosResource> consumeDynamicPort(ResourceRequirement resReq){

public Optional<MesosResource> consume(ResourceRequirement resourceRequirement) 

public Optional<MesosResource> consumeDynamicPort(ResourceRequirement resReq)

PR #244, opened 2016-10-26
Update serialization utilities, fixes to sdk exceptions, config updater

nickbp:
- Misc cleanup resulting from config management work
- Replaces previous config validation interfaces and adds underlying config update class which calls those validators


sdk/core/build.gradle
```diff
@@ -9,8 +9,9 @@ ext {
     dropwizardVer = "0.9.2"
     findbugsAnnotationsVer = "3.0.1u2"
     guiceVer = "3.0"
-    jacksonVer = "2.6.3"
     hamcrestVer = "1.3"
+    jacksonVer = "2.6.3"
+    jacksonProtobufVer = "0.9.3"
```
nickbp: Invoked by SerializationUtils. Technically not used yet, will be used soon for serialization of ServiceSpecs.

sdk/core/src/main/java/org/apache/mesos/config/ConfigStore.java
```diff
@@ -26,14 +26,15 @@
...
      *
      * @param id The UUID of the configuration to be fetched
-     * @param factory The deserializer implementation for the configuration data
      * @return The deserialized configuration
-     * @throws ConfigStoreException if retrieval or deserialization fails
+     * @throws ConfigStoreException if retrieval or deserialization fails, or if the requested
+     *                              config is missing
      */
-    T fetch(UUID id, ConfigurationFactory<T> factory) throws ConfigStoreException;
+    T fetch(UUID id) throws ConfigStoreException;
```
nickbp: The factory param was originally because only this fetch() command actually used the factory. Moved it to the CuratorConfigStore constructor since:
- it's leaking a bit of implementation detail into the interface
- when are you going to want a write-only ConfigStore anyway? may as well just pass it to the constructor up-front.

sdk/core/src/main/java/org/apache/mesos/config/ConfigurationComparer.java
```diff
@@ -0,0 +1,14 @@
+package org.apache.mesos.config;
+
+/**
+ * This interface defines a way to compare two Configuration objects for equality. Intentionally
+ * separate from {@link Object#equals(Object)} to be very explicit about this requirement.
+ *
+ * @param <T> The {@code Configuration} type that will be parsed.
+ */
+public interface ConfigurationComparer<T extends Configuration> {
```
gabrielhartmann: `ConfigurationComparator` as here: https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html

nickbp: Good point, I'd thought of doing the same thing but wasn't sure. Renamed.

sdk/core/src/main/java/org/apache/mesos/config/ConfigurationUpdater.java
```diff
@@ -0,0 +1,230 @@
...
+                // one, and we'd be stuck with this config forever! Hopefully the new config will fix things...
+            }
+        } else {
+            LOGGER.info("Old target config: <null>");
+        }
+        try {
+            LOGGER.info("New prospective config: {}", candidateConfig.toJsonString());
+        } catch (Exception e) {
+            LOGGER.error("Unable to get JSON representation of new prospective config object: " + candidateConfig, e);
+            errors.add(ConfigurationValidationError.valueError("NewConfigAsJson", "jsonString",
```
gabrielhartmann: Should we short-circuit here?  We continue to attempt validation anyway right now.

gabrielhartmann: I see we want to continue to the if cases below.  Ok.  Maybe skip validation though?

sdk/core/src/main/java/org/apache/mesos/config/ConfigurationUpdater.java
```diff
@@ -0,0 +1,230 @@
...
+                    "configuration {}, with {} errors across {} validators:\n{}",
+                    targetConfigId, errors.size(), validators.size(), sj.toString());
+            if (targetConfig == null) {
+                throw new ConfigStoreException(String.format(
+                        "Configuration failed validation without any prior target configuration" +
+                        "available for fallback. Initial launch with invalid configuration? " +
+                        "%d Errors: %s", errors.size(), sj.toString()));
+            }
+        } else if (targetConfig == null || !configComparer.equals(targetConfig, candidateConfig)) {
+            LOGGER.info("Changes detected between current target configuration '{}' and new " +
```
gabrielhartmann: Is it easy to get a diff in String form of the config change here?  Would be really helpful in logs.  If it's a major effort forget I asked.  I can be done later, and can be learned anyway by hitting config endpoints.

sdk/core/src/main/java/org/apache/mesos/config/ConfigurationComparator.java
```diff
@@ -0,0 +1,15 @@
+package org.apache.mesos.config;
+
+/**
+ * This interface defines a way to compare two {@link Configuration} objects for equality. This
+ * interface is intentionally separate from {@link Object#equals(Object)} in order to have very
+ * explicit requirements around equality checks when they're needed.
+ *
+ * @param <T> The {@code Configuration} type that will be parsed.
+ */
+public interface ConfigurationComparator<T extends Configuration> {
```
mohitsoni: Nit: Did you consider using Java's Comparator instead ?

nickbp: Comparator is slightly different, it also requires `int compare(T a, T b)`

This is why I'd gone with 'Comparer' to begin with but that creates its own confusion.

mohitsoni: Got it. Java's interface is a bit weird, there's `compare` and there's a `equals` which orverrides Object's `equals`. 
This will do for now. I was just trying to figure out if we can avoid duplication of interfaces.

sdk/core/src/main/java/org/apache/mesos/config/ConfigurationUpdater.java
nickbp: Old and new configs are both outputted in JSON form earlier in the file. (lines 98-115)

sdk/core/src/main/java/org/apache/mesos/config/ConfigurationUpdater.java
```diff
@@ -0,0 +1,230 @@
...
+        // Update config IDs on tasks whose config contents match the current target, then clean up
+        // leftover configs which are not the target and which are not referenced by any tasks.
+        cleanupDuplicateAndUnusedConfigs(targetConfig, targetConfigId);
+
+        return new UpdateResult(targetConfigId, errors);
+    }
+
+    /**
+     * Searches for any task configurations which are already identical to the target configuration
+     * and updates the accounting for those tasks to point to the current target configuration.
```
nickbp: Updated to clarify with 'the embedded config version label in those tasks'

sdk/core/src/main/java/org/apache/mesos/config/ConfigurationUpdater.java
nickbp: May as well still invoke the validators even with this toJson call failing -- all validation errors including this one will be logged in a nice format further down (see StringJoiner usage)

sdk/core/src/main/java/org/apache/mesos/config/ConfigurationUpdater.java
```diff
@@ -0,0 +1,230 @@
...
+            LOGGER.info("New prospective config: {}", candidateConfig.toJsonString());
+        } catch (Exception e) {
+            LOGGER.error("Unable to get JSON representation of new prospective config object: " + candidateConfig, e);
+            errors.add(ConfigurationValidationError.valueError("NewConfigAsJson", "jsonString",
+                    String.format("Unable to serialize new config to JSON for logging: %s", e.getMessage())));
+        }
+
+        // Check for any validation errors (including against the prior config, if one is available)
+        // NOTE: We ALWAYS run validation regardless of config equality. This allows the developer
+        // to e.g. require that a value be increased, whereas a short-circuit on equality would
```
nickbp: Clarified comment.

PR #245, opened 2016-10-26
Proactive Asset Dirtying

mohitsoni:
This PR determines all known dirty assets pro-actively. This is used to ensure that PlanManagers that are presented with offers first, does not accidentally schedule an asset that's actively being worked upon by another PlanManager that is presented offers later.


sdk/core/src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -151,4 +148,21 @@ private Block createBlock(Protos.TaskInfo taskInfo)
                 recoveryRequirements.get(0),
                 launchConstrainer);
     }
+
+    @Override
+    public Set<String> getDirtyAssets() {
+        Set<String> dirtyAssets = new HashSet<>();
+        if (plan != null) {
+            final List<? extends Phase> phases = plan.getChildren();
```
gabrielhartmann: I know this is rich coming from me, but can't this be done more succinctly in a streaming style?

sdk/core/src/test/java/org/apache/mesos/scheduler/plan/DefaultPlanManagerTest.java
```diff
@@ -208,6 +205,95 @@ public void testUpdate() {
...
+                Arrays.asList(block1, block2),
+                new SerialStrategy<>(),
+                Collections.emptyList());
+
+        DefaultPlan waitingPlan = new DefaultPlan(
+                "test-plan",
+                Arrays.asList(phase),
+                new SerialStrategy<>(),
+                Collections.emptyList()
+        );
```
gabrielhartmann: inconsistent style.  Generally we leave that last `);` on the previous line as on line 218

mohitsoni:
Addressed both comments!

PR #246, opened 2016-10-26
Support JSON serialization of ServiceSpecification (and all members)

nickbp:
Mainly boilerplate work to make things serializable by jackson, particularly with the PlacementConstraints.

Includes tests to validate that serialization and equality work as expected.


sdk/core/src/main/java/org/apache/mesos/offer/constrain/AgentRule.java
```diff
@@ -34,52 +40,162 @@ public String toString() {
         return String.format("AgentRule{agentId=%s}", agentId);
     }
 
+    @Override
+    public boolean equals(Object o) {
+        return EqualsBuilder.reflectionEquals(this, o);
```
mohitsoni: This is neat! I didn't know there's a library function for this.

Although, reading through the [documentation](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java#L69) suggests that `reflectionEquals` should be avoided as it can lead to a SecurityException when JVM's SecurityManager is enabled, as `reflectionEquals` uses reflection to gain access to private fields.

sdk/core/src/main/java/org/apache/mesos/offer/constrain/AgentRule.java
```diff
@@ -34,52 +40,162 @@ public String toString() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        return HashCodeBuilder.reflectionHashCode(this);
```
mohitsoni: Same as above.

sdk/core/src/main/java/org/apache/mesos/offer/constrain/AgentRule.java
```diff
@@ -34,52 +40,162 @@ public String toString() {
...
+        }
+
+        @Override
+        public String toString() {
+            return String.format("RequireAgentGenerator{agentId=%s}", agentId);
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            return EqualsBuilder.reflectionEquals(this, o);
```
mohitsoni: See earlier comments about `reflectionEquals`.

sdk/core/src/main/java/org/apache/mesos/offer/constrain/AgentRule.java
```diff
@@ -34,52 +40,162 @@ public String toString() {
...
+        }
+
+        @Override
+        public int hashCode() {
+            return HashCodeBuilder.reflectionHashCode(this);
```
mohitsoni: See earlier comments about `reflectionEquals`

sdk/core/src/main/java/org/apache/mesos/offer/constrain/AgentRule.java
mohitsoni: Maybe you can do something like (as the doc suggests):
```
    if (o == null) { return false; }
    if (o == this) { return true; }
    if (o.getClass() != getClass()) { return false; }
    MyClass rhs = (MyClass) o;
    return new EqualsBuilder()
    .append(agentId, o.agentId)
    .isEquals();
```

sdk/core/src/main/java/org/apache/mesos/offer/constrain/AttributeSelector.java
```diff
@@ -69,9 +94,33 @@ public boolean select(String attribute) {
...
+        public boolean equals(Object obj) {
+            // Custom comparison required: Pattern doesn't implement equals().
+            if (this == obj) {
+                return true;
+            }
+            if (obj == null || getClass() != obj.getClass()) {
+                return false;
+            }
+            RegexSelector other = (RegexSelector) obj;
+            return pattern.pattern().equals(other.pattern.pattern());
```
mohitsoni: Can `other.pattern` be null ?

sdk/core/src/main/java/org/apache/mesos/offer/constrain/AgentRule.java
nickbp: Are we expecting to run our JVM in that configuration? A hand-coded check risks going out of sync when new members are added.

gabrielhartmann: I would defer the change to the equals method until we encounter a situation where we need the SecurityManager enabled.

sdk/core/src/main/java/org/apache/mesos/offer/constrain/AgentRule.java
```diff
@@ -34,52 +40,162 @@ public String toString() {
...
      * Ensures that the given Offer is NOT located on the specified agent ID.
      */
+    @JsonIgnoreProperties(PassthroughGenerator.RULE_NAME) // don't include in serialization
     public static class AvoidAgentGenerator extends PassthroughGenerator {
 
-        public AvoidAgentGenerator(String agentId) {
+        private final String agentId;
+
+        @JsonCreator
+        public AvoidAgentGenerator(@JsonProperty("agent_id") String agentId) {
```
gabrielhartmann: Now might be a good opportunity to address this issue: https://mesosphere.atlassian.net/browse/INFINITY-637

PR #248, opened 2016-10-27
Allow custom RecoveryRequirementProvider

gabrielhartmann:
* Allow children of the DefaultScheduler to specify custom RecoveryRequirementProvider (Allows Kafka to start using the the DefaultScheduler)
* Fixed bug where the DefaultOfferRequirementProvider was being used as the RecoveryRequirementProvider.
* Reorganized member variables so they're grouped together a little sanely.

### DIFF
--- a/sdk/scheduler/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
+++ b/sdk/scheduler/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
@@ -57,6 +57,7 @@ public class DefaultScheduler implements Scheduler, Observer {
     protected final StateStore stateStore;
     protected final Optional<Integer> permanentFailureTimeoutSec;
     protected final Integer destructiveRecoveryDelaySec;
+    private final RecoveryRequirementProvider recoveryRequirementProvider;
 
     protected SchedulerDriver driver;
     protected Reconciler reconciler;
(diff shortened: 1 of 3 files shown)


sdk/core/src/main/java/org/apache/mesos/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -143,12 +142,16 @@ private Block createBlock(Protos.TaskInfo taskInfo)
             recoveryRequirements = offerReqProvider.getTransientRecoveryRequirements(Arrays.asList(taskInfo));
         }
 
-        return new DefaultRecoveryBlock(
-                taskSpec.getName(),
-                offerRequirementProvider.getExistingOfferRequirement(taskInfo, taskSpec),
-                Status.PENDING,
-                recoveryRequirements.get(0),
-                launchConstrainer);
+        if (recoveryRequirements.size() == 1) {
```
gabrielhartmann: In principle, when asking the RecoveryRequirementProvider to provide RecoveryRequirements for a single failed Task it isn't necessarily true that a single requirement must be returned.  However in this case we are explicitly creating a single Step which can only support encapsulation of a single requirement.  So the alternative to this situation would be to drop some requirements on the floor.  Instead we throw here to indicate something unexpected has happened.

`createStep` is not part of the interface but is part of the implementation of the _default_ implementation of a PlanManager (for recovery).  If more subtle interesting behavior were desired, users should provide their own PlanManager.

Make sense?

PR #249, opened 2016-10-27
Rename Block to Step

gabrielhartmann:
This PR is annoying.  I know.  The intent is just to change the name of Block to Step.  I cleaned up any old documentation I saw in passing as well.  Nothing serious.  Should have zero functional change.   Chose you @nickbp because your'e on-call.


nickbp:
Any changes to `.md` docs in this repo that'd be relevant?

PR #251, opened 2016-10-27
Add documentation around Tardigrade (automatic permanent failure reco…

gabrielhartmann:
…very)


sascala:
@gabrielhartmann I made line edits that I don't think prevent merging.

PR #253, opened 2016-10-27
Add default config validators (and associated tests)

nickbp:
- TaskSetsCannotShrink: disallows shrinking a deployment
- TaskVolumesCannotChange: disallows changes to VolumeSpecs


gabrielhartmann:
@nickbp:  Can you incorporate this validation into the `sdk-reference-framework`?

mohitsoni:
LGTM!

nickbp:
Am planning on including them by default in the DefaultScheduler itself (with support for developer configuration). Something like https://github.com/mesosphere/dcos-commons/blob/spec-serialization/sdk/core/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java#L93

PR #256, opened 2016-10-28
Many nitpicks, and some cleanup/restructuring of DefaultSched.create()

nickbp:
Only functional change is using the caching state store by default in DefaultStateStore, and avoiding creating several StateStore instances. (Previously had 2 in DefaultScheduler and a 3rd in DefaultService)

The create() cleanup will make it easier to add ConfigStore to DefaultScheduler construction.


sdk/core/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -157,12 +177,11 @@ private void initialize(SchedulerDriver driver) throws InterruptedException {
                 deployPlanManager);
         planCoordinator = new DefaultPlanCoordinator(planManagers, planScheduler);
         planCoordinator.subscribe(this);
-        logger.info("Done initializing.");
+        LOGGER.info("Done initializing.");
     }
 
     private void initializeGlobals(SchedulerDriver driver) {
-        logger.info("Initializing globals...");
-        stateStore = new CuratorStateStore(frameworkName, zkConnectionString);
```
gabrielhartmann: This was actually by design.  To avoid split brain problems we only allow work to be done post-registration.  The best guarantee we have that metadata isn't mutated in unexpected ways is to not even construct the `StateStore` until after registration.

nickbp: What about the additional CuratorStateStore instance being created and passed to DefaultStepFactory in a create() method? (See L72 in original DefaultScheduler.java)

PR #269, opened 2016-10-31
Release 0.8.1

nickbp:
Moves version to root and adds instructions.


nickbp:
Now includes #271 (requires JAVA_HOME in scheduler env)

PR #271, opened 2016-10-31
Wired DCOSCertInstaller into DefaultScheduler

mohitsoni:


sdk/scheduler/src/main/java/org/apache/mesos/util/JavaHomeDetector.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Map;
+
+/**
+ * Determines the location of JAVA_HOME for a given task.
+ */
+public class JavaHomeDetector {
+    public static final String JAVA_HOME = "JAVA_HOME";
+    public static final String DEFAULT_JAVA_HOME = "jre1.8.0_91";
+    public static final String DEFAULT_JAVA_URI =
```
nickbp: Isn't this already in DefaultOfferRequirementProvider.java?

mohitsoni: It is and I will send out another PR to make `DefaultOfferRequirementProvider` use this.

nickbp:
Is there a case where JAVA_HOME shouldn't exist in a JRE process? If not I'd say just throw an exception if it's missing and skip the assumed fallback path.

sdk/scheduler/src/main/java/org/apache/mesos/util/JavaHomeDetector.java
nickbp: Alternately could just remove these defaults entirely per other comment. Feels risky having this custom behavior dependent on a specific JRE version, assuming it's not expected to actually occur in practice.

mohitsoni:
There's a case, like right now. Also, as long as `java` executable is in `PATH`, `JAVA_HOME` isn't really required. 

In our scheduler, `JAVA_HOME` is required specifically for strict mode.

sdk/scheduler/src/main/java/org/apache/mesos/util/JavaHomeDetector.java
```diff
@@ -0,0 +1,45 @@
...
+     * the it returns the {@link JavaHomeDetector.DEFAULT_JAVA_HOME}.
+     *
+     * @return String representing the JAVA_HOME
+     */
+    public static String getJavaHomeFromEnv() {
+        return getJavaHomeFromEnv(System.getenv());
+    }
+
+    /**
+     * Similar to {@code getJavaHomeFromEnv}.
```
mohitsoni: The reason I added this is because `DefaultOfferRequirementProvider`, sets an additional `DEFAULT_JAVA_URI` in the CommandInfo, if JAVA_HOME was not present in the environment.

sdk/scheduler/src/main/java/org/apache/mesos/util/JavaHomeDetector.java
mohitsoni: I think it does occur today, that's why `DefaultOfferRequirementProvider` falls back to these defaults.

I think we can certainly not fallback on the default JRE. Let me make those changes.

mohitsoni:
I think if light of above comments, I'm going to delete JavaHomeDetector class, as it doesn't help much, if we are not going to include the fall back logic.

nickbp:
Ok, makes sense. So one thing I'm concerned about is that this effectively locks existing schedulers to a specific JRE version.

The current behavior in DefaultOfferRequirementProvider isn't optimal but it's at least just assigning the value to the executor's list of fetched URIs, so it's effectively enforcing the version internally anyway.

Maybe instead of hardcoding the version here (and causing potential sync issues between schedulers and their packaging), maybe just autodetect the JRE directory by searching for a directory that starts with 'jre' within the sandbox dir that starts with 'jre'? For example Kafka is doing this with a `jre*` path in the shell: https://github.com/mesosphere/dcos-kafka-service/blob/fb118b73a59836eef4db4e0bbed092eae37d3046/kafka-scheduler/src/main/java/com/mesosphere/dcos/kafka/offer/PersistentOfferRequirementProvider.java#L356

mohitsoni:
The auto-detection logic feels very risky. I would rather go very explicit route OR a reasonable library standard default.
Although, the more I think about it, explicit route is best as it leads to deterministic behavior.

nickbp:
It might be safer/simpler to just require that the scheduler have JAVA_HOME defined

mohitsoni:
Exactly! By explicitly specifying `JAVA_HOME` we can expect a safe behavior with simpler code.

sdk/scheduler/src/main/java/org/apache/mesos/scheduler/DefaultScheduler.java
```diff
@@ -272,6 +273,7 @@ private void initialize(SchedulerDriver driver) throws InterruptedException {
         initializeGlobals(driver);
         initializeRecoveryPlanManager();
         initializeResources();
+        DCOSCertInstaller.installCertificate(System.getenv("JAVA_HOME"));
```
nickbp: Maybe add an explicit null check at the start of installCertificate()

PR #275, opened 2016-10-31
Basic sanity test integration test for HDFS

susanxhuynh:
Copied over basic sanity test (install hdfs) from dcos_hdfs_service project. @NimaVaziri

### DIFF
--- a/frameworks/hdfs/integration/tests/test_sanity.py
+++ b/frameworks/hdfs/integration/tests/test_sanity.py
@@ -1,0 +1,33 @@
+import pytest
+import shakedown
+import inspect
+import os
+
+from tests.test_utils import (
+    PACKAGE_NAME,
+    check_health,
+    uninstall,
+)
+
+
+strict_mode = os.getenv('SECURITY', 'permissive')
+
+
+def setup_module(module):
+    uninstall()
+
+    if strict_mode == 'strict':
+        shakedown.install_package_and_wait(package_name=PACKAGE_NAME, options_file=os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))) + "/strict.json")
+    else:
+        shakedown.install_package_and_wait(package_name=PACKAGE_NAME, options_file=None)
+
+    check_health()
+
+
+def teardown_module(module):
+    uninstall()
+
+
+@pytest.mark.sanity
+def test_install_worked():
+    pass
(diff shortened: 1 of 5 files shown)


frameworks/hdfs/integration/tests/test_sanity.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+import shakedown
+import inspect
+import os
+
+from tests.test_utils import (
```
NimaVaziri: This should be `test_utils` instead of `tests.test_utils` since we're importing the other module from the same directory.

NimaVaziri:
@susanxhuynh Once that one comment is addressed, this should be good to merge.

frameworks/hdfs/integration/tests/test_utils.py
```diff
@@ -0,0 +1,69 @@
...
+def uninstall():
+    print('Uninstalling/janitoring {}'.format(PACKAGE_NAME))
+    try:
+        shakedown.uninstall_package_and_wait(PACKAGE_NAME, app_id=PACKAGE_NAME)
+    except (dcos.errors.DCOSException, ValueError) as e:
+        print('Got exception when uninstalling package, continuing with janitor anyway: {}'.format(e))
+
+    shakedown.run_command_on_master(
+        'docker run mesosphere/janitor /janitor.py '
+        '-r hdfs-role -p hdfs-principal -z dcos-hdfs/hdfs '
```
nickbp: the -z value here should be `dcos-service-hdfs` for the new version of the service

frameworks/hdfs/integration/tests/test_sanity.py
nickbp: For whatever reason all paths are relative to integration/ rather than integration/tests. For reference scheduler tests just now, I tried dropping `tests.` and found that python couldn't find the module anymore.

NimaVaziri: Hmm, yeah this seemed to only happen for me when I tried running it. @susanxhuynh ran it fine with `tests` included in the import statement. Haven't gotten to the bottom of this yet...

susanxhuynh: @NimaVaziri I added the __init__.py file in this directory (which I had omitted earlier.) I don't know if it's related to the problem you saw, though.

PR #283, opened 2016-11-03
Restructure placement constraints, fix TaskType/MaxPerAttribute logic

nickbp:
- Merges PlacementRuleGenerator into PlacementRule. PlacementRules are now passed task state directly, creating a single class for all placement logic.
- Adds OfferRequirement as a parameter to the PlacementRule interface. This allows placement rules to consider information about the task that's about to be launched, and in the config rollout case allows them to ignore the prior version of the same task that's about to be launched.
- Updates/fixes TaskType and MaxPerAttribute rules to correctly deal with this 'replacing task that's still listed' situation described above.
- Updates serialization handling to just use the class name, e.g. "TaskTypeRule", instead of the full class path, e.g. "org.apache.mesos......TaskTypeRule". Adds registry access so that service devs could insert their own types if needed.
- Ignores placement rules when performing transient recovery (which should target the previously used location anyway)
- While in OfferEvaluator, updates logging of evaluation outcomes to be a bit more readable.


sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MaxPerAttributeGenerator.java
```diff
@@ -124,4 +95,84 @@ public boolean equals(Object o) {
...
+        private final AttributeSelector attributeSelector;
+        private final Collection<TaskInfo> tasks;
+
+        private MaxPerAttributeRule(
+                int maxTasksPerSelectedAttribute,
+                AttributeSelector attributeSelector,
+                Collection<TaskInfo> tasks) {
+            this.maxTasksPerSelectedAttribute = maxTasksPerSelectedAttribute;
+            this.attributeSelector = attributeSelector;
+            this.tasks = tasks;
```
gabrielhartmann: I'm a little confused by Tasks are being passed up to a PlacementRule.  Tasks are available at the PlacementRuleGenerator level.  Couldn't this implementation reduce to a PlacementRuleGenerator that generates an ORed list of AgentRules?

nickbp: PlacementRuleGenerator gets the current state of the system, then PlacementRule is used to evaluate offers given that state. In this case, the task list passthrough is done because we want to avoid double-counting the task that's about to be launched.

Example scenario:
- Limit per rack: 3
- Current state (Collection<TaskInfo>):
 - rack:1: tasks A, B, C
 - rack:2: tasks D, E, F
- Offer being evaluated (OfferRequirement):
 - (Re)deploy B

Without the duplicate detection, B would be unable to deploy at all because both racks are fully populated. But with the duplicate detection the old instance of B against rack:1 is filtered out because it matches the task being deployed and we'd be able to deploy B against rack:1.

nickbp:
I actually wonder if maybe PlacementRule+PlacementRuleGenerator could just be united into a single PlacementRule class that gets all the data at once with a single method: `Offer filter(Offer offer, OfferRequirement offerRequirement, Collection<TaskInfo> currentTasks)`.

Given the need for this dedupe checking anywhere that actually looks at task state, resulting in passing the data from 'phase 1' directly to 'phase 2', there isn't that much benefit in the two phases we have today. The placement rule handling (invoking PlacementRuleGenerator and PlacementRule) is all contained in OfferEvaluator anyway. What do you think?

gabrielhartmann:
@nickbp:  I think your latest suggestion is a great idea.  The separation now is actually a little difficult to understand and I think it leads to racy code.  Step 1. Filter based on Tasks.  time passes.  Step 2. Filter based on Offers.  Ideally both steps are working of the same snapshot of information, so fine, but I think unification is a better idea.  I think it will also lead to a smoother transition when this gets merged with the new OfferEvalutor design later.

nickbp:
OK will do. As discussed offline I'll also ensure that placement constraints are disabled when handling transient recovery.

sdk/scheduler/src/main/java/org/apache/mesos/offer/OfferEvaluator.java
```diff
@@ -41,47 +40,68 @@ public OfferEvaluator(StateStore stateStore) {
 
     public List<OfferRecommendation> evaluate(OfferRequirement offerRequirement, List<Offer> offers)
             throws StateStoreException {
-        Optional<PlacementRule> placementRule = getPlacementRule(offerRequirement, stateStore);
-        if (placementRule.isPresent()) {
+
+        // First, check placement constraints (to filter offers)
+        List<Offer> filteredOffers = new ArrayList<>();
+        Optional<PlacementRule> placementRuleOptional = offerRequirement.getPlacementRuleOptional();
+        if (placementRuleOptional.isPresent()) {
```
gabrielhartmann: Can we please factor out this Offer/Resource filtering into a separate method to keep `evaluate(`)'s size down?

nickbp: Good point, done.

sdk/scheduler/src/main/java/org/apache/mesos/offer/OfferRequirement.java
```diff
@@ -3,75 +3,94 @@
...
+    }
+
+    /**
+     * Creates and returns a new {@link OfferRequirement} with any placement rules removed.
+     */
+    public OfferRequirement withoutPlacementRules() {
+        return new OfferRequirement(taskType, taskRequirements, executorRequirementOptional, Optional.empty());
+    }
+
+    private OfferRequirement(
```
nickbp: Eh I think it's fine to keep this private until a usecase comes up.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/PlacementRule.java
```diff
@@ -10,10 +14,12 @@
...
  * may return the original Offer without copying if everything passes the filter. If the entire
- * Offer is rejected, the return value will be a non-{@code null} Offer with zero Resources. A
- * PlacementRule may contain multiple internal checks, and may even be composed of other sub-Rules.
+ * Offer is rejected, the return value will be a non-{@code null} Offer with zero Resources.
+ *
+ * PlacementRules may contain multiple internal checks, and may even be a composition of other
+ * PlacementRules.
  */
-@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS)
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME)
```
gabrielhartmann: What's the consequence of this?  No writing rules that have the same class name in different packages?

nickbp: - Only the class name needs to be specified in serialized representations
- Classes need to be [registered](http://massapi.com/method/com/fasterxml/jackson/databind/ObjectMapper.registerSubtypes.html) against the Jackson ObjectMapper, either by java Class (to just use the class name, eg "AndRule"), or by a NamedType object (to specify a custom name -- or to get around name collisions). See default registry (and access to add custom classes) in DefaultServiceSpecification.Factory.

sdk/scheduler/src/main/java/org/apache/mesos/specification/DefaultServiceSpecification.java
```diff
@@ -29,19 +33,48 @@
...
-        private Factory() { }
+        private static final Collection<Class<?>> defaultRegisteredSubtypes = Arrays.asList(
+                AgentRule.class,
+                AndRule.class,
+                AttributeRule.class,
+                HostnameRule.class,
+                MaxPerAttributeRule.class,
+                NotRule.class,
+                OrRule.class,
+                TaskTypeRule.class);
```
gabrielhartmann: What are the requirements for a developer that wishes to write their own `PlacementRule` that doesn't make sense as an addition to the SDK?

sdk/scheduler/src/test/java/org/apache/mesos/scheduler/DefaultSchedulerTest.java
```diff
@@ -587,22 +587,6 @@ private void register() {
                 .build();
     }
 
-    private Protos.Offer getSufficientOfferForUpdatedTaskA() {
```
gabrielhartmann: Thanks

sdk/scheduler/src/main/java/org/apache/mesos/specification/DefaultServiceSpecification.java
```diff
@@ -98,9 +131,13 @@ public String toJsonString() throws Exception {
     /**
      * Returns a {@link ConfigurationFactory} which may be used to deserialize
      * {@link DefaultServiceSpecification}s.
+     *
+     * @param additionalSubtypesToRegister any class subtypes which should be registered with
+     *     Jackson for deserialization. any custom placement rule implementations must be provided
      */
-    public static ConfigurationFactory<ServiceSpecification> getFactoryInstance() {
-        return FACTORY;
+    public static ConfigurationFactory<ServiceSpecification> getFactory(
```
gabrielhartmann: It would be very cool, if a unit test could be added validating that an attempt to use an unregistered (but properly constructed) custom PlacementRule in a ServiceSpecification caused some kind of validation error.  Otherwise I think people will run into faults at deserialization time i.e. after an unplanned Task crash or something and find out that things are never recoverable.

sdk/scheduler/src/main/java/org/apache/mesos/specification/DefaultServiceSpecification.java
nickbp: In their types they need to:
- Ensure that they specify `@JsonCreator` and `@JsonProperty` in their constructor
- Ensure that their members are accessible via 'get' methods with `@JsonProperty` annotations.

In the Scheduler they need to:
- Add the Classes of their custom types into DefaultServiceSpecification.Factory() at the time of constructing the DefaultScheduler:
```
DefaultScheduler.create(
    serviceSpecification,
    DefaultScheduler.createStateStore(
        serviceSpecification.getName(),
        DcosConstants.MESOS_MASTER_ZK_CONNECTION_STRING),
    DefaultScheduler.createConfigStore(
        serviceSpecification.getName(),
        DcosConstants.MESOS_MASTER_ZK_CONNECTION_STRING),
        Arrays.asList(Custom.class, Types.class, Here.class)));
```

sdk/scheduler/src/main/java/org/apache/mesos/specification/DefaultServiceSpecification.java
nickbp: Added automatic validation (and unit tests to check that it works appropriately): Using the provided classes, serializes and deserializes the provided servicespecification. Then checks that the serialized/deserialized version equals() the original.

Validates:
- Any necessary classes have been provided up-front
- Equal()ity behaves consistently

PR #285, opened 2016-11-05
Add static VIP support in task and executor requirements

mrbrowning:
Similar to the `DynamicPortRequirement` class, the solution for adding this functionality involves tracking named-VIP-ports as a separate type of resource from regular the `ResourceRequirement` list. Although this involves changing many method signatures, it's preferable to blending this logic in with regular resource accounting with excessive branching. Ultimately, this logic will be encapsulated by an upcoming change to the `OfferEvaluator` which will introduce a pipeline of resource requirement/`TaskInfo` modification stages. Note that the final consequence of this change from Mesos' perspective is that either the `TaskInfo` or the `ExecutorInfo` will have a `DiscoveryInfo` attached, with a hostname and `Port` resource labeled with the intended service address.

### DIFF
--- a/sdk/scheduler/src/test/java/org/apache/mesos/offer/NamedVIPPortRequirementTest.java
+++ b/sdk/scheduler/src/test/java/org/apache/mesos/offer/NamedVIPPortRequirementTest.java
@@ -1,0 +1,22 @@
+package org.apache.mesos.offer;
+
+import org.apache.mesos.Protos;
+import org.apache.mesos.testutils.TestConstants;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class NamedVIPPortRequirementTest {
+
+    @Test
+    public void testNamedVIPPortRequirementConstruction() throws NamedVIPPortRequirement.NamedVIPPortException {
+        Protos.Resource vipPortResource = NamedVIPPortRequirement.getDesiredNamedVIPPort(
+                TestConstants.VIP_KEY,
+                TestConstants.VIP_NAME,
+                (long) 10000,
+                TestConstants.ROLE,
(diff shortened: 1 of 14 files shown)


sdk/scheduler/src/test/java/org/apache/mesos/offer/NamedVIPPortRequirementTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.apache.mesos.Protos;
+import org.apache.mesos.testutils.TestConstants;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class NamedVIPPortRequirementTest {
+
+    @Test
+    public void testNamedVIPPortRequirementConstruction() throws NamedVIPPortRequirement.NamedVIPPortException {
```
gabrielhartmann: At least one test that actually throws the `NamedVIPPortException` is probably wanted.

loren:
It seems like many (most?) FW's will want a named VIP for each taskset, as each task in the taskset is likely there to provide HA among other things. Rather than have the reference FW hook up named VIPs for `data` and `meta-data` tasksets, and the Elastic FW hook up 5 named VIPs for each of the 5 tasksets, and so on, would it make sense to just set them up like this by default and let FW devs override the behavior when needed?

Barring that, showing this used in the reference FW would be great. Looking forward to getting this functionality!

gabrielhartmann:
@loren:  Thanks for the feedback.  We'll be constantly iterating towards something which puts fewer and fewer burdens on the framework developer.  We would like to make VIPs settable on TaskSets (in the future Pods), and will work towards this as priorities allow.

PR #288, opened 2016-11-09
Rename AttributeSelector to StringMatcher, clean up serialization

nickbp:
- StringMatchers will also be used for hostname comparisons
- Have StringMatchers support class-name serialization
- Have TaskTypeConverters support class-name serialization


sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/HostnameRule.java
```diff
@@ -20,85 +20,109 @@
...
 
     /**
      * Requires that a task NOT be placed on any of the provided hostnames.
      *
-     * @param hostnames hostnames of the mesos agents to avoid
+     * @param matchers matchers for hostnames of the mesos agents to avoid
      */
-    public static PlacementRule avoid(String... hostnames) {
-        return avoid(Arrays.asList(hostnames));
+    public static PlacementRule avoid(StringMatcher... matchers) {
```
mrbrowning: Are these methods meant to be used by framework authors? It might be a kindness to retain the old `String...` interface in addition to this, and instantiate an `ExactMatcher` for each by default, so that the user can continue to just supply hostname strings.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MaxPerAttributeRule.java
```diff
@@ -48,22 +48,56 @@
...
+ * look at ALL tasks in the service, but with a task matcher we can only count e.g. tasks named
+ * 'index-.*'. This allows us to only enforce the rule against certain task types or task instances
+ * within the service.
  */
 public class MaxPerAttributeRule implements PlacementRule {
 
     private final int maxTasksPerSelectedAttribute;
-    private final AttributeSelector attributeSelector;
+    private final StringMatcher attributeMatcher;
+    private final StringMatcher taskFilter;
```
mrbrowning: Filtering by task seems pretty general, would it be worth recommending the composition of several rules (e.g. an `AttributeRule` that filters by task name combined with the `MaxPerAttributeRule` here with an `AndRule`) as the way to achieve this?

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/RegexMatcher.java
```diff
@@ -0,0 +1,57 @@
...
+    private final Pattern pattern;
+
+    @JsonCreator
+    private RegexMatcher(@JsonProperty("pattern") String pattern) {
+        this.pattern = Pattern.compile(pattern);
+    }
+
+    @Override
+    public boolean match(String value) {
+        return pattern.matcher(value).matches();
```
mrbrowning: Looking at the `Pattern` interface, I kind of like `matches()` as the name for the current `match()` method, since it implies a return type of `boolean` rather than `void`. Might be worth changing if it's low-effort.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/HostnameRule.java
nickbp: I had indeed tried to have that but ended up getting type conflicts with java. As a workaround, I added the `toStringMatchers()` functions further up for devs to use like this: `HostnameRule.avoid(toStringMatchers(exactValues))`

An alternate method would be to add new differently-named functions, e.g. `avoidExact()`/`requireExact()`, to avoid the type collision issues. Any preference?

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MaxPerAttributeRule.java
nickbp: The tricky part is that taskFilter is being applied to the current state of known tasks (3rd parameter), while `*Rule` operate against the Offer being evaluated (1st parameter). They're each effectively performing work against different sources of information.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/RegexMatcher.java
nickbp: Good point. IMO this is the best time to fix it in any case. Renamed to `matches()`.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MaxPerAttributeRule.java
mrbrowning: Ah, yeah, I see how that's not really a possibility here. If we do decide that more rules might want task-filtering behavior, I wonder if we could add a default method to the `PlacementRule` interface that performs that filtering with a supplied matcher, and then the `filter` method could always dependably delegate to it if it wanted to. Anyway, that can probably wait.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/HostnameRule.java
mrbrowning: It's a shame that won't type-check, in that case maybe the second method? It doesn't really duplicate any logic, and I like the idea of providing low-friction access to the most intuitive behavior.

nickbp: Sounds good. Done. There's now avoidExact()/requireExact() for Strings in addition to avoid()/require() for StringMatchers

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MaxPerAttributeRule.java
nickbp: Yeah it somewhat depends on the context. Some Rules care about what other tasks have been launched while others don't

PR #291, opened 2016-11-15
Fixed HDFS and common CLI pointing to incorrect endpoints.

susanxhuynh:
Commands affected:

- hdfs node list
- hdfs state status datanode-0
- hdfs state task datanode-0
- hdfs node restart datanode-0 (affects only HDFS)
- hdfs node replace datanode-0 (affects only HDFS)


NimaVaziri:
LGTM

PR #292, opened 2016-11-15
Implement support for placement constraints from Marathon-style strings

nickbp:
- Implements Marathon String => PlacementRule
- Adds rule implementations to fill out marathon support: MaxPerHostnameRule, PassthroughRule, RoundRobinByAttributeRule, RoundRobinByHostnameRule


sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/RoundRobinByAttributeRule.java
```diff
@@ -0,0 +1,197 @@
...
+ *  attribute |     tasks
+ * -----------+---------------
+ *    rack:1  | a-1, b-1, c-1
+ *    rack:2  | a-2, c-2, c-3
+ *    rack:3  | b-2, c-4, b-3
+ * Result:
+ *  allow any of rack:1/rack:2/rack:3, unless we know that there's >3 racks via the attribute_count
+ *  parameter.
+ */
+public class RoundRobinByAttributeRule implements PlacementRule {
```
gabrielhartmann: Strictly speaking these RoundRobin rules are not taking into account very much Cluster state (which we could know).  Correct me if I'm wrong, but as long as this Offer has the attribute in question and is below the maxCount for this attribute we're going to to say it's OK even if the next Offer, in the Offers we've received would be better.  e.g.  max is 3, and this one has 2 on it already, but the next Offer in the stream has none.  I wouldn't block on this, but it's worth filing a ticket for future improvement such that we look at a wider snapshot of the world and make smarter decisions.  Of course optimistic offers will alter this greatly, as we will have very close to a global view of the cluster...

sdk/scheduler/src/test/java/org/apache/mesos/offer/constrain/MarathonConstraintParserTest.java
```diff
@@ -0,0 +1,224 @@
...
+        constraintStr = MarathonConstraintParser.parse(unescape("[['rack_id', 'UNIQUE']]")).toString();
+        assertEquals("AndRule{rules=[AttributeRule{matcher=RegexMatcher{pattern='rack_id:.*'}}, " +
+                "MaxPerAttributeRule{max=1, matcher=RegexMatcher{pattern='rack_id:.*'}, task_filter=AnyMatcher{}}]}", constraintStr);
+        assertEquals(constraintStr, MarathonConstraintParser.parse(unescape("['rack_id', 'UNIQUE']")).toString());
+        assertEquals(constraintStr, MarathonConstraintParser.parse("rack_id:UNIQUE").toString());
+    }
+
+    @Test
+    public void testClusterOperator() throws IOException {
+        // marathon doc:
```
gabrielhartmann: These comments are odd.  Should probably be removed.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/RoundRobinByAttributeRule.java
nickbp: That case is meant to be covered by having the user specify the optional attributeValueCount in their filter. This allows us to know that there are other nodes out there which don't have anything on them at all yet. In case (known task nodes < user-provided expected node count) we avoid launching until we see an offer from a node which has that attribute with no tasks. See comment at L128 for handling that case.

If the user refrains from providing a number of nodes then we indeed will do that.

nickbp: See also RoundRobinByAttributeRuleTest.java:L65, where the rule refuses to launch a 2nd task against value1 because it knows there are a total of 3 values out there (and it should instead put the task against one of those empty values).

gabrielhartmann: @nickbp Fantastic.

nickbp:
TODO:
- Add to reference scheduler
- Check that implementations align with [marathon's implementation](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/mesos/Constraints.scala)

sdk/scheduler/src/test/java/org/apache/mesos/offer/constrain/MarathonConstraintParserTest.java
nickbp: These were meant to highlight the cases where the tests are against something from the marathon doc at https://mesosphere.github.io/marathon/docs/constraints.html. Updated them to be clearer about this.

nickbp:
Support is now integrated with the reference scheduler (including config in packaging) with per-node-type granularity. If end-user specifies custom constraints for either node type, they will be ANDed with the hardcoded constraint.

sdk/scheduler/src/main/java/org/apache/mesos/offer/AttributeStringUtils.java
```diff
@@ -46,7 +60,43 @@ public static String toString(List<Attribute> attributes) throws IllegalArgument
...
+        return new NameValue(split[0], split[1]);
+    }
+
+    /**
+     * Converts the provided attribute's name + value into a string which follows the format defined
+     * by Mesos.
+     *
+     * @see #valueString(Attribute)
+     */
+    public static String toString(Attribute attribute) throws IllegalArgumentException {
```
mrbrowning: Reimplement in terms of `join`?

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MarathonConstraintParser.java
```diff
@@ -0,0 +1,321 @@
...
+        SUPPORTED_OPERATORS.put("UNLIKE", new UnlikeOperator());
+        SUPPORTED_OPERATORS.put("MAX_PER", new MaxPerOperator());
+    }
+
+    private MarathonConstraintParser() {
+        // do not instantiate
+    }
+
+    /**
+     * ANDs the provided marathon-style constraint string onto the provided hard-coded
```
mrbrowning: What does "hard-coded" mean here?

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MarathonConstraintParser.java
```diff
@@ -0,0 +1,321 @@
...
+     */
+    public static PlacementRule parse(String marathonConstraints) throws IOException {
+        if (marathonConstraints == null || marathonConstraints.isEmpty()) {
+            // nothing to enforce
+            return new PassthroughRule();
+        }
+        List<List<String>> rows = splitConstraints(marathonConstraints);
+        if (rows.size() == 1) {
+            return parseRow(rows.get(0));
+        } else {
```
mrbrowning: Matter of preference, but you can omit the `else` clause and move that logic into the parent scope given that control flow never gets here if `rows.size() == 1`.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MarathonConstraintParser.java
```diff
@@ -0,0 +1,321 @@
...
+     * @return
+     * @throws IOException
+     */
+    @VisibleForTesting
+    static List<List<String>> splitConstraints(String marathonConstraints) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        // The marathon doc uses a format like: '[["a", "b", "c"], ["d", "e"]]'
+        // Meanwhile the marathon web interface uses a format like: 'a:b:c,d:e'
+        try {
+            // First try: ["a", "b", "c"] (not technically correct but lets be lenient here)
```
mrbrowning: This comment is a little difficult to understand, I'd either flesh it out more or remove it.

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MarathonConstraintParser.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /**
+     * {@code GROUP_BY} can be used to distribute tasks evenly across racks or datacenters for high
+     * availability: {@code [["rack_id", "GROUP_BY"]]}
+     *
+     * Marathon only knows about different values of the attribute (e.g. "rack_id") by analyzing
+     * incoming offers from Mesos. If tasks are not already spread across all possible values,
+     * specify the number of values in constraints. If you don't specify the number of values, you
+     * might find that the tasks are only distributed in one value, even though you are using the
```
mrbrowning: Might be worth noting which circumstances could lead to this. I assume it's a function of the distribution of $ATTRIBUTE among the incoming offers?

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MarathonConstraintParser.java
```diff
@@ -0,0 +1,321 @@
...
+                        AttributeRule.require(matcher), new MaxPerAttributeRule(max, matcher));
+            }
+        }
+    }
+
+    private static boolean isHostname(String fieldName) {
+        return HOSTNAME_FIELD.equalsIgnoreCase(fieldName);
+    }
+
+    private static String validateRequiredParameter(
```
mrbrowning: What's the value of wrapping an `isPresent` check in a new function?

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/MaxPerHostnameRule.java
```diff
@@ -0,0 +1,138 @@
...
+        this.maxTasksPerHostname = maxTasksPerSelectedHostname;
+        if (taskFilter == null) { // null when unspecified in serialized data
+            taskFilter = AnyMatcher.create();
+        }
+        this.taskFilter = taskFilter;
+    }
+
+    @Override
+    public Offer filter(Offer offer, OfferRequirement offerRequirement, Collection<TaskInfo> tasks) {
+        int offerHostnameTaskCounts = 0;
```
mrbrowning: I don't quite have a full picture of the logic in my head here -- this rule doesn't keep any state about tasks that have already been scheduled, so is `tasks` a comprehensive list of all tasks that have already been placed as well as those that need to be placed? How is the constraint maintained?

sdk/scheduler/src/main/java/org/apache/mesos/offer/constrain/RoundRobinByHostnameRule.java
```diff
@@ -0,0 +1,164 @@
...
+            @JsonProperty("task_filter") StringMatcher taskFilter) {
+        this.agentCountOptional = agentCount;
+        if (taskFilter == null) { // null when unspecified in serialized data
+            taskFilter = AnyMatcher.create();
+        }
+        this.taskFilter = taskFilter;
+    }
+
+    @Override
+    public Offer filter(Offer offer, OfferRequirement offerRequirement, Collection<TaskInfo> tasks) {
```
mrbrowning: It feels like this logic should be generalizable between this and the attribute round-robin, maybe by making this a higher-order function.

nickbp: It was! Done.

tobilg:
Can this used with a YAML definition? If so, how? @gabrielhartmann

keithchambers:
Hi @tobilg, we do plan to expose these placement constraints in the YAML API at the pod level.

tobilg:
@keithchambers Thanks for the fast feedback. It's a bit confusing to me that some things are already exposed, and some aren't. Is there a way to quickly check which features are available in through the YAML? Thanks for the great work so far!

PR #294, opened 2016-11-17
Added package upgrade test.

susanxhuynh:
0. - (After normal test tools have already automatically added the infinity-artifacts stub universe...)
1. - Add the master custom universe at index 0
2. - Install data-store master build
4. - Upgrade data-store to infinity-artifacts build
5. - Verify that rollout occurred as expected, and that upgraded data-store is healthy

### DIFF
--- a/examples/reference/integration/tests/test_upgrade.py
+++ b/examples/reference/integration/tests/test_upgrade.py
@@ -1,0 +1,89 @@
+import dcos.http
+import pytest
+import re
+import shakedown
+
+from tests.test_utils import (
+    PACKAGE_NAME,
+    check_health,
+    install,
+    marathon_api_url_with_param,
+    request,
+    run_dcos_cli_cmd,
+    spin,
+    uninstall,
+)
+
+
+MASTER_CUSTOM_NAME='Master Custom'
+# TODO: replace this URL once we have a stock version from which to upgrade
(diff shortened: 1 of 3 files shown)


examples/reference/integration/tests/test_utils.py
```diff
@@ -11,6 +14,9 @@
 DEFAULT_TASK_COUNT = 5 # 2 metadata nodes, 3 data nodes
 
 
+strict_mode = os.getenv('SECURITY', 'permissive')
```
nickbp: could just check within install(), looks like nothing else is using it

examples/reference/integration/tests/test_upgrade.py
```diff
@@ -0,0 +1,89 @@
...
+def test_upgrade():
+    test_version = get_pkg_version()
+    print('Found test version: {}'.format(test_version))
+    add_repo(test_version)
+    master_version = get_pkg_version()
+    print('Found master version: {}'.format(master_version))
+    print('Installing master version')
+    install(master_version)
+    check_health()
+    print('Upgrading to test version')
```
nickbp: shouldn't remove_repo(master_version) be done here? otherwise it looks like it'd just reinstall the test version

susanxhuynh: No, when the reinstall is called, it specifies the version to install rather than taking the default value:
L43: install(test_version)

PR #304, opened 2016-11-18
Updated Quick Start to add versions

NimaVaziri:
This is based on the requirements for [dcos-vagrant](https://github.com/dcos/dcos-vagrant/blob/master/docs/deploy.md#software) which is used by dcos-docker.


NimaVaziri:
@keithchambers @nickbp doesn't look like there's a workaround to having specific versions being required...

PR #313, opened 2016-11-19
Fixed project name

NimaVaziri:


triclambert:
This script should probably include a .sh suffix, like the rest of our scripts.

PR #319, opened 2016-11-21
Upgrade test

gabrielhartmann:
Re-add the `hello-world` upgrade test.  Got lost on promotion of the pods branch to master.


nickbp:
Could also delete the now-duplicate copy in examples/

frameworks/helloworld/integration/tests/test_utils.py
```diff
@@ -68,10 +71,13 @@ def success_predicate(response):
     return spin(fn, success_predicate)
 
 
-def install(additional_options = {}):
+def install(package_version=None, package_name=PACKAGE_NAME, additional_options = {}):
```
susanxhuynh: You can remove the second param, "package_name". It's not used.

gabrielhartmann: It is being used in logging and in the `install_package_and_wait` command.

NimaVaziri: Looks like the default value of the argument is being used (`PACKAGE_NAME`), and not any other value that could be passed in via that parameter.

PR #321, opened 2016-11-22
Use a pre-built/uploaded VM image, instead of requiring DIY builds

nickbp:
Upside(s):
- Environment creation is much simpler, mainly consisting of convenince logic surrounding a plain ol' `vagrant up` command.

Downside:
- Current up-front image download approaches 5GB (many easy optimizations left to be made by DC/OS). This is ameliorated somewhat by the USB-stick option.

Main changes:
- Adds new `get-dcos-docker.sh` and updates quickstart to use it. This replacement for `build-dcos-docker.sh` just grabs the latest version of the uploaded package. Supports passing a local path to the package using a `BOX_PATH` envvar (USB stick support).
- Moves `build-dcos-docker.sh` into `tools/vagrant/` while adding some automation around creating box packages. This tool should now only be needed when the box package needs an update.

### DIFF
--- a/get-dcos-docker.sh
+++ b/get-dcos-docker.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+# Launches a DC/OS dev environment based containing a running 3-agent cluster and dev tools.
+#
+# Prerequisites:
+# - 8GB RAM
+# - Vagrant
+# - Virtualbox
+#
+# Arguments:
+# - BOX_PATH: Local path to use for .box image (useful if installing from USB stick)
+
+# capture anonymous metrics for reporting
+curl --fail https://mesosphere.com/wp-content/themes/mesosphere/library/images/assets/sdk/create-dev-env-start.png >/dev/null 2>&1
+
+# abort script at first error:
+set -e
+
+BOX_NAME=mesosphere/dcos-docker-sdk
+
+error_msg() {
+    echo "---"
+    echo "Failed to build the cluster: Exited early at $0:L$1"
+    echo "To try again, re-run this script."
+    echo "---"
+}
+trap 'error_msg ${LINENO}' ERR
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $SCRIPT_DIR
+
+cd tools/vagrant/
+VAGRANT_DIR=$(pwd)
+
+# Image-on-USB-stick scenario: Manually add the provided .box file directly
+if [ $(vagrant box list | grep $BOX_NAME | wc -l) -eq 0 ]; then
+    if [ -n "$BOX_PATH" ]; then
+        echo "### Installing provided local box image as $BOX_NAME: $BOX_PATH"
+        vagrant box add --name $BOX_NAME ${BOX_PATH}
+    fi
+elif [ -n "$BOX_PATH" ]; then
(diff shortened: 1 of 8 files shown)


nickbp:
Rebased to include `--fail` removal, PTAL

get-dcos-docker.sh
```diff
@@ -0,0 +1,71 @@
...
+# Prerequisites:
+# - 8GB RAM
+# - Vagrant
+# - Virtualbox
+#
+# Arguments:
+# - BOX_PATH: Local path to use for .box image (useful if installing from USB stick)
+
+# capture anonymous metrics for reporting
+curl https://mesosphere.com/wp-content/themes/mesosphere/library/images/assets/sdk/create-dev-env-start.png >/dev/null 2>&1
```
nickbp: Open question: This is reusing the location used by build-dcos-image.sh, should it be different?

get-dcos-docker.sh
```diff
@@ -0,0 +1,71 @@
...
+
+cd tools/vagrant/
+VAGRANT_DIR=$(pwd)
+
+# Image-on-USB-stick scenario: Manually add the provided .box file directly
+if [ $(vagrant box list | grep $BOX_NAME | wc -l) -eq 0 ]; then
+    if [ -n "$BOX_PATH" ]; then
+        echo "### Installing provided local box image as $BOX_NAME: $BOX_PATH"
+        vagrant box add --name $BOX_NAME ${BOX_PATH}
+    fi
```
NimaVaziri: What about the case where `$BOX_NAME` isn't in the list and `$BOX_PATH` isn't set?

tools/publish_http.py
```diff
@@ -228,8 +228,8 @@ def main(argv):
         logger.info('Install your package using the following command:')
     else:
         logger.info('Install your package using the following commands:')
-        logger.info('dcos package repo add --index=0 {}-aws {}'.format(self._pkg_name, universe_url))
-    logger.info('dcos package install {}'.format(self._pkg_name))
+        logger.info('dcos package repo add --index=0 {}-aws {}'.format(package_name, universe_url))
+    logger.info('dcos package install {}'.format(package_name))
```
NimaVaziri: For consistency, should we also put this "add-package-and-install" log message for the `aws` script?

get-dcos-docker.sh
nickbp: In that case `vagrant up` further down will automatically fetch the .box file referenced in the `Vagrantfile` before starting it:
- Looks at http://downloads.mesosphere.com/dcos-docker-sdk/metadata.json (or `DCOS_BOX_URL` if a custom one is set)
- Json in turn (currently) points to https://downloads.mesosphere.com/dcos-docker-sdk/dcos-docker-sdk-20161121-213020.box

This section is effectively an override of the default behavior, allowing the user to specify a different local copy of the .box file (hence USB-stick case)

NimaVaziri:
Besides the consistency comment, LGTM

tools/publish_http.py
nickbp: It does currently log a message like this: https://github.com/mesosphere/dcos-commons/blob/master/tools/publish_aws.py#L168

Difference is `publish_aws` doesn't automatically attempt to add the universe, reasoning being that publish_aws is more often called in CI where automatically adding the universe at build-time may be undesirable (nothing says this couldnt be changed later though)

PR #326, opened 2016-11-23
Allow frameworks to reference the latest CustomExecutor build.

gabrielhartmann:


frameworks/helloworld/build.sh
```diff
@@ -5,7 +5,10 @@ set -e
 curl https://mesosphere.com/wp-content/themes/mesosphere/library/images/assets/sdk/build-sh-start.png >/dev/null 2>&1
 
 FRAMEWORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-${FRAMEWORK_DIR}/../../tools/build_framework.sh hello-world $FRAMEWORK_DIR $1
+ROOT_DIR=$FRAMEWORK_DIR/../..
+BUILD_DIR=$ROOT_DIR/build
+${ROOT_DIR}/gradlew distZip -p ${ROOT_DIR}/sdk/executor
+${FRAMEWORK_DIR}/../../tools/build_framework.sh hello-world $FRAMEWORK_DIR $BUILD_DIR $1
```
nickbp: FRAMEWORK_DIR/../.. => ROOT_DIR

frameworks/kafka/build.sh
```diff
@@ -2,4 +2,6 @@
 set -e
 
 FRAMEWORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-${FRAMEWORK_DIR}/../../tools/build_framework.sh kafka $FRAMEWORK_DIR $1
+ROOT_DIR=$FRAMEWORK_DIR/../..
+BUILD_DIR=$ROOT_DIR/build
+${FRAMEWORK_DIR}/../../tools/build_framework.sh kafka $FRAMEWORK_DIR $BUILD_DIR $1
```
nickbp: FRAMEWORK_DIR/../.. => ROOT_DIR

tools/build_framework.sh
```diff
@@ -10,7 +10,8 @@ fi
 
 FRAMEWORK_NAME=$1
 FRAMEWORK_DIR=$2
```
nickbp: how about having the args to build_framework.sh:
```
FRAMEWORK_DIR=$2
EXECUTOR_DIR=$3 # path to executor to build (parallels FRAMEWORK_DIR), or an empty string ("" will register as an arg for this indexing)
PUBLISH_STEP=$4
```

if EXECUTOR_DIR is empty then no additional work, or if populated then run `gradlew -p ${EXECUTOR_DIR}` and also include `${EXECUTOR_DIR}/build/distributions/*` in the publish step at the end

gabrielhartmann: I'm not sure I understand.  Conditional compilation is accomplished by describing a dependency in the build.gradle file for a framework.  In `hello-world`'s case it wants to be on HEAD all the time so it describes a dependency on the executor project.  On the other hand `kafka` wants to stay on a stable released version of the Executor, so it's resources.json file refers to the published executor and no dependency on the executor project is listed in the build.gradle file.

@mohitsoni has also indicated that as part of the development experience we should probably be moving towards something that relies to as great a degree as possible on gradle and as little as possible on scripts.  @mohitsoni, do you have any thoughts here?  I'll list you as a reviewer.

nickbp: Actually I tried switching cassandra to use this just a bit ago and found that it's pretty close as-is. Disregard earlier comment as it wouldn't really make things easier or more flexible. Instead, how about doing the following:
- Make publish step required, maybe even have it be the first parameter. Have a value of `none` mean to skip the publish step.
- Change `BUILD_DIR` to `ARTIFACT_FILES`, and have it be a variable length argument at the end.

With these changes this section could look like this (with weird handling to make the varargs work):

```
PUBLISH_STEP=$1
shift
FRAMEWORK_NAME=$1
shift
FRAMEWORK_DIR=$1
shift
ARTIFACT_FILES=$@
```

With this, `${ARTIFACT_FILES}` could be passed directly to the publish step at the end, allowing the caller to pass whatever artifact files they need to publish in the package.

Usage would look like:
```
build_framework.sh \
  none \
  kafka \
  frameworks/kafka/ \
  frameworks/kafka/scheduler/scheduler.zip \
  frameworks/kafka/executor/executor.zip
  [...]
```

mohitsoni:
LGTM!

PR #327, opened 2016-11-23
Commons subproject

NimaVaziri:
Refactored repo hierarchy to introduce the commons subproject.

This factors out only what's necessary to export executor content into the `executor` subproject and have the `executor` and `scheduler` depend on `common`.

This also renames the packages from "org.apache.mesos" to "com.mesosphere.sdk".


sdk/commons/src/test/java/org/apache/mesos/offer/CommonTaskUtilsTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+    @Test
+    public void testSetTargetConfiguration() throws Exception {
+        Protos.TaskInfo taskInfo = CommonTaskUtils.setTargetConfiguration(
+                getTestTaskInfo().toBuilder(), testTargetConfigurationId).build();
+        Assert.assertEquals(testTargetConfigurationId, CommonTaskUtils.getTargetConfiguration(taskInfo));
+    }
+
+
+    /*
```
NimaVaziri: Will have to uncomment and refactor these (and other) tests.

PR #348, opened 2016-11-29
a more complicated helloworld example with plan, serial/parallel execution, port, resourceSet, etc.

mbdcos:


mbdcos:
taking so much time to complete

PR #349, opened 2016-11-29
Add support for setting rlimits on a per-pod basis

mrbrowning:
This change adds an `rlimits` field to the `container` YAML specification, with the ability to specify each rlimit by name (e.g. `RLIMIT_NOFILE`) with soft and hard limits attached. To specify "unlimited," one supplies `-1` for both soft and hard limits.


build.gradle
```diff
@@ -31,6 +31,9 @@ allprojects {
     targetCompatibility = '1.8'
 
     repositories {
+        maven {
```
nickbp: Can this go into sdk/scheduler/build.gradle?

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * An exception for errors pertaining to {@link RLimit}.
+     */
+    public static class InvalidRLimitException extends Exception {
+        public InvalidRLimitException(String s) {
+            super(s);
+        }
+    }
+
+
```
nickbp: MEGA-NIT: extra newline

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
```diff
@@ -0,0 +1,117 @@
...
+        return soft;
+    }
+
+    public Long getHard() {
+        return hard;
+    }
+
+    private void validate() throws InvalidRLimitException {
+        if (!RLIMITS.contains(name)) {
+            throw new InvalidRLimitException(name + " is not a valid rlimit");
```
nickbp: Error could list the known-valid values in RLIMITS

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Set;
+
+/**
+ * Representation of an individual rlimit, consisting of a name and optional soft/hard limits.
+ *
+ * A valid instance of this class has either both limits set or neither, with the further constraint that the soft limit
+ * must be less than or equal to the hard limit.
+ */
+public class RLimit {
+    public static final String UNKNOWN = "RLIMIT_UNKNOWN";
```
nickbp: Do these need to be `public`?

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
```diff
@@ -0,0 +1,117 @@
...
+        RLIMITS.add(STACK);
+    }
+
+    private final String name;
+    private final Long soft;
+    private final Long hard;
+
+    public RLimit(
+            @JsonProperty("name") String name,
+            @JsonProperty("soft") Long soft,
```
nickbp: Going by 'both set or both unset' below, could be more explicit here and have soft/hard be `Optional<Long>`'s

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
```diff
@@ -0,0 +1,117 @@
...
+        RLIMITS.add(RTTIME);
+        RLIMITS.add(SIGPENDING);
+        RLIMITS.add(STACK);
+    }
+
+    private final String name;
+    private final Long soft;
+    private final Long hard;
+
+    public RLimit(
```
nickbp: Couldn't hurt to add a `@JsonCreator`

sdk/scheduler/src/main/java/org/apache/mesos/specification/DefaultRLimitSpec.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Default implementation of an {@link RLimitSpec}.
+ */
+public class DefaultRLimitSpec implements RLimitSpec {
+    @NotNull
+    @Size(min = 1)
+    private Collection<RLimit> rlimits;
+
+    public DefaultRLimitSpec(@JsonProperty("rlimits") Collection<RLimit> rlimits) {
```
nickbp: Couldn't hurt to add a `@JsonCreator`

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
```diff
@@ -0,0 +1,117 @@
...
+    private void validate() throws InvalidRLimitException {
+        if (!RLIMITS.contains(name)) {
+            throw new InvalidRLimitException(name + " is not a valid rlimit");
+        }
+
+        if (!(soft == -1 && hard == -1) && !(soft != -1 && hard != -1)) {
+            throw new InvalidRLimitException("soft and hard rlimits must be either both set or both unset");
+        }
+
+        if (soft > hard || soft < hard && soft == -1) {
```
nickbp: Add parens to be clear on expected precedence

sdk/scheduler/src/main/java/org/apache/mesos/specification/yaml/RawContainer.java
```diff
@@ -16,4 +20,13 @@ public String getImageName() {
     public void setImageName(String imageName) {
         this.imageName = imageName;
     }
+
+    public LinkedHashMap<String, RawRLimit> getRLimits() {
+        return rlimits;
+    }
+
+    @JsonProperty("rlimits")
```
nickbp: The `set*()` methods could be removed and replaced by a `@JsonCreator`-style constructor. Then the `imageName`/`rlimits` members could be marked `final`.

sdk/scheduler/src/main/java/org/apache/mesos/specification/yaml/RawRLimit.java
```diff
@@ -0,0 +1,29 @@
+package org.apache.mesos.specification.yaml;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Raw YAML individual rlimit specification.
+ */
+public class RawRLimit {
+    private Long soft;
```
nickbp: Same comment as `RawContainer`: Drop setters in favor of `@JsonCreator` constructor+`final`

sdk/scheduler/src/main/java/org/apache/mesos/specification/ContainerSpec.java
```diff
@@ -3,11 +3,16 @@
...
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
+import java.util.Optional;
+
 /**
  * Spec for defining a Container.
  */
 @JsonDeserialize(as = DefaultContainerSpec.class)
 public interface ContainerSpec {
     @JsonProperty("image-name")
```
mohitsoni: Not part of this PR, but `image-name` has a hyphen and `rlimit_spec` has an underscore.

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
mohitsoni: I think `Optional` isn't intended to be used for input parameters of methods or constructors. It should be for return params of `get` methods. So, in this particular case if soft and hard are optional params, then `getSoft` and `getHard` should return `Optional<Long>`.

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
mohitsoni: 👍

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
mohitsoni: +1

sdk/scheduler/src/main/java/org/apache/mesos/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -113,7 +114,7 @@ public static PodSpec from(RawPod rawPod, String role, String principal) throws
         if (container != null) {
             System.out.println("HELLO");
             System.out.println(container.toString());
```
NimaVaziri: These output lines shouldn't be here...lol.

sdk/scheduler/src/main/java/org/apache/mesos/specification/ContainerSpec.java
mrbrowning: Yeah, I followed the style with respect to every other member with the word "spec" in it, which all have underscores, but that's probably getting too specific -- we should probably standardize on one or the other.

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
mrbrowning: A user hopefully shouldn't be touching the `RLimit` class directly, but if they do it'd be nice to have the constants available on instantiation. I also forgot to use them in `ExecutorUtils`, but making that change now.

sdk/scheduler/src/main/java/org/apache/mesos/specification/ContainerSpec.java
NimaVaziri: Overall looking at other instances of `JsonProperty` it looks like underscores are used more often in spite of a few pre-existing uses of dashes - let's go with underscores?

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
mrbrowning: Weird, this isn't showing up for me. Even a hexdump confirms that the second-to-last character in my local file is `}`...

build.gradle
mrbrowning: This didn't seem to work, it complained about not being able to find the mesos 1.2 snapshot when I moved it into the scheduler's `build.gradle`.

sdk/scheduler/src/main/java/org/apache/mesos/executor/ExecutorUtils.java
```diff
@@ -1,14 +1,39 @@
...
 
 /**
  * Various utility methods for manipulating data in {@link org.apache.mesos.Protos.ExecutorInfo}s.
  */
 public class ExecutorUtils {
     private static final String EXECUTOR_NAME_DELIM = "__";
+    private static final Map<String, Protos.RLimitInfo.RLimit.Type> RLIMIT_TYPE_MAP = new HashMap<>();
+
+    static {
+        RLIMIT_TYPE_MAP.put(RLimit.AS, Protos.RLimitInfo.RLimit.Type.RLMT_AS);
```
nickbp: Could do this here:
```
for (Protos.RLimitInfo.RLimit.Type type : Protos.RLimitInfo.RLimit.Type.values()) {
  RLIMIT_TYPE_MAP.put(type.toString(), type);
}
```

sdk/scheduler/src/main/java/org/apache/mesos/offer/DefaultOfferRequirementProvider.java
```diff
@@ -317,19 +318,30 @@ private void validateTaskRequirements(List<TaskRequirement> taskRequirements) th
         return volumes;
     }
 
+    private Protos.ContainerInfo getContainerInfo(ContainerSpec containerSpec) {
+        Protos.ContainerInfo.Builder containerInfo = Protos.ContainerInfo.newBuilder()
+                .setType(Protos.ContainerInfo.Type.MESOS);
+
+        if (containerSpec.getImageName().isPresent()) {
+                containerInfo.setDocker(Protos.ContainerInfo.DockerInfo.newBuilder()
```
nickbp: `containerInfo.getDockerBuilder().setImage(containerSpec.getImageName().get());`

sdk/scheduler/src/main/java/org/apache/mesos/specification/DefaultContainerSpec.java
```diff
@@ -5,27 +5,35 @@
...
     private String imageName;
+    @Valid
+    private RLimitSpec rLimitSpec;
 
     @JsonCreator
     public DefaultContainerSpec(
-            @JsonProperty("image-name")
-            String imageName) {
+            @JsonProperty("image-name") String imageName,
+            @JsonProperty("rlimit_spec") RLimitSpec rLimitSpec) {
```
nickbp: You _might_ need to have these be optionals as well for de/reserialization to work as expected? It'd be good to have a unit test that checks de/reserialization works

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
nickbp: Actually, could just do `Protos.RLimitInfo.RLimit.Type.values().toString()` here.

sdk/scheduler/src/main/java/org/apache/mesos/specification/yaml/YAMLToInternalMappers.java
nickbp: This looks like something I'd have added...

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Set;
+
+/**
+ * Representation of an individual rlimit, consisting of a name and optional soft/hard limits.
+ *
+ * A valid instance of this class has either both limits set or neither, with the further constraint that the soft limit
+ * must be less than or equal to the hard limit.
+ */
+public class RLimit {
+    public static final String UNKNOWN = "RLIMIT_UNKNOWN";
```
nickbp: Given the `values()` iteration suggestion in ExecutorUtils.java, are these needed anymore? Looks like they can be removed.

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
mrbrowning: They're slightly different - Mesos spells it `RLMT`, which I didn't want to duplicate because it's not the friendliest abbreviation for new users.

nickbp:
Ping, what's left here? (other than a rebase)

sdk/scheduler/src/main/java/org/apache/mesos/specification/util/RLimit.java
nickbp: Eh I feel like mesos' names will be pretty easy for someone digging into rlimits to understand, and we won't need to maintain our own list as their list evolves

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/util/RLimit.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Set;
+
+/**
+ * Representation of an individual rlimit, consisting of a name and optional soft/hard limits.
+ *
+ * A valid instance of this class has either both limits set or neither, with the further constraint that the soft limit
+ * must be less than or equal to the hard limit.
+ */
+public class RLimit {
+    private static final Set<String> RLIMITS = new HashSet<>();
```
nickbp: Could do the following to consolidate the enum<->String handling that's currently spread across here and `DefaultOfferRequirementProvider`:
- `DefaultOfferRequirementProvider`'s `RLIMIT_TYPE_MAP` could be moved here, replacing this `RLIMITS` Set
- `DefaultOfferRequirementProvider.getRLimitInfo()` can then call a `@JsonIgnore` `getEnum()` function which returns the `RLimit.Type`

(I was about to say 'what if the name isn't present in RLIMIT_TYPE_MAP?' in DefaultOfferRequirementProvider, and then realized that that check was effectively happening in `validate()` over here)

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/util/RLimit.java
```diff
@@ -0,0 +1,87 @@
...
+        return Optional.ofNullable(soft);
+    }
+
+    public Optional<Long> getHard() {
+        return Optional.ofNullable(hard);
+    }
+
+    private void validate() throws InvalidRLimitException {
+        if (!RLIMITS.contains(name)) {
+            throw new InvalidRLimitException(name + " is not a valid rlimit of " + RLIMITS.toString());
```
nickbp: Maybe an error more like '<name> is not a valid rlimit name, expected one of: <names>'

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/util/RLimitTest.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @Test(expected = RLimit.InvalidRLimitException.class)
+    public void testRLimitRequiresValidName() throws RLimit.InvalidRLimitException {
+        new RLimit("NONSENSE", 0L, 1L);
+    }
+
+    @Test(expected = RLimit.InvalidRLimitException.class)
+    public void testRLimitRequiresBothLimits() throws RLimit.InvalidRLimitException {
+        new RLimit("NONSENSE", 0L, -1L);
```
nickbp: Should have a valid name to ensure that the bad limit is what's throwing the exception? (here and below)

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/util/RLimit.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * A valid instance of this class has either both limits set or neither, with the further constraint that the soft limit
+ * must be less than or equal to the hard limit.
+ */
+public class RLimit {
+    private static final Set<String> RLIMITS = new HashSet<>();
+
+    static {
+        for (Protos.RLimitInfo.RLimit.Type rLimitType : Protos.RLimitInfo.RLimit.Type.values()) {
+            RLIMITS.add(rLimitType.toString());
```
nickbp: If you wanted to support e.g. `AS` instead of requiring `RLMT_AS` for convenience, the map (once moved from `DefaultOfferRequirementProvider`) could strip out the `RLMT_` on names when present:

```
for (Protos.RLimitInfo.RLimit.Type rLimitType : Protos.RLimitInfo.RLimit.Type.values()) {
    String typeName = rLimitType.toString();
    if (typeName.startsWith("RLMT_")) {
        typeName = typeName.substring(5);
    }
    RLIMIT_TYPE_MAP.put(typeName, rLimitType);
}
```

Or (maybe better) the mesos `RLMT_` prefix could be replaced with `RLIMIT_` so that names align with docs: https://linux.die.net/man/2/setrlimit

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/util/RLimit.java
nickbp: Assuming the above `RLMT_` => `RLIMIT_` conversion in expected names is performed, this error could also directly cite the docs with e.g. `See man setrlimit(2)`.

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/validation/CapabilityValidatorTest.java
```diff
@@ -0,0 +1,74 @@
...
+        CapabilityValidator capabilityValidator = new CapabilityValidator(
+                new Capabilities(mockDcosCluster.getDcosCluster()));
+
+        ClassLoader classLoader = getClass().getClassLoader();
+        File file = new File(classLoader.getResource("valid-exhaustive.yml").getFile());
+        DefaultServiceSpec serviceSpec = YAMLServiceSpecFactory.generateServiceSpec(generateRawSpecFromYAML(file));
+
+        capabilityValidator.validate(serviceSpec);
+    }
+}
```
nickbp: missing newline

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultContainerSpec.java
```diff
@@ -5,27 +5,35 @@
...
 
     @Override
-    public String getImageName() {
-        return imageName;
+    public Optional<String> getImageName() {
+        return Optional.ofNullable(imageName);
+    }
+
+    @Override
+    public Optional<RLimitSpec> getRLimitSpec() {
```
nickbp: Maybe just a `Collection<RLimit>` named `rlimits` here? It looks like the current schema would work out as:

```
"container" : {
  "rlimit-spec": {
    "rlimits": [
      ...
    ]
  }
}
```

When it could just be:

```
"container" : {
  "rlimits": [
    ...
  ]
}
```

sdk/scheduler/src/test/resources/valid-exhaustive.yml
```diff
@@ -9,6 +9,11 @@ pods:
   meta-data:
     placement: "avoid-type: meta-data"
     count: 2
+    container:
+      rlimits:
+        RLIMIT_AS:
```
nickbp: Maybe include an example of an rlimit without specified `soft`/`hard`

PR #351, opened 2016-11-30
Fix regression of bug where ports were being added as individual scalar resources

mrbrowning:


sdk/scheduler/src/test/java/org/apache/mesos/specification/DefaultServiceSpecTest.java
```diff
@@ -51,6 +53,30 @@ public void validSimple() throws Exception {
...
+        ClassLoader classLoader = getClass().getClassLoader();
+        File file = new File(classLoader.getResource("valid-multiple-ports.yml").getFile());
+        RawServiceSpecification rawServiceSpecification = YAMLServiceSpecFactory.generateRawSpecFromYAML(file);
+        DefaultServiceSpec serviceSpec = YAMLServiceSpecFactory
+                .generateServiceSpec(rawServiceSpecification);
+
+        List<ResourceSpecification> portsResources = serviceSpec.getPods().get(0).getTasks().get(0).getResourceSet()
+                .getResources()
+                .stream()
+                .filter(r -> r.getName() == "ports")
```
mohitsoni: This should be `"ports".equals(r.getName())` OR `Objects.equals("ports", r.getName())`. In it's current form it just does reference equality.

mrbrowning:

PR #359, opened 2016-12-01
Move classes out of utils/ into the packages that use them

nickbp:
Every class in `utils` could be in the package that uses it:

- Algorithms (now RangeAlgorithms): only used in `offer`
- ValidationUtils: only used in `specification`
- WriteOnceLinkedHashMap: only used in `specification.yaml`

With that, the `utils` package goes away!


docs/publish.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+
+# Publishes docs changes to the gh-pages branch.
+
+# abort script at first error:
+set -e
```
mrbrowning: For maximum defensiveness, maybe `set -eu`? I always add `-o pipefail` reflexively too, but it's not strictly necessary here since there are no pipes.

nickbp: Whoops, actually `publish.sh` was something I was still working on locally for auto-publishing doc artifacts to `gh-pages`. I'll take it out of this PR (but will make the change you suggested)

PR #362, opened 2016-12-01
Move HTTP API classes into a common package tree

nickbp:
Resolves current situation where HTTP APIs are distributed throughout the source tree.

- `*Resource.java` => `api` package
- `*Info.java` => `api.types` package


sdk/scheduler/src/main/java/org/apache/mesos/api/PlansResource.java
```diff
@@ -193,4 +196,17 @@ public Response restartCommand(
             return Optional.empty();
         }
     }
+
+    static class CommandResultInfo {
```
gabrielhartmann: Is this new or moved from somewhere else?  Why the `Info` suffix?  To match protobufs?  But it isn't a protobuf.

nickbp:
The separate API package code should only change when a user/CLI-facing interface change is expected. Having a separate dedicated package for these endpoints provides some assurance that changes to that API only occur when expected.

Having a common package for all the API endpoints also makes them a bit more discoverable by developers/operators that may be browsing the source tree.

gabrielhartmann:
@mrbrowning:  We tried having the APIs defined on a per sub-component basis, but it was scattering the API definition so it was hard to figure out if everything that should be exposed etc. was being covered.  I think this is @nickbp's discoverability point.

PR #363, opened 2016-12-02
Remove Element.setStatus(), various other fixes/nitpicks via Cassandra bump

nickbp:
Was also seeing local errors about a missing `LIBMESOS_URI` which are now fixed.


sdk/scheduler/src/main/java/org/apache/mesos/scheduler/plan/DefaultStep.java
```diff
@@ -137,16 +105,18 @@ public Status getStatus() {
     }
 
     @Override
-    public List<Element> getChildren() {
+    public List<Element<?>> getChildren() {
```
gabrielhartmann: Why change this?

sdk/scheduler/src/main/java/org/apache/mesos/scheduler/plan/DefaultStep.java
```diff
@@ -137,16 +105,18 @@ public Status getStatus() {
...
         return Collections.emptyList();
     }
 
     @Override
-    @SuppressWarnings("unchecked")
-    public Strategy<? extends Element> getStrategy() {
+    public Strategy<Step> getStrategy() {
```
gabrielhartmann: A Phase has a strategy to deploy Steps.  A Step doesn't have Steps as children.  I suppose this gets rid of the unchecked warning...

sdk/scheduler/src/main/java/org/apache/mesos/scheduler/plan/DefaultStep.java
nickbp: Just fixed a warning in eclipse

sdk/scheduler/src/main/java/org/apache/mesos/scheduler/plan/DefaultStep.java
nickbp: Yes this just was to remove the need for the `SuppressWarnings`

sdk/scheduler/src/main/java/org/apache/mesos/scheduler/plan/PlanUtils.java
```diff
@@ -27,99 +35,135 @@
...
                 element.getClass().getName(), element.getName(), element.getId(), element.getStatus());
     }
 
-    @SuppressWarnings("unchecked")
-    public static final List<String> getErrors(List<String> errors, Element parent) {
+    /**
+     * Returns all of the errors from the parent and all its child {@link Element}s. Intended for
+     * use by implementations of {@Link Element#getErrors()}.
+     *
+     * @param parentErrors Errors to be included from the parent itself, will be modified in-place.
```
nickbp: Correct, this was once the case until it was fixed. Updated the javadoc

nickbp:
Filed a ticket about refactoring PlanUtils.

PR #370, opened 2016-12-08
a quick fix for cli, update in README, and removal of 'health' command in cli 

mbdcos:
* removing 'health' command from cli.  no admin/healthcheck  in com.mesosphere.sdk.api
 
     ./dcos-hello-world-darwin hello-world health
    HTTP GET Query for http://35.164.209.100/service/hello-world/admin/healthcheck failed: 404 Not Found

 *   updated README.md, added a note about api-port

  *  api-port is dynamically set by Marathon.  In yml file, it should be “api-port: {{PORT0}}”
    
    Marathon dynamically selects a port, and pass this info to the framework, so framework can start the service on that selected port.

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -92,7 +92,6 @@ func HandleCommonSections(app *kingpin.Application, connectionTypes []string) {
 	HandleConnectionSection(app, connectionTypes)
 	HandlePlanSection(app)
 	HandleStateSection(app)
-	HandleHealthSection(app)
 }
 
 // Config section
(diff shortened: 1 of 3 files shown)


nickbp:
The health check endpoint is used by Cassandra and Kafka. *If* we want to keep the command around for those services (I'm not sure we do), there should be separate PRs which add the command back in their respective CLIs:
- https://github.com/mesosphere/dcos-kafka-service/blob/master/cli/dcos-kafka/main.go
- https://github.com/mesosphere/dcos-cassandra-service/blob/master/cli/dcos-cassandra/main.go

PR #375, opened 2016-12-10
"uris" keyword is added in pod level for yaml file.

mbdcos:
- DefaultPodSpec,
  - YAMLToInternalMap,
  - RawPod,
  - PodSpec are updated.

While pasing to the TaskSpec, pod uris' are appended to task uris'

added  simple_uri.yml in hello-world framework

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
@@ -81,6 +81,18 @@ public class YAMLToInternalMappers {
         WriteOnceLinkedHashMap<String, RawResourceSet> rawResourceSets = rawPod.getResourceSets();
         String user = rawPod.getUser();
         LinkedHashMap<String, RawTask> tasks = rawPod.getTasks();
+ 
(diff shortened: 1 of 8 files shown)


mbdcos:
example svc_uri.yml:

"README.md" is given as pod uri, another file "generate.sh" is given as task uri. Both are downloaded!

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -81,6 +81,12 @@ public static PodSpec from(RawPod rawPod, String role, String principal) throws
         WriteOnceLinkedHashMap<String, RawResourceSet> rawResourceSets = rawPod.getResourceSets();
         String user = rawPod.getUser();
         LinkedHashMap<String, RawTask> tasks = rawPod.getTasks();
+ 
+        Collection<String> rawTaskUris = rawPod.getUris();
+        Collection<URI> uris = new ArrayList<>();
+        for (String uriStr : rawTaskUris) {
+            uris.add(new URI(uriStr));
```
gabrielhartmann: `uris.addAll()` may be used instead of a loop.

PR #376, opened 2016-12-13
YAML: Add 'colocate-types'/'avoid-types' for devs, make 'placement' Marathon-formatted for ops

nickbp:
In practice, type colocate/avoid is the main thing used in the yaml specs, so this optimizes for that common case. Devs may still implement their own fully custom placement rule trees, but that now requires using the Java API to do. Requiring the Java API for custom placement rules may be useful in its own right as authoring placement rule trees in yaml, particularly from scratch, is tricky/error-prone.

YAML:
- `avoid-tasks`/`colocate-tasks` are added to pod spec in yaml as shortcut for common placement use-case
- fully custom `placement` option is removed. devs who need custom placement logic are likely to be using java anyway, and writing the string form correctly is difficult.

Java:
- `getAvoidTasks()`/`getColocateTasks()` are added to pod spec to map from the yaml shortcut (or may also be used as a shortcut when building `PodSpec`s).
- fully custom placement setting is retained in the Java API, allowing devs to implement their own placement logic, and/or implement and use their own custom PlacementRule classes.

### DIFF
--- a/frameworks/helloworld/src/main/dist/svc_plan.yml
+++ b/frameworks/helloworld/src/main/dist/svc_plan.yml
@@ -5,6 +5,9 @@ api-port: {{PORT0}}
 pods:
   hello:
     count: {{HELLO_COUNT}}
+    avoid-types:
+      - hello
+      - world
     tasks:
       server:
         goal: RUNNING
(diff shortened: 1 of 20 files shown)


nickbp:
Had some more conversations on formatting, going to make some changes then reopen for review.

nickbp:
OK, it feels like it'd make more sense to instead just omit these from the yml and have framework devs do any advanced placement rules like cross-type avoid/colocate in Java-land. Meanwhile the yml will still have a marathon-style placement constraint which can be used for the straightforward cases (like 'avoid-self' which can be done with `hostname:UNIQUE`)

PR #384, opened 2016-12-16
Added data integrity tests to HDFS

NimaVaziri:
Mostly done, still a bit WIP.

### DIFF
--- a/frameworks/hdfs/integration/tests/test_data_integrity.py
+++ b/frameworks/hdfs/integration/tests/test_data_integrity.py
@@ -11,25 +11,24 @@ from tests.test_utils import (
 )
 
 TEST_CONTENT_SMALL = "This is some test data"
-# TEST_CONTENT_LARGE = Give a large file as input to the write/read commands...
+# TODO: TEST_CONTENT_LARGE = Give a large file as input to the write/read commands...
 TEST_FILE_1_NAME = "test_1"
 TEST_FILE_2_NAME = "test_2"
 HDFS_CMD_TIMEOUT_SEC = 60
 
 
 def setup_module(module):
-    #uninstall()
-    #install()
+    uninstall()
+    install()
     check_health()
 
 
 def teardown_module(module):
-    #uninstall()
-    shakedown.wait_for(lambda: delete_data_from_hdfs("data-0-node.hdfs.mesos", TEST_FILE_1_NAME), HDFS_CMD_TIMEOUT_SEC)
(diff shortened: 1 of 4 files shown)


susanxhuynh:
- Removed the hardcoded JRE and Hadoop versions from shell commands.
- Replaced sleep() with polling.
- Fixed a bug in the janitor cleanup command.

frameworks/hdfs/integration/tests/test_data_integrity.py
```diff
@@ -0,0 +1,112 @@
...
+    delete_command = "./bin/hdfs dfs -rm -r /{}".format(filename)
+    rc, _ = run_hdfs_command(host, delete_command)
+    return rc
+
+
+def run_hdfs_command(host, command):
+    """
+    Go into the Data Node hdfs directory, set JAVA_HOME, and execute the command.
+    """
+    java_home = find_java_home(host)
```
nickbp: Should be able to just have something like this, without needing to run commands on the agent at all. The `framework_id` check ensures that we select the right data node in the event multiple `hdfs` instances are installed (or in the event some other service has an executor named 'data'):

```
framework_id = shakedown.get_service_framework_id("hdfs")
[...]
"JAVA_HOME=$(ls -d /var/lib/mesos/slave/slaves/*/frameworks/{}/executors/data__*/runs/latest/jre*/)".format(framework_id)
```

The `ls -d` command will be run on the agent system and will resolve the correct path.

frameworks/hdfs/integration/tests/test_data_integrity.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # Find hdfs home directory by looking up the Data Node process.
+    # Hdfs directory is found in an arg to the java command.
+    hdfs_dir_cmd = """ps -ef | grep hdfs | grep DataNode \
```
nickbp: Similar to the above, this lookup could be something like:
```
cd $(ls -d /var/lib/mesos/slave/slaves/*/frameworks/{}/executors/data__*/runs/latest/hadoop-*/)".format(framework_id)
```
(this gives us the explicit framework id filtering as above)

frameworks/hdfs/universe/config.json
```diff
@@ -10,8 +10,13 @@
           "type":"string",
           "default":"hdfs"
         },
+        "role": {
```
nickbp: Does this need a corresponding change in marathon.json.mustache?

frameworks/hdfs/integration/tests/test_data_integrity.py
susanxhuynh: I tried to implement this suggestion but the "ls -d" returns multiple directories after the DataNode restart. I will leave the command as is.

PR #388, opened 2016-12-20
Implement 'pods' endpoint for info/restart/replace

nickbp:
Also makes several improvements to:
- logging in offer evaluation code, which was done while debugging an issue during 'pods replace' development.
- launching 'world' pods in the default hello-world instance (for some reason the 'world' section was missing?), while also hooking up all package config values correctly


nickbp:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/OfferAccepter.java
```diff
@@ -106,8 +106,9 @@ private void record(List<OfferRecommendation> recommendations) throws Exception
     }
 
     private static void logOperations(List<Operation> operations) {
+        logger.info("Performing {} operations:", operations.size());
```
mrbrowning: Does this info have value for an operator?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -478,9 +475,10 @@ private void initializeResources() throws InterruptedException {
 
     private void declineOffers(SchedulerDriver driver, List<Protos.OfferID> acceptedOffers, List<Protos.Offer> offers) {
         final List<Protos.Offer> unusedOffers = OfferUtils.filterOutAcceptedOffers(offers, acceptedOffers);
+        LOGGER.info("Declining {} unused offers:", unusedOffers.size());
         unusedOffers.stream().forEach(offer -> {
             final Protos.OfferID offerId = offer.getId();
-            LOGGER.info("Declining offer: " + offerId.getValue());
+            LOGGER.info("  {}", offerId.getValue());
```
mrbrowning: Is the two spaces before each offer intentional? They'll be on their own line anyway, right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/PodInstance.java
```diff
@@ -22,6 +22,10 @@ default String getName() {
     }
 
     static String getName(PodSpec podSpec, int index) {
-        return String.format("%s-%d", podSpec.getType(), index);
+        return getName(podSpec.getType(), index);
+    }
+
+    public static String getName(String podType, int index) {
```
mrbrowning: What case might have this being called with something other than `podSpec.getType()`? This might be a small instance of YAGNI.

nickbp: It's called at PodsResource.java:L306, where only the launched TaskInfo is available.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
mrbrowning: OK, fair enough. My only concern here is that individual log lines become way more dependent on context for their meaning, which can make gleaning info from them a little bit more involved, but it's a matter of taste.

nickbp:
retest this please

PR #389, opened 2016-12-21
[WIP] [INFINITY-706] Add ZK lock to prevent multiple running schedulers

mgummelt:
cc @nickbp for review

blocked on https://github.com/dcos/shakedown/pull/101

### DIFF
--- a/frameworks/helloworld/integration/tests/test_sanity.py
+++ b/frameworks/helloworld/integration/tests/test_sanity.py
@@ -199,27 +199,34 @@ def test_lock():
     So in order to verify that the scheduler fails immediately, we ensure
     that the ZK config state is unmodified.'''
 
-    zk_path = "dcos-service-{}/ConfigTarget".format(PACKAGE_NAME)
+    marathon_client = dcos.marathon.create_client()
 
     # Get ZK state from running framework
+    zk_path = "dcos-service-{}/ConfigTarget".format(PACKAGE_NAME)
     zk_config_old = shakedown.get_zk_node_data(zk_path)
 
-    # Install second framework
-    options = {"service": {"name": "hello-world-lock"}, "hello": {"count": 2}}
-    install(additional_options=options, wait_for_completion=False)
+    # Get marathon app
+    app_id = "/{}".format(PACKAGE_NAME)
+    app = marathon_client.get_app(app_id)
+    old_timestamp = app.get("lastTaskFailure", {}).get("timestamp", None)
 
-    # Wait for second scheduler to terminate
-    app_id = "/hello-world-lock"
-    client = dcos.marathon.create_client()
-    tasks = client.get_tasks(app_id)
-    task_id = tasks[0]["id"]
-    shakedown.wait_for_task_completion(task_id)
+    # Scale to 2 instances
+    labels = app["labels"]
+    labels.pop("MARATHON_SINGLE_INSTANCE_APP")
+    marathon_client.update_app(app_id, {"labels": labels})
+    shakedown.deployment_wait()
+    marathon_client.update_app(app_id, {"instances": 2})
+
+    # Wait for second scheduler to fail
+    fn = lambda: marathon_client.get_app(app_id).get(
+        "lastTaskFailure", {}).get("timestamp", None)
+    success = lambda timestamp: (timestamp != old_timestamp,
+                                 "second scheduler has not yet failed")
+    spin(fn, success)
 
     # Verify ZK is unchanged
-    zk_config_new = get_zk_node_data(zk_path)
+    zk_config_new = shakedown.get_zk_node_data(zk_path)
     assert zk_config_old == zk_config_new
-
-    marathon.remove_app(app_id)
 
 
 def get_task_ids(prefix):
--- a/frameworks/helloworld/integration/tests/test_sanity.py
+++ b/frameworks/helloworld/integration/tests/test_sanity.py
@@ -1,4 +1,5 @@
 import dcos.http
+import dcos.marathon
 import json
 import pytest
 import re
@@ -190,6 +191,36 @@ def test_pods_replace():
     #assert old_agent != new_agent
 
 
+@pytest.mark.sanity
+def test_lock():
+    '''This test verifies that a second scheduler fails to startup when
+    an existing scheduler is running.  Without locking, the scheduler
+    would fail during registration, but after writing its config to ZK.
+    So in order to verify that the scheduler fails immediately, we ensure
+    that the ZK config state is unmodified.'''
+
+    ZK_PATH = "dcos-service-{}/ConfigTarget".format(PACKAGE_NAME)
+
+    # Get ZK state from running framework
+    zk_config_old = shakedown.get_zk_node_data(ZK_PATH)
+
+    # Install second framework
+    options = {"service": {"name": "hello-world-lock"}, "hello": {"count": 2}}
+    install(additional_options=options, wait_for_completion=False)
+
+    # Wait for second scheduler to terminate
+    client = dcos.marathon.create_client()
+    tasks = client.get_tasks("/hello-world-lock")
+    task_id = tasks[0]["id"]
+    shakedown.wait_for_task_completion(task_id)
+
+    # Verify ZK is unchanged
+    zk_config_new = get_zk_node_data("ZK_PATH")
+    assert zk_config_old == zk_config_new
+
+    uninstall()
+
+
 def get_task_ids(prefix):
     tasks = shakedown.get_service_tasks(PACKAGE_NAME)
     prefixed_tasks = [t for t in tasks if t['name'].startswith(prefix)]
@@ -253,4 +284,3 @@ def tasks_not_updated(prefix, old_task_ids):
         )
 
     return spin(fn, success_predicate)
-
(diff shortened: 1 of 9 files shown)


mgummelt:
retest this please

frameworks/helloworld/integration/tests/test_sanity.py
```diff
@@ -71,6 +72,37 @@ def test_bump_hello_nodes():
...
+    So in order to verify that the scheduler fails immediately, we ensure
+    that the ZK config state is unmodified.'''
+
+    zk_path = "dcos-service-{}/ConfigTarget".format(PACKAGE_NAME)
+
+    # Get ZK state from running framework
+    zk_config_old = shakedown.get_zk_node_data(zk_path)
+
+    # Install second framework
+    options = {"service": {"name": "hello-world-lock"}, "hello": {"count": 2}}
```
nickbp: Does this name result in storing the zk data at `/dcos-service-hello-world-lock`? Or does something else ensure that it tries to use `/dcos-service-hello-world`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorUtils.java
```diff
@@ -20,14 +20,14 @@
 
     private static final String PATH_DELIM = "/";
 
-    static final int DEFAULT_CURATOR_POLL_DELAY_MS = 1000;
-    static final int DEFAULT_CURATOR_MAX_RETRIES = 3;
+    public static final int DEFAULT_CURATOR_POLL_DELAY_MS = 1000;
```
nickbp: Looks like `CuratorConfigStore` + `CuratorStateStore` were both using these to do the following:

```
new ExponentialBackoffRetry(
                CuratorUtils.DEFAULT_CURATOR_POLL_DELAY_MS,
                CuratorUtils.DEFAULT_CURATOR_MAX_RETRIES)
```

Could:
- Mark these private
- Add a static `getDefaultRetryPolicy()` function
- Switch `CuratorConfigStore`/`CuratorStateStore`/`DefaultScheduler` to call that function

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -75,9 +85,43 @@ public DefaultService(ServiceSpec serviceSpec, Collection<Plan> plans) {
         register(serviceSpec, this.plans);
     }
 
+    private void lock() {
+        String rootPath = CuratorUtils.toServiceRootPath(serviceSpec.getName());
+        String lockPath = CuratorUtils.join(rootPath, "LOCK_PATH");
+        curatorMutex = new InterProcessMutex(curatorClient, lockPath);
+
+        LOGGER.info("Acquiring ZK lock...");
```
nickbp: Could include lockPath here and in the error messages below, to help debuggability just in case.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -75,9 +85,43 @@ public DefaultService(ServiceSpec serviceSpec, Collection<Plan> plans) {
...
+        try {
+            if (!curatorMutex.acquire(10, TimeUnit.SECONDS)) {
+                LOGGER.error("Failed to acquire ZK lock.  Exiting.");
```
nickbp: Could include some more contextual info in this error, something like:
`Are you running another instance named '<serviceSpec.getName()>'?`

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -75,9 +85,43 @@ public DefaultService(ServiceSpec serviceSpec, Collection<Plan> plans) {
...
+                System.exit(1);
```
nickbp: May as well use the hardExit() wrapper call and add a SchedulerErrorCode value for it.

Optional: Replace SchedulerErrorCode enum with a class that has explicit integer values, as enum values as ints can be brittle. Something like:

```
public class SchedulerExitCode {

  [... list of public static final int's for the error code values? or an ImmutableMap? ...]

  public SchedulerExitCode(int value) {
    this.value = value;
  }

  public void exit() {
    LOGGER.error("Exiting with code {}", value);
    System.exit(value);
  }
}
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -75,9 +85,43 @@ public DefaultService(ServiceSpec serviceSpec, Collection<Plan> plans) {
...
+        } catch (Exception ex) {
+            LOGGER.error("Error acquiring ZK lock.", ex);
+        }
+    }
+
+    private void unlock() {
+        try {
+            curatorMutex.release();
+        } catch (Exception ex) {
+            LOGGER.error("Error releasing ZK lock.", ex);
```
nickbp: Should be able to provide the zk path using `curatorMutex.getLockPath()`

frameworks/helloworld/integration/tests/test_sanity.py
mgummelt: "hello-world-lock" is just the service name for this test, like "kafka".  So whatever dcos-commons does normally to name the ZK directory is what it does here.  The lock itself is a child znode.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorUtils.java
mgummelt: I don't see any issue with making a constant public.  Do you?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
mgummelt: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
mgummelt: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
mgummelt: Done, though I didn't change the enum.

I had to move `hardExit` to `SchedulerUtils`

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
mgummelt: It doesn't seem so.  There's a `getParticipantNodes`, but it's not clear what it returns.

mgummelt:
cc @nickbp ready for re-review

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorUtils.java
nickbp: It's just that the constants are being used in the same way in three places, could just have those places call a function instead.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
nickbp: Oh whoops it looks like `getLockPath()` is `protected`. Nevermind. Worst case someone can scroll back to when `lock()` was called to figure out the path.

frameworks/helloworld/integration/tests/test_sanity.py
nickbp: Looks like the ZK path is derived from `name` in the service spec yaml. Hello-world's service name is currently hardcoded:

```
dcos-commons/frameworks/helloworld/src/main/dist$ cat svc.yml 
name: "hello-world"
[...]
```

So it should be fine. This might need revisiting if that field stops being hardcoded though. Would it be possible to increase the scheduler task count in marathon, to ensure we get two identical schedulers even if that `name` field becomes templated out (like it probably should be)?

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorUtils.java
mgummelt: Oh I see what you mean.  Yea, done.

frameworks/helloworld/integration/tests/test_sanity.py
mgummelt: Done

mgummelt:
@nickbp ready for re review

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorUtils.java
```diff
@@ -20,8 +21,14 @@
 
     private static final String PATH_DELIM = "/";
 
-    public static final int DEFAULT_CURATOR_POLL_DELAY_MS = 1000;
-    public static final int DEFAULT_CURATOR_MAX_RETRIES = 3;
+    private static final int DEFAULT_CURATOR_POLL_DELAY_MS = 1000;
+    private static final int DEFAULT_CURATOR_MAX_RETRIES = 3;
+
+    public static ExponentialBackoffRetry getDefaultRetry() {
```
nickbp: Mega-nitpick: Could have the return type be `RetryPolicy`?

mgummelt: done

PR #399, opened 2016-12-24
Put the SDK in a single project.

gabrielhartmann:
Executables should consume the SDK.  This includes the CustomExecutor and all framework schedulers.


nickbp:
Are executors supposed to have access to things like the `api` package with it's REST endpoints?

gabrielhartmann:
@nickbp:  I'm not 100% sure I understand your question.  Ideally the executor wouldn't have access to API code as it won't be serving a REST API.  The problem stems initially from the consumption of  *Spec code by both Scheduler and Executor side code.  It's possible moving just that into a different package would work, but I doubt it.

I ran into the need for this change when starting to implement readiness checks on the CustomExecutor.  As I started moving things needed by the Executor in order to run readiness checks, it snowballed and it started to be very unclear why something would be in commons, or scheduler or executor packages.  Code that should live together was getting spread around.

Downsides are size of executor dependencies which are already too large.  Happily we are moving off the custom executor soonish, so that should become a moot point.

With the readiness check working going in Mesos right now and [Default Executor epic](https://mesosphere.atlassian.net/secure/RapidBoard.jspa?rapidView=187&view=planning.nodetail&selectedIssue=INFINITY-799&epics=visible&selectedEpic=INFINITY-797), this problem should go away.

nickbp:
I mean, in the case of the `api` package, it's being moved into `commons` as of this PR, and as a result it'll be visible to anything linking in `commons`, including the custom executor, right? The `api` case is just an example where it looks especially weird. Offer handling is a lot of the scheduler code (probably the largest single component) and it wouldn't make sense to make that visible to the executor either.

I'm expecting we'll have still the current custom executor for a long time as a fallback when a cluster is running an older version of Mesos (other option being to deal with 'legacy' branches of every service for the forseeable future, which would likely be worse). Given that likely lifespan, I'm worried we'll end up making it difficult to keep the 'legacy' custom executor partitioned off from the rest of the code, affecting future maintainability. Say we did a refactor of X in the scheduler code but it turned out the custom executor had its claws in X due to being too intertwined. Suddenly we need to do a lot of extra testing on an old cluster version to make sure we didn't cause a regression in the custom executor, or worse we don't realize the code affected the custom executor and we forget to do that testing at all.

gabrielhartmann:
ok.  I'll see what I can do.

PR #407, opened 2016-12-30
Many SDK fixes from dogfooding the API

nickbp:
- Rename `Hdfs/Kafka/HelloWorldServiceSpecTest` to just `ServiceSpecTest`. The HelloWorld one gets copied into new projects, leading to awkward naming. May as well rename the others to line up.
- Move `TASKCFG_*` routing into `CommandSpec` construction. This ensures that envvars specified via `TASKCFG_*` (eg by end users) get the same treatment as envvars specified via a `ServiceSpec`, and ensures that changes to `TASKCFG_*` envvars are correctly detected/logged in the spec change detection.
- Move Executor-specific functionality in `CommonTaskUtils` into the executor's `ProcessTask`
- Update the executor to include ALL container envvars, not just `MESOS_SANDBOX`. Took a while to find out why `LIBPROCESS_IP` wasn't working in a config template..
- Add a nice diff to the `ServiceSpec` change detection. New behavior is to show the full new config, and its diff vs the old config.
- Internal refactoring in `DefaultOfferRequirementProvider`, dropping `TaskConfigRouter` as that's now handled in `CommandSpec`. Shouldn't be any functional changes otherwise.
- Remove pod type from `TaskSpec` as nothing was actually reading it.


frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -22,10 +22,11 @@ public static void main(String[] args) throws Exception {
                     .apiPort(8080)
                     .addPod(DefaultPodSpec.newBuilder()
                             .count(COUNT)
+                            .type("hello")
                             .addTask(DefaultTaskSpec.newBuilder()
                                     .name("hello")
                                     .goalState(GoalState.RUNNING)
-                                    .commandSpec(DefaultCommandSpec.newBuilder()
+                                    .commandSpec(DefaultCommandSpec.newBuilder("hello")
```
gabrielhartmann: Two uses of the string `"hello"`.  Time to make s constant?  A constant named `HELLO_POD_TYPE` also might make things clearer.  I was confused by the newBuilder with an argument like that.

nickbp: Done.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
```diff
@@ -83,6 +78,31 @@ public UpdateResult updateConfiguration(ServiceSpec candidateConfig) throws Conf
                     String.format("Unable to serialize new config to JSON for logging: %s", e.getMessage())));
         }
 
+        if (targetConfig == null) {
+            LOGGER.info("Skipping config diff: There is no old config target to diff against");
+        } else if (candidateConfigJson == null) {
+            LOGGER.info("Skipping config diff: New target couldn't be represented as JSON");
```
gabrielhartmann: When could this happen?  Should we fail hard here, or is there YAML in the world which can validly not be expressed as JSON?

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
```diff
@@ -83,6 +78,31 @@ public UpdateResult updateConfiguration(ServiceSpec candidateConfig) throws Conf
...
+        } else {
+            // Print a diff of this new config vs the prior config:
```
gabrielhartmann: Could we factor out this diff printing code into a method outside of `updateConfiguration`?

sdk/executor/src/main/java/com/mesosphere/sdk/executor/ProcessTask.java
```diff
@@ -251,4 +258,61 @@ private boolean waitExit(final Duration timeout) {
         }
     }
 
+    /**
+     * Sets up the config files on the executor side.
+     *
+     * @param taskInfo The {@link TaskInfo} to extract the config file data from
+     * @throws IOException if the data in the taskInfo is not valid or the config can't be written to disk
+     */
+    private static void setupConfigFiles(TaskInfo taskInfo) throws IOException {
```
gabrielhartmann: Generally don't like adding code to the CustomExecutor as I'm trying to get rid of it.  But we can rip this code out later for a standalone tool.  Taking it from here or a Utils class makes no difference in the end really.

nickbp: Yeah this is really just moving it out of the commons package since it's not used by the scheduler.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
nickbp: This could happen if candidateConfig.toJsonString() threw an exception (couple lines up). Error handling in that case is to add an error to the plan, but we wouldnt want that to be short-circuited by an NPE here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
nickbp: Good point, done.

PR #409, opened 2016-12-31
Update CanaryStrategy.java

mbdcos:
instead of parentElement.getChildren(), it should be children list (after filter) because it can be empty


gabrielhartmann:
@mbdcos:  Good catch, thanks.  Are you going to add the `|| isWaiting()` change as well?  I would.

PR #413, opened 2017-01-03
Refactor OfferEvaluator to use more general pipeline approach, add dynamic/named vip ports

mrbrowning:
This PR removes `FulfilledRequirement` and changes logic in `OfferEvaluator` to evaluate offers using a general queue based on a series of stages in a pipeline. Each stage takes a `MesosResourcePool` and `OfferRequirement` as input and generates an `OfferRecommendation` and/or modifies the `TaskInfo`s  and `ExecutorInfo` in the `OfferRequirement` as necessary. Stages for creating dynamic ports based on the available ports in the offer and for creating a DC/OS named VIP mapping to a port are included in this PR.

Aside: tests for `OfferEvaluator` have significant added complexity as a result of the fact that it is impossible with the current design to evaluate an offer exclusively from an `OfferRequirement`, since only implementors of `ResourceSpecification` know how to create evaluation stages for resources. An alternative design I considered and half-started involves having `ResourceSpecification` instead know how to create its own `ResourceRequirement`, which then knows how to create its own evaluation stage; this would make constructing test inputs easier (the fake `StateStore` wouldn't need as much preparation, since offer requirements for "existing" pods could be supplied directly to `OfferEvaluator` as is currently done) but would involve another level of indirection and significant changes in the semantics of `ResourceRequirement`. Thoughts on this are welcome.


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -318,11 +315,56 @@ private static void validateTaskRequirements(List<TaskRequirement> taskRequireme
...
+    private static List<Protos.Resource> coalescePorts(Collection<Protos.Resource> resources) {
+        // Within the SDK, each port is handled as its own resource, since they can have extra meta-data attached, but
+        // we can't have multiple "ports" resources on a task info, so we combine them here. Since ports are also added
+        // back onto TaskInfos during the evaluation stage (since they may be dynamic) we actually just clear the ranges
+        // from that resource here to make the bookkeeping easier.
+        Protos.Resource portsResource = null;
+        List<Protos.Resource> coalescedResources = new ArrayList<>();
+
+        for (Protos.Resource r : resources) {
+            if (r.getName().equals("ports") && portsResource == null) {
```
gabrielhartmann: There's a `Constants.java` file.  Might be good to make a constant for "ports" there.  If not there, somewhere.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -318,11 +315,56 @@ private static void validateTaskRequirements(List<TaskRequirement> taskRequireme
             }
         }
 
-        return resources;
+        return coalescePorts(resources);
+    }
+
+    private static List<Protos.Resource> coalescePorts(Collection<Protos.Resource> resources) {
+        // Within the SDK, each port is handled as its own resource, since they can have extra meta-data attached, but
+        // we can't have multiple "ports" resources on a task info, so we combine them here. Since ports are also added
```
gabrielhartmann: I don't understand.  This function still returns a list.  Was the intent to generate a single `Resource`?  Also there's a `mergeRanges` in `RangeAlgorithms.java` method for merging ports, might be useful.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -318,11 +315,56 @@ private static void validateTaskRequirements(List<TaskRequirement> taskRequireme
...
+            }
+        }
+
+        return volumes;
     }
 
     private static Protos.ExecutorInfo getNewExecutorInfo(PodSpec podSpec) throws IllegalStateException {
-        Protos.ExecutorInfo.Builder executorInfoBuilder = Protos.ExecutorInfo.newBuilder()
+        Protos.CommandInfo.URI executorURI;
+        Protos.CommandInfo.URI libmesosURI;
```
gabrielhartmann: Are these URIs used?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
```diff
@@ -115,6 +119,10 @@ public Offer getOffer() {
         return Optional.empty();
     }
 
+    public void retainOnly(Offer offer) {
```
gabrielhartmann: Javadoc comment please.  Not immediately obvious to me why the word `retain` is used.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,189 @@
...
+    protected void setResourceRequirement(ResourceRequirement resourceRequirement) {
+        this.resourceRequirement = resourceRequirement;
+    }
+
+    protected ResourceRequirement getResourceRequirement() {
+        return resourceRequirement;
+    }
+
+    protected Optional<String> getTaskName() {
+        return taskName;
```
gabrielhartmann: `return Optional.ofNullable(taskName);`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * This class evaluates an offer against a given {@link OfferRequirement}, ensuring that it contains a sufficient amount
+ * or value of the supplied {@link Resource}, and creating a {@link ReserveOfferRecommendation} or
+ * {@link UnreserveOfferRecommendation} where necessary.
+ */
+public class ResourceEvaluationStage implements OfferEvaluationStage {
+    private static final Logger logger = LoggerFactory.getLogger(OfferEvaluator.class);
+
+    private ResourceRequirement resourceRequirement;
+    private Optional<String> taskName;
```
gabrielhartmann: I think the best practice for Optionals is that they aren't fields.  I know we have broken this rule a lot.  Functions should return Optionals as is done in the `getTaskName` method.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,189 @@
...
+        logger.info("Satisfying resource requirement: {}\nwith resource: {}",
+                TextFormat.shortDebugString(resourceRequirement.getResource()),
+                TextFormat.shortDebugString(mesosResource.getResource()));
+
+        validateRequirements(offerRequirement);
+        setProtos(offerRequirement, fulfilledResource);
+    }
+
+    protected void validateRequirements(OfferRequirement offerRequirement) throws OfferEvaluationException {
+        if (!getTaskName().isPresent() && offerRequirement.getExecutorRequirementOptional().isPresent()) {
```
gabrielhartmann: I don't understand the use of taskName here.  Can you explain please?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,189 @@
...
+                            unreserveResource, resourceRequirement.getResourceId());
+
+                    offerResourcePool.release(new MesosResource(
+                            ResourceUtils.getUnreservedResource(resourceRequirement.getName(), unreserveValue)));
+                    offerRecommendationSlate.addUnreserveRecommendation(
+                            new UnreserveOfferRecommendation(offerResourcePool.getOffer(), unreserveResource));
+                    fulfilledResource = getFulfilledResource(new MesosResource(resourceRequirement.getResource()));
+                }
+
+                if (ValueUtils.compare(reserveValue, ValueUtils.getZero(reserveValue.getType())) > 0) {
```
gabrielhartmann: Should this be an `else if` clause?  Not possible to reserve and unreserve for a single ResourceRequirement right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/VolumeEvaluationStage.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.mesos.Protos.Resource.DiskInfo;
+
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * This class evaluates an offer against a given {@link com.mesosphere.sdk.offer.OfferRequirement}, ensuring that it
+ * contains an appropriately-sized volume, and creating any necessary instances of
+ * {@link com.mesosphere.sdk.offer.ReserveOfferRecommendation},
+ * {@link com.mesosphere.sdk.offer.UnreserveOfferRecommendation}, and
```
gabrielhartmann: Once a volume is created we should never change its size.  So UNRESERVE should never occur for disks outside of the `ResourceCleaner` or similar.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,189 @@
...
+                }
+            }
+        } else {
+            if (resourceRequirement.reservesResource()) {
+                logger.info("Reserves Resource");
+                offerRecommendationSlate.addReserveRecommendation(
+                        new ReserveOfferRecommendation(offerResourcePool.getOffer(), fulfilledResource));
+            }
+
+            if (resourceRequirement.createsVolume()) {
```
gabrielhartmann: Should this be in `VolumeEvaluationStage`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
mrbrowning: It's meant to return a list. The use case is this: we have a `ResourceSpecification` per resource type usually, but in the case of ports there's a specification per port, so a naive translation of spec to proto will give you a list of `Resource` that has multiple port-type resources in it. This just returns a list of resources where all the ports have been coalesced into a single `Resource` (slightly misleading, since it actually just clears the ranges, but that's the idea).

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
mrbrowning: Looks like a merge artifact, I'll remove.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
mrbrowning: Yeah, further documentation is probably warranted -- the idea is that evaluation stages have a task affinity, since e.g. a dynamic port may need to modify the task that it belongs to, so that's what `taskName` does. If it's not present, then the evaluation stage instead has executor affinity and modifies the `ExecutorInfo` instead, although that functionality is dormant right now since there's not an `ExecutorSpecification`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
mrbrowning: It could -- I didn't put it there initially because the super class's `evaluate` already consumes the resource, and so it only has one chance to make decisions based on the `MesosResource` that the pool returns, so putting this logic in `VolumeEvaluationStage` would entail duplicating that code, but of course almost none of the code in `ResourceEvaluationStage` applies since it's mostly dealing with bookkeeping for non-atomic resources. I'll move this.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -0,0 +1,152 @@
...
+
+        for (int i = 0; i < offers.size(); ++i) {
+            if (!recommendations.isEmpty()) {
+                break;
+            }
+
+            Offer offer = offers.get(i);
+            MesosResourcePool resourcePool = new MesosResourcePool(offer);
+            OfferRecommendationSlate recommendationSlate = new OfferRecommendationSlate();
+            try {
```
nickbp: Ideally we'd be logging _all_ reasons why a given Offer doesn't work, instead of just the first one. This could accumulate a list of `OfferEvaluationException`s in a list and then log them all at once. I'm not sure if that'd cover all cases though, as e.g. the placement constraint stage is more granular than that (per-resource rather than per-offer).

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/LaunchEvaluationStage.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.mesos.Protos;
+
+import java.util.Optional;
+
+/**
+ * This class sets pod metadata on a {@link org.apache.mesos.Protos.TaskInfo} in an {@link OfferRequirement}, ensuring
+ * that this metadata is available in the task's environment and creating an {@link LaunchOfferRecommendation}.
+ */
+public class LaunchEvaluationStage implements OfferEvaluationStage {
+    private String taskName;
```
nickbp: could be `final`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorEvaluationStage.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.mesos.Protos;
+
+import java.util.Optional;
+
+/**
+ * This class evaluates an offer against a given {@link OfferRequirement}, ensuring that executor IDs match between
+ * the two and setting the executor ID for a newly-launching pod.
+ */
+public class ExecutorEvaluationStage implements OfferEvaluationStage {
+    private Optional<Protos.ExecutorID> executorId;
```
nickbp: could be `final`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/NamedVIPEvaluationStage.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.mesos.Protos;
+
+/**
+ * This class evaluates an offer against a given {@link OfferRequirement} for port resources as in
+ * {@link PortEvaluationStage}, additionally setting {@link org.apache.mesos.Protos.DiscoveryInfo} properly for
+ * DC/OS to pick up the specified named VIP mapping.
+ */
+public class NamedVIPEvaluationStage extends PortEvaluationStage implements OfferEvaluationStage {
+    private String vipName;
+    private Integer vipPort;
```
nickbp: both could be `final`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluationStage.java
```diff
@@ -0,0 +1,27 @@
...
+     * @param offerResourcePool the available resources left available for this stage to consume
+     * @param offerRequirement the tasks and executor to launch, with their associated resource requirements
+     * @param offerRecommendationSlate the current set of all {@link com.mesosphere.sdk.offer.OfferRecommendation}s
+     *                                 for this offer
+     * @throws OfferEvaluationException
+     */
+    void evaluate(
+            MesosResourcePool offerResourcePool,
+            OfferRequirement offerRequirement,
+            OfferRecommendationSlate offerRecommendationSlate) throws OfferEvaluationException;
```
nickbp: Should later stages need to see the contents of `OfferRecommendationSlate`? If not, could this interface be changed to just return an `Optional<OfferRecommendation>` and omit the slate parameter?

The sorting property of `OfferRecommendationSlate` could be implemented in `OfferEvaluator` using `instanceof` checks against the `OfferRecommendation`s returned by the Stages.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PlacementRuleEvaluationStage.java
```diff
@@ -0,0 +1,53 @@
...
+    public PlacementRuleEvaluationStage(Collection<Protos.TaskInfo> deployedTasks) {
+        this.deployedTasks = deployedTasks;
+    }
+
+    @Override
+    public void evaluate(
+            MesosResourcePool offerResourcePool,
+            OfferRequirement offerRequirement,
+            OfferRecommendationSlate offerRecommendationSlate) {
+        if (offerRequirement.getPlacementRuleOptional().isPresent()) {
```
nickbp: Could `return` early if `!isPresent()`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluationStage.java
```diff
@@ -0,0 +1,27 @@
...
+     * variable), deducting used resources from the pool, and creating the appropriate
+     * {@link com.mesosphere.sdk.offer.OfferRecommendation} in the given {@link OfferRecommendationSlate} if relevant.
+     * @param offerResourcePool the available resources left available for this stage to consume
+     * @param offerRequirement the tasks and executor to launch, with their associated resource requirements
+     * @param offerRecommendationSlate the current set of all {@link com.mesosphere.sdk.offer.OfferRecommendation}s
+     *                                 for this offer
+     * @throws OfferEvaluationException
+     */
+    void evaluate(
+            MesosResourcePool offerResourcePool,
```
nickbp: Could make sense to name this `mesosResourcePool` (both here and in the concrete implementations) to align with the matching names of the other params.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PlacementRuleEvaluationStage.java
```diff
@@ -0,0 +1,53 @@
...
+            Protos.Offer offer = offerRequirement.getPlacementRuleOptional().get().filter(
```
nickbp: It'd be a bit clearer what's going on here if this was named `filteredOffer`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PlacementRuleEvaluationStage.java
```diff
@@ -0,0 +1,53 @@
...
+            if (filteredCount == originalCount) {
+                logger.info(
+                        "- Fully passed placement constraint, {} resources remain for evaluation: {}",
+                        filteredCount, offer.getId().getValue());
+            } else if (filteredCount > 0) {
+                logger.info(
+                        "- {}: Partially passed placement constraint, {} of {} resources remain for evaluation: {}",
+                        filteredCount, originalCount, offer.getId().getValue());
+            } else {
+                logger.info(
```
nickbp: Could throw `OfferEvaluationException` in this case

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -0,0 +1,125 @@
...
+ * This class evaluates an offer for its port resources against an {@link OfferRequirement}, finding ports dynamically
+ * in the offer where specified by the framework and modifying {@link org.apache.mesos.Protos.TaskInfo} and
+ * {@link org.apache.mesos.Protos.ExecutorInfo} where appropriate so that the ports are available in their respective
+ * environments.
+ */
+public class PortEvaluationStage extends ResourceEvaluationStage implements OfferEvaluationStage {
+    private String portName;
+    private Integer port;
+
+    public PortEvaluationStage(Protos.Resource resource, String taskName, String portName, Integer port) {
```
nickbp: Should `port` ever be null? If so, then add a null check on assignedPort at L42. If not, then use an `int` instead.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.stream.IntStream;
+
+/**
+ * This class evaluates an offer for its port resources against an {@link OfferRequirement}, finding ports dynamically
+ * in the offer where specified by the framework and modifying {@link org.apache.mesos.Protos.TaskInfo} and
+ * {@link org.apache.mesos.Protos.ExecutorInfo} where appropriate so that the ports are available in their respective
+ * environments.
+ */
+public class PortEvaluationStage extends ResourceEvaluationStage implements OfferEvaluationStage {
+    private String portName;
```
nickbp: These could be final

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PlacementRuleEvaluationStage.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Collection;
+
+/**
+ * This class evaluates an offer against a given {@link OfferRequirement}, ensuring that its resources meet the
+ * constraints imposed by the supplied {@link com.mesosphere.sdk.offer.constrain.PlacementRule}.
+ */
+public class PlacementRuleEvaluationStage implements OfferEvaluationStage {
+    private static final Logger logger = LoggerFactory.getLogger(PlacementRuleEvaluationStage.class);
+
+    private Collection<Protos.TaskInfo> deployedTasks;
```
nickbp: This could be final

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -0,0 +1,125 @@
...
+                offerRequirement.getTaskRequirement(getTaskName().get()).getTaskInfo().getCommand() :
+                offerRequirement.getExecutorRequirementOptional().get().getExecutorInfo().getCommand();
+        String taskPort = CommandUtils.getEnvVar(commandInfo, portName);
+        Integer assignedPort = port;
+        if (assignedPort == 0 && taskPort != null) {
+            assignedPort = Integer.parseInt(taskPort);
+        }
+
+        if (assignedPort == 0) {
+            // We don't want to dynamically consume a port that's explicitly claimed by this pod accidentally, so
```
nickbp: Could break this out into e.g. `private static int selectDynamicPort(MesosResourcePool, OfferRequirement) { ... }`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    protected void setProtos(OfferRequirement offerRequirement, Protos.Resource resource) {
+        long port = resource.getRanges().getRange(0).getBegin();
+
+        if (getTaskName().isPresent()) {
+            String taskName = getTaskName().get();
+            Protos.TaskInfo taskInfo = offerRequirement.getTaskRequirement(taskName).getTaskInfo();
+            Protos.TaskInfo.Builder taskInfoBuilder = taskInfo.toBuilder();
```
nickbp: It looks like we don't need to copy the `TaskInfo` here, as the `ports` being manipulated below is itself a copy coming from the `TaskInfo`. Could move this `toBuilder()` call immediately above `setResource()` below at L90.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -0,0 +1,125 @@
...
+
+            taskInfoBuilder.setCommand(
+                    CommandUtils.addEnvVar(
+                            taskInfoBuilder.getCommand(), getPortEnvironmentVariable(portName), Long.toString(port)));
+            offerRequirement.updateTaskRequirement(taskName, taskInfoBuilder.build());
+        } else {
+            Protos.ExecutorInfo executorInfo = offerRequirement.getExecutorRequirementOptional()
+                    .get()
+                    .getExecutorInfo();
+            Protos.ExecutorInfo.Builder executorInfoBuilder = executorInfo.toBuilder();
```
nickbp: Same as `taskInfoBuilder` above

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -0,0 +1,125 @@
...
+                            Long.toString(port)));
+            offerRequirement.updateExecutorRequirement(executorInfoBuilder.build());
+        }
+    }
+
+    private static String getPortEnvironmentVariable(String portName) {
+        return "PORT_" + portName;
+    }
+
+    private static ResourceRequirement getDynamicPortRequirement(
```
nickbp: This function isn't specific to dynamic ports, appears to be called regardless of dynamic vs fixed?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,182 @@
...
+ * or value of the supplied {@link Resource}, and creating a {@link ReserveOfferRecommendation} or
+ * {@link UnreserveOfferRecommendation} where necessary.
+ */
+public class ResourceEvaluationStage implements OfferEvaluationStage {
+    private static final Logger logger = LoggerFactory.getLogger(ResourceEvaluationStage.class);
+
+    private ResourceRequirement resourceRequirement;
+    // Resources frequently have necessary task affinities -- a dynamic port, for example, will need to modify the
+    // TaskInfo it belongs to. We express that affinity with this field, which if left null means that this evaluation
+    // stage modifies the ExecutorInfo instead.
```
nickbp: This information could be moved to javadocs in the constructors.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,182 @@
...
+    private String taskName;
```
nickbp: `taskName` could be final (looks like `resourceRequirement` couldn't)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,182 @@
...
+        if (!mesosResourceOptional.isPresent()) {
+            throw new OfferEvaluationException(String.format(
+                    "Failed to satisfy resource requirement: %s",
+                    TextFormat.shortDebugString(resourceRequirement.getResource())));
+        }
+
+        final MesosResource mesosResource = mesosResourceOptional.get();
+        Resource fulfilledResource = getFulfilledResource(mesosResource);
+        if (resourceRequirement.expectsResource()) {
+            logger.info("Expects Resource");
```
nickbp: Maybe port the log messages from the current/old OfferEvaluator? They made it easier to tell what was/wasn't passing evaluation. See e.g. old OfferEvaluator.java:L291 onwards.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -0,0 +1,182 @@
...
+                        offerRecommendationSlate.addReserveRecommendation(
+                                new ReserveOfferRecommendation(offerResourcePool.getOffer(), reserveResource));
+                        fulfilledResource = getFulfilledResource(new MesosResource(resourceRequirement.getResource()));
+                    } else {
+                        throw new OfferEvaluationException("Insufficient resources to increase resource usage.");
+                    }
+                }
+            }
+        } else if (resourceRequirement.reservesResource()) {
+            logger.info("Reserves Resource");
```
nickbp: Old OfferEvaluator.java:L352

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/VolumeEvaluationStage.java
```diff
@@ -0,0 +1,109 @@
...
+            throw new OfferEvaluationException(String.format(
+                    "Failed to satisfy resource requirement: %s",
+                    TextFormat.shortDebugString(resourceRequirement.getResource())));
+        }
+
+        final MesosResource mesosResource = mesosResourceOptional.get();
+        Resource fulfilledResource = getFulfilledResource(mesosResource);
+
+        if (resourceRequirement.reservesResource()) {
+            logger.info("Reserves Resource");
```
nickbp: Old OfferEvaluator.java:L352

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/VolumeEvaluationStage.java
```diff
@@ -0,0 +1,109 @@
...
+            offerRecommendationSlate.addReserveRecommendation(
+                    new ReserveOfferRecommendation(offerResourcePool.getOffer(), fulfilledResource));
+        }
+
+        if (resourceRequirement.createsVolume()) {
+            logger.info("Creates Volume");
```
nickbp: Old OfferEvaluator.java:L357

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/NamedVIPSpecification.java
```diff
@@ -0,0 +1,82 @@
...
+            @JsonProperty("principal") String principal,
+            @JsonProperty("env_key") String envKey) {
+        super(name, value, role, principal, envKey);
+        this.portName = portName;
+        this.vipName = vipName;
+        this.vipPort = vipPort;
+
+        ValidationUtils.validate(this);
+    }
+
```
nickbp: `portName` getter? or alternately this class could `extend PortSpecification`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/ResourceSpecification.java
```diff
@@ -10,7 +12,7 @@
  * A ResourceSpecification encapsulates a Mesos Resource that may be used by a Task and therefore specified in a
  * TaskSpecification.
  */
-@JsonDeserialize(as = DefaultResourceSpecification.class)
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME)
```
nickbp: This may be an incompatible change if this data is stored in zk. Not sure though.

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/ExecutorEvaluationStageTest.java
```diff
@@ -0,0 +1,64 @@
...
+                        TestConstants.EXECUTOR_ID, Arrays.asList(expectedExecutorMem, expectedTaskCpu)));
+
+        ExecutorEvaluationStage executorEvaluationStage = new ExecutorEvaluationStage(execInfo.getExecutorId());
+        executorEvaluationStage.evaluate(resources, offerRequirement, new OfferRecommendationSlate());
+
+        Assert.assertEquals(
+                execInfo.getExecutorId(),
+                offerRequirement.getExecutorRequirementOptional().get().getExecutorInfo().getExecutorId());
+    }
+}
```
nickbp: missing newline

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/ExecutorEvaluationStageTest.java
```diff
@@ -0,0 +1,64 @@
+package com.mesosphere.sdk.offer.evaluate;
```
nickbp: Missing tests for other `Stage`s

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
```diff
@@ -18,15 +18,19 @@
...
+    private Map<String, MesosResource> reservedPool;
 
     /**
      * Creates a new pool of resources based on what's available in the provided {@link Offer}.
      */
     public MesosResourcePool(Offer offer) {
+        init(offer);
+    }
+
+    private void init(Offer offer) {
```
mrbrowning: Not easily -- the pipeline keeps the same `MesosResourcePool` and passes it to each stage in turn, so this change would get lost to later stages, since the evaluation stage in question would just overwrite its local reference to that pool with a new reference, rather than the original object being mutated.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -204,139 +204,61 @@ public static Resource getExpectedRanges(
...
-    /**
-     * Populates {@link DiscoveryInfo} with necessary information to create VIP if the provided {@link TaskInfo}
-     * has a port associated with a named VIP.
-     */
-    public static TaskInfo.Builder setDiscoveryInfo(TaskInfo.Builder taskBuilder, Collection<Resource> resources) {
-        for (Resource r : resources) {
-            if (RequirementUtils.isNamedVIPPort(r)) {
-                taskBuilder.setDiscovery(getVIPDiscoveryInfo(r, taskBuilder.getName()));
-            }
+    public static ExecutorInfo.Builder addVIP(
```
mrbrowning: Potentially -- it's moot currently, since executors will never have resources attached to them (since they don't have an associated specification type), but presumably this will change in the future. Not sure if it makes sense to defer adding that to `EndpointsResource` until that happens.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
gabrielhartmann: @nickbp:  I generally think the approach you outline is a good pattern, but I think mutating the ResourcePool is fine as it simplifies the pipeline and that is the intent of this object (that it be mutated).

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorEvaluationStage.java
gabrielhartmann: Let's not make fields Optional going forward, but instead have the getters return Optionals.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/OfferRequirement.java
```diff
@@ -59,42 +62,39 @@ public static OfferRequirement create(
...
-     * Creates a new {@link OfferRequirement} with provided executor requirement and empty placement
-     * constraints.
+     * Creates a new {@link OfferRequirement} with provided executor requirement and empty placement constraints.
      *
      * @see #OfferRequirement(String, Integer, Collection, Optional, Optional)
      */
     public static OfferRequirement create(
             String taskType,
-            int index,
+            Integer index,
```
mrbrowning: That's from master, it looks like? Not sure what's going on here. I can change it to `int`, though.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
nickbp: Makes sense, if it's not actually being used there isn't much reason to add effectively-dead logic to deal with it before we've found out whether it makes to have in the first place. Maybe just add a TODO to EndpointsResource's current TaskInfo extraction for later reference?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/NamedVIPEvaluationStage.java
```diff
@@ -0,0 +1,61 @@
...
+    public NamedVIPEvaluationStage(
+            Protos.Resource resource, String portName, Integer port, String vipName, Integer vipPort) {
+        this(resource, null, portName, port, vipName, vipPort);
+    }
+
+    @Override
+    protected void setProtos(OfferRequirement offerRequirement, Protos.Resource resource) {
+        super.setProtos(offerRequirement, resource);
+
+        if (getTaskName().isPresent() &&
```
nickbp: Could add comments describing the difference between these cases

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -0,0 +1,162 @@
...
+                            resourcePool,
+                            offerRequirement,
+                            recommendationSlate);
+                }
+            } catch (OfferEvaluationException e) {
+                failureNotifications.add(e.getMessage());
+            }
+
+            if (!failureNotifications.isEmpty()) {
+                logger.info("- {}: did not pass resource requirements for the following reasons:", i + 1);
```
nickbp: super nits:
- could update this and the below log entries to say `evaluation stages` instead of `resource requirements`
- could include `evaluationStages.size()` in the message, e.g. '... did not pass one or more of {} evaluation stages ...`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluationStage.java
```diff
@@ -0,0 +1,27 @@
...
+     * {@link OfferRequirement} where appropriate to ensure they have necessary metadata (e.g. a populated environment
+     * variable), deducting used resources from the pool, and creating the appropriate
+     * {@link com.mesosphere.sdk.offer.OfferRecommendation} in the given {@link OfferRecommendationSlate} if relevant.
+     * @param mesosResourcePool the available resources left available for this stage to consume
+     * @param offerRequirement the tasks and executor to launch, with their associated resource requirements
+     * @param offerRecommendationSlate the current set of all {@link com.mesosphere.sdk.offer.OfferRecommendation}s
+     *                                 for this offer
+     * @throws OfferEvaluationException
+     */
+    void evaluate(
```
nickbp: Feels like this could return a success/error object instead of throwing -- failure to pass an offer isn't an exceptional case. This may also make it slightly clearer in the Stage implementations when a stage succeeds (right now it's just exiting the function, while this change would imply explicitly returning a success object)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -0,0 +1,162 @@
...
+                break;
+            }
+
+            Offer offer = offers.get(i);
+            MesosResourcePool resourcePool = new MesosResourcePool(offer);
+            OfferRecommendationSlate recommendationSlate = new OfferRecommendationSlate();
+            List<String> failureNotifications = new ArrayList<>();
+            try {
+                for (OfferEvaluationStage evaluationStage : evaluationStages) {
+                    evaluationStage.evaluate(
```
nickbp: Shouldn't short-circuit when one stage fails. Could also collect the following information for each stage:
- A name for the stage (class name? or a new `OfferEvaluationStage.getName()`?)
- Whether the stage passed or failed (could then include what passed as well as what failed in the failure log below)

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/NamedVIPSpecification.java
nickbp: (looks like this was resolved)

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/ExecutorEvaluationStageTest.java
nickbp: (resolved)

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/ExecutorEvaluationStageTest.java
nickbp: ping

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/OfferEvaluatorTest.java
```diff
@@ -1,14 +1,14 @@
 package com.mesosphere.sdk.offer;
```
nickbp: File should be moved to `evaluate` package

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
mrbrowning: I think I'll omit any explicit handling of the passing case, since each stage logs what passes itself, and it would complicate things unnecessarily (you'd have to pass the success notification string out in addition to whatever else, I don't think we should reify logging concerns to the point that they become part of the actual interface), but will add the names and fix the short-circuiting.

mrbrowning: Nevermind, saw below comment, which I agree with. Since I plan to alter that interface significantly in an upcoming refactor (with `evaluate` returning at least an `Optional<OfferRecommendation>`) I think it would make sense to defer this until that happens, though.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/NamedVIPEvaluationStage.java
```diff
@@ -29,14 +29,17 @@ public NamedVIPEvaluationStage(
...
         if (getTaskName().isPresent() &&
                 !isVIPSet(offerRequirement.getTaskRequirement(getTaskName().get()).getTaskInfo().getDiscovery())) {
+            // Set the VIP on the TaskInfo.
             Protos.TaskInfo.Builder taskInfoBuilder = offerRequirement
                     .getTaskRequirement(getTaskName().get()).getTaskInfo().toBuilder();
 
             ResourceUtils.addVIP(taskInfoBuilder, vipName, vipPort, resource);
             offerRequirement.updateTaskRequirement(getTaskName().get(), taskInfoBuilder.build());
         } else if (offerRequirement.getExecutorRequirementOptional().isPresent() &&
+                // Set the VIP on the ExecutorInfo.
```
nickbp: looks like this comment is one line too early

PR #415, opened 2017-01-04
Added data integrity check to upgrade test

susanxhuynh:
Added data integrity check to upgrade test: write data, do upgrade, read data. Minor refactoring and code style changes.

### DIFF
--- a/frameworks/hdfs/integration/tests/test_upgrade.py
+++ b/frameworks/hdfs/integration/tests/test_upgrade.py
@@ -32,7 +32,9 @@ def setup_module(module):
     uninstall()
 
 
+@pytest.mark.skip(reason="Waiting for released version from which to upgrade")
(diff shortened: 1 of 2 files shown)


frameworks/hdfs/integration/tests/test_upgrade.py
```diff
@@ -26,32 +34,36 @@ def setup_module(module):
 
 @pytest.mark.upgrade
```
nickbp: Should this be marked `sanity`+`skip` like hello-world's version?: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/helloworld/integration/tests/test_upgrade.py#L31

susanxhuynh: Yes, thanks. I will mark it with `sanity` + `skip`.

PR #416, opened 2017-01-04
Switch to Builder-based construction of DefaultScheduler

nickbp:
Results in a greatly simplified bootstrapping flow in `DefaultService` and `Main`s.

Also:
- Fixes customization of role/principal in `DefaultService` (was previously ignoring `ServiceSpec`)
- Exposes ability to define custom endpoints, including a couple stubs for HDFS xml file endpoints.
- Fixes potential for an NPE if MinReplaceDelayMs is unset in policy (now treated as 'zero min delay')
- Renames `ConfigurationValidator` => `ConfigValidator` and `ConfigurationValidationError` => `ConfigValidationError` to align with usage of `Config` elsewhere.
- Resumes printing the full old config when configs change (in addition to the newer diff)
- Retries ZK lock up to 3 times, to reduce superfluous errors when the scheduler is recently restarted (seen often during reconfiguration)


nickbp:
retest this please

loren:
+1
This builder-based approach will help me DRY up my existing Elastic service implementation around specifying custom config validators.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -207,21 +156,35 @@ public void run() {
...
                 .setCheckpoint(true);
 
+        // Use provided role if specified, otherwise default to "<svcname>-role".
+        if (StringUtils.isEmpty(serviceSpec.getRole())) {
+            fwkInfoBuilder.setRole(SchedulerUtils.nameToRole(serviceName));
+        } else {
+            fwkInfoBuilder.setRole(serviceSpec.getRole());
+        }
+
+        // Use provided role if specified, otherwise default to "<svcname>-principal".
```
gabrielhartmann: Use provided principal...

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -30,9 +30,11 @@
     private final StateStore stateStore;
     private final UUID targetConfigurationId;
 
-    public DefaultOfferRequirementProvider(
-            StateStore stateStore,
-            UUID targetConfigurationId) {
+    /**
+     * Creates a new instance which relies on the provided {@link StateStore} for storing known tasks, and which expects
```
mrbrowning: Not clear what "expects" means here? It will tag tasks with that configuration ID, but it doesn't really _expect_ it as far as I can tell.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -70,116 +53,82 @@ public DefaultService(File pathToYamlSpecification) throws Exception {
...
-        this.serviceSpec = serviceSpec;
-        init();
-        this.plans = plans;
-        register(serviceSpec, this.plans);
+    public DefaultService(DefaultScheduler.Builder schedulerBuilder) throws Exception {
+        this.schedulerBuilder = schedulerBuilder;
+
+        CuratorFramework curatorClient = CuratorFrameworkFactory.newClient(
+                schedulerBuilder.getServiceSpec().getZookeeperConnection(), CuratorUtils.getDefaultRetry());
+        curatorClient.start();
```
mrbrowning: Ok for the reference to `curatorClient` to become unreachable after we leave this function scope, since we'll never be able to close it? I guess there's no issue with sockets being held open here, since a crash will take care of that anyway.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -70,116 +53,82 @@ public DefaultService(File pathToYamlSpecification) throws Exception {
...
-                offerRequirementProvider);
-
-        startApiServer(defaultScheduler, apiPort);
-        registerFramework(defaultScheduler, getFrameworkInfo(), "zk://" + zkConnectionString + "/mesos");
-    }
-
-    public ServiceSpec getServiceSpec() {
-        return serviceSpec;
-    }
+    public void register() throws Exception {
```
mrbrowning: Who throws? Could we narrow the exception type?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -70,116 +53,82 @@ public DefaultService(File pathToYamlSpecification) throws Exception {
...
+        DefaultScheduler defaultScheduler = schedulerBuilder.build();
+        ServiceSpec serviceSpec = schedulerBuilder.getServiceSpec();
```
mrbrowning: Is there a place where the `schedulerBuilder` member is frequently modified after this class is instantiated? Why not build during the init and have that member's type be `DefaultScheduler`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
nickbp: Rephased to '...updates tasks which are not tagged with...'

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
nickbp: The builder.build() call does some extra work internally, specifically creating the statestore/configstore (if not manually provided by the developer) and then validating/updating the target config. Figured it'd make sense to keep that work out of the constructor as it's effectively network access.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
nickbp: This was inheriting the `Service` interface's `throws` but doesn't need to. Dropped `throws` entirely.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
nickbp: Added a close() call.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
nickbp: Fixed

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
mrbrowning: Hm, that's a good point. In something like Python I'd use a memoizing `@property` descriptor or something like that, but that would probably be a lot of boilerplate here to little effect. Seems reasonable as-is.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -70,116 +53,83 @@ public DefaultService(File pathToYamlSpecification) throws Exception {
...
         LOGGER.info("Acquiring ZK lock on {}...", lockPath);
+        final String failureLogMsg = String.format("Failed to acquire ZK lock on %s. " +
+                "Duplicate service named '%s', or recently restarted instance of '%s'?",
+                lockPath, serviceName, serviceName);
         try {
-            if (!curatorMutex.acquire(10, TimeUnit.SECONDS)) {
-                LOGGER.error("Failed to acquire ZK lock. Are you running another framework named {}? Exiting.",
-                        serviceSpec.getName());
-                SchedulerUtils.hardExit(SchedulerErrorCode.LOCK_UNAVAILABLE);
+            for (int i = 0; i < LOCK_ATTEMPTS; ++i) {
```
mgummelt: What's the point in trying 3 times?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -70,116 +53,83 @@ public DefaultService(File pathToYamlSpecification) throws Exception {
...
+                    return curatorMutex;
+                }
+                LOGGER.error("{}/{} {} Retrying lock...", i + 1, LOCK_ATTEMPTS, failureLogMsg);
             }
+            LOGGER.error(failureLogMsg + " Restarting scheduler process to try again.");
+            SchedulerUtils.hardExit(SchedulerErrorCode.LOCK_UNAVAILABLE);
         } catch (Exception ex) {
-            LOGGER.error("Error acquiring ZK lock.", ex);
+            LOGGER.error(String.format("Error acquiring ZK lock on path: %s", lockPath), ex);
+            SchedulerUtils.hardExit(SchedulerErrorCode.LOCK_UNAVAILABLE);
```
mgummelt: This error code is incorrect.  If an exception was thrown, that doesn't mean the lock was unavailable.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
nickbp: After a scheduler kill (e.g. restart for config update) the old lock apparently isn't cleared right away and is still held on zk's end until some timeout. Without the retry I was seeing pretty consistent scheduler task failures after reconfiguration because the (first) lock attempt would fail.

mgummelt: This is likely because the ephemeral locks are held until the TCP connection expires, which might take a few seconds.  Regardless, it makes more sense to just increase the timeout to 3*N rather than looping 3 times over a timeout of N.

nickbp: It's been a while, but I vaguely remember the lock still failing with a timeout increase. Could be misremembering though.

PR #417, opened 2017-01-05
Move example yamls into an examples/ directory, exercise TASKCFG_x/POD_INSTANCE envvars

nickbp:


nickbp:
retest this please

frameworks/helloworld/src/main/dist/svc.yml
```diff
@@ -9,11 +9,11 @@ pods:
     tasks:
       server:
         goal: RUNNING
-        cmd: "echo hello >> hello-container-path/output && sleep $SLEEP_DURATION"
+        cmd: "echo hello-$POD_INSTANCE_INDEX >> $OUTPUT_VOLUME/$OUTPUT_FILENAME && sleep $SLEEP_DURATION"
```
gabrielhartmann: The changes here in svc.yml are illustrative, but I think a little complex for the default hello-world example.  Do you have objections to moving this to one of the example yamls, or a new example yaml (with an integration test that actually exercises code)?

nickbp: I think it'd make sense to just put this in a separate yml that's exercised by an integration test. The `TASKCFG_*` stuff is only really useful in specific cases, and for newcomers it'd likely be confusing to have both env methods in the same place.

nickbp:
Done

nickbp:
retest this please

PR #418, opened 2017-01-05
Use proper mocks to remove network dependency from unit tests

nickbp:
Avoid requiring a specific port to be available.


sdk/scheduler/src/test/java/com/mesosphere/sdk/dcos/CapabilitiesTest.java
```diff
@@ -1,148 +1,88 @@
...
-        Capabilities capabilities = new Capabilities(mockDcosCluster.getDcosCluster());
-        Assert.assertTrue(capabilities.supportsRLimits());
-    }
-
-    @Test
-    public void testHasRLimitsSucceeds_2_0dev() throws URISyntaxException, IOException {
-        mockDcosCluster =  MockDcosCluster.create("2.0-dev");
-        Capabilities capabilities = new Capabilities(mockDcosCluster.getDcosCluster());
-        Assert.assertTrue(capabilities.supportsRLimits());
+    private void testClusterVersion(String version, boolean expectNamedVips, boolean expectRLimits) throws IOException {
```
mrbrowning: I think this may be a case where keeping things DRY will make adding test cases more work than necessary down the line -- maybe keep the mock setup in this function, but perform the asserts on a per-test basis? That way, if there's some new capability we need to add tests for in the future, we don't have to modify this function signature as well as each existing test in this module. I think some redundancy in tests makes them more readable, anyway -- I want to be able to read a test and see plainly what behavior is being asserted as much as possible, instead of having to chase function invocations around.

nickbp: Good point, switched to asserts in each case. Done.

PR #423, opened 2017-01-06
Refactor plan elements, Strategies, and interrupt handling

nickbp:
Solves the situation with `Step`s confusingly having `getChildren()`/`getStrategy()` calls despite not having any children.
`Step`s now directly implement `Element`, while `Phase`s/`Plan`s now implement `ParentElement<C>`, which has the prior `getChildren()`/`getStrategy()` calls.

`CanaryStrategy` needed to be reimplemented to work without `Step.getStrategy()`. In the process of doing this, it was updated to support arbitrary other strategies after the canary stage has completed. For example a canary stage may now be followed by either serial or parallel deployment. Additional tests are included for this extra functionality.


nickbp:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultStep.java
```diff
@@ -143,16 +140,6 @@ public Status getStatus() {
         }
     }
 
-    @Override
-    public List<Element<?>> getChildren() {
-        return Collections.emptyList();
-    }
-
-    @Override
```
mbdcos: how do you interrupt a step?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/CanaryStrategy.java
```diff
@@ -1,58 +1,196 @@
...
+     */
+    private static final int DEFAULT_PROCEED_COUNT = 2;
+
     private final Logger logger = LoggerFactory.getLogger(getClass());
-    private AtomicBoolean initialized = new AtomicBoolean(false);
-    private List<Element> children;
+    private final int requiredProceeds;
+    private final Strategy<Step> postCanaryStrategy;
+
+    private List<InterruptableStep> canaryProceedSteps;
```
mbdcos: himm are we making InterruptableStep available only for CanaryStrategy?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/CanaryStrategy.java
```diff
@@ -1,58 +1,196 @@
...
+
+        /**
+         * Creates a new generator which will require a specified number of {@link #proceed()} calls from a user before
+         * following the provided {@code postCanaryStrategy}.
+         *
+         * @param postCanaryStrategy the strategy to use after the canary stage has completed
+         * @param requiredProceeds the number of {@link #proceed()} calls to require before the provided strategy is
+         *     executed
+         */
+        public Generator(Strategy<Step> postCanaryStrategy, int requiredProceeds) {
```
mbdcos: great

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultStep.java
nickbp: Steps cannot be interrupted in the middle of performing work. Interrupt takes effect at the phase level, after the current step has completed. This has always effectively been the case but with this change it's more explicit.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/CanaryStrategy.java
nickbp: Yes it's just bookeeping of the steps that need to be interrupted for the canary itself before falling back to the provided strategy. Interrupt isn't quite accurate as these steps just haven't been started yet when interrupted=true.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultStep.java
mbdcos: I am not 100% sure whether it is a good idea to get rid of interrupts from steps. I think every element should be interruptible. Interrupt is similar to status. getStatus()/update() ~ pause()/proceed().

mbdcos:
I am not 100% sure whether it is a good idea to get rid of interrupts from steps. I think every element should be interruptible. Interrupt is similar to status. getStatus()/update() ~ pause()/proceed(). 
Other than that it looks good to me.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultStep.java
nickbp: Interrupting a step implies halting the thread of execution within that step, which we haven't ever supported. In practice when you interrupt a plan it results in the current step finishing its operation and then the following step sitting in an interrupted (`WAITING`?) state

mbdcos: I kind of think it as  pause/proceed loop. I am fine with either we keep or dont keep interrupts in steps.

One benefit of it in canary phase: phase is in waiting state, and first two steps are also waiting state, rest of the steps are pending. When you say continue, only one of those steps will change to pending then in_progress, but second one will be waiting. So you know that you need to say continue one more time (0.8.1.2 DefaultStep returns waiting if interrupted and pending).

gabrielhartmann: @nickbp:  Steps should be interruptable.  If a step is failing to succeed it will constantly retry.  In this case and interrupted step would stop retrying.  Consider the case where many steps are being deployed in parallel but one or two are having trouble and the operator would like to stop them, she might want to interrupt individual steps.

gabrielhartmann:
I agree with @mbdcos.  All steps should be interruptible, but otherwise fine.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultStep.java
nickbp: It doesn't look like that's the case though? `DependencyStrategyHelper` currently filters out steps which have entered an error state when selecting candidates: https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/DependencyStrategyHelper.java#L62

Similarly we don't offer a way to interrupt at step granularity, only at the plan level: https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PlansResource.java#L91

gabrielhartmann: PlansResource isn't correct anymore.  As in the [restart example](https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PlansResource.java#L159-L162), so should things be interruptable and continuable.

You're comment regarding the error state is correct.  However it is common that a Step not enter the Error state, but continuously loop through Pending, Prepared, Starting, Pending, Prepared, Starting ...  In all cases an end user needs the ability to stop the processing of a Step regardless of its status.

nickbp: Aha, makes sense. OK I'll add an interrupt bit into `Step`.

nickbp:
Ready for review. Steps are interruptable once more. Ended up doing a lot of refactoring of interrupt handling and steps in the process.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/PlanUtils.java
```diff
@@ -30,9 +30,9 @@ private PlanUtils() {
     }
 
     public static final Collection<? extends Step> getCandidates(Plan plan, Collection<String> dirtyAssets) {
-        Collection<Phase> candidatePhases = plan.getStrategy().getCandidates(plan, dirtyAssets);
+        Collection<Phase> candidatePhases = plan.getStrategy().getCandidates(plan.getChildren(), dirtyAssets);
```
mbdcos: still getCandidates is first checking whether parent is interrupted or not, right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/AbstractStep.java
```diff
@@ -0,0 +1,91 @@
...
+        synchronized (statusLock) {
+            if (interrupted && status == Status.PENDING) {
+                return Status.WAITING;
+            }
+            return status;
+        }
+    }
+
+    @Override
+    public void interrupt() {
```
mbdcos: you overwrite interrupt and proceed, ok

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/DependencyStrategy.java
```diff
@@ -19,7 +18,8 @@ public DependencyStrategy(DependencyStrategyHelper<C> helper) {
     }
 
     @Override
-    public Collection<C> getCandidates(Element<C> parentElement, Collection<String> dirtyAssets) {
-        return helper.getCandidates(dirtyAssets);
+    public Collection<C> getCandidates(Collection<C> elements, Collection<String> dirtyAssets) {
+        // Fixed prerequites as defined in the provided helper:
+        return helper.getCandidates(isInterrupted(), dirtyAssets);
```
mbdcos: ok

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/PlanUtils.java
nickbp: Yes, via `DependencyStrategyHelper`. Added an interrupted parameter that the `Strategy`s have to pass.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/CanaryStrategy.java
```diff
@@ -1,60 +1,165 @@
...
+        if (getNextCanaryStep() != null) {
+            // Still in canary. Only return subset of canary steps which are now eligible due to proceed() calls.
+            return canarySteps.stream()
+                    .filter(step -> PlanUtils.isEligibleCandidate(step, dirtyAssets))
+                    .collect(Collectors.toList());
         }
 
-        return super.getCandidates(parentElement, dirtyAssets);
+        // Canary period has ended. Revert to underlying strategy for the rest of the operation.
+        return strategy.getCandidates(steps, dirtyAssets);
```
mbdcos: what if strategy is parallel?

nickbp: That's allowed and tested: https://github.com/mesosphere/dcos-commons/pull/423/files#diff-464e0ec863aaf1ece38b21da2237b5f5R133

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/CanaryStrategy.java
```diff
@@ -1,60 +1,165 @@
...
+        Step canaryStep = getNextCanaryStep();
+        if (canaryStep != null) {
+            canaryStep.proceed();
             return;
         }
+        strategy.proceed();
+    }
+
+    @Override
+    public boolean isInterrupted() {
```
mbdcos: looks good to me

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/CanaryStrategy.java
```diff
@@ -1,60 +1,165 @@
...
+    }
+
+    @Override
+    public void interrupt() {
+        Step canaryStep = getNextCanaryStep();
+        if (canaryStep != null) {
+            // Ignoring interrupt as we are still in canary and therefore already interrupted.
+            return;
+        }
+        strategy.interrupt();
```
mbdcos: If I interrupt the CanaryStrategy, before it calls getCandidates, it will interrupt its strategy.  Also, if first (at most two) steps are already proceeded, it will interrupt its strategy.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/strategy/CanaryStrategy.java
```diff
@@ -1,60 +1,165 @@
...
     public void proceed() {
-        super.proceed();
-        
-        if (!initialized.get())  {
-            logger.warn("Proceed has no effect to children before strategy initialization.");
+        Step canaryStep = getNextCanaryStep();
+        if (canaryStep != null) {
+            canaryStep.proceed();
             return;
         }
```
mbdcos: same for proceed.  all is well

PR #424, opened 2017-01-06
Move all YAML keys to kebab-case

gabrielhartmann:


nickbp:
In dcos-commons:kebab:

```
dcos-commons$ grep -R JsonProperty | grep _ | grep src/ | awk '{print $1}' | uniq
sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultResourceSpecification.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/PortSpecification.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/NamedVIPSpecification.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultHealthCheckSpec.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/ReplacementFailurePolicy.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpecification.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/constrain/RoundRobinByAttributeRule.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/constrain/AgentRule.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/constrain/MaxPerAttributeRule.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/constrain/MaxPerHostnameRule.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/constrain/RoundRobinByHostnameRule.java:
sdk/scheduler/src/main/java/com/mesosphere/sdk/config/RecoveryConfiguration.java:
```

Despite Cassandra now using placement constraints, it's currently only storing the marathon string (not the placement constraints themselves) so those should be safe to change for now as well. (Ref: `CassandraSchedulerConfiguration.java`)

keithchambers:
I see "specification" in some places and "spec" in other places. Which should we use? I'm OK with abbreviating or not abbreviating ... but we should form an opinion and then drive consistency across the API.

gabrielhartmann:
@nickbp @keithchambers @mrbrowning:  Should be completely kebabed and speced.  Will be surprised if upgrade tests pass...

mohitsoni:
Are these changes upgrade compliant ?

nickbp:
Didn't see any missing items...

gabrielhartmann:
@mohitsoni:  Almost certainly not.  Hence my comment above.  We'll have to decide if we want to go through the upgrade or force it.

mohitsoni:
@gabrielhartmann 👍

gabrielhartmann:
@loren:  You may want to be aware of this PR.  Changes serialization/deserialization for a bunch of stuff.  Only matters for upgrade, but since nothing other than hello-world has been released, we should be cool.

triclambert:
This is new and experimental.  No one besides @loren should be using this atm, so I don't think we need to worry about upgrades beyond our tests.  @susanxhuynh we may want a way to blacklist certain version jumps.

PR #425, opened 2017-01-06
remove create_service_account.sh call from launch_ccm_cluster.py

mgummelt:
cc @nickbp 

create_service_account.sh requires a core.dcos_url, which isn't set at the time it's called in this script


nickbp:
Any test jobs that are gonna work in strict mode will need to run this manually before starting tests

mgummelt:
Yea, but those were already broken (I've just updated the "Framework Test" project to explicitly create service accounts, which was broken before due to this bug).

mgummelt:
retest this please

PR #426, opened 2017-01-06
Make zookeeper connection string optional

mohitsoni:
This PR makes configuration of zookeeper connection string optional for both YAML and Java interfaces. By default, the value of zookeeper connection string is "master.mesos:2181". This value can be over-ridden in the YAML interface via:

```yaml
...
zookeeper: custom.master.mesos:2181
...
```

or in Java interface via:

```java
new DefaultService(DefaultServiceSpec.newBuilder()
                    .zookeeperConnection("custom.master.mesos:2181")
```

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
@@ -66,7 +66,9 @@ public class DefaultServiceSpec implements ServiceSpec {
         this.role = role;
         this.principal = principal;
         this.apiPort = apiPort;
-        this.zookeeperConnection = zookeeperConnection;
+        // If no zookeeperConnection string is configured, fallback to the default value.
+        this.zookeeperConnection = StringUtils.isBlank(zookeeperConnection)
+                ? DEFAULT_ZK_CONNECTION : zookeeperConnection;
         this.pods = pods;
         this.replacementFailurePolicy = replacementFailurePolicy;
     }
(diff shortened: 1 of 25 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -116,7 +119,8 @@ public int getApiPort() {
 
     @Override
     public String getZookeeperConnection() {
-        return zookeeperConnection;
+        // If no zookeeperConnection string is configured, fallback to the default value.
+        return StringUtils.isNotBlank(zookeeperConnection) ? zookeeperConnection : DEFAULT_ZK_CONNECTION;
```
nickbp: Maybe just use isBlank() and switch the cases? Could see it being slightly easier to parse

sdk/scheduler/src/test/resources/valid-minimal.yml
```diff
@@ -1,6 +1,5 @@
 name: "hello-world"
 principal: "hello-world-principal"
-zookeeper: master.mesos:2181
```
nickbp: Maybe remove default cases from the other .yml's as well

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
mohitsoni: Sure

sdk/scheduler/src/test/resources/valid-minimal.yml
mohitsoni: Sure

mohitsoni: Done

PR #429, opened 2017-01-09
return Status.WAITING when step is PENDING but interrupted 

mbdcos:


mbdcos:
retest this please

mbdcos:
retest this please

mbdcos:
retest this please

mbdcos:
retest this please

mbdcos:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultStep.java
```diff
@@ -139,6 +139,9 @@ public String getName() {
     @Override
     public Status getStatus() {
         synchronized (statusLock) {
+            if (status == Status.PENDING && getStrategy().isInterrupted()) {
+                return Status.WAITING;
```
mrbrowning: Should `status` be set to `WAITING` by whatever interrupted the plan? This is the only case where we don't return the "true" status. Or is it not possible to access the step from whatever that code path is?

mrbrowning:
One comment. Could you update the PR title to match the general format for this repo? E.g. "Return Status.WAITING when plan is interrupted"

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultStep.java
mbdcos: interrupt (pause) and status are independent. Only step has status variable and setStatus method.  Step is the last one in PlanUtil.getStatus lookup chain.  This PR will make output (plan show) show that Step is waiting. Other than that it should have no effect.

mrbrowning: Well, but status does depend on whether or not the phase interrupted, thus this change. I'm just wondering if there's a point at which we can actively set `status` to `WAITING`, since I could see another method in the future accessing the member directly, bypassing this logic and getting the wrong answer.

mbdcos: WAITING is a return value only. If you set status to WAITING, then you need to store previous value to revert back.  We have two variables(one is "status", and the other one is "interrupted"). "interrupted" is in strategy, so in all plan, step, phase.  "status" is in DefaultStep only. There is no setStatus in Element interface.

mrbrowning: Alright, sounds good.

PR #433, opened 2017-01-09
Added a script to save build properties

susanxhuynh:
A script to save the <version>.properties build file in S3.


nickbp:
Assuming this is being run in a jenkins job, could the jenkins job invoke the `aws` call directly?

PR #436, opened 2017-01-09
Let FW developer easily specify FrameworkInfo

loren:


nickbp:
add to whitelist

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -194,4 +190,12 @@ private static void registerAndRunFramework(
 
         return fwkInfoBuilder.build();
     }
+
+    protected static Protos.FrameworkInfo.Builder getFrameworkInfoBuilder(String serviceName) {
```
nickbp: - Should be non-`static` to allow overriding in a subclass?
- Probably should add a comment that this being exposed so that it may be overridden in subclasses.

nickbp:
retest this please

nickbp:
retest this please

PR #440, opened 2017-01-10
YAML Polish

nickbp:
- Move principal/role/zk/api-port under a new 'scheduler' section, and make them all optional with reasonable defaults
- Add optional `web-url` to root allowing the service to specify a user-facing UI location.
- Turn `ports` into a dict keyed on the name of the ports.
- Add optional `protocol` and `visible` fields to `vip`.
- Turn `volumes` from a list into a dict against arbitrary names
- Add an alternate single `volume` to allow less verbosity in the common single-volume case.
- Turn `configurations` templates into a dict against arbitrary config names (these may soon be used for http template serving).
- Add an alternate single `configuration` to allow less verbosity in the common single-config case.
- Turn custom `steps` into dicts keyed against pod instance IDs.
- Update port name in `PORT_` envvar to avoid surprise envvar failures due to invalid chars. For example `some.sample-name` will be result in an envvar of `PORT_SOME_SAMPLE_NAME`.

To see how all these changes look in practice, see updates to the example yml files.


loren:
With the web-url option, you could probably revert https://github.com/mesosphere/dcos-commons/commit/3095d937c3d4c3f58379624d2d235dc5407fc55e if you wanted to. I wouldn't need it anymore.

triclambert:
Can one specify more than one `web-url`?  We will need this.

nickbp:
Mesos only allows at most 1: https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L277

triclambert:
We're going to need to figure out a story for supporting more than one.  I
suppose this is fine, for now.  /cc @mrbrowning
ᐧ

mrbrowning:
I'm still pretty opposed to `endpoints` -- it feels like it's maximally intelligible for someone looking to get their Django app running on DC/OS on a lark over the weekend, kind of confusing for anyone else, and that the "anyone else" is kind of our target audience here. I'll defer to public opinion, but could we weigh in on it one more time?

nickbp:
This only came up relatively recently. Could also go with instead renaming the child `port`:
```
ports:
  http:
    value: 80
```

nickbp:
Was trying out the above change and found that it ended up looking weird when a `vip` is defined:

```
ports:
  http:
    value: 8080
      vip:
        port: 80
```

Sticking with `endpoints` for the moment as I'm out of other ideas:

```
endpoints:
  http:
    port: 8080
      vip:
        port: 80
```

mrbrowning:
Could maybe modify vip like so:
```
vip:
  address: myvip:8080
  advertise: true
```
What do you think?

nickbp:
Wouldn't the `8080` in that `address` also be specified in the parent port?

nickbp:
retest this please

mrbrowning:
No, the parent port and the VIP port can differ; e.g. `broker.kafka.l4lb.thisdcos.directory:8080` that would result from a VIP address of `broker:8080` could map to `broker-${INDEX}.kafka.mesos:9092`.

nickbp:
Oh I see, ok. Looks like the VIP only has `prefix` and `port` fields so we could shorten it to something like this, how's this look?:

```
ports:
  http:
    value: 8080
    vip: http:80
```

We could also allow this shorthand when the vip prefix is the same as the port name:

```
ports:
  http:
    value: 8080
    vip: 80
```

mrbrowning:
Sure. Do we still need `advertise` on VIPs?

nickbp:
Oh whoops I was looking at an old version. Scratch that. There are currently `port`, `prefix`, `protocol` and `visible` fields, so that shortcut wouldn't really work. I guess it'd look more like:

```
ports:
  http:
    value: 8080
    vip:
      address: other-prefix:80
      protocol: udp
      visible: false
```

or

```
ports:
  http:
    value: 8080
    vip:
      address: 80
```

nickbp:
retest this please

frameworks/helloworld/src/main/dist/examples/taskcfg.yml
```diff
@@ -25,7 +22,7 @@ pods:
         cmd: "echo world-$POD_INSTANCE_INDEX >> world-container-path/$OUTPUT_FILENAME && sleep $SLEEP_DURATION"
         cpus: {{WORLD_CPUS}}
         memory: {{WORLD_MEM}}
-        volumes:
-          - path: "world-container-path"
-            type: ROOT
-            size: {{WORLD_DISK}}
+        volume:
+          path: "world-container-path"
```
keithchambers: Does this prevent (or require an API change) to supporting multiple volumes in the future?

keithchambers: Maybe it would be better to use the same approached used in `configurations`?  Of course this assumes multiple volumes would work, which might not be the case.

keithchambers: Nevermind.

sdk/scheduler/src/test/resources/valid-exhaustive.yml
```diff
@@ -1,7 +1,9 @@
 name: "data-store"
-principal: "data-store-principal"
-zookeeper: master.mesos:2181
-api-port: {{PORT0}}
+scheduler:
+  principal: "some-other-principal"
+  zookeeper: non-default-zk:2181
+  api-port: {{PORT_API}}
+  web-url: http://webui.data-store.mesos
```
keithchambers: This does not make sense here to me.

nickbp: Done. Moving back to parent level (adjacent to service name) per offline discussion.

sdk/scheduler/src/test/resources/full-manual-plan.yml
```diff
@@ -11,20 +9,20 @@ pods:
       hello-resources:
         cpus: {{HELLO_CPUS}}
         memory: 256
-        ports:
-          - name: http
+        endpoints:
```
nickbp: Done. Back to `ports` (and `port` per offline discussion)

sdk/scheduler/src/test/resources/full-manual-plan.yml
```diff
@@ -1,6 +1,4 @@
 name: "hello-world"
-principal: "hello-world-principal"
-api-port: 8080
 replacement-failure-policy:
   permanent-failure-timeout-ms: 10
   min-replace-delay-ms: 10
```
nickbp: Done

frameworks/hdfs/src/main/dist/hdfs_svc.yml
```diff
@@ -133,24 +144,26 @@ pods:
...
-            size: {{DATA_DISK}}
-            type: {{DATA_DISK_TYPE}}
+        volume:
+          path: "data-data"
+          size: {{DATA_DISK}}
+          type: {{DATA_DISK_TYPE}}
         uris:
           - https://downloads.mesosphere.com/hdfs/assets/hadoop-2.6.0-cdh5.7.1-dcos.tar.gz
           - https://downloads.mesosphere.com/dcos-commons/artifacts/resolve.zip
         configurations:
```
nickbp: Done

frameworks/helloworld/src/main/dist/examples/taskcfg.yml
nickbp: :+1:

frameworks/hdfs/src/main/dist/hdfs_svc.yml
```diff
@@ -12,21 +11,23 @@ pods:
         cpus: {{JOURNAL_CPUS}}
         memory: {{JOURNAL_MEM}}
         ports:
-          - name: journal_rpc
+          journal-rpc:
             port: {{JOURNAL_NODE_RPC_PORT}}
```
mrbrowning: I thought `port` here was changing to `value`?

frameworks/hdfs/src/main/dist/hdfs_svc.yml
```diff
@@ -166,24 +179,14 @@ plans:
...
         steps:
-          - pod-instance: 0
-            tasks: [format]
-          - pod-instance: 0
-            tasks: [node]
-          - pod-instance: 1
-            tasks: [bootstrap]
-          - pod-instance: 1
-            tasks: [node]
+          0: [format, node]
```
mrbrowning: Sorry, I'm actually not too clear on what the indices here are doing. Is it a pod-instance index, or a step index?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -18,13 +18,20 @@
     private final String portName;
     private final int port;
 
-    public PortEvaluationStage(Protos.Resource resource, String taskName, String portName, int port) {
+    public PortEvaluationStage(
```
mrbrowning: Nit: prefer not breaking lines unless the character limit is exceeded? Vertical screen real estate is always the comparatively limited one.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPlanGenerator.java
```diff
@@ -37,26 +38,17 @@ public Plan generate(RawPlan rawPlan, String planName, Collection<PodSpec> podsS
 
     @VisibleForTesting
     protected Phase from(RawPhase rawPhase, String phaseName, Collection<PodSpec> podsSpecs) {
-        String pod = rawPhase.getPod();
-        List<RawStep> rawSteps = rawPhase.getSteps();
-        String strategy = rawPhase.getStrategy();
-
-        Optional<PodSpec> podSpecOptnl = filter(pod, podsSpecs);
+        Optional<PodSpec> podSpecOptnl = filter(rawPhase.getPod(), podsSpecs);
```
mrbrowning: Might as well spell it out instead of disemvoweling.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawVip.java
```diff
@@ -24,4 +30,12 @@ public int getPort() {
     public String getPrefix() {
         return prefix;
     }
+
+    public String getProtocol() {
+        return protocol;
+    }
+
+    public Boolean isAdvertise() {
```
mrbrowning: Does the serialization logic require this name? If not, maybe `shouldAdvertise` or `isAdvertised`?

frameworks/hdfs/src/main/dist/hdfs_svc.yml
nickbp: Ended up just going with it as-is in discussion w/ Keith

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawVip.java
nickbp: It's not a serialization requirement, I'd just been naming things in `Raw*.java` ala java beans, with the philosophy being to have as little logic in `Raw*.java` as possible.

Done. Went with `isAdvertised`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPlanGenerator.java
nickbp: Done (had actually done this locally just now, prior to seeing this comment)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
nickbp: Done, back to one line.

frameworks/hdfs/src/main/dist/hdfs_svc.yml
nickbp: This section is fully manual, with the keys being pod indexes.

Example:

```
- 0: [format-index, format-data]
- 1: [networking]
- 0: [server]
- 1: [server]
```

In this case, the above ordering will be followed to the letter: Pod 0 will run `format-index` and `format-data` in parallel, then Pod 1 will run `networking`, then Pod 0 will run `server, then Pod 1 will run `server` as well.

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -12,21 +11,23 @@ pods:
...
+          journal-rpc:
             port: {{JOURNAL_NODE_RPC_PORT}}
-          - name: journal_http
+          journal-http:
             port: {{JOURNAL_NODE_HTTP_PORT}}
-        volumes:
-          - path: "journal-data"
-            size: {{JOURNAL_DISK}}
-            type: {{JOURNAL_DISK_TYPE}}
+        volume:
```
mrbrowning: Is this moving to a format like:
```
volumes:
  volume-name:
     ...
```
?

nickbp: Yes. There are two versions available:

- singular `volume` (without labels as seen here), shorthand for presumed common case where just one volume is needed
- multiple `volumes` (with labels as in your example) longer form for two or more volumes

The volume labels used in the multi case don't actually do anything for now. Can see both forms used in the exhaustive example.

nickbp:
FYI should be all caught up now, ready for (further) review

PR #441, opened 2017-01-10
use proper docker protobufs and fix rlimits NPE

mgummelt:
cc @mrbrowning

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawContainer.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawContainer.java
@@ -15,7 +15,12 @@ public class RawContainer {
             @JsonProperty("image-name") String imageName,
             @JsonProperty("rlimits") WriteOnceLinkedHashMap<String, RawRLimit> rlimits) {
         this.imageName = imageName;
-        this.rlimits = rlimits;
+
+        if (rlimits != null) {
+            this.rlimits = rlimits;
+        } else {
+            this.rlimits = new WriteOnceLinkedHashMap<>();
+        }
     }
 
     public LinkedHashMap<String, RawRLimit> getRLimits() {
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawTask.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawTask.java
@@ -14,7 +14,6 @@ public class RawTask {
 
     private final String goal;
     private final String cmd;
-    private final String image;
     private final Map<String, String> env;
     private final Collection<RawConfiguration> configurations;
     private final Collection<String> uris;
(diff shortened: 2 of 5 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawContainer.java
```diff
@@ -15,7 +16,12 @@ private RawContainer(
             @JsonProperty("image-name") String imageName,
             @JsonProperty("rlimits") WriteOnceLinkedHashMap<String, RawRLimit> rlimits) {
         this.imageName = imageName;
-        this.rlimits = rlimits;
+
+        if (rlimits != null) {
```
mgummelt: There may be a way of more gracefully handling this through the YAML mapper, but this was quicker.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawTask.java
```diff
@@ -76,10 +73,6 @@ public String getCmd() {
         return cmd;
     }
 
-    public String getImage() {
```
mgummelt: was never used, and I don't believe it makes sense to specify an image on a task

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawContainer.java
mrbrowning: From what I've seen, the typical pattern for dealing with potentially `null` members in the raw YAML specs has been to check for emptiness at the site of access, e.g. [YAMLToInternalMappers.java#L248](https://github.com/mesosphere/dcos-commons/blob/13165ea10fe67f4cc49ffaf09f95617eebe7a2c7/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java#L248). I'm not married to it, but it'd be good to have the approach be consistent.

mgummelt: Those checks are for emptiness, not non-nullness, right?  All else equal, I think it's better to get rid of nulls early.

nickbp: The util functions being used in there also check for null FYI

(I'm accustomed to `isEmptyOrNull()` but I can see why they'd opt for more brevity)

mgummelt: Yea, I know.  My point is that those fields aren't null.  They're empty collections.  That's what I'm trying to do here as well.

mgummelt: Jackson will auto-translate missing fields to empty collections if they're a `Collection`, but a `HashMap` is not a collection, which is why we need to set it here explicitly for consistency.

mrbrowning: Makes sense. I agree about reducing the scope of a null field, just wanted to ensure consistency in one direction or another.

PR #442, opened 2017-01-11
return WAITING in Plan and Phase, if interrupted and not completed.

mbdcos:
-Plan and Phase return WAITING, if interrupted and not completed.
-IN_PROGRESS status only for plans and phases

DefaultPlan and DefaultPhase used to check if interrupted in getStatus. That is omitted when we remove Element.setStatus. Putting back this feature in PlanUtils.getStatus (only Plan and Phase call PlanUtils.getStatus)!

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
@@ -757,6 +757,7 @@ public class DefaultScheduler implements Scheduler, Observer {
                 try {
                     stateStore.storeStatus(status);
                     planCoordinator.getPlanManagers().stream()
+                            //TODO: why do you ignore waiting plan here?
                             .filter(planManager -> !planManager.getPlan().isWaiting())
                             .forEach(planManager -> planManager.update(status));
                     reconciler.update(status);
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
@@ -757,8 +757,6 @@ public class DefaultScheduler implements Scheduler, Observer {
                 try {
                     stateStore.storeStatus(status);
                     planCoordinator.getPlanManagers().stream()
-                            //TODO: why do you ignore waiting plan here?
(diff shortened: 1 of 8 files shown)


mbdcos:
retest this please

mbdcos:
Retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -726,6 +726,7 @@ public void run() {
                 try {
                     stateStore.storeStatus(status);
                     planCoordinator.getPlanManagers().stream()
+                            //TODO: why do you ignore waiting plan here?
```
gabrielhartmann: I agree, this looks like a bug.

mbdcos:
retest this please

mbdcos:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Status.java
```diff
@@ -49,6 +44,17 @@
...
+    /**
+     * RETURN VALUE: only returned when execution has been interrupted.
+     */
+    WAITING,
 
+    /**
+     * RETURN VALUE: only returned by Phase and Plan getStatus, to state that at lease one child is complete and at
+     * least one child is in progress (either PENDING or prepared PREPARED).
+     */
+    IN_PROGRESS
```
gabrielhartmann: Let's avoid ordering changes on enums.  It's potentially dangerous for upgrades.  It doesn't matter since we don't persist these right now, but let's get in the habit of just appending to the enum list OK?

gabrielhartmann:
Other than one minor comment, it looks good to me.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Status.java
mbdcos: sure

mbdcos:
OK, merging after Jenkins

mbdcos:
retest this please

mbdcos:
enum ordering kept same. merging...

mbdcos:
retest this please

mbdcos:
enum ordering is same

rebased with master

merging....

PR #447, opened 2017-01-11
Fixed OfferEvaluation pack/unpack bug.

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -284,7 +284,7 @@ public void clearTask(String taskName) throws StateStoreException {
         try {
             byte[] bytes = curator.get(path);
             if (bytes.length > 0) {
-                return Optional.of(Protos.TaskInfo.parseFrom(bytes));
+                return Optional.of(CommonTaskUtils.unpackTaskInfo(Protos.TaskInfo.parseFrom(bytes)));
```
mrbrowning: StateStore is storing packed `TaskInfo`? Shouldn't we only pack it where we have to, e.g. where we send it to Mesos?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -64,6 +66,7 @@ public OfferEvaluator(StateStore stateStore, OfferRequirementProvider offerRequi
             }
 
             if (!failureNotifications.isEmpty()) {
+                recommendations.clear();
```
mrbrowning: Could also just make a new `recommendations` at the head of each loop run, up to you.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -256,13 +251,34 @@ public Builder setPlans(Collection<Plan> plans) {
         }
 
         /**
+         * Allow setting the capabilities of the DC/OS cluster.  Generally this should not be used except in test
+         * environments as it may return incorrect information regarding the capabilities of the DC/OS cluster.
+         * @param capabilities the capabilities used to validate the ServiceSpec
+         */
+        @VisibleForTesting
+        public Builder setCapabilities(Capabilities capabilities) {
```
mrbrowning: Nice, this is like a rider on a defense spending bill, but less insidious.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
gabrielhartmann: I thought about this and went back and forth on it.  In some sense I'd like the StateStore to reflect exactly what we told Mesos, so we can debug if necessary.  So I think storing the packed information is the right thing to do.  We send packed info to Mesos, so that's what we store.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
gabrielhartmann: This line comes from an abundance of caution.  It's a variable derived from the recommendationSlate, which is recreated anew on ever pass of the loop.  In fact the variable could probably go away altogether, let me look at that.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
mrbrowning: We also use what's in the state store to reconstitute TaskInfos when we're relaunching an existing pod, so there's some tension between the state store's dual role of "record of what we told Mesos" versus "repository for internal domain objects," which our unpacked illegal TaskInfos kind of are by virtue of representing an internal model that Mesos won't accept. The process for packing them and unpacking is deterministic, so I think they still serve their purpose as a debugging tool if you trust the packing/unpacking logic. I guess if the state store never lets them leave its scope in packed form then I'm okay with it.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
mohitsoni: LOL!

failure
```diff
@@ -0,0 +1,253 @@
+Tue Jan 10 23:17:46 2017 ip-10-0-1-8.us-west-2.compute.internal hello [9642] I0110 23:17:46.305068  9738 exec.cpp:161] Version: 1.0.1
```
nickbp: Probably shouldn't be in the PR

gabrielhartmann: AHH!

PR #450, opened 2017-01-12
Add web-url to ServiceSpec

gabrielhartmann:
This is needed for the consumption of proxylite as a custom
proxy to multiple service APIs (e.g. the Scheduler API and Kibana)

Proxylite is a container under development which is deployable as a pod
in each service which will provide routing through admin router to 
framework author specified HTTP endpoints.


nickbp:
This is also done in YAML Polish, fwiw: https://github.com/mesosphere/dcos-commons/pull/440

PR #451, opened 2017-01-12
Add ability to provide custom RestartHook to DefaultScheduler.Builder

nickbp:
Allows developers to specify custom teardown logic when 'pod restart' and/or 'pod replace' is invoked.


gabrielhartmann:
@nickbp:  Looks reasonable and of small scope.  In general I'm wary of adding too many hooks in different lifecycle locations, but this is fine.  Would be good to get the community members who requested this to review before we merge.

nickbp:
The community member in DC/OS slack sounded satisfied, looks like we're set.

PR #453, opened 2017-01-12
Resync swagger spec to current state

nickbp:
- Add `/state`
- Move `/tasks` to `/pods` and `/endpoints`
- Other misc fixes, e.g. updated response codes


sascala:
LGTM

PR #460, opened 2017-01-13
Refactor framework YAML validation tests.

gabrielhartmann:
* Refactor framework YAML validation tests.
* Move test utilities into a test project with test compile rules.


nickbp:
Superset of #459?

gabrielhartmann:
@nickbp, yes it's a superset.  They won't conflict and I can test in parallel.

nickbp:
- Should be able to rebase now
- Binaries in `frameworks/proxylite/cli/...`

frameworks/kafka/src/main/resources/kafka_svc.yml
```diff
@@ -0,0 +1 @@
+../dist/kafka_svc.yml
```
nickbp: why the rename? I'd have thought this would stay `svc.yml` if anything

nickbp: (in yaml-polish I was actually renaming `hdfs_svc.yml` => `svc.yml`)

gabrielhartmann: Oh, this isn't necessary.  You can rename back everything to svc.yml.  When I was trying to fix tests, things were picking up the wrong resource objects because they had the same name and I was having trouble figuring out what was going on.  That isn't a problem now.

nickbp: OK I'll rename it back in #440. Would be good to get this merged and no reason to bump the CI for this.

PR #463, opened 2017-01-17
Enable recovery tests as sanity tests.

gabrielhartmann:


gabrielhartmann:
@NimaVaziri:  Can you take this PR over please?

PR #464, opened 2017-01-18
DON'T Use YAML for ServiceSpecs in DefaultSchedulerTest

gabrielhartmann:


sdk/scheduler/src/test/resources/test-scale-pod-a.yml
```diff
@@ -0,0 +1,36 @@
...
+scheduler:
+  api-port: 8080
+pods:
+  pod-a:
+    count: 2
+    resource-sets:
+      resource-set-a:
+        cpus: 1
+        memory: 1000
+        volumes:
```
nickbp: If you want you could also just use `volume:` in single-blind cases like these, see e.g. https://github.com/mesosphere/dcos-commons/blob/master/frameworks/hdfs/src/main/dist/svc.yml#L18

sdk/testing/src/main/java/com/mesosphere/sdk/testing/BaseServiceSpecTest.java
```diff
@@ -27,7 +27,7 @@
     public static final EnvironmentVariables ENV_VARS = new EnvironmentVariables();
 
     @Before
-    public void beforeEach() {
+    public void beforeEach() throws Exception {
```
nickbp: does it though?

sdk/scheduler/src/test/resources/test-update-pod-a.yml
```diff
@@ -0,0 +1,36 @@
+name: "test"
```
nickbp: Something to think about: dealing with minor changes between these files for tests. As it stands it's not easy to tell what's different between these files at a glance. Was thinking patch files but those would be annoying to maintain as the base files are updated. Maybe have a single common `RawServiceSpec` that gets parsed in from a single file, and then the `RawServiceSpec` is what gets modified for the tests?

mrbrowning: Is there anything like Python's [factory_boy](https://factoryboy.readthedocs.io/en/latest/) for creating fixtures in code and automating the generation of valid values for fields you don't care about in a test? I hate fixtures with a passion, and if we can nip this proliferation of slightly-differing YAMLs now, all the better.

mrbrowning: Alternatively, if not that, then maybe a Builder class for `RawSpec` that handles the generation of sensible defaults for unspecified fields? I reaaaally don't want to end up with a folder full of literally 100s of test YAMLs, which I *promise* will be the end result here, and which will be a real maintainability burden should we end up adding or modifying features that bubble up to YAML.

sdk/testing/src/main/java/com/mesosphere/sdk/testing/BaseServiceSpecTest.java
gabrielhartmann: It doesn't.  I had to add it so overriding in sub-classes could do it.

gabrielhartmann:
@nickbp and @mrbrowning:  I spent a ton of time doing this and we've lost maybe 200 lines of YAML and gained ~2000 lines of code.  I'm not sure I like this trade off at all.  Also we're not exercising the YAML to Java translation layer as much.  Also it's harder to write tests...

Not sure about this direction.  I think a lot of YAML to Java translation test code needs to be written now and then I'm not sure what we've gained.

mrbrowning:
So, I'll let this rest for now, but it's still something I'm interested in pursuing. I think this effort failed because I didn't really communicate the nature of what I meant by using builders -- test builders aren't meant for general purpose use and should in general be testing-only infrastructure, and as such their purpose isn't to serve as general-purpose builders but to serve as builders that give you *by default*, and with no other parameters supplied, a minimum viable valid instance of the object in question. So if I were create an instance of `RawServiceSpecTestBuilder` with no parameters, it would give me a service with e.g. a name, a principal, and a pod. The `RawPodSpecTestBuilder` would be invoked in the process of creating that default pod, and so on in turn, with either reasonable defaults being supplied for simple types or another test builder being invoked for compound types. This way, you don't have to specify an entire spec from the ground up each time -- you only supply the fields relevant to your test, however far down the tree you are, and you can be assured that everything else about the instance will be valid. The rest of my comments will be with reference to this hypothetical implementation.

> I spent a ton of time doing this and we've lost maybe 200 lines of YAML and gained ~1000 lines of code.

Think about this like TCO. The present-day tradeoff is less the right comparison than the number of lines of YAML that will proliferate as the feature base grows (it has been my experience that number of lines devoted to fixtures increases roughly quadratically in the number of features added).

> Also we're not exercising the YAML to Java translation layer as much.

If this is reduces our confidence in the YAML translation layer to an unacceptable level, then those tests are insufficient.

> Also it's harder to write tests.

This would hopefully be assuaged by the model I describe above.

> Also, now the YAML translation layer should be refactored to use the builders, I guess.

No, test builders shouldn't be used in the rest of the codebase, since they should have test-specific functionality.

> Not sure about this direction. I think a lot of YAML to Java translation test code needs to be written now to make sure it's still being exercised as fully as it was before and then I'm not sure what we've gained.

We've gained test isolation -- exercising multiple layers of a stack in a test makes those tests less informative, since they are subject to fail even when the logic they are specifically testing remains correct, but something in the lower level has changed behavior. Tests should be like a map for finding where in your codebase an assumption has failed to hold after a change: if my change broke something in the YAML layer, I want only the YAML tests to fail. If the ServiceSpec tests fail too, then I have comparatively little in the way of guidance as to where the problem is, hindering my effectiveness in making changes to the codebase.

> Using RawSpecBuilders is far more difficult to write, and more importantly to read.

This should also be ameliorated with the model described above. But in general: code in front of me, assuming reasonable code (e.g. a builder where I only need to specify the fields relevant to my test, and can assume validity for everything else), is always much easier to read than a fixture far away, especially when most of that fixture is boilerplate devoted to ensuring that we get a valid instance of the object and isn't directly concerned with the behavior I'm testing.

Anyway, I might give this another shot myself. It may be that my confidence in this approach is shaded by how easy it is in dynamically-typed languages, but I think a reasonable test builder infrastructure can be created here.

PR #465, opened 2017-01-18
Extend LD_LIBRARY_PATH instead of a complete override

mohitsoni:
`LD_LIBRARY_PATH` of the task should be extended to also include libmesos, instead of overriding it to just include libmesos. Not extending `LD_LIBRARY_PATH` causes other system utilities to not work as expected.

Basically, we should be doing:
```
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<libmesos-stuff>
```
v/s
```
export LD_LIBRARY_PATH=<libmesos-stuff>
```


mohitsoni:
Closing in favor of @gabrielhartmann's https://github.com/mesosphere/dcos-commons/pull/468 PR.

PR #467, opened 2017-01-18
All task env-vars including ports are now also available on health and readiness checks

mohitsoni:
Both health checks and readiness checks will now get all the env vars of the task they are associated with.

/cc: @loren

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
@@ -2,6 +2,8 @@ package com.mesosphere.sdk.offer.evaluate;
 
 import com.mesosphere.sdk.offer.*;
 import org.apache.mesos.Protos;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashSet;
 import java.util.Optional;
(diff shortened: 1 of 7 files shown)


mrbrowning:
Code looks good, could you add a test? Also, nit, but the PR title could be more descriptive of the fact that all task envvars are now available in the executor environment.

loren:
I tried running this against my test case and it still does not seem to work. The only difference I see is that the dynamic port env var that is passed into the command env vars is now capitalized, so what used to be `PORT_http` is now `PORT_HTTP`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -82,6 +85,36 @@ protected void setProtos(OfferRequirement offerRequirement, Protos.Resource reso
...
+                            getPortEnvironmentVariable(portName),
+                            Long.toString(port));
+                    Protos.HealthCheck readinessCheckWithPort = Protos.HealthCheck.newBuilder(readinessCheckToMutate)
+                            .setCommand(readinessCommandWithPort).build();
+                    CommonTaskUtils.setReadinessCheck(taskInfoBuilder, readinessCheckWithPort);
+                } else {
+                    LOGGER.info("Readiness check is not defined for task: {}", taskName);
+                }
+            } catch (TaskException e) {
+                LOGGER.error("Got exception while adding PORT env vars to ReadinessCheck", e);
```
loren: s/vars/var

PR #468, opened 2017-01-19
Extend, not override LD_LIBRARY_PATH

gabrielhartmann:
Additionally, simplify generic prefixes on Scheduler cmds

### DIFF
--- a/frameworks/hdfs/universe/config.json
+++ b/frameworks/hdfs/universe/config.json
@@ -14,6 +14,11 @@
           "description":"The principal for the HDFS service instance.",
           "type":"string",
           "default":"hdfs-principal"
+        },
+        "cmd_prefix" : {
+          "description":"A generic prefix to start the scheduler.",
+          "type":"string",
+          "default":"LD_LIBRARY_PATH=$MESOS_SANDBOX/libmesos-bundle/lib:$LD_LIBRARY_PATH MESOS_NATIVE_JAVA_LIBRARY=$(ls $MESOS_SANDBOX/libmesos-bundle/lib/libmesos-*.so) PATH=$(ls -d $MESOS_SANDBOX/jre*/bin):$PATH"
         }
       }
     },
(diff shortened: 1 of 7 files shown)


gabrielhartmann:
@mohitsoni:  We should remove the hack in Gilbert's dse implementation regarding LD_LIBRARY_PATH once this is merged.

gabrielhartmann:
@gmflau: You will want to rebase on master after this PR goes in.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -470,7 +470,7 @@ private static boolean isPortResource(Protos.Resource resource) {
         // command and user:
 
         Protos.CommandInfo.Builder commandInfoBuilder = executorInfoBuilder.getCommandBuilder()
-                .setValue("export LD_LIBRARY_PATH=$MESOS_SANDBOX/libmesos-bundle/lib && " +
+                .setValue("export LD_LIBRARY_PATH=$MESOS_SANDBOX/libmesos-bundle/lib:$LD_LIBRARY_PATH && " +
```
mrbrowning: Not for this PR, but should we consider making the executor command prefix configurable with a reasonable default? Most people won't need or want to touch it, but those that do would probably appreciate not being forced into using our bundle.

gabrielhartmann:

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
gabrielhartmann: @mrbrowning: It's definitely a thought worth discussion.  This changes something pretty fundamental to the operation of a Framework.  We try to keep each version of a package very well defined.  As long as a change here were statically built into a package I'd be happy, but I worry about N customers defining N different ways to start the executor.  Could be a support problem

mohitsoni:
I going to close https://github.com/mesosphere/dcos-commons/pull/465 as this PR fixes the same problems.

frameworks/hdfs/universe/config.json
```diff
@@ -14,6 +14,11 @@
           "description":"The principal for the HDFS service instance.",
           "type":"string",
           "default":"hdfs-principal"
+        },
+        "cmd_prefix" : {
```
nickbp: Feels a little iffy putting this option front and center in the `service` section but there currently isn't an obviously better place to put it. If more options like this are added we may want to break them out into an `advanced` section at some point but that's more of a UI/product thing.

gabrielhartmann: Yeah, I don't want this here, but as you say, I have nowhere else to put it.  I think it's better than in Marathon.json.mustache which was getting to be a huge ugly command.  Also the Template framework will just get this, things will stay a bit neater.

PR #469, opened 2017-01-20
Template framework

gabrielhartmann:
A simple template framework has been added to clone for the creation of new frameworks.  I tested by running:
```
$ ./new-service.sh frameworks/test
$ ./test.sh test
```

Integration tests passed, for the simplest install case.


mohitsoni:

gabrielhartmann:
@mohitsoni @mrbrowning ping

frameworks/template/src/main/resources/logback.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mrbrowning: This file doesn't need to exist, nothing uses it. Populate `simplelogger.properties` (see [here](https://www.slf4j.org/api/org/slf4j/impl/SimpleLogger.html)) until we move to a real logging backend.

settings.gradle
```diff
@@ -20,3 +20,6 @@ project(":frameworks/kafka").name = "kafka"
 
 include 'frameworks/hdfs'
 project(":frameworks/hdfs").name = "hdfs"
+
+include 'frameworks/template'
```
mrbrowning: Should this be here? Presumably someone adding a framework would add these lines when they actually create a new one in the repository, but doesn't seem worth including it by default.

gabrielhartmann: The `new-service` script does indeed add new lines here when generating a new project.  Adding the template project here makes it get built and tested like all other projects, so we prove it's a valid project.  Otherwise it may not be kept up to date as a valid project which others should be based on.

PR #470, opened 2017-01-20
Replace label-based config templates with a host-based method

nickbp:
Entirely replaces the current Label-based config distribution system.

Scheduler:
- Adds a new `ArtifactsResource` to the scheduler to host the config templates.
- Updates current `ConfigResource`/`StateResource` to be more accurate in errors (see summary of fixes in `swagger-spec.yaml` -- e.g. don't return 500 when it should really be a 404).
- Updates artifact URLs in YAML to only be defined at a per-pod level, not at a per-task level. This avoids potential for confusing disk state across tasks in a pod and solidifies the idea that pods represent a set of shared resources (included pre-downloaded artifacts).

Executor:
- Adds a `bootstrap` process (with many available customizations) to be run in executors before tasks start up. At the moment it supports printing the current env, waiting for host(s) to resolve (replaces `resolve-dns.sh`), and performing the mustache templating work on config files (advertised via envvars provided by the scheduler). Each of these steps is configurable (or can be disabled entirely) via commandline args. Developers are expected to explicitly provide the bootstrap process in their `cmd`, along with any customization via args, to avoid being too "magical" (see updated yamls for kafka and hdfs).

Config process steps:
- Scheduler includes URLs to itself (against ArtifactsResource) for the mesos fetcher to retrieve the config template(s). The URLs are against the `api` named VIP in order to get the port right (mesos-dns would require a SRV lookup which mesos fetcher doesn't do). Scheduler also includes envvars which reference those config templates: `CONFIG_TEMPLATE_FOO=<src-path>,<dest-path>`.
- When executor starts, the mesos fetcher retrieves the config templates and puts them as-is in a directory named `config-templates/` in the sandbox. Note that this separate staging directory is intentional: The mesos fetcher scrambles the order by which it downloads URLs so we can't guarantee that our config templates won't be overwritten by unzipping of a stock distribution (e.g. kafka.tgz's default `server.properties` overwriting a templated `server.properties`). The separate staging directory also makes it easy for users to examine their templates.
- The executor launches `cmd`, to which the developer has included a command to run `bootstrap`. `bootstrap` looks for the `CONFIG_TEMPLATE_*` envvars and performs the template work, copying the result into the defined destination paths.


nickbp:
retest this please

nickbp:
retest this please

frameworks/helloworld/universe/resource.json
```diff
@@ -1,7 +1,7 @@
 {
   "assets": {
     "uris": {
-      "jre-tar-gz": "https://downloads.mesosphere.com/dcos-commons/artifacts/jre-8u91-linux-x64.tar.gz",
+      "jre-tar-gz": "https://downloads.mesosphere.com/java/jre-8u112-linux-x64.tar.gz",
```
triclambert: FYI, 8u121 is out.

sdk/bootstrap/build.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo "Binary CLI compression enabled: $($UPX_BINARY -V | head -n 1)"
+else
+    echo "Binary CLI compression disabled"
+fi
+
+# Configure GOPATH with dcos-commons symlink (rather than having it pull master):
+echo "Creating GOPATH symlink into dcos-commons: $GOPATH"
+rm -rf $GOPATH_MESOSPHERE/$REPO_NAME
+mkdir -p $GOPATH_MESOSPHERE
+pushd $GOPATH_MESOSPHERE
```
mrbrowning: Can also achieve the same effect of a `pushd/popd` by running in a subshell with `( cd $dir ; ... )`. I like this especially when defining a function that mucks with the cwd, but it's a matter of taste.

sdk/bootstrap/build.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
```
mrbrowning: Could do `/usr/bin/env bash`, although we're probably not portable across systems for many other more fundamental reasons.

sdk/bootstrap/build.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# exit immediately on failure
+set -e
```
mrbrowning: `set -eu` to also disallow the reference of unset envvars? I'd normally also plug `-o pipefail` but there are no pipes in this script.

sdk/bootstrap/build.sh
```diff
@@ -0,0 +1,73 @@
...
+BOOTSTRAP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_ROOT_DIR=$(dirname $(dirname $BOOTSTRAP_DIR))
+
+if [ -z "$GOPATH" -o -z "$(which go)" ]; then
+  echo "Missing GOPATH environment variable or 'go' executable. Please configure a Go build environment."
+  exit 1
+fi
+
+REPO_NAME=dcos-commons # CI dir does not match repo name
+GOPATH_MESOSPHERE=$GOPATH/src/github.com/mesosphere
```
mrbrowning: For maximal defensiveness, quote all envvars like `$GOPATH`, and envvars that contain those envvars, that could be set by a user to contain spaces.

frameworks/kafka/src/main/dist/svc.yml
```diff
@@ -14,9 +17,7 @@ pods:
           type: ROOT
           size: {{BROKER_DISK}}
         goal: RUNNING
-        cmd: "env && exec $MESOS_SANDBOX/kafka_2.11-0.10.0.0/bin/kafka-server-start.sh $MESOS_SANDBOX/kafka_2.11-0.10.0.0/config/server.properties"
-        uris:
-          - "https://downloads.mesosphere.com/kafka/assets/kafka_2.11-0.10.0.0.tgz"
+        cmd: "./bootstrap -resolve=false && exec $MESOS_SANDBOX/kafka_2.11-0.10.0.0/bin/kafka-server-start.sh $MESOS_SANDBOX/kafka_2.11-0.10.0.0/config/server.properties"
```
gabrielhartmann: YAML supports multline strings like this:
https://github.com/mesosphere/dcos-commons/blob/master/frameworks/helloworld/src/main/dist/svc.yml#L31-L34

Also we're very inconsistent throughout our YAML files.  Generally speaking all of our strings do not need to use `"` marks, so if you could remove them when you see them, that'd be cool.

sdk/bootstrap/main.go
```diff
@@ -0,0 +1,276 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	// TODO switch to upstream once https://github.com/hoisie/mustache/pull/57 is merged:
```
gabrielhartmann: Neat.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/ArtifactResource.java
```diff
@@ -0,0 +1,130 @@
...
+        this.configStore = configStore;
+    }
+
+    /**
+     * Produces the content of the requested configuration template, or returns an error if that template doesn't exist
+     * or the data couldn't be read.
+     */
+    @Path("/template/{configurationId}/{podType}/{taskName}/{configurationName}")
+    @GET
+    public Response getTemplate(
```
gabrielhartmann: Can you please update the javadoc with the parameters to this method?  All are obvious to me except `configurationId`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -470,53 +476,75 @@ private static boolean isPortResource(Protos.Resource resource) {
...
             javaUri = DEFAULT_JAVA_URI;
         }
-        commandInfoBuilder.addUrisBuilder().setValue(javaUri);
+        executorCommandBuilder.addUrisBuilder().setValue(javaUri);
 
-        for (TaskSpec taskSpec : podSpec.getTasks()) {
-            if (!taskSpec.getCommand().isPresent()) {
-                continue;
+        // Any URIs defined in PodSpec itself.
+        if (podSpec.getUris() != null) {
```
gabrielhartmann: Why do we have null checks here and below?  Shouldn't these be guaranteed to be empty collections?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -56,6 +58,15 @@ public static PodSpec getPodSpec(ServiceSpec serviceSpec, TaskInfo taskInfo) thr
     }
 
     /**
+     * Converts the provided string to a valid environment variable name.
+     *
+     * For example: {@code hello.There!} => {@code HELLO_THERE_}
+     */
+    public static String toEnvName(String str) {
+        return ENVVAR_INVALID_CHARS.matcher(str.toUpperCase()).replaceAll("_");
```
gabrielhartmann: What's wrong with lowercase characters?  It's not conventional for sure, but valid right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -271,37 +224,33 @@ public static boolean volumesEqual(TaskSpec first, TaskSpec second) {
...
-            String prevValue =
-                    configMap.put(configSpecification.getRelativePath(), configSpecification.getTemplateContent());
-            if (prevValue != null) {
+    private static Map<String, ConfigFileSpec> getConfigTemplateMap(Collection<ConfigFileSpec> configSpecs)
+            throws IllegalArgumentException {
+        Set<String> configPaths = new HashSet<>();
+        Map<String, ConfigFileSpec> configMap = new HashMap<>();
+        for (ConfigFileSpec configSpec : configSpecs) {
+            if (!configPaths.add(configSpec.getRelativePath())) {
+                throw new IllegalArgumentException(String.format(
```
gabrielhartmann: Is there a test that exercises this path?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -271,37 +224,33 @@ public static boolean volumesEqual(TaskSpec first, TaskSpec second) {
...
+                        configSpec.getRelativePath()));
+
+            }
+            ConfigFileSpec prevSpec = configMap.put(configSpec.getName(), configSpec);
+            if (prevSpec != null) {
                 throw new IllegalArgumentException(String.format(
-                        "Config templates for a given task may not share the same path. " +
-                                "path:'%s' oldContent:'%s' newContent:'%s'",
-                        configSpecification.getRelativePath(), prevValue, configSpecification.getTemplateContent()));
+                        "Config templates for a given task may not share the same name: '%s'",
```
gabrielhartmann: Is this the correct error message?

nickbp: It is, added a comment to the code to explain what's going on: `A config of this name is already present in the map`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
nickbp: Yes. `TaskUtilsTest` has both `testConfigsSamePathFailsValidation` and `testConfigsSameNameFailsValidation`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
nickbp: Lowercase is legal but unusual. I've only seen lowercase in shell scripts for internal use (ie unexported variables). Updated the javadoc to clarify.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
nickbp: They can be null. Updated the specs to ensure that they're always non-null, and removed the null checks here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/ArtifactResource.java
nickbp: Done

sdk/bootstrap/build.sh
nickbp: Oh interesting I didn't know about that version. I'll leave this as is for now with `pushd`/`popd`, if only because there's a second `pushd`/`popd` pair further down that's less trivial. Figure it may as well stick to the same strategy across the file.

sdk/bootstrap/build.sh
nickbp: Added quotes around the path vars throughout.

sdk/bootstrap/build.sh
nickbp: I'd add this but there an explicit check for `$GOPATH` which fails in the unset case:
```
if [ -z "$GOPATH" -o -z "$(which go)" ]; then
  echo "Missing GOPATH environment variable or 'go' executable. Please configure a Go build environment."
  exit 1
fi
```

Result with `set -eu` (no pretty error):
```
./build.sh: line 9: GOPATH: unbound variable
```

sdk/bootstrap/build.sh
nickbp: Eh we're at least targeting mac/linux, so thats two. Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
gabrielhartmann: Oh I see.  I read the two error messages as identical.  One is for `name`, one is for `path`.  Got it.

nickbp:
retest this please

nickbp:
whoops wrong button

nickbp:
retest this please

PR #473, opened 2017-01-22
Fix multi-volume support

gabrielhartmann:
The main error was a failure to differentiate disk resouces on container
path.  Other resources may be safely differentiated based on name (e.g.
cpu, mem).  However disk resources must be further compared with
reference to their container path.


sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorBaseTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.contrib.java.lang.system.EnvironmentVariables;
+
+import java.io.File;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * A BaseTest for use in writing offer evaluation tests.
+ */
+public class OfferEvaluatorBaseTest {
```
nickbp: How about `OfferEvaluatorTestBase` (`Test` suffix implies it contains tests)

PR #477, opened 2017-01-24
Replace local spin() method with shakedown version

loren:
- Optimized imports
- Removed unused nested dict options code around service installation
- Added simple test around DCOS service health as an example


nickbp:
FYI there's similar cleanup (larger in scope) WIP here:

https://github.com/mesosphere/dcos-commons/compare/test-cleanup?expand=1

PR #478, opened 2017-01-24
Making StateStoreCache's fetchXXX similar to CuratorStateStore

mohitsoni:
StateStoreCache was not unpacking TaskInfo for fetch calls, which was causing problems elsewhere in `PortEvaluationStage` class (kind of like butterfly effect). More specifically, PortEvaluationStage expects certain env vars in TaskInfo's CommandInfo proto which are only available once the TaskInfo it belongs to is unpacked.

This PR does following:
1. Makes StateStoreCache unpack TaskInfo for fetch calls just like CuratorStateStore
2. Adds corresponding tests inside StateStoreCacheTest
3. Minor fix to a logging statement.

cc: @loren

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/state/PersistentOperationRecorder.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/state/PersistentOperationRecorder.java
@@ -1,5 +1,6 @@
 package com.mesosphere.sdk.state;
 
+import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.TextFormat;
 import org.apache.mesos.Protos;
 import org.apache.mesos.Protos.Offer;
@@ -11,6 +12,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 
 /**
@@ -34,7 +36,13 @@ public class PersistentOperationRecorder implements OperationRecorder {
     private void recordTasks(List<Protos.TaskInfo> taskInfos) throws StateStoreException {
         logger.info(String.format("Recording %d updated TaskInfos/TaskStatuses:", taskInfos.size()));
         List<Protos.TaskStatus> taskStatuses = new ArrayList<>();
-        for (Protos.TaskInfo taskInfo : taskInfos) {
+        Collection<Protos.TaskInfo> unpackedTaskInfos;
+        try {
+            unpackedTaskInfos = CommonTaskUtils.unpackTaskInfos(taskInfos);
+        } catch (InvalidProtocolBufferException e) {
+            throw new StateStoreException(e);
+        }
+        for (Protos.TaskInfo taskInfo : unpackedTaskInfos) {
             if (!taskInfo.getTaskId().getValue().equals("")) {
                 Protos.TaskStatus.Builder taskStatusBuilder = Protos.TaskStatus.newBuilder()
                         .setTaskId(taskInfo.getTaskId())
@@ -51,7 +59,7 @@ public class PersistentOperationRecorder implements OperationRecorder {
             }
         }
 
-        stateStore.storeTasks(taskInfos);
+        stateStore.storeTasks(unpackedTaskInfos);
         for (Protos.TaskStatus taskStatus : taskStatuses) {
             recordTaskStatus(taskStatus);
         }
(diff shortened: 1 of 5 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreCache.java
```diff
@@ -216,7 +222,13 @@ public void clearTask(String taskName) throws StateStoreException {
     public Optional<TaskInfo> fetchTask(String taskName) throws StateStoreException {
         RLOCK.lock();
         try {
-            return Optional.ofNullable(nameToTask.get(taskName));
+            if (nameToTask.containsKey(taskName)) {
```
nickbp: Could just do `nameToTask.get()` and check for a null value. Single map lookup instead of two.

nickbp:
Feel free to merge (modulo the above nitpick) to fix the issue.

---

It feels a little off that CuratorStateStore/StateStoreCache need to know about unpacking taskinfos to begin with. Medium/long term I think the current packing workaround should be revisited as keeping it working is getting increasingly difficult. Is there some other way around the issue that doesn't require two versions of `TaskInfo` that look alike but are incompatible with each other? If we could have a separate type for packed vs unpacked versions that'd at least make things clearer from the code. cc @gabrielhartmann

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreCache.java
mohitsoni: Because we don't allow storing `null` tasks in `nameToTask` map. I think it's fine to drop the `containsKey` check.

mohitsoni:
@nickbp I had similar thoughts about packing/unpacking taskInfos inside CuratorStateStore/StateStoreCache. If we keep our current design, `TaskInfo` should just be packed right before they are dispatched to be run, and shouldn't be stored inside ZK or Cache in "packed" mode. If we can adhere to that contract, we should we fine. As executor already know about unpacking them. And doesn't rely on ZK or Cache.

mohitsoni:

mohitsoni:

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/PersistentOperationRecorder.java
```diff
@@ -34,7 +36,13 @@ public void record(Operation operation, Offer offer) throws Exception {
     private void recordTasks(List<Protos.TaskInfo> taskInfos) throws StateStoreException {
         logger.info(String.format("Recording %d updated TaskInfos/TaskStatuses:", taskInfos.size()));
         List<Protos.TaskStatus> taskStatuses = new ArrayList<>();
-        for (Protos.TaskInfo taskInfo : taskInfos) {
+        Collection<Protos.TaskInfo> unpackedTaskInfos;
+        try {
+            unpackedTaskInfos = CommonTaskUtils.unpackTaskInfos(taskInfos);
```
nickbp: Instead of unpacking from `OfferRecommendation.getOffer()`, how about passing the original TaskInfo directly into `LaunchOfferRecommendation`, and then performing the packing inside of there? Then `PersistentOperationRecorder` can do something like:

```
if (operation instanceof LaunchOfferRecommendation) {
  LaunchOfferRecommendation lro = (LaunchOfferRecommendation) operation;
  recordTasks(lro.getOriginalTaskInfos());
}
```

This should make it possible to ensure that only the executor is calling `unpackTaskInfos()`

mohitsoni: I see the point that you are making. But, we cannot do `operation instanceof LaunchOfferRecommendation` as `LaunchOfferRecommendation` IS-NOT-A `Operation` but IS-A `OfferRecommendation` which HAS-A `Operation`. Let me see what we can do here to simplify this further.

mohitsoni:

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/PersistentOperationRecorder.java
nickbp: Oh right, well, that's the idea anyway -- PersistentOfferRecorder would retrieve the original unpacked TaskInfos directly from LaunchOfferRecommendation

nickbp:
OK, I ended up getting a fix locally, but it touches a lot of code and I don't want to delay your fix any longer. I'll merge your fix for the immediate issue, and then start a separate PR for cleaning up the packed taskinfo handling.

PR #484, opened 2017-01-25
Dedupe and clean integration tests

nickbp:
- Better logging across the board, allowing ability to tell what tests are actually doing/waiting for -- Development is MUCH faster when you can see a problem live instead of waiting >15m for a sleep timer to fail.
- Extracts common logic from each service and moves it into a new `/testing` package which each of the per-service tests include via their `__init__.py`. This common logic should later be upstreamed into shakedown, but for now it's all in one distinct and easy-to-export location.
- Flattens directory structure from `framework/<SVC>/integration/tests/` to just `framework/<SVC>/tests/`.
- Makes `requirements.txt` argument optional in `run_tests.py`, instead using common defaults if left unspecified (currently always the case).
- Updates the `py.test` command in `run_tests.py` to enable printing a stacktrace upon keyboard interrupt (likewise useful when running tests locally)
- 600+ LOC removed!


gabrielhartmann:
@nickbp:  Did you change the github notifications in CI in this PR?  Why is the `build:sdk` step yellow.  Looks like it completed successfully.

nickbp:
`mesosphere-ci` was seeing errors when updating PR statuses due to github rate limiting (probably another job on `mesosphere-ci` gone rogue?), figured it made sense for that to not disrupt the build itself

PR #488, opened 2017-01-26
Only use unpacked TaskInfos throughout the Scheduler

nickbp:
The packed `TaskInfo` is now *only* stored within the `LaunchOfferRecommendation`'s Operation to be sent to Mesos. All other scheduler code has been moved to exclusively using the original unpacked `TaskInfo`, by updating interfaces to accept the parent `OfferRecommendation` with the unpacked `TaskInfo` rather than the `Operation` with the packed `TaskInfo`.

For compatibility with prior releases that store packed `TaskInfo`s, this keeps a speculative unpack in `CuratorStateStore`. Added a deprecation notice for removing this speculative unpack after a few releases.

Renamed `PersistentOperationRecorder` to `PersistentLaunchRecorder` to reflect that it only cares about LAUNCH operations.


sdk/common/src/main/java/com/mesosphere/sdk/offer/CommonTaskUtils.java
```diff
@@ -448,8 +462,6 @@ public static TaskInfo packTaskInfo(TaskInfo taskInfo) {
 
             if (taskInfo.hasCommand()) {
                 executorInfoBuilder.setCommand(taskInfo.getCommand());
-            } else {
-                executorInfoBuilder.setCommand(CommandInfo.getDefaultInstance());
```
nickbp: (ensures that if the original TaskInfo lacked a command, the packed+unpacked version lacks it too)

mrbrowning:
Mostly looks good to me, but I'm not sure about the move from `Operation` to `*Recommendation` as the input to a bunch of functions. I know it makes it easier to not have to call `recommendation::getOperation`, but I kind of like the idea of having those functions operate on the minimal object necessary to do their jobs. This is subjective, so not a blocker.

PR #489, opened 2017-01-26
Deal with reregistration correctly

nickbp:
- Port base reregister() support from cassandra (assumption that cassandra does things correctly)
- Port of reregistration-specific fix recently made to cassandra: https://github.com/mesosphere/dcos-cassandra-service/issues/372

Also:
- Replace enum.ordinal() calls with explicit integer values.
- Cleanup of logs and unused code in CustomExecutor.


sdk/executor/src/main/java/com/mesosphere/sdk/executor/CustomExecutor.java
```diff
@@ -44,15 +43,13 @@ public void registered(
...
+        LOGGER.info("Registered executor: {}", TextFormat.shortDebugString(executorInfo));
         this.slaveInfo = slaveInfo;
-        this.executorInfo = executorInfo;
-        this.frameworkInfo = frameworkInfo;
     }
 
     @Override
     public void reregistered(ExecutorDriver driver, Protos.SlaveInfo slaveInfo) {
-        LOGGER.info("Re-registered on slave: {}", slaveInfo.getId());
+        LOGGER.info("Re-registered on slave: {}", TextFormat.shortDebugString(slaveInfo));
```
mrbrowning: This might be a case where I'd prefer just the ID, since the rest of the agent info isn't really relevant, can be cross-referenced anyway, and there are probably other logs that contain the meat of the issue that I wouldn't want obscured by this full proto debug string. What do you think?

sdk/executor/src/main/java/com/mesosphere/sdk/executor/CustomExecutor.java
```diff
@@ -168,20 +165,20 @@ public void killTask(ExecutorDriver driver, Protos.TaskID taskId) {
         // TODO(mohit): Implement SIGKILL shutdown. Currently only perform SIGTERM.
         try {
             if (!launchedTasks.containsKey(taskId)) {
-                LOGGER.info("Unknown TaskId = {}. Cannot destroy something that is unknown.", taskId);
+                LOGGER.error("Unable to kill unknown TaskID: {}", TextFormat.shortDebugString(taskId));
```
mrbrowning: Going to miss the philosophical implications of the original phrasing, but this is probably for the best. 🤔

sdk/executor/src/main/java/com/mesosphere/sdk/executor/CustomExecutor.java
```diff
@@ -192,61 +189,20 @@ public void shutdown(ExecutorDriver driver) {
         for (Map.Entry<Protos.TaskID, LaunchedTask> entry : launchedTasks.entrySet()) {
             final Protos.TaskID taskId = entry.getKey();
             try {
-                LOGGER.info("Stopping task as part of executor shutdown: {}", taskId);
+                LOGGER.info("Stopping task as part of executor shutdown: {}", TextFormat.shortDebugString(taskId));
```
mrbrowning: Maybe `taskId.getValue()`? Same info but without unnecessary brackets and field names, I assume.

sdk/executor/src/main/java/com/mesosphere/sdk/executor/CustomExecutor.java
```diff
@@ -192,61 +189,20 @@ public void shutdown(ExecutorDriver driver) {
...
                 killTask(driver, taskId);
             } catch (Throwable t) {
-                LOGGER.error("Error killing task {}. Reason: {}", taskId, t);
+                LOGGER.error(String.format("Error killing task {}", TextFormat.shortDebugString(taskId)), t);
```
mrbrowning: Same as above re formatting.

sdk/executor/src/test/java/com/mesosphere/sdk/executor/CustomExecutorTest.java
```diff
@@ -48,7 +48,7 @@ public void testSimple() {
                         .setEnvironment(Protos.Environment
                                 .newBuilder()
                                 .addVariables(
-                                        TaskTestUtils.createEnvironmentVariable(DcosTaskConstants.TASK_TYPE, "TEST")))
+                                        TaskTestUtils.createEnvironmentVariable("TASK_TYPE", "TEST")))
```
mrbrowning: One benefit to leaving this as a constant, even though it leads to redundant typing, is that it can be changed in one place only if it needs to be changed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerErrorCode.java
```diff
@@ -3,14 +3,26 @@
...
-    SUCCESS,
-    INITIALIZATION_FAILURE,
-    REGISTRATION_FAILURE,
-    RE_REGISTRATION,
-    OFFER_RESCINDED,
-    DISCONNECTED,
-    ERROR,
-    PLAN_CREATE_FAILURE,
-    LOCK_UNAVAILABLE
+public class SchedulerErrorCode {
```
mrbrowning: Why move from enum?

nickbp: Avoids surprise instability when values are added/removed/reordered. A similar use of enums in cassandra led to bugs.

In this case I was removing the REGISTRATION_FAILURE enum as it's no longer being used, and I didn't want that to cause other values to change their ordinals.

sdk/executor/src/test/java/com/mesosphere/sdk/executor/CustomExecutorTest.java
nickbp: Switched back to named variables.

sdk/executor/src/main/java/com/mesosphere/sdk/executor/CustomExecutor.java
nickbp: Switched to taskId.getValue()

sdk/executor/src/main/java/com/mesosphere/sdk/executor/CustomExecutor.java
nickbp: Switched to taskId.getValue()

sdk/executor/src/main/java/com/mesosphere/sdk/executor/CustomExecutor.java
nickbp: Switched to slaveInfo.getId().getValue()

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerErrorCode.java
mrbrowning: Ah, sounds good.

PR #490, opened 2017-01-27
Raise an error if any field has duplicates in the yaml

nickbp:
- Specified 'memory' or 'count' twice? That's an error
- Also an explicit check disallowing both 'volume' and 'volumes' at once

### DIFF
--- a/sdk/scheduler/src/test/resources/invalid-pod-name.yml
+++ b/sdk/scheduler/src/test/resources/invalid-pod-name.yml
@@ -6,7 +6,7 @@ pods:
     resource-sets:
       meta-data-resources:
         cpus: 0.1
-        mem: 512
+        memory: 512
         volume:
           path: "/meta-data-container-path"
           type: ROOT
@@ -22,7 +22,7 @@ pods:
     resource-sets:
       data-store-resources:
         cpus: 0.2
-        mem: 1024
+        memory: 1024
         volume:
           path: "/data-store-container-path"
           type: ROOT
--- a/sdk/scheduler/src/test/resources/invalid-task-name.yml
+++ b/sdk/scheduler/src/test/resources/invalid-task-name.yml
@@ -6,7 +6,7 @@ pods:
     resource-sets:
       meta-data-resources:
         cpus: 0.1
-        mem: 512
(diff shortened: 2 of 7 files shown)


sdk/scheduler/src/test/resources/invalid-pod-name.yml
```diff
@@ -6,7 +6,7 @@ pods:
     resource-sets:
       meta-data-resources:
         cpus: 0.1
-        mem: 512
+        memory: 512
```
mrbrowning: Should this file be called `invalid-resource-name.yml`?

sdk/scheduler/src/test/resources/invalid-task-name.yml
```diff
@@ -6,7 +6,7 @@ pods:
     resource-sets:
       meta-data-resources:
         cpus: 0.1
-        mem: 512
+        memory: 512
```
mrbrowning: Similar issue with YAML file name.

sdk/scheduler/src/test/resources/invalid-pod-name.yml
nickbp: (These misspelled fields were being missed in the tests -- they were throwing a parse exception but not the one the tests were meant for)

nickbp: I think it's accurate -- the duplicate 'meta-data' pod name is the issue

sdk/scheduler/src/test/resources/invalid-task-name.yml
nickbp: In this case it's two tasks both named 'meta-data-task'

sdk/scheduler/src/test/resources/invalid-pod-name.yml
gabrielhartmann: Are the changes you made to the tests sufficient to make them fail now if we put in unknown properties?

nickbp: No the tests were just written to check for ANY JsonMappingException:

PR #491, opened 2017-01-28
Refactor evaluation logic to simplify spec creation and evaluator testing

mrbrowning:
The changes that were made:
- `*Spec`s now know how to create `*Requirement`s which know how to create `*EvaluationStages`. This is more approachable to the developer and avoids the exposure of evaluation logic at the spec level.
- The old `evaluate(OfferRequirement, List<Offer>)` interface has been restored in `OfferEvaluator` to account for this, and all tests of the evaluation logic now do so against fabricated `OfferRequirement`s, making testing of evaluation significantly easier in light of the fact that existing requirements can be simulated without modifying the state store.
- `PortSpec`, `PortRequirement`, and `PortEvaluationStage` all have a parent `Ports*` version that holds a collection of those children and aggregates their output where appropriate. This simplifies some accounting of ports, since we no longer have to clear the ports on a `ResourceRequirement` and thereby lose record of what ports existed in the case of an existing requirement.
- `OfferRequirement` is immutable again, and the `TaskInfo` and `ExecutorInfo` builders that are modified in the process of performing offer evaluation are managed by a `PodInfoBuilder` class that is now the only other argument to `EvaluationStage.evaluate()` after `MesosResourcePool`.
- `EvaluationOutcome` now also contains the `OfferRecommendation`, if any, generated by the evaluation stage returning it. These are aggregated as required by `OfferEvaluator`.


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -51,11 +52,116 @@ public OfferRequirement getNewOfferRequirement(PodInstance podInstance, Collecti
...
+                continue;
+            }
+
+            if (!usedResourceSets.contains(taskSpec.getResourceSet().getId())) {
+                LOGGER.info("Generating taskInfo to launch for: {}, with resource set: {}",
+                        taskSpec.getName(), taskSpec.getResourceSet().getId());
+                usedResourceSets.add(taskSpec.getResourceSet().getId());
+                taskRequirements.add(getNewTaskRequirement(
+                        podInstance, taskSpec, serviceName, targetConfigurationId, false));
+            }
```
mohitsoni: Can you also please add a test, if it's not already there ?

mrbrowning: The DefaultOfferRequirementProvider tests already check for this behavior, I believe -- I just adapted it to return TaskRequirements instead of TaskInfos, but it was already existing.

PR #499, opened 2017-02-01
Makes Port config updates work

mohitsoni:
With this PR following port config updates are supported:
* Static to Static
* Static to Dynamic
* Dynamic to Static
* ~Dynamic to Dynamic~


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -123,9 +123,17 @@ public static Resource getExpectedRootVolume(
     }
 
     public static Resource getDesiredResource(String role, String principal, String name, Value value) {
+        return getDesiredResource(role, principal, name, value, "");
+    }
+
+    public static Resource getDesiredResource(String role,
```
mrbrowning: Should the naming here change? I figure `Desired` necessarily implies that `resourceId == ""`, otherwise it's expected, since we know the `resourceId` we want.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -63,40 +66,49 @@ public EvaluationOutcome evaluate(
...
+        final ResourceRequirement resourceRequirement = getResourceRequirement();
+        final String resourceId = resourceRequirement.getResourceId();
 
-        final MesosResource mesosResource = mesosResourceOptional.get();
-        Resource fulfilledResource = getFulfilledResource(mesosResource);
+        Resource fulfilledResource = getFulfilledResource(new MesosResource(resourceRequirement.getResource()));
         if (resourceRequirement.expectsResource()) {
             logger.info("Expects Resource");
 
+            Optional<MesosResource> existingResourceOptional = mesosResourcePool.getReservedResourceById(resourceId);
```
mrbrowning: It wouldn't work to fold this into the `consume` call below? I thought `MesosResourcePool` already handles this logic internally.

sdk/scheduler/src/test/resources/static-port.yml
```diff
@@ -0,0 +1,13 @@
...
+    count: 1
+    tasks:
+      test-task:
+        goal: RUNNING
+        cmd: "./task-cmd"
+        ports:
+          test-port-name:
+            port: 555
+        env:
+          PORT_TEST_PORT_NAME: 555
```
mrbrowning: Nit, missing newline.

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorTest.java
```diff
@@ -23,6 +23,107 @@
...
+        Resource offeredReservedResource = ResourceTestUtils.getExpectedRanges("ports", 555, 555, resourceId);
+        Resource offeredUnReservedResource = Resource.newBuilder(ResourceTestUtils.getDesiredRanges("ports", 666, 666))
+                .clearReservation().build();
+
+        List<Resource> offeredResources = Arrays.asList(offeredReservedResource, offeredUnReservedResource);
+        List<OfferRecommendation> recommendations = evaluator.evaluate(
+                getExistingPodInstanceRequirement(updatedResource, false, "static-port.yml"),
+                Arrays.asList(getOffer(offeredResources)));
+        // UNRESERVE, RESERVE, LAUNCH
+        Assert.assertEquals(3, recommendations.size());
```
mrbrowning: Also verify that the first two operations are `RESERVE` and `UNRESERVE`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
gabrielhartmann: Agree.  This should be `getExpectedResource`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
gabrielhartmann: `consume` gives you back what you asked for.  This is checking what exists already, so a determination can be made about whether we need to modify the reservation.  This is the heart of the bug.  For ports it was impossible to know solely from the resource returned by `consume` whether the reservation required modification.

e.g. I want to change from port 5 to 6.  `consume` will return `6`.  That's what I asked for!  Do I need to reserve this?  Unknown based on just the `consume` call, hence the change.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -198,7 +217,7 @@ protected Resource getFulfilledResource(MesosResource mesosResource) {
         }
     }
 
-    private boolean expectedValueChanged(MesosResource mesosResource) {
-        return !ValueUtils.equal(getResourceRequirement().getValue(), mesosResource.getValue());
+    private boolean expectedValueChanged(Value requestedResource, Value existingResource) {
```
gabrielhartmann: I'm not convinced this function needs to exist.  Its kind of helping by changing the name of the function it calls.  This could be equally well accomplished by a comment.

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorTest.java
```diff
@@ -23,6 +23,107 @@
...
+        Resource updatedResource = ResourceTestUtils.getExpectedRanges("ports", 666, 666, resourceId);
+        Resource offeredReservedResource = ResourceTestUtils.getExpectedRanges("ports", 555, 555, resourceId);
+        Resource offeredUnReservedResource = Resource.newBuilder(ResourceTestUtils.getDesiredRanges("ports", 666, 666))
+                .clearReservation().build();
+
+        List<OfferRecommendation> recommendations = evaluator.evaluate(
+                getExistingPodInstanceRequirement(updatedResource, false, "dynamic-port.yml"),
+                Arrays.asList(getOffer(Arrays.asList(offeredReservedResource, offeredUnReservedResource))));
+        // UNRESERVE, RESERVE, LAUNCH
+        Assert.assertEquals(3, recommendations.size());
```
gabrielhartmann: Same as above.  Validate at least types of operations.

mohitsoni:
@mrbrowning @gabrielhartmann I have addressed the review comments.

PR #505, opened 2017-02-02
Enable the propagation of additional envvars to tasks at plan start

mrbrowning:
This change allows plans to keep parameters, which are subject to change over their lifetimes, and which are propagated to all `Phase`s and `Step`s belong to them. It is ultimately up to the `Step` instances at the leaves of a `Plan` what it wants to do with those parameters -- `DeploymentStep` will make sure to incorporate them as environment variables into the `PodInstanceRequirement` returned by `start()`, for example (`PodInstanceRequirement` now has a `environment` member, which extends the environments of all tasks contained by it). This allows sidecar plans to have parameters that are only supplied when the `{plan}/start` endpoint is invoked (via JSON in the POST request body), which is useful for e.g. backup operations that may have a changing or unknown-at-install-time backup location.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PlansResource.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PlansResource.java
@@ -2,6 +2,8 @@ package com.mesosphere.sdk.api;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.mesosphere.sdk.api.types.PlanInfo;
+import com.mesosphere.sdk.offer.evaluate.placement.RegexMatcher;
+import com.mesosphere.sdk.offer.evaluate.placement.StringMatcher;
 import com.mesosphere.sdk.scheduler.plan.*;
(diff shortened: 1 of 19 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PlansResource.java
```diff
@@ -22,6 +24,7 @@
     static final Response PLAN_ELEMENT_NOT_FOUND_RESPONSE = Response.status(Response.Status.NOT_FOUND)
             .entity("Element not found")
             .build();
+    private static final StringMatcher ENVVAR_MATCHER = RegexMatcher.create("[A-Za-z_][A-Za-z0-9_]*");
```
mrbrowning: I know we already have this somewhere, but a cursory search didn't reveal it, so...#lazyweb

mrbrowning:
@gabrielhartmann had to rebase, but I added a parameterized sidecar integration test, and fixed up the unit tests.

PR #512, opened 2017-02-05
Define a single way to define custom Steps for Phases

gabrielhartmann:
There used to be two ways to define custom plans.  One way allowed defining the same set of tasks be launched for every instance a pod type.  Another allowed specification of entirely custom tasks to be launched for each instance of a pod.

These two approaches have been unified and extended such that custom definitions for individual pod instances are still possible while allowing for the definition of `default` steps for every other pod instance.  See `custom-phases.yml` for an exhaustive list of the possible examples.  I've included the plan selection below for ease of review:

```yaml
plans:
  deploy:
    strategy: serial
    phases:
      server-deploy:
        strategy: parallel
        pod: hello
        steps:
          - default: [[server]]
      once-deploy:
        strategy: parallel
        pod: hello
        steps:
          - default: [[once]]
      interleave-deploy:
        strategy: serial
        pod: hello
        steps:
          - default: [[once], [server]]
      full-custom-deploy:
        strategy: serial
        pod: hello
        steps:
          - 0: [[once], [server]]
          - 1: [[server], [once]]
      partial-custom-deploy:
        strategy: serial
        pod: hello
        steps:
          - 1: [[once], [server]]
          - default: [[server], [once]]
```


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPlanGenerator.java
```diff
@@ -56,31 +55,62 @@ protected Phase from(RawPhase rawPhase, String phaseName, Collection<PodSpec> po
...
+
+            // Convert from map to list
+            Map<String, List<List<String>>> validatedSteps =
+                    rawSteps.stream()
+                            .map(stepMap -> stepMap.entrySet().stream().findFirst().get())
+                            .collect(Collectors.toMap(
+                                    stringListEntry -> stringListEntry.getKey(),
+                                    stringListEntry -> stringListEntry.getValue()));
+
+            for (int i = 0; i < podSpec.getCount(); i++) {
```
mrbrowning: Nit, prefer pre-increment (`++i`) always for consistency.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPlanGenerator.java
```diff
@@ -56,31 +55,62 @@ protected Phase from(RawPhase rawPhase, String phaseName, Collection<PodSpec> po
...
+                    taskLists = validatedSteps.get("default");
+
+                    if (taskLists == null) {
+                        // Fall back to default behavior
+                        List<String> taskNames = podSpec.getTasks().stream()
+                                .map(taskSpec -> taskSpec.getName())
+                                .collect(Collectors.toList());
+                        steps.add(from(new DefaultPodInstance(podSpec, i), taskNames));
+                    } else {
+                        // Use wildcard defined behavior (e.g. * : [[foo, bar], [bazz]])
```
mrbrowning: I don't understand the comment? That's not actual yaml you would see, right?

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultPlanGeneratorTest.java
```diff
@@ -66,57 +68,61 @@ public void testFullManualPlan() throws Exception {
...
+                            Arrays.asList("server"),
+                            Arrays.asList("once")));
+
+            validatePhase(
+                    partialCustomPhase,
+                    Arrays.asList(
+                            Arrays.asList("server"),
+                            Arrays.asList("once"),
+                            Arrays.asList("once"),
+                            Arrays.asList("server")));
```
mrbrowning: Maybe bump the pod size for this one, so that we can validate that the `default` step holds for all steps after the first?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPlanGenerator.java
gabrielhartmann: My mistake.  Tried to use `*` instead of `default` initially.  I'll fix the comment.

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultPlanGeneratorTest.java
gabrielhartmann: Done.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPlanGenerator.java
gabrielhartmann: Done.

PR #513, opened 2017-02-06
Standardize and provide types to code snippets

gabrielhartmann:
@sascala:  Any reason we shouldn't move to this style of code snippets.  I think the style we had up until now was an result of the conversion tool we used to move from Google docs right?


sascala:
No, there's no reason not to except I was able to get the document up faster if I didn't change it. You're right that it's better to have the type specified.

PR #522, opened 2017-02-07
Add support for joining overlay networks with CNI

mrbrowning:
This change adds a new field under `container` in each pod called `networks` that allows the user to specify which overlay networks to join by name:
```
container:
  networks:
    dcos: {}
```
An empty dictionary must be passed for the body right now, due to the fact that we only currently support joining networks by name, but was designed this way to allow backwards-compatible addition of e.g. port mappings in the future.

This was tested by creating a container that joins the `dcos` virtual network included by default in DC/OS installs and verifying that the IP for interface eth0 is in the 9.0.0.0/24 subnet.

### DIFF
--- a/docs/pages/dev-guide/developer-guide.md
+++ b/docs/pages/dev-guide/developer-guide.md
@@ -798,7 +798,7 @@ The most basic set of features present in the YAML representation of the `Servic
 
 ### Containers
 
-Each pod runs inside a single container. The `ServiceSpec` specifies the Docker image to run for that container and the POSIX resource limits for every task that runs inside that container. In the example below, the soft limit for number of open file descriptors for any task in the "hello" pod is set to 1024, and the hard limit to 2048:
+Each pod runs inside a single container. The `ServiceSpec` specifies the Docker image to run for that container, the virtual network memberships, and the POSIX resource limits for every task that runs inside that container. In the example below, the soft limit for number of open file descriptors for any task in the "hello" pod is set to 1024, and the hard limit to 2048:
 
 ```yaml
 name: "hello-world"
@@ -807,6 +807,8 @@ pods:
     count: 1
     container:
       image-name: ubuntu
+      networks:
+        dcos: {}
       rlimits:
         RLIMIT_NOFILE:
           soft: 1024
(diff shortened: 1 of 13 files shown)


mrbrowning:

mrbrowning:

docs/pages/dev-guide/developer-guide.md
```diff
@@ -819,6 +821,8 @@ pods:
         memory: 256
 ```
 
+Currently an empty YAML dictionary is passed as the body for each network definition under `networks`, since we only support joining virtual networks by name, but in the future it will be possible to specify port mappings and other information in a network definition.
```
mgummelt: We really need some equivalent of a json schema for YAML so we can normalize this documentation.

frameworks/helloworld/tests/test_networking.py
```diff
@@ -0,0 +1,37 @@
...
+            "spec_file": "examples/cni.yml"
+        }
+    }
+
+    install.install(PACKAGE_NAME, 1, additional_options=options)
+    print('done installing...')
+
+
+@pytest.mark.sanity
+def test_deploy():
```
mgummelt: Can you add a comment explaining what this is testing?

frameworks/helloworld/tests/test_networking.py
```diff
@@ -0,0 +1,37 @@
...
+    deployment_plan = plan.get_deployment_plan(PACKAGE_NAME).json()
+    print("deployment_plan: " + str(deployment_plan))
+
+    assert(len(deployment_plan['phases']) == 1)
+    assert(deployment_plan['phases'][0]['name'] == 'all-deploy')
+    assert(len(deployment_plan['phases'][0]['steps']) == 2)
+
+
+@pytest.mark.sanity
+def test_joins_overlay_network():
```
mgummelt: This doesn't really test that the containers have joined the overlay network, right?  Is there an easy way to test this?

mrbrowning: It does. Unfortunately the actual logic is in the YAML definition itself (see `cni.yml` in this PR), since there's no real sensible way to get information about whether a property holds for a task's environment (such as "is my IP address what I expect?") other than constructing the command so that it fails if that property is false. In this case, the command greps for `9\.[0-9\.]+` in the output of `ip addr show`. If the plan makes it to `COMPLETE`, the task finished and all is well.

docs/pages/dev-guide/developer-guide.md
mrbrowning: I've been thinking about this, especially because there's a lot of static validation of the YAML we could do (beyond just whether all required fields are present and their types are correct) to save developers the pain of having to go through a full stub universe deployment just to find out they made a trivial error in their spec.

frameworks/helloworld/tests/test_networking.py
mgummelt: oooh I missed that in the svc.yml

PR #526, opened 2017-02-08
custom resource endpoints. review: 

mbdcos:
taskKiller (from defaultScheduler)
stateStore (from builder)

I am adding here only two custom resources: BrokerController and InterruptProceed. InterruptProceed's functionality can be achieved with the default PlanResource. 

BrokerController is application specific, it lists all brokers and also provides replace interface (restart task). For restart we need to access to the taskKiller and stateStore.


mbdcos:
Rest will be added after we decide on getter methods in DefaultScheduler.

mbdcos:
see PR 527. I prefer 527

PR #529, opened 2017-02-08
Update tasks sharing resource-sets on launch

gabrielhartmann:


mrbrowning:
LGTM, what's the context? Do tasks sharing resource sets get out of sync in the state store currently?

PR #532, opened 2017-02-09
Allow tasks to override mesos-DNS name with explicit prefix

mrbrowning:
This change allows individual tasks in a pod to specify their own DNS names, with the result being that their Mesos-DNS address will be `{dns-name}.{framework-name}.mesos` instead of the usual `{task-name}.{framework-name}.mesos`. Two tasks that belong to different pods are not allowed to share the same DNS name, and a step may not simultaneously deploy two tasks in the same pod that share the same DNS name.


mrbrowning:

frameworks/helloworld/tests/test_dns.py
```diff
@@ -0,0 +1,32 @@
...
+
+def setup_module(module):
+    install.uninstall(PACKAGE_NAME)
+    options = {
+        "service": {
+            "spec_file": "examples/dns.yml"
+        }
+    }
+
+    # this yml has 2 hello's + 0 world's:
```
gabrielhartmann: Why note the `0 worlds`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -338,6 +342,13 @@ private static String getConfigTemplateDownloadPath(ConfigFileSpec config) {
         return String.format("%s%s", CONFIG_TEMPLATE_DOWNLOAD_PATH, config.getName());
     }
 
+    private static Protos.DiscoveryInfo getDiscoveryInfo(String dnsName) {
+        return Protos.DiscoveryInfo.newBuilder()
+                .setVisibility(Protos.DiscoveryInfo.Visibility.CLUSTER)
```
gabrielhartmann: This is fine.  I wonder whether we'd ever want to expose this as configurable.  This would change the YAML format right?

frameworks/helloworld/tests/test_dns.py
mrbrowning: Cargo-culted from another test, but I figured it was there to indicate why the total number of tasks expected was 2. Will remove.

mrbrowning:

frameworks/helloworld/tests/test_dns.py
```diff
@@ -0,0 +1,32 @@
...
+    install.install(PACKAGE_NAME, 2, additional_options=options)
+
+
+@pytest.mark.sanity
+def test_discovery_is_set():
+    pod_info = dcos.http.get(
+        shakedown.dcos_service_url(PACKAGE_NAME) +
+        "/v1/pods/{}/info".format("hello-0")).json()
+
+    assert(all(p["info"]["discovery"]["name"] == "hello" for p in pod_info))
```
gabrielhartmann: @mrbrowning: My apologies, should have brought this up before.  Can we add a test or extend this test, which does a `dig` or something and actually validates that `init` is reachable at `hello-0.hello-world.mesos`?

mrbrowning:
@gabrielhartmann Updated tests to check for the existence of the expected A record.

frameworks/helloworld/tests/test_discovery.yml
```diff
@@ -0,0 +1,35 @@
+import dcos
```
mgummelt: not a yml file

PR #535, opened 2017-02-09
[HDFS] Recover permanently failed name nodes

gabrielhartmann:


mohitsoni:
@gabrielhartmann Can you please resolve conflicts on this one ?

gabrielhartmann:
@mohitsoni:  Thanks for having a look, should be ready for review now.

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -242,3 +242,21 @@ plans:
       dn-deploy:
         strategy: {{DATA_STRATEGY}}
         pod: data
+
+  replace-nn-0:
```
mohitsoni: 👍

mrbrowning: Since we have to write custom Java for this anyway, couldn't it construct two separate `Plan` instances from the same raw YAML, so there could just be a single `replace-nn` plan here?

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
```diff
@@ -102,6 +103,11 @@ private static ServiceSpec serviceSpecWithCustomizedPods(RawServiceSpec rawServi
 
         return DefaultServiceSpec.newBuilder(serviceSpec)
                 .pods(Arrays.asList(journal, name, zkfc, data))
+                .replacementFailurePolicy(
+                        ReplacementFailurePolicy.newBuilder()
+                                .permanentFailureTimoutMs(null)
```
mohitsoni: I understand why you have used a `null` here, but in general passing `null` around is bad. In this particular case, it will work out just fine. But, we should avoid doing this. For example, we can have `ReplacementFailurePolicy.disablePermanentFailureReplacement()` and `ReplacementFailurePolicy.enablePermanentFailureReplacement(). permanentFailureTimoutMs(1234)`.

Also, `permanentFailureTimoutMs` maps to `permanentFailureTimoutMins` internally, something that should also be addressed in a separate PR.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsRecoveryPlanManager.java
```diff
@@ -0,0 +1,167 @@
...
+        return new DefaultPlan(
+                RECOVERY_ELEMENT_NAME,
+                phases,
+                new ParallelStrategy<>(),
+                Collections.emptyList());
+    }
+
+    @Override
+    public Collection<? extends Step> getCandidates(Collection<String> dirtyAssets) {
+        // Always allow the default recovery manager to update its plan.
```
mrbrowning: This comment feels like it was the result of some pain, if so, maybe clarify what the commented line prevents?

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
gabrielhartmann: @mohitsoni:  Thanks, this is an excellent point.  I'll file a ticket.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsRecoveryPlanManager.java
```diff
@@ -0,0 +1,167 @@
...
+                .collect(Collectors.toList());
+    }
+
+    private void setNameNodePhase(Phase phase) {
+        this.nameNodeRecoveryPhase = phase;
+    }
+
+    private Phase initNNRecoveryPhase(Plan inputPlan) {
+        Phase inputPhase = inputPlan.getChildren().get(0);
+        Step inputBootStrapStep = inputPhase.getChildren().get(0);
```
mrbrowning: NIT: BootStrap -> Bootstrap

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsRecoveryPlanManager.java
```diff
@@ -0,0 +1,167 @@
...
+        Step bootstrapStep =
+                new DefaultRecoveryStep(
+                        inputBootStrapStep.getName(),
+                        Status.PENDING,
+                        inputBootStrapStep.start().get().getPodInstance(),
```
mrbrowning: I might be missing something, but how much is the YAML really doing if we need to recapitulate its structure here?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -114,6 +103,7 @@
      * such as destroying a failed task.
      */
     protected final int destructiveRecoveryDelayMs;
+    private final Optional<RecoveryPlanManagerFactory> recoveryPlanManagerFactoryOptional;
```
mrbrowning: This doesn't look like an HDFS-only change :)

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
gabrielhartmann: https://mesosphere.atlassian.net/browse/INFINITY-967

frameworks/hdfs/src/main/dist/svc.yml
gabrielhartmann: Done.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsRecoveryPlanManager.java
gabrielhartmann: Done.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsRecoveryPlanManager.java
gabrielhartmann: We're not recapitulating its structure.  Just converting from Deploy to Recovery steps.  Much easier than constructing plan from Java entirely.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: It is not.  I was going to break it out, but I thought this PR provided some context, and proof that it works etc.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsRecoveryPlanManager.java
mrbrowning: 👍

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsRecoveryPlanManager.java
gabrielhartmann: Done.

mbdcos:
LGTM

PR #555, opened 2017-02-10
[CASSANDRA] Prototype framework

loren:


loren:

frameworks/cassandra/src/main/dist/cassandra.yaml
```diff
@@ -0,0 +1,100 @@
...
+commitlog_sync: periodic
+index_summary_capacity_in_mb:
+trickle_fsync: false
+trickle_fsync_interval_in_kb: 10240
+rpc_server_type: sync
+incremental_backups: false
+snapshot_before_compaction: false
+auto_snapshot: true
+cross_node_timeout: false
+endpoint_snitch: SimpleSnitch
```
verma7: We need to support `GossipingPropertyFileSnitch` and expose the DC and rack name in a configuration file passed to Cassandra.

frameworks/cassandra/src/main/dist/cassandra.yaml
```diff
@@ -0,0 +1,100 @@
...
+listen_address: "{{TASK_NAME}}.{{FRAMEWORK_NAME}}.mesos"
+hints_directory: {{MESOS_SANDBOX}}/container-path/hints
+data_file_directories:
+  - {{MESOS_SANDBOX}}/container-path/data
+commitlog_directory: {{MESOS_SANDBOX}}/container-path/commitlog
+saved_caches_directory: {{MESOS_SANDBOX}}/container-path/saved_caches
+seed_provider:
+    - class_name: org.apache.cassandra.locator.SimpleSeedProvider
+      parameters:
+          - seeds: "node-0-server.{{FRAMEWORK_NAME}}.mesos,node-1-server.{{FRAMEWORK_NAME}}.mesos"
```
verma7: We need to support seed discovery from a different datacenter and periodically synchronize them.

frameworks/cassandra/src/main/dist/cassandra_service.yml
```diff
@@ -0,0 +1,103 @@
...
+        goal: RUNNING
+        resource-set: server-resources
+        cmd: ./bootstrap && exec ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/cassandra -f
+        configs:
+          cassandra:
+            template: {{CONFIG_TEMPLATE_PATH}}/cassandra.yaml
+            dest: apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra.yaml
+          jvm:
+            template: {{CONFIG_TEMPLATE_PATH}}/jvm_{{CASSANDRA_HEAP_GC}}.options
+            dest: apache-cassandra-{{CASSANDRA_VERSION}}/conf/jvm.options
```
verma7: Please add conf/cassandra-rackdc.properties

frameworks/cassandra/src/main/dist/cassandra_service.yml
```diff
@@ -0,0 +1,103 @@
...
+        goal: FINISHED
+        cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool repair -pr -p {{TASKCFG_ALL_JMX_PORT}} -- $CASSANDRA_KEYSPACE_TABLES
+        resource-set: sidecar-resources
+      cleanup:
+        goal: FINISHED
+        cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool cleanup -p {{TASKCFG_ALL_JMX_PORT}} -- $CASSANDRA_KEYSPACE_TABLES
+        resource-set: sidecar-resources
+      snapshot:
+        goal: FINISHED
+        cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool snapshot -p {{TASKCFG_ALL_JMX_PORT}} -t $CASSANDRA_SNAPSHOT_NAME -- $CASSANDRA_KEYSPACES
```
verma7: I am not sure I understand how $CASSANDRA_KEYSPACES gets filled in. Does this happen at scheduler startup? Shouldn't this be different for every HTTP request to snapshot?

loren: From the CLI, like this: `dcos cassandra plan start snapshot CASSANDRA_KEYSPACES="demo",CASSANDRA_SNAPSHOT_NAME="20170210"`

frameworks/elastic/tests/config.py
```diff
@@ -55,7 +55,7 @@ def expected_nodes_success_predicate():
     if result is None:
         return False
     node_count = result["number_of_nodes"]
-    print('Waiting for {} healthy nodes, got {}'.format(node_count, DEFAULT_NODE_COUNT))
+    print('Waiting for {} healthy nodes, got {}'.format(DEFAULT_NODE_COUNT, node_count))
```
verma7: Should the following changes to elastic be a part of this PR?

mohitsoni:
@loren Have you considered adding readiness checks for Cassandra nodes, so that we only move forward with the installation when the current node JOINS the ring. The way it is currently right now, things will eventually work but will cause unnecessary noise as @verma7 points out.

loren:

gabrielhartmann:
As @verma7 @mohitsoni, have said, a readiness check will be needed to add more than one C* node at a time.

@verma7:  Please create issues for the features you need before this framework can be considered to have reached feature parity.  In particular I am interested in detailed reasons why node replacement **requires** the use of the `replace_address` instead of a replace being the removal of a node from the cluster and the addition of a new node (a policy also not implemented yet).

Finally, all the above are good valid things to work on, but I don't think they should block a PR clearly labeled as "Prototype".  A fully featured solution should not be the goal of this PR.

gabrielhartmann:
retest this please

frameworks/elastic/tests/config.py
gabrielhartmann: Good catch.  This shouldn't be here.

frameworks/cassandra/src/main/dist/cassandra_service.yml
```diff
@@ -0,0 +1,103 @@
+name: {{FRAMEWORK_NAME}}
```
gabrielhartmann: Can you rename this file to `svc.yml` please.  So far all services use this as the default name for their ServiceSpec YAML files.

loren:
@verma7 I just added the readiness check as part of this PR and verified that the task for node N+1 does not transition from PENDING to STARTING until node N is COMPLETE, based on nodetool reporting it as "UN".

PR #558, opened 2017-02-12
Add read-only access to the StateStore

gabrielhartmann:
Components (Tasks) other than the scheduler may wish to understand the state of a given service without having access to modifying service state.  The addition of a read-only StateStore interfaces allows for this.

Example uses include:
* Cassandra seed provider
* Cross-service locality

This PR looks like it has many changes when in fact it is a pure lift of methods from `StateStore` to `StateStoreReadOnly` in both interfaces and implementation.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
@@ -5,14 +5,11 @@ import com.mesosphere.sdk.dcos.DcosConstants;
 import com.mesosphere.sdk.offer.CommonTaskUtils;
 import com.mesosphere.sdk.offer.TaskException;
 import com.mesosphere.sdk.state.*;
-import com.mesosphere.sdk.storage.Persister;
 import com.mesosphere.sdk.storage.StorageError.Reason;
 
 import org.apache.curator.RetryPolicy;
 import org.apache.mesos.Protos;
 import org.apache.zookeeper.KeeperException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.*;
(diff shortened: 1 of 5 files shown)


gabrielhartmann:
retest this please

gabrielhartmann:
retest this please

mrbrowning:
The code seems fine, but rather than deepen the hierarchy now rooted at `StateStoreReadOnly`, what I'd propose is to take a page from the Java collections book and retain just a single interface -- `StateStore` -- with different implementors for the read-only and read/write cases (and maybe, in deference to the Rule of Least Power, with names that make mutability sound like the exceptional case, e.g. `CuratorStateStore` and `MutableCuratorStateStore`), and where the read-only variant throws `java.lang.UnsupportedOperationException` for any write calls. This retains the same level of code reuse (the mutable concrete implementation can extend the read-only implementation) with the only downside being that the compiler won't catch some contract violations until runtime, but I foresee that being rare in practice and worth the simplification of the type hierarchy here.

nickbp:
I'd lean towards the type-aware enforcement, but how about this structure?:

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -39,27 +36,7 @@
...
-    private static final String FWK_ID_PATH_NAME = "FrameworkID";
-    private static final String PROPERTIES_PATH_NAME = "Properties";
-    private static final String TASKS_ROOT_NAME = "Tasks";
-    private static final String SUPPRESSED_KEY = "suppressed";
-
-    private final Persister curator;
-    private final TaskPathMapper taskPathMapper;
-    private final String fwkIdPath;
-    private final String propertiesPath;
+public class CuratorStateStore extends CuratorStateStoreReadOnly implements StateStore {
```
nickbp: `extends CuratorStateStoreReadOnly` should be enough here?

nickbp: nevermind see other comment

gabrielhartmann:
@nickbp: Thought about the casting approach, but wanted to enforce things such that one couldn't escape from readonly mode to writable mode.  I guess this isn't that big a deal.  If a developer wants to break stuff with code, there are easier ways to do it.

@mrbrowning: I agree with @nickbp, I prefer types.  Throwing exceptions on unsupported operations seems weird.  Why expose what can't be used?

mrbrowning:
I think the nature of Java's type system is such that all of the available options have glaring drawbacks. We know the drawbacks of throwing in unsupported methods, but as for the approach here: inheritance implies that the subclass satisfies at least all of the invariants of the superclass; so we can pass a mutable `StateStore` to any site expecting a read-only `StateStore`. The response is, of course, "don't mutate the state store in read methods," but that's exactly the same logic underlying "don't call write methods on a read-only state store" except that the invariant is broken silently, in a much less obvious way, and maybe much later! So as far as this dilemma goes, the type system is mostly pageantry no matter how you slice it in terms of what behavior it guarantees. I'd much rather runtime checks that fail aggressively and loudly than weak contracts that are violated behind closed doors.

nickbp:
Having two separate state store implementations separately accessing the same underlying storage will make caching semantics more complicated. At the moment there's a thread-safe singleton StateStoreCache wrapping a singleton CuratorStateStore.

PR #560, opened 2017-02-12
[ELASTIC] Updated README quick start

loren:


frameworks/elastic/README.md
```diff
@@ -26,13 +26,19 @@ DC/OS Elasticsearch provides the following features:
...
 
-1. Install an Elasticsearch cluster and log on.
+1. Install an Elasticsearch cluster with Kibana and log on to the Mesos master node.
 
   ```bash
   dcos package install --app elastic
   dcos node ssh --master-proxy --leader
   ```
 
+1. Wait until the cluster is deployed and the nodes are all running. This may take 5-10 minutes. If you try to hit the cluster too soon, you may get an empty response or an authentication error like this:
```
gabrielhartmann: `hit` --> `reach` or `access`

frameworks/elastic/README.md
```diff
@@ -26,13 +26,19 @@ DC/OS Elasticsearch provides the following features:
...
 
-1. Install an Elasticsearch cluster and log on.
+1. Install an Elasticsearch cluster with Kibana and log on to the Mesos master node.
 
   ```bash
   dcos package install --app elastic
   dcos node ssh --master-proxy --leader
   ```
 
+1. Wait until the cluster is deployed and the nodes are all running. This may take 5-10 minutes. If you try to access the cluster too soon, you may get an empty response or an authentication error like this:
```
gabrielhartmann: Might want to point to `service/elastic/v1/plans/deploy` which when `COMPLETE` indicates the cluster is fully deployed.  There should also be a CLI command `dcos elastic plan show` which will show the same information.  This way, people can know when the deployment is actually one, instead of guessing.

gabrielhartmann: In fact, plan completeness probably doesn't signal that Kibana is ready to go right now.  It looks like the health-check should be changed to a readiness check or replicated as a readiness check, as it's testing whether the login endpoint is ready yet.  Killing Kibana if it fails to be ready within the given timeout may or may not be a good idea.

loren:
Incorporated suggestions and wrote things that will be true once https://github.com/mesosphere/dcos-commons/pull/565 merges.

PR #561, opened 2017-02-12
Renamed elastic_service.yml to svc.yml

loren:


gabrielhartmann:
@loren:  Needs conflict resolution

PR #562, opened 2017-02-13
[SDK] Fix config updates of multi-disk pods

gabrielhartmann:
Configuration updates for pods with multiple volumes don't work due to a map key collision on the resource name: `disk`.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
@@ -135,9 +135,22 @@ public class DefaultOfferRequirementProvider implements OfferRequirementProvider
     private static Collection<ResourceRequirement> getResourceRequirements(
             TaskSpec taskSpec, Collection<Protos.Resource> resources) {
         ResourceSet resourceSet = taskSpec.getResourceSet();
+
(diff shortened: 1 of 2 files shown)


gabrielhartmann:
@Gilbert88 You will want to consume this fix.

gabrielhartmann:
@gmflau: You'll want to consume this fix.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -135,17 +135,30 @@ private TaskRequirement getNewTaskRequirement(
...
-                resources.stream().collect(Collectors.toMap(r -> r.getName(), Function.identity()));
+                resources.stream()
+                        .filter(resource -> !resource.hasDisk())
+                        .collect(Collectors.toMap(r -> r.getName(), Function.identity()));
+
+        Map<String, Protos.Resource> volumeMap = resources == null ?
+                Collections.emptyMap() :
+                resources.stream()
+                        .filter(resource -> resource.hasDisk())
+                        .filter(resource -> resource.getDisk().hasVolume())
```
mohitsoni: Nit: I think this can be merged with the previous filter:
`.filter(resource -> resource.hasDisk() && resource.getDisk().hasVolume())`

PR #563, opened 2017-02-13
[SDK] Kill tasks based on resource-sets

gabrielhartmann:
Tasks should be killed based on whether a resource-set is being consumed by a task from their pod already, not based on task name.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultPlanScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultPlanScheduler.java
@@ -4,6 +4,7 @@ import com.google.inject.Inject;
 import com.mesosphere.sdk.offer.*;
 import com.mesosphere.sdk.offer.evaluate.OfferEvaluator;
 import com.mesosphere.sdk.scheduler.TaskKiller;
+import com.mesosphere.sdk.specification.TaskSpec;
 import com.mesosphere.sdk.state.StateStore;
 
 import org.apache.mesos.Protos;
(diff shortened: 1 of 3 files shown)


gabrielhartmann:
@mrbrowning: This is rebased on #532. So that should go in first, so ignore the first two commits please?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultPlanScheduler.java
```diff
@@ -133,26 +134,43 @@ public DefaultPlanScheduler(
...
+                .filter(taskSpec -> resourceSetsToConsume.contains(taskSpec.getResourceSet().getId()))
+                .map(taskSpec -> TaskSpec.getInstanceName(podInstanceRequirement.getPodInstance(), taskSpec))
                 .collect(Collectors.toList());
+        logger.info("Tasks to kill: {}", tasksToKill);
 
-        for (String taskName : taskNames) {
+        for (String taskName : tasksToKill) {
             TaskInfo taskInfo = taskInfoMap.get(taskName);
-            Optional<Protos.TaskStatus> taskStatusOptional = stateStore.fetchStatus(taskInfo.getName());
+            if (taskInfo != null) {
```
mrbrowning: Why not just filter in the Stream sequence above and remove this if?

gabrielhartmann: I wanted the logging.  I wanted to see what the system wanted to kill, pre-filtered.

gabrielhartmann:
@mrbrowning: I will add the following to documentation:

At any given point in time only a single `task` may consume a `resource-set`.  Any attempt to launch a `task` consuming an already consumed `resource-set`.  Will result in the killing of the `task` which is currently running and the launch of the new task.

Clear?

Finally, the scenario addresses is one I ran into when testing manual recovery of HDFS clusters.  If you want to `restart` a `bootstrap` `Step`, you can set it to pending, but it doesn't kill the appropriate running `server` task so that it can run.

gabrielhartmann:
@mrbrowning:  See #566 for the docs change.

PR #565, opened 2017-02-13
Added readiness check to elasticsearch and kibana tasks

loren:
Also increased grace period for kibana, since 10 minutes might not actually be enough for it.


gabrielhartmann:
retest this please

PR #569, opened 2017-02-14
dev-guide: Add instructions for Repoxy

nlsun:
A fallback if proxylite is not compatible with the application.


nlsun:
@mesosphere-ci Retest this please.

nlsun:
I'm not even sure if we need to wait for the tests on this one, it's only a documentation change.

PR #572, opened 2017-02-14
Turn off automated failure recovery by default

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -592,11 +572,18 @@ protected void initializeDeploymentPlanManager() {
...
-        FailureMonitor failureMonitor =
-                permanentFailureTimeoutMs.isPresent()
-                        ? new TimedFailureMonitor(Duration.ofMillis(permanentFailureTimeoutMs.get()))
-                        : new NeverFailureMonitor();
+        LaunchConstrainer launchConstrainer;
+        FailureMonitor failureMonitor;
+        if (failurePolicyOptional.isPresent()) {
+            ReplacementFailurePolicy failurePolicy = failurePolicyOptional.get();
+            launchConstrainer = new TimedLaunchConstrainer(
+                    Duration.ofMillis(failurePolicy.getMinReplaceDelayMins()));
```
nickbp: There was a null check on `getMinReplaceDelayMins` before. Looks like `ofMillis()` expects a (primitive) `long`. null check still needed?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -142,8 +139,8 @@ public String getZookeeperConnection() {
     }
 
     @Override
-    public ReplacementFailurePolicy getReplacementFailurePolicy() {
-        return replacementFailurePolicy;
+    public Optional<ReplacementFailurePolicy> getReplacementFailurePolicy() {
```
nickbp: Just in case, may be fine as-is: Does the constructor need an update as well? (for de/reserialization)

If there isn't one already, there should probably be a spec exercised by the tests that contains a replacement policy, specifically exercised via the reserialize operation in `BaseServiceSpecTest` => `DefaultServiceSpec.getFactory()`

keithchambers: Let's remove/reword this from the main README `Available and Durable` benefits section too. Down the line we can readdress this.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: `null` was being used as a signal to disable automated permanent failure recovery.  No longer necessary.

gabrielhartmann:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
nickbp: :+1:

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/ServiceSpec.java
```diff
@@ -31,5 +32,5 @@
     String getZookeeperConnection();
 
     @JsonProperty("replacement-failure-policy")
-    ReplacementFailurePolicy getReplacementFailurePolicy();
+    Optional<ReplacementFailurePolicy> getReplacementFailurePolicy();
```
nickbp: To clarify from earlier comment -- does this represent a compatibility break in ServiceSpec serialization? I don't remember for sure but an optional may be serialized differently from a nullable.

gabrielhartmann:
retest this please

nickbp:
retest this please (ephemeral build failure)

nickbp:
retest this please

PR #574, opened 2017-02-15
[SDK] Clear permanent failure mark on successful task launch

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -2,7 +2,7 @@
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.TextFormat;
-import com.mesosphere.sdk.scheduler.recovery.RecoveryPlanManagerFactory;
+import com.mesosphere.sdk.scheduler.recovery.*;
```
mrbrowning: Can this move back to qualified imports? It makes code easier to read and the move from qualified imports to unqualified seems to confuse diff algorithms.

nickbp: I think there's some lint check that complains if too many classes from a package are imported separately. I don't remember what the limit is, but that could be why this was switched to a wildcard?

While in here it could make sense to just clean up the import ordering since it looks pretty haphazard at this point.

mrbrowning: IntelliJ also has a configurable limit beyond which it'll move a bunch of imports over to `*`, which I assumed was the culprit here.

nickbp:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: Yeah, it's IntelliJ.  All the imports are managed by it.  I haven't manually imported anything...ever.  In any case I can do whatever anyone likes.

gabrielhartmann:
@mrbrowning: Regarding metadata lifecycle, I think what is actually needed is Task lifecycle.  If we had a lifecycle metadata element that had states: ACTIVE, INACTIVE, and FAILED, a lot of logic could revolve around that, including resource garbage collection.  The explicit transitioning of tasks through these states on certain events would have solved the problem fixed here, and will also provide a good mechanism for controlling scale-in.  Instead of maintaining many metadata elements they're either generated entirely anew, or not depending on their state.  Transition tasks to INACTIVE.  If the last task in the index range is marked INACTIVE, it can be removed from the StateStore altogether.  Anyway, point taken, it's a good idea to re-examine this and neaten it up a bit.

gabrielhartmann:
retest this please

PR #576, opened 2017-02-15
Be more permissive of unknown characters in JSON responses

nickbp:
Go's JSON library doesn't know what `\v` means. Instead of returning an error for the whole command, just log a complaint to stderr.

### DIFF
--- a/cli/response.go
+++ b/cli/response.go
@@ -18,11 +18,15 @@ func PrintJSONBytes(responseBytes []byte, request *http.Request) {
 	var outBuf bytes.Buffer
 	err := json.Indent(&outBuf, responseBytes, "", "  ")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mrbrowning:
Looks like the emitter (see the syntax for strings): http://www.json.org Why in tarnation is a vertical tab ending up in our JSON responses?

nickbp:
The specific case was a protobuf with `bytes` fields (specifically mesos likes to pass around UUIDs as `bytes`). We're using a jackson module to interpret protobuf java objects to json, so that's probably the origin of the problem.

https://github.com/mesosphere/dcos-commons/blob/48e193618d5e461130903f0e7adba3dd56211139/sdk/scheduler/build.gradle#L80

nickbp:
retest this please

PR #579, opened 2017-02-15
[DOCS] Delete duplicate tutorial

nickbp:
This appears to be an older version of https://github.com/mesosphere/dcos-commons/blob/master/docs/pages/tutorials/quick-start-java.md


sascala:
It looks like we'd have to update the links here (at the bottom) as well: https://github.com/mesosphere/dcos-commons/blob/master/README.md

In general, do we want the README to just point to the GH pages site, or is there a reason information is duplicated there?

nickbp:
Updated root `README.md` to point to the correct version in GH pages, good catch! I think it'd make sense to just link directly into the GH pages site, that'd at least make the GH pages site slightly more discoverable.

I think this duplicate file was likely left behind due to a bad merge in the repo. The two files are otherwise identical and this is the only thing listed in that docs/tutorials/ directory, the others are all in docs/pages/tutorials/.

gabrielhartmann:
retest this please

PR #580, opened 2017-02-16
Added Strict mode support

mohitsoni:
Summary of changes:
* Updated libmesos binary
* Installs injected certs into JRE
* Adds Strict mode support for:
  * cassandra
  * elastic
  * hdfs
  * kafka
  * hello-world
  * template
* Currently, strict mode is only supported with V0 (non-http) Mesos API. Once Mesos adds support for auth-n-auth for V1 (http) Mesos API, the frameworks can just switch to that via `MESOS_API_VERSION` variable.
* Also, refactors Kafka so that it does that right thing w.r.t. registration. This was required to ensure that FrameworkInfo was created in a consistent way like other frameworks.


frameworks/hdfs/universe/marathon.json.mustache
```diff
@@ -40,6 +46,13 @@
     "LIBMESOS_URI": "{{resource.assets.uris.libmesos-bundle-tar-gz}}",
     "HDFS_URI": "{{resource.assets.uris.hdfs-tar-gz}}",
     "BOOTSTRAP_URI": "{{resource.assets.uris.bootstrap-zip}}",
+
+    {{#service.secret_name}}
+    "DCOS_SERVICE_ACCOUNT_CREDENTIAL": { "secret": "serviceCredential" },
+    "MESOS_MODULES": "{\"libraries\": [{\"file\": \"libdcos_security.so\", \"modules\": [{\"name\": \"com_mesosphere_dcos_ClassicRPCAuthenticatee\"}]}]}",
```
mrbrowning: This should be unnecessary with the bundle that provides `modules.json` out of the box, right? Or is there something preventing us from using that bundle here?

frameworks/kafka/src/main/dist/svc.yml
```diff
@@ -1,8 +1,6 @@
 name: {{FRAMEWORK_NAME}}
 scheduler:
-  principal: {{FRAMEWORK_PRINCIPLE}}
-  zookeeper: {{MESOS_ZOOKEEPER_URI}}
-  api-port: {{PORT_API}}
+  principal: {{FRAMEWORK_PRINCIPAL}}
```
mrbrowning: Does this one need a user too?

frameworks/hdfs/universe/marathon.json.mustache
mohitsoni: As per an offline conversation with @karya0, this is something that we need to include now.

frameworks/kafka/src/main/dist/svc.yml
mohitsoni: Good catch. Let me update the PR.

frameworks/hdfs/universe/marathon.json.mustache
mrbrowning: Okay. @karya0 , maybe this is a good place to ask: why is this necessary? I thought modules were linked on agent launch, so how is this value being used? And will schedulers running in strict mode always have to care about what seems like a Mesos implementation detail?

loren:
Juts curious, what sets elastic apart w/r/t strict mode? Is this one more way that it's a special snowflake?

Also, what about the cassandra FW?

frameworks/hdfs/universe/marathon.json.mustache
gabrielhartmann: We should address this ASAP.  @mohitsoni: Can you ping @karya0 out of band please?

mohitsoni:
@loren Elastic needs a 1.9 cluster, and at the time of sending this PR there was a bug in 1.9-dev strict mode. And, because of that I was unable to verify my changes, and hence decided to not include elastic in this PR (until the bug gets fixed).

As far as the strict-mode support changes are concerned, elastic is not a snowflake.

frameworks/hdfs/universe/marathon.json.mustache
karya0: The `MESOS_MODULES` env requires an absolute path of the form `file:///` and for those frameworks that untar the bundle in their MESOS_SANDBOX directory, we can't predict the absolute path.

For framework that untar the bundle at a known location, i.e., `/opt/libmesos-bundle`, we can include a `mesos_scheduler_authenticatee_module.json` file in the bundle and let the app point to it.

gabrielhartmann: @karya0:  Why is the absolute path required?  Can support for relative path be added?

frameworks/kafka/src/main/dist/svc.yml
mohitsoni: Done!

mohitsoni:
@loren Added strict mode support for Cassandra.

mohitsoni:
Added strict mode support for elastic.

frameworks/hdfs/universe/marathon.json.mustache
karya0: @gabrielhartmann: It's because Mesos doesn't support it. We can try to get a patch into Mesos. Currently, Mesos just hands over the string to the 'open' call. What we need is to change this behavior to resolve the env vars before passing it to open. I'll lookup open semantics around env var resolution and write back with an update on the possibilities.

nickbp: Out of curiosity, what does passing file:///dev/urandom do?

karya0: @nickbp: The way Mesos stout processes the filepaths, it will first try to read the *entire* file contents into a buffer (8KB at a time) and then will parse it to JSON parser and will follow it by a JSON->Proto conversion. As you can guess, the reading will never finish up and at some point, you'll run out of memory+swap and the process will get killed.

mohitsoni:
retest this please

frameworks/hdfs/universe/marathon.json.mustache
nickbp: I assume 'the process' in this case is the agent? When it restarts will it just try to do the same thing again?

karya0: It would be whatever process was trying to load the module manifest. It could be the master, agent, or the scheduler.

PR #587, opened 2017-02-17
Use per-service 'test:<fwk>' github label instead of default 'test:shakedown' each time

nickbp:
At the moment, each successive test run hides the result of prior runs in github

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -13,7 +13,7 @@ function proxylite_preflight {
 
 function run_framework_tests {
     framework=$1
-    FRAMEWORK_DIR=${REPO_ROOT_DIR}/frameworks/$framework
+    FRAMEWORK_DIR=${REPO_ROOT_DIR}/frameworks/${framework}
 
     if [ "$framework" = "proxylite" ]; then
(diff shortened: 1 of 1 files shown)


loren:
Great idea! I'm seeing it as `test:test:<fwk>` though.
screen shot 2017-02-17 at 1 32 35 pm

nickbp:
Oh whoops, fixed

test.sh
```diff
@@ -41,7 +41,7 @@ function run_framework_tests {
     fi
 
     # Run shakedown tests in framework directory:
-    TEST_GITHUB_LABEL="test:${framework}" ${REPO_ROOT_DIR}/tools/run_tests.py shakedown ${FRAMEWORK_DIR}/tests/
+    TEST_GITHUB_LABEL="${framework}" ${REPO_ROOT_DIR}/tools/run_tests.py shakedown ${FRAMEWORK_DIR}/tests/
```
loren: nit: ditch the quotes?

nickbp: wouldn't expect framework dirs to have spaces, but just in case...

PR #589, opened 2017-02-18
[SDK:INFINITY-976] Support resetting and disabling the state store cache

nickbp:
- Cache reset may be performed via a new /state/reset endpoint
- Cache may be manually disabled by setting a DISABLE_STATE_CACHE envvar in the scheduler.

Other fixes:
- Catches up CLIs to reflect the current `/state` endpoints, while dropping per-service boilerplate (particularly the demo 'example' command).
- Fixes a state cache issue: underlying 'suppressed' property isn't displayed in the property list, as it was being written as a side effect within `CuratorStateStore` (and not reflected in the list of properties in the cache). Fixed by refactoring out the side effect behavior to pass through the cache as a property write.


sdk/scheduler/src/test/java/com/mesosphere/sdk/state/StateStoreCacheTest.java
```diff
@@ -401,6 +405,56 @@ public void testStoreTaskStatusFailureThenSuccess() throws Exception {
...
+            assertEquals("hello (reason: UNKNOWN)", e.getMessage());
+        }
+
+        // cache content is all left as-is following the failed refresh.
+        // data is not cleared as it was in the underlying storage:
+        assertEquals(Optional.of(FRAMEWORK_ID), mockedCache.fetchFrameworkId());
+        assertEquals(TASK, mockedCache.fetchTasks().iterator().next());
+        assertEquals(STATUS, mockedCache.fetchStatuses().iterator().next());
+        assertEquals(PROP_KEY, mockedCache.fetchPropertyKeys().iterator().next());
+
```
mgummelt: remove newline

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreCache.java
```diff
@@ -41,13 +41,18 @@
 
     private static StateStoreCache instance = null;
 
+    @VisibleForTesting
     protected final StateStore store;
 
+    @VisibleForTesting
```
mgummelt: Why does this need to be @VisibleForTesting?  A `fetchFrameworkId()` method already exists.

sdk/scheduler/src/test/java/com/mesosphere/sdk/state/StateStoreCacheTest.java
mgummelt: remove newline

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -180,18 +180,4 @@
      * @throws StateStoreException if key validation fails or clearing the entry fails
      */
     void clearProperty(final String key) throws StateStoreException;
-
-    /**
-     * @return true if the offers are suppressed.
-     * @throws StateStoreException
-     */
-    boolean isSuppressed() throws StateStoreException;
```
mgummelt: This may solve our particular problem with suppress/revive, but it only papers over the real problem, which is that `StateStore` is too high level.  We should make it a key/value store.  Why does it need to know about tasks/statuses/properties/etc.?

nickbp: INFINITY-1072

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreCache.java
nickbp: Just annotating that these are only non-private so that tests can use them (specifically by `StateStoreCacheTest.TestStateStoreCache.consistencyCheckForTests()`)

mgummelt: StateStoreCache already has a `fetchFrameworkId()` method we can use in the tests.  Why do the tests need to access the instance variable directly?

nickbp: consistencyCheckForTests() is checking that the cache members remain internally consistent throughout multithreaded tests

nickbp:
Updated, PTAL

nickbp:
retest this please

gabrielhartmann:
@nickbp:  Can you rebase this on `master` please?

nickbp:
Another hour, another merge conflict

PR #592, opened 2017-02-21
Shakedown frameworks in random order instead of alphabetical order

loren:
Useful for identifying upstream problems


gabrielhartmann:
retest this please

loren:
Kind of meta, but it appears this PR is **failing** because it's _working_. In this [last CI run](https://jenkins.mesosphere.com/service/jenkins/job/dcos-commons/job/infinity-dcos-commons-pr/2074/console), the randomization caused the frameworks to be ordered like this for shakedown: kafka, hdfs, template, hello-world, prototype, proxylite, elastic.

Elastic got all the way to its last shakedown test, where it bumps the node count by 3 (one each for data, coordinator, and ingest). The ingest node got added but not the data or coordinator nodes so it timed out. This behavior is consistent with a lack of suitable offers to fulfill offer requirements. But shakedown runs on a 5-node cluster while most of my local testing was done on a 3-node cluster, so where are the resources?

I don't think kafka, template, hello-world, prototype, or proxylite uninstalled the framework and released their resources when they finished. Only hdfs did this. Since elastic's shakedown ran after kafka, template, hello-world, prototype, and proxylite, my hunch is the cluster's resources were diminished enough to cause the timeout.

nickbp:
It'd probably make sense to uninstall services if their tests passed, leaving them installed (for debugging) only if the tests failed

gabrielhartmann:
retest this please

gabrielhartmann:
@nickbp:  Agreed.  Don't know how to do that.  Pytest doesn't support conditional teardown, not that I'm aware of anyway.

gabrielhartmann:
The latest retest loop here is with 8 private agents.  Let's see if that's sufficient.

loren:
I think there's more to be done here. Janitoring the frameworks is a good first step, but there is a potentially bigger problem with stale framework directories being left around in `/var/lib`. I added a [little helper](https://github.com/mesosphere/dcos-commons/blob/master/frameworks/elastic/tests/config.py#L23-L25 <https://github.com/mesosphere/dcos-commons/blob/master/frameworks/elastic/tests/config.py#L23-L25>) for myself in the Elastic FW so I could GC old frameworks before running tests. If you run shakedown many (~20) times on the same CCM cluster as I did in order to track down an intermittent failure, the node volumes get pretty full (but at different rates), shakedown tests start failing intermittently, and it becomes even harder to track down the original intermittent failures. The default Mesos GC behavior is too slow to account for this. So it's possible with a default `/var/lib` partition of 37GB that `/var/lib/mesos/slave/slaves/*/frameworks/*` will fill up enough on one or more nodes to make it unfit for a framework task's volume requirement.

So I think I should invoke that `gc_frameworks()` method at least once per framework, like I do in the `elastic` framework. What do you think?

gabrielhartmann:
@loren:  Let's not work around that bug.  DC/OS needs to support volume clean up.

PR #595, opened 2017-02-22
Different fix for allowing mustache within dev guide

nickbp:
Other fix didn't work, but this one does (manually verified locally): http://www.digiguru.co.uk/blog/2015/03/31/how-to-output-mustache-templates-in-a-jekyll-site.html

Other changes are just cleaning up trailing whitespace.


nickbp:
YOLO

PR #597, opened 2017-02-23
Adds basic TLS Support

mohitsoni:
* Adds TLS support to HDFS.
* TLS can be enabled by setting `service.tls_enabled` flag to true.
* Added a new install certificate feature inside the bootstrap go binary. This feature will allow installing DC/OS injected certificate inside the JRE's keystore that's present inside the task sandbox.

Note about current implementation:
* It relies on the default JDK keystore, and doesn't generate a separate keystore and truststore for HDFS. This is something that can be re-visited in future. This is not the most secure way for doing this.
* Current implementation also turns off ssl host verification, as the DC/OS generated certs that are injected in the sandbox doesn't have Common Name set to task's A-Record. Once DC/OS starts injecting certs with correct Common Name, we can turn on ssl host verification.


frameworks/hdfs/src/main/dist/core-site.xml
```diff
@@ -32,7 +32,12 @@
         <name>ha.zookeeper.parent-znode</name>
         <value>/{{SERVICE_ZK_ROOT}}/hadoop-ha</value>
     </property>
-
+    {{#TLS_ENABLED}}
+    <property>
+        <name>hadoop.ssl.hostname.verifier</name>
+        <value>ALLOW_ALL</value>
```
gabrielhartmann: What are the other possible values? Should this be templated too?

frameworks/hdfs/src/main/dist/hdfs-site.xml
```diff
@@ -17,7 +17,12 @@
         <name>dfs.cluster.administrators</name>
         <value>{{ADMINISTRATORS}}</value>
     </property>
-
+    {{#TLS_ENABLED}}
+    <property>
+        <name>dfs.http.policy</name>
+        <value>HTTP_AND_HTTPS</value>
```
gabrielhartmann: Same here.

frameworks/hdfs/src/main/dist/ssl-client.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+<configuration>
+    <property>
+        <name>ssl.client.truststore.password</name>
+        <value>changeit</value>
```
gabrielhartmann: Are these passwords hard-coded forever?

frameworks/hdfs/src/main/dist/ssl-client.xml
```diff
@@ -0,0 +1,16 @@
...
+    </property>
+    <property>
+        <name>ssl.client.truststore.type</name>
+        <value>jks</value>
```
gabrielhartmann: Is this an option that should be templated?

frameworks/hdfs/src/main/dist/core-site.xml
mohitsoni: It can be templated. Other values are:
- DEFAULT
- STRICT
- STRICT_I6
- DEFAULT_AND_LOCALHOST 
- ALLOW_ALL

frameworks/hdfs/src/main/dist/hdfs-site.xml
mohitsoni: It can be templated. Other values are:
- HTTP_ONLY
- HTTPS_ONLY
- HTTP_AND_HTTPS

frameworks/hdfs/src/main/dist/ssl-client.xml
mohitsoni: These are well-known passwords. And making them configurable won't make sense until we start generating keystores and truststores.

frameworks/hdfs/src/main/dist/ssl-client.xml
mohitsoni: I think JKS is what everyone one uses in Java world. So, this might not be something that can be templated. Although, I might be wrong.

gabrielhartmann:
@mohitsoni:  Once you're happy that everything which can be usefully templated today has been, it looks good to me.

frameworks/hdfs/src/main/dist/core-site.xml
jsrodman: STRICT_IE6, right?  (I misread this as STRICT_16 and was pretty confused).

sdk/bootstrap/main.go
```diff
@@ -251,6 +260,49 @@ func renderTemplates(templateMaxBytes int64) {
...
+		return
+	}
+
+	javaHome := os.Getenv("JAVA_HOME")
+	if len(javaHome) == 0 {
+		log.Fatalf("No JAVA_HOME provided. Will not install certs.")
+		return
+	}
+	certPath := filepath.Join(mesosSandbox, ".ssl", "ca.crt")
+	//keytoolPath := path.filepath.Join(javaHome, "bin", "keytool")
```
jsrodman: Cruft?

sdk/bootstrap/main.go
```diff
@@ -251,6 +260,49 @@ func renderTemplates(templateMaxBytes int64) {
 	}
 }
 
+func installCerts() {
+	mesosSandbox := os.Getenv("MESOS_SANDBOX")
+	sslDir := filepath.Join(mesosSandbox, ".ssl");
+
+	sslDirExists, err := _pathExists(sslDir)
```
jsrodman: I think you want to check if this is a dir, not if it exists.  In other words, right now this seems to be okay with a .ssl file, device file, named pipe, etc.

sdk/bootstrap/main.go
```diff
@@ -251,6 +260,49 @@ func renderTemplates(templateMaxBytes int64) {
...
+	if !sslDirExists || err != nil {
```
jsrodman: Can we log the err if there is one?  Or is this level of nitpicking foolish at this point in the project?

sdk/bootstrap/main.go
```diff
@@ -272,5 +324,9 @@ func main() {
 		log.Printf("Template handling disabled via -template=false: Skipping any config templates")
 	}
 
-	log.Printf("SDK bootstrap successful.")
+	if (args.installCerts) {
+		installCerts()
+	}
```
jsrodman: Does it make sense for this to continue on with success  when cert installation was requested but fails?

jsrodman:
Are things like TLS versions supported or choice of ciphersuites likely to be needed here, or are those controlled via some other mechanism?

frameworks/hdfs/src/main/dist/core-site.xml
mohitsoni: @jsrodman You are right! I made a typo, it should be `STRICT_IE6`.

The deal with STRICT_IE6 is that it will match either the CN or any of the alternate subject names. Currently, the certificate that's provided to us only contains CN == agent's hostname, and != the task's A-Record.

There's ongoing work to address that, and once it's made available to us, we can move to `STRICT`.

sdk/bootstrap/main.go
mohitsoni: Thanks for pointing this out. I'll fix it.

frameworks/hdfs/src/main/dist/core-site.xml
mgummelt: Let's not template these until we support the other values.

frameworks/hdfs/src/main/dist/ssl-client.xml
```diff
@@ -0,0 +1,16 @@
...
+    </property>
+    <property>
+        <name>ssl.client.truststore.location</name>
+        <value>{{MESOS_SANDBOX}}/jre1.8.0_121/lib/security/cacerts</value>
```
mgummelt: jre version should use {{JRE_VERSION}}

frameworks/hdfs/src/main/dist/ssl-server.xml
```diff
@@ -0,0 +1,32 @@
...
+        <name>ssl.server.keystore.keypassword</name>
+        <value>changeit</value>
+    </property>
+    <property>
+        <name>ssl.server.keystore.password</name>
+        <value>changeit</value>
+    </property>
+    <property>
+        <name>ssl.server.keystore.location</name>
+        <value>{{MESOS_SANDBOX}}/jre1.8.0_112/lib/security/cacerts</value>
```
mgummelt: et al.

frameworks/hdfs/src/main/dist/ssl-client.xml
```diff
@@ -0,0 +1,16 @@
...
+    </property>
+</configuration>
```
mgummelt: newline at end of all files

frameworks/hdfs/src/main/dist/ssl-client.xml
```diff
@@ -0,0 +1,16 @@
...
+    <property>
+        <name>ssl.client.truststore.password</name>
+        <value>changeit</value>
+    </property>
+    <property>
+        <name>ssl.client.truststore.type</name>
+        <value>jks</value>
+    </property>
+    <property>
+        <name>ssl.client.truststore.location</name>
```
mgummelt: Why add these client configs?  Clients won't have access to this JRE, for example.

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -38,6 +38,11 @@ pods:
           hadoop-metrics2:
             template: {{CONFIG_TEMPLATE_PATH}}/journal-hadoop-metrics2.properties
             dest: {{HDFS_VERSION}}/etc/hadoop/hadoop-metrics2.properties
+          {{#TLS_ENABLED}}
```
mgummelt: @gabrielhartmann We really need a solution for this type of redundancy.  It occurs frequently.

frameworks/hdfs/universe/config.json
```diff
@@ -35,6 +35,11 @@
           "type":"string",
           "default":"svc.yml"
         },
+        "tls_enabled": {
```
mgummelt: Can you split this up into `{"tls": {"enabled": ...}}`.  For 1) consistency with `kerberos`, and 2) forwards compatibility with future TLS configuration variables.

frameworks/hdfs/src/main/dist/ssl-client.xml
mohitsoni: Thanks!

mohitsoni: Done

frameworks/hdfs/src/main/dist/ssl-server.xml
mohitsoni: Done

frameworks/hdfs/src/main/dist/ssl-client.xml
mohitsoni: `ssl-client.xml` is an example, it's actually not actually used by service.

frameworks/hdfs/universe/config.json
mohitsoni: 👍

frameworks/hdfs/src/main/dist/ssl-client.xml
mgummelt: Can you mention that in a comment in this file?

mohitsoni:
retest this please

sdk/bootstrap/main.go
```diff
@@ -251,6 +260,78 @@ func renderTemplates(templateMaxBytes int64) {
...
+	err = cmd.Run()
+	if err != nil {
+		log.Println("Failed to install the certificate.")
+		log.Fatal(err)
+		return
+	}
+	log.Println("Successfully installed the certificate.")
+}
+
+func _isDir(path string) (bool, error) {
```
nickbp: I'd drop the underscore prefixes in these, doesn't fit go style

sdk/bootstrap/main.go
```diff
@@ -251,6 +260,78 @@ func renderTemplates(templateMaxBytes int64) {
...
 
+func installDCOSCertIntoJRE() {
+	mesosSandbox := os.Getenv("MESOS_SANDBOX")
+	sslDir := filepath.Join(mesosSandbox, ".ssl")
+
+	// Check if .ssl directory is present
+	sslDirExists, err := _isDir(sslDir)
+	if !sslDirExists || err != nil {
+		log.Println("No $MESOS_SANDBOX/.ssl directory found. Cannot install certificate.")
+		log.Fatal(err)
```
nickbp: `Fatal` exit the process. No need for the following `return` (here and below)

nickbp: Also may as well include the `err` on the same line, via a format parameter. see e.g. L210

mohitsoni:
retest this please

mohitsoni:
retest this please

frameworks/hdfs/src/main/dist/ssl-client.xml
mohitsoni: Done!

PR #604, opened 2017-02-24
[SDK] Expose PodInstanceRequirement from Step

gabrielhartmann:
It is desirable to be able to inspect what Task related work a Step wishes to accomplish without actually initiating that work.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Step.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Step.java
@@ -29,6 +29,12 @@ public interface Step extends Element, Interruptible {
     Optional<PodInstanceRequirement> start();
 
     /**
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Step.java
```diff
@@ -29,6 +29,12 @@
     Optional<PodInstanceRequirement> start();
 
     /**
+     * Returns an {@link OfferRequirement}, or an empty Optional if obtaining/updating resource requirements are not
```
mgummelt: s/OfferRequirement/PodInstanceRequirement

mgummelt:
Please always include the motivation for the PR in the description.

gabrielhartmann:
@mgummelt: done

mgummelt:
There is no client code.  What's the application?

PR #610, opened 2017-02-27
[KAFKA] Add TopicController endpoint API and Command Executor class

mbdcos:


frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Topic Controller executing commands through command executor.
+ * Kafka package should also be deployed with framework to enable those kafka commands.
+ */
+
+@Path("/v1/topics")
+public class TopicController {
```
mrbrowning: Aren't other classes of this type named `*Resource`? I agree that `Controller` is probably the more intuitive name, but for consistency's sake, they should be the same.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
```diff
@@ -0,0 +1,171 @@
...
+            int replFactor = Integer.parseInt(replicationFactor);
+            JSONObject result = cmdExecutor.createTopic(name, partCount, replFactor);
+            return Response.ok(result.toString(), MediaType.APPLICATION_JSON).build();
+        } catch (Exception ex) {
+            log.error("Failed to create topic: " + name + " with exception: " + ex);
+            return Response.serverError().build();
+        }
+    }
+
+    /* Below is copied from dcos-kafka-service */
```
mrbrowning: I don't think this comment or the blank line after need to be present.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
```diff
@@ -0,0 +1,171 @@
...
+            @QueryParam("key") String key,
+            @QueryParam("value") String value,
+            @QueryParam("partitions") String partitions,
+            @QueryParam("messages") String messages) {
+
+        try {
+            JSONObject result = null;
+            List<String> cmds = null;
+
+            if (operation == null) {
```
mrbrowning: I think this setup is a little confusing for users. With every other endpoint, if they get the name wrong, they'll get a 404. Here, they'll get a 5XX, which means that `/{name}` is implicitly the path for `/operation?name={name}`, which shouldn't be the case. I would either split these out into their own (concrete) endpoint methods (e.g. `/partitions`) or put this in its own subpath like the `/operation` example above, with the operation to run as query parameter.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
```diff
@@ -0,0 +1,171 @@
...
+
+        } catch (Exception ex) {
+            log.error("Failed to perform operation: " + operation + " on Topic: " + name + " with exception: " + ex);
+            return Response.serverError().build();
+        }
+    }
+
+    @DELETE
+    @Path("/{name}")
+    public Response deleteTopic(
```
mrbrowning: Same as above.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+        } catch (Exception ex) {
+            log.error("Failed to delete Topic: " + name + " with exception: " + ex);
+            return Response.serverError().build();
+        }
+    }
+
+    @GET
+    @Path("/{name}/offsets")
```
mrbrowning: `/{name}...` could mean that `{name}` is a topic or an operation depending on context, these really need to be in different subpaths to avoid that ambiguity.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/cmd/CmdExecutor.java
```diff
@@ -0,0 +1,252 @@
...
+import org.apache.commons.lang3.time.StopWatch;
+
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+
+
+/**
+ * A simple command executor class, copied from dcos-kafka-service.
```
mrbrowning: Can remove this and surrounding newlines.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/cmd/CmdExecutor.java
```diff
@@ -0,0 +1,252 @@
...
+    public CmdExecutor(KafkaZKClient kafkaZkClient, String kafkaStandBoxPath) {
+        if (kafkaStandBoxPath == null){
+            throw new InvalidParameterException("KAFKA_VERSION_PATH is not set. Can not start CmdExecutor");
+        }
+        this.binPath = kafkaStandBoxPath + "/bin/";
+        this.kafkaZkClient = kafkaZkClient;
+        this.zkPath = kafkaZkClient.getZKString();
+    }
+
+    /* TODO: below is copied from dcos-kafka-service */
```
mrbrowning: Can remove this and empty line below.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/scheduler/KafkaService.java
```diff
@@ -27,14 +29,18 @@ protected void startApiServer(DefaultScheduler defaultScheduler,
                                   int apiPort,
                                   Collection<Object> additionalResources) {
         final ServiceSpec serviceSpec = super.getServiceSpec();
+
```
mrbrowning: Line break probably not necessary, seems reasonable to group all these declarations in a single paragraph.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mbdcos: It was named as TopicController in dcos-kafka-service

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mbdcos: ok

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/cmd/CmdExecutor.java
mbdcos: ok

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/cmd/CmdExecutor.java
mbdcos: ok

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/scheduler/KafkaService.java
mbdcos: ok

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mbdcos: I am not sure whether changing endpoints is a good idea but you are right.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mrbrowning: I think `Resource` is referring to what `Controller` is referring to (e.g. a grouping of endpoint handler logic) so `TopicResource` or `TopicsResource` seems most consistent with the other classes in this area.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mrbrowning: @triclambert should the HTTP API be a drop-in replacement, or can we move paths around?

mrbrowning: If each operation has its own endpoint, the paths will be the same, though anything automated being run by a user somewhere that depends on the error codes being the same would break.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mbdcos: done

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mbdcos: Done. Need to update cli.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mbdcos: /delete/{name}

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
mbdcos: it is an operation

topics/{name}/offsets

topics/operations/....

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicController.java
nickbp: It's already a `DELETE` HTTP operation though?

mbdcos: yep, both are fine for me

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
```diff
@@ -0,0 +1,169 @@
...
+        try {
+            return Response.ok(kafkaZkClient.getTopic(topicName), MediaType.APPLICATION_JSON).build();
+        } catch (Exception ex) {
+            log.error("Failed to fetch topic: " + topicName + " with exception: " + ex);
+            return Response.serverError().build();
+        }
+    }
+
+    @POST
+    @Path("/create")
```
mrbrowning: No need for `/create`, RESTful semantics dictate that creating a resource in `/resources` just requires a `POST` to that path.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
```diff
@@ -0,0 +1,169 @@
...
+            JSONObject obj = cmdExecutor.underReplicatedPartitions();
+            return Response.ok(obj.toString(), MediaType.APPLICATION_JSON).build();
+        } catch (Exception ex) {
+            log.error("Failed to fetch topics with exception: " + ex);
+            return Response.serverError().build();
+        }
+    }
+
+    @PUT
+    @Path("/operations/{type}")
```
mrbrowning: Now that I'm thinking about it, do you think it makes more sense to have this be at `/{topic}/operation/{type}`? That's what I'd go for intuitively if I were constructing URLs by hand. Also, this should probably be `POST`?

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
```diff
@@ -0,0 +1,169 @@
...
+            JSONObject result = cmdExecutor.createTopic(name, partCount, replFactor);
+            return Response.ok(result.toString(), MediaType.APPLICATION_JSON).build();
+        } catch (Exception ex) {
+            log.error("Failed to create topic: " + name + " with exception: " + ex);
+            return Response.serverError().build();
+        }
+    }
+
+    @GET
+    @Path("/unavailable_partitions")
```
mrbrowning: It is slightly unfortunate that a topic can never be named `unavailable_partitions` without stepping on toes here... @nickbp what do you think?

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
```diff
@@ -0,0 +1,169 @@
...
+            return Response.ok(result.toString(), MediaType.APPLICATION_JSON).build();
+
+        } catch (Exception ex) {
+            log.error("Failed to perform operation: " + type + " on Topic: " + topicName + " with exception: " + ex);
+            return Response.serverError().build();
+        }
+    }
+
+    @DELETE
+    @Path("/delete/{name}")
```
mrbrowning: Since the `DELETE` method makes this unambiguous, the path should be `/{topic}` for RESTfulness.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
mrbrowning: Alternatively, since we index by name, that means you know the resource's path beforehand (e.g. `/v1/topics/{name}`). In that case, it would also be acceptable to have this method be a `PUT` to `/v1/topics/{name}`. Either way seems fine.

mbdcos:
TopicController is for demos to show examples, not related to scheduler.   I prefer removing them totally but keeping now for compatibility.  Especially: installing Kafka binary along with framework and running Kafka commands through web api. Nobody should create topics using frameworks topics endpoint interface. You should directly connect to Kafka.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
mbdcos: if POST you do not give {type}

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
mbdcos: Reverting back to original. Changed it to /delete/{name} due to first comment here.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
mbdcos: Cli : "gets info for any unavailable partitions"

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
mrbrowning: That's true when you're _creating_ a resource in RESTful style, but this is effectively an RPC call, nothing is being created. The convention there is just to use `POST`.

mbdcos:
retest this please

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/TopicResource.java
mbdcos: POST did not work with CLI.

PR #616, opened 2017-02-27
Drop unwanted embedded quotes that break scripts in some circumstances.

jsrodman:
https://jira.mesosphere.com/browse/INFINITY-1037

I didn't add the --raw-output to the .id line, because it's an integer and thus won't have quotes.  Of course maybe that will be false in the future?


jsrodman:
retest this please

jsrodman:
@susanxhuynh I'm taking your verbal response to my mentioning this as OKing it, since this is at the "trivial" level of change.

PR #618, opened 2017-02-28
Faster service rebuilds via caching of go artifacts

nickbp:
Adds a common `build_go_exe.sh` script which implements caching logic to avoid unnecessary rebuilds of go executables. This is done by comparing the SHA1 of a native/local build (faster than cross-builds) to detect changes in the go build artifact.

This script is now used for building per-service CLI executables as well as the common `bootstrap` executable.

As a result, rebuilding a service goes from ~40s to ~5s (plus the following artifact upload, which mainly depends on bandwidth)

### DIFF
--- a/sdk/bootstrap/build.sh
+++ b/sdk/bootstrap/build.sh
@@ -3,71 +3,19 @@
 # exit immediately on failure
 set -e
 
-BOOTSTRAP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-REPO_ROOT_DIR="$(dirname $(dirname $BOOTSTRAP_DIR))"
-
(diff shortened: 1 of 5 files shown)


sdk/bootstrap/build.sh
```diff
@@ -3,71 +3,19 @@
...
-        ;;
-esac
+cd $CUR_DIR
 
-if [ -n "$UPX_BINARY" ]; then
-    echo "Binary CLI compression enabled: $($UPX_BINARY -V | head -n 1)"
-else
-    echo "Binary CLI compression disabled"
-fi
+../../tools/build_go_exe.sh sdk/bootstrap/ linux
```
mohitsoni: nit: We should make tools directory a env var, that's injected from outside. If not present we should default to the monorepo specific path.

PR #621, opened 2017-03-01
[SDK] Introduce storeStatus(name, status) method in StateStore interface

mbdcos:
See comments in PR #640 

 Kafka Upgrade changes are in Kafka.  And major SDK changes such as this one should be in SDK.

StateStore is a **core** functionality and I am not comfortable extending it _**outside of sdk.curator**_. 

There is of course always other ways. If introducing a new method in StateStore interface will break some other design requirements, we could ignore. Otherwise, **what is the reason to not have storeStatus(name,status) method in StateStore interface?** What is it breaking?

* **_Why I was proposing to introduce storeStatus(name, status) in StateStore interface:_**

   - It _complements_ StateStore. If you do not know taskName, storeStatus(status) will search and find for you. If you know taskName, no need to search, you can use storeStatus(name, status) directly. 
Other than KafkaConfigUpgrade, there might be other use cases in the future.
    
  - Reimplementing stateStore functionality of CuratorStateStore externally, _outside of sdk.curator_, **breaks** the design.
     - You get lock-in to CuratorStateStore, you can not use any other StateStore
     - If CuratorStateStore is dropped, or changed, you need to update your NewCuratorStateStore class
     - Since NewCuratorStateStore is outside, any maintainer will not know that you extend CuratorStateStore and you depend on it.
 
This is our SDK and we should be able to update as necessary.


sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,14 +192,45 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
 
     @Override
     public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
-        String taskName;
+        Optional<String> optionalTaskName = Optional.empty();
+
+        //TODO(MB): keep the commented code below: original approach for getting TaskName
```
gabrielhartmann: Please don't leave any commented code.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,14 +192,45 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
         }
+        */
+
+        for (Protos.TaskInfo taskInfo : fetchTasks()) {
+            if (taskInfo.getTaskId().equals(status.getTaskId())) {
+                if (optionalTaskName.isPresent()) {
+                    logger.error("Found identical task ids in Task {} and  Task {}",
+                            optionalTaskName.get(), taskInfo.getName());
+                    throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
+                            "There are more than one tasks with the same TaskStatus.task_id: %s", status));
```
gabrielhartmann: "Duplicate TaskID encountered."

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,14 +192,45 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
     @Override
     public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
-        String taskName;
+        Optional<String> optionalTaskName = Optional.empty();
+
+        //TODO(MB): keep the commented code below: original approach for getting TaskName
+        /* We used to get TaskName from TaskID's prefix, instead of comparing TaskIDs.
+           TaskID is our unique identifier, can not change, but we can modify TaskName
+
+           Getting TaskName from prefix may be more efficient, however it prevents us from changing TaskName.
```
mbdcos: If you think toTasName() has no performance benefit and no further effort needed in this, I will delete the comments

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,14 +192,45 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
+                            optionalTaskName.get(), taskInfo.getName());
+                    throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
+                            "There are more than one tasks with the same TaskStatus.task_id: %s", status));
+                }
+                    optionalTaskName = Optional.of(taskInfo.getName());
+            }
+        }
+        if (!optionalTaskName.isPresent()) {
+            throw new StateStoreException(Reason.NOT_FOUND, String.format(
+                    "Failed to find a task with matching TaskStatus.task_id: %s", status));
```
gabrielhartmann: Please don't use this notation `TaskStatus.task_id`.  Just say `TaskID`.  e.g. `Failed to find a task with TaskID: %s`

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,14 +192,45 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
     public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
-        String taskName;
+        Optional<String> optionalTaskName = Optional.empty();
+
+        //TODO(MB): keep the commented code below: original approach for getting TaskName
+        /* We used to get TaskName from TaskID's prefix, instead of comparing TaskIDs.
+           TaskID is our unique identifier, can not change, but we can modify TaskName
+
+           Getting TaskName from prefix may be more efficient, however it prevents us from changing TaskName.
+           For further investigation, keep the commented code below:
```
mbdcos: If you think toTasName() has no performance benefit and no further effort needed in this, I will delete the comments

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -78,6 +78,19 @@
 
 
     /**
+     * Stores the TaskStatus of a particular Task. The {@link TaskInfo} for this exact task MUST have already been
```
gabrielhartmann: if it must already have been written the name should be `updateStatus` probably.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -226,6 +256,7 @@ public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
                     status.getTaskId().getValue(), optionalTaskInfo.get().getTaskId().getValue(),
                     status, optionalTaskInfo));
         }
+        /*  All checks till here are redundant if we get TaskName via looking at matching TaskIDs */
```
gabrielhartmann: I don't understand this comment.  If we have redundant checks they should be removed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,14 +192,45 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
 
     @Override
     public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
-        String taskName;
+        Optional<String> optionalTaskName = Optional.empty();
+
+        //TODO(MB): keep the commented code below: original approach for getting TaskName
+        /* We used to get TaskName from TaskID's prefix, instead of comparing TaskIDs.
+           TaskID is our unique identifier, can not change, but we can modify TaskName
```
gabrielhartmann: TaskID changes on every launch.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -78,6 +78,19 @@
...
+     * written via {@link #storeTasks(Collection)} beforehand.
+     *
+     * @param name The name of the task, which has same TaskId with the status
+     * @param status The status to be stored
+     * @throws StateStoreException if storing the TaskStatus fails, or if its TaskId is malformed, or if its matching
+     *                             TaskInfo wasn't stored first
+     */
+    void storeStatus(String name, TaskStatus status) throws StateStoreException;
```
gabrielhartmann: We should have a single `storeStatus` method, not two.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,14 +192,45 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
+           Getting TaskName from prefix may be more efficient, however it prevents us from changing TaskName.
+           For further investigation, keep the commented code below:
+         */
+        /*
         try {
-            taskName = CommonTaskUtils.toTaskName(status.getTaskId());
+            optionalTaskName = Optional.of(CommonTaskUtils.toTaskName(status.getTaskId()));
         } catch (TaskException e) {
             throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
                     "Failed to parse the Task Name from TaskStatus.task_id: '%s'", status), e);
```
mbdcos: It was TaskStatus.task_id:

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
mbdcos: you want to call it something else?

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
mbdcos: L202

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
mbdcos: ok

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
mbdcos: see: L209 I kept the original

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
mbdcos: ok

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
mbdcos: ok

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
mbdcos: Yes of course.  You can change taskName without a launch because it is just a key to stateStore. You still get status of your task with taskID.

mbdcos:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
gabrielhartmann: There should never be duplicate TaskIDs in the state store.

gabrielhartmann:
* It complements StateStore. If you do not know taskName, storeStatus(status) will search and find for you. If you know taskName, no need to search, you can use storeStatus(name, status) directly.
Other than KafkaConfigUpgrade, there might be other use cases in the future.

No such cases currently exist.  Changing the interface speculatively isn't a good idea.

* Reimplementing stateStore functionality of CuratorStateStore externally, outside of sdk.curator, breaks the design.

I don't see any "re-implementation" occurring.  If special one time logic is needed for the framework migration, it should be done in the old frameworks.

    * You get lock-in to CuratorStateStore, you can not use any other StateStore

This is fine for the old frameworks.  They are deprecated code.

    * If CuratorStateStore is dropped, or changed, you need to update your NewCuratorStateStore class

I would agree to all of the above if there were non migration related reasons for this change to the interface to exist.  This interface isn't referenced within the SDK or by any of its service, so doesn't belong here.

    * Since NewCuratorStateStore is outside, any maintainer will not know that you extend CuratorStateStore and you depend on it.

The whole idea is to move maintainers to the new services to ease support/maintenance burden.  Introduction of interface/implementation elements which only apply to migration scenarios adds to that burden.

PR #622, opened 2017-03-01
[SDK][INFINITY-966] Correctly reuse existing VIP port on discovery info after task restart

mrbrowning:
`NamedVIPEvaluationStage` was using the wrong criterion to determine whether a VIP for a given port was already set on the `DiscoveryInfo` (i.e. whether a VIP label existed on the `DiscoveryInfo` itself rather than on a particular `Port`) leading it to unnecessarily re-add those port definitions on task restart.

### DIFF
--- a/sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/NamedVIPEvaluationStageTest.java
+++ b/sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/NamedVIPEvaluationStageTest.java
@@ -1,7 +1,9 @@
 package com.mesosphere.sdk.offer.evaluate;
 
+import com.mesosphere.sdk.offer.InvalidRequirementException;
 import com.mesosphere.sdk.offer.MesosResourcePool;
 import com.mesosphere.sdk.offer.OfferRequirement;
+import com.mesosphere.sdk.offer.ResourceUtils;
 import com.mesosphere.sdk.testutils.OfferRequirementTestUtils;
 import com.mesosphere.sdk.testutils.OfferTestUtils;
 import com.mesosphere.sdk.testutils.ResourceTestUtils;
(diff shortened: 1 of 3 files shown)


mrbrowning:

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/NamedVIPEvaluationStageTest.java
```diff
@@ -45,4 +49,40 @@ public void testDiscoveryInfoPopulated() throws Exception {
         Assert.assertTrue(vipLabel.getKey().startsWith("VIP_"));
         Assert.assertEquals(vipLabel.getValue(), "test-vip:80");
     }
+
+    @Test
+    public void testVIPIsReused() throws InvalidRequirementException {
```
gabrielhartmann: Is a test needed which evaluates multiple offers?  I'm not grasping the bug exactly.

mrbrowning: `NamedVIPEvaluationStage` will look to see if a VIP for the port/name pair in question already exists on the task's `DiscoveryInfo` by checking its labels. This will be the case when we're evaluating an existing offer requirement, since that offer requirement is created from the task definitions that were retrieved from the state store. The bug was that `NamedVIPEvaluationStage` was looking for the relevant VIP label on `DiscoveryInfo` itself, but VIP labels are actually on each `Port`, so every time an existing offer requirement was evaluated it would create a new port since it didn't find anything in `DiscoveryInfo.getLabels()` that matched what it expected.

mrbrowning: As for multiple offers, each offer evaluation round starts with a fresh `PodInfoBuilder`, so no state is retained between rounds.

PR #624, opened 2017-03-01
Ccm cluster launch log template url

jsrodman:
Log the actual url that launch_ccm_cluster.py is using, and whether we're taking it from the environment (thus ignoring other logged values).


jsrodman:
@susanxhuynh I'm mostly trying to eliminate one of the questions I have while continuing to work on parameterizing things.

jsrodman:
FWIW, the failures here appear to be the tests undesirably running my WIP project for changing parameterization.  Thus the failure isn't in any way relevant, and isn't even supposed to be running at all.

susanxhuynh:
LGTM

PR #625, opened 2017-03-01
[SDK] added examples directory with worked example of RethinkDB implementation

ArtRand:
This PR includes an example framework (RethinkDB). The implementation has a complexity that lies somewhere between the hello-world example and one of the official supported frameworks (i.e. Kafka, HDFS). It may be useful for demonstration purposes, but it may also die on the vine as a ramp-up task.


ArtRand:
retest this please

keithchambers:
Thanks for the PR -- but we do not want to add additional example services at this time as each example service needs to be maintained.

PR #628, opened 2017-03-01
[SDK] Add VolumePathCannotBeEmpty validator

mbdcos:


sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/VolumePathCannotBeEmpty.java
```diff
@@ -0,0 +1,37 @@
...
+import com.mesosphere.sdk.specification.TaskSpec;
+import com.mesosphere.sdk.specification.VolumeSpec;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.*;
+
+/**
+ * Validates that volume path is not empty if new ServiceSpec has a volume
+ */
+public class VolumePathCannotBeEmpty implements ConfigValidator<ServiceSpec> {
```
mohitsoni: Can you please add some unit tests ?

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/VolumePathCannotBeEmpty.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public Collection<ConfigValidationError> validate(ServiceSpec nullableOldConfig, ServiceSpec newConfig) {
+        List<ConfigValidationError> errors = new ArrayList<>();
+
+        for (PodSpec podSpec : newConfig.getPods()) {
+            for (TaskSpec taskSpec : podSpec.getTasks()) {
+                Collection<VolumeSpec> volumeSpecs = taskSpec.getResourceSet().getVolumes();
+                if (volumeSpecs.isEmpty()) {
```
mohitsoni: Can volumeSpecs be `null` ?

mbdcos: get ResourceSet returns a Collection. If no VolumeSpec, it should be an empty list.

mohitsoni: Okay

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/VolumePathCannotBeEmpty.java
mbdcos: OK

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/VolumePathCannotBeEmpty.java
gabrielhartmann: Why have this check at all?  If it's empty the for loop will be a noop anyway.

mbdcos: It was checking null before that is why left here!

mbdcos:
```@NotEmpty``` does not detect blank string

mbdcos:
```@NotBlank``` works
We should also have ```@Pattern```

PR #630, opened 2017-03-01
Made explicit the DC/OS 1.9 requirement

Neuw84:
This framework requires DC/OS 1.9 with Mesos 1.2 as it depends on rlimit support.  
Just added this "requirement" to the documentation of the framework (have spend some time trying to install the framework on 1.8)

### DIFF
--- a/frameworks/elastic/docs/index.md
+++ b/frameworks/elastic/docs/index.md
@@ -25,3 +25,8 @@ DC/OS Elastic provides the following features:
 *   High availability runtime configuration and software updates
 *   Storage volumes for enhanced data durability, known as Mesos Dynamic Reservations and Persistent Volumes
 *   Automatic reporting of Elasticsearch metrics to DC/OS statsd collector
+
+# Requirements
+
+This framework requires DC/OS 1.9 as it depends on rlimit support provided by Mesos 1.2 [MESOS-6402](https://issues.apache.org/jira/browse/MESOS-6402).
+


keithchambers:
Thank you for the PR.

This repo holds the source code for our DC/OS Elastic Service product, but the product is not intended to be consumed from here. We provide the source code as an example for partners developers (like any good SDK does). It sounds like we need to make this clearer in our SDK docs.

For those consuming our product outside of this repo -- DC/OS Elastic Service is only available for install from Universe on a 1.9 clusters. And the documentation first appeared with the DC/OS 1.9 documentation: https://docs.mesosphere.com/1.9/usage/service-guides/ -- so it there is a low chance of confusion.

We plan to rename frameworks to examples. As part of that, we can add a readme explaining the purpose of the examples.

Neuw84:
Hi: 
Thanks for your feedback.  The thing is that we have a 1.8 cluster and the service is available for installing from the universe repo. 
We have the elasticsearch framework option from the Container Solution guys and the elastic framework as install options. 

Maybe, there is a mistake on that side?

keithchambers:
Sorry I accidentally clicked close on this from my iPhone. 

I just noticed this issue this morning. Also, on install we point you to these docs. We'll get both of these issues fixed ASAP.

keithchambers:
This should be fixed now. Closing one more time. :-)

PR #638, opened 2017-03-02
proxylite: Fix bug with stale backends

nlsun:
Previously HAProxy was not re-resolving dns addresses, leaving it pointing to stale and potentially incorrect backends. This change re-resolves dns on a 2 second interval.


nlsun:
@nickbp @loren This change affects you guys, this fixes a bug in proxylite.

nlsun:
@nickbp  ok, tests have passed

PR #639, opened 2017-03-02
[SDK] Add VolumeSpec path validation

mbdcos:
* Regex string to verify path is "[a-zA-Z0-9]+([a-zA-Z0-9_-]*[/\\\\]*)*"
* See examples in VerifyVolumePath.java

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
@@ -9,6 +9,8 @@ import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
 import org.apache.mesos.Protos;
 import com.mesosphere.sdk.specification.validation.ValidationUtils;
+import javax.validation.constraints.NotNull;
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
```diff
@@ -18,6 +22,11 @@
     public static final String RESOURCE_NAME = "disk";
 
     private final Type type;
+
+    @NotEmpty
+    @NotBlank
+    @NotNull
```
gabrielhartmann: `@NotNull` is not needed.  both `@NotEmpty` and `@NotBlank` already enforce that.

gabrielhartmann: Actually maybe `@NotNull` and the `@Pattern` are all that's really needed.

mbdcos: ```@NotNull``` and ```@Pattern``` are enough. regex detects blank

mbdcos: See examples please

mbdcos:
retest this please

PR #640, opened 2017-03-02
[SDK] Modify storeStatus to compare TaskIDs to search/find task name and TaskInfo

mbdcos:
StateStore keeps TaskInfo and TaskStatus and index information based on TaskName.
```
<TaskName>/<TaskInfo>
<TaskName>/<TaskStatus>
```
Each TaskInfo has a unique TaskID. Each TaskStatus also has a unique TaskID. TaskID changes at launch. Every time you restart you get a new TaskID.

When we receive a request to update status in ```StateStore.storeStatus(TaskStatus)```,  we first find its TaskName.  You need to know which task's status to update:  see the path "TaskName/TaskStatus".

- Conventional way of finding TaskName in Status update:
Normally you look existing TaskInfos and find the task that has same TaskID (TaskIDs in TaskStatus and TaskInfo should match if they belong to same task).

- Old method of generating TaskName during Status update:
However, what we were doing was extracting TaskName from TaskID. TaskID is generated as ```TaskName_<randomUUID>```. We assume that TaskName is TaskID prefix always.

This old approach would work unless you change TaskName after generating TaskID. 

We need to use conventional way of finding path. In *Config Upgrade*, we change TaskName. Status Store can not generate TaskName from TaskID and can not reconcile and perform Status update, till task is relaunched.

It is not clear whether there is any performance benefit in old method. StateStore is also cached in StateStoreCache, so performance benefits could be minimal if there is any.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
@@ -6,7 +6,6 @@ import com.mesosphere.sdk.offer.CommonTaskUtils;
 import com.mesosphere.sdk.state.*;
 import com.mesosphere.sdk.storage.Persister;
 import com.mesosphere.sdk.storage.StorageError.Reason;
-
 import org.apache.curator.RetryPolicy;
 import org.apache.mesos.Protos;
 import org.apache.zookeeper.KeeperException;
@@ -192,61 +191,26 @@ public class CuratorStateStore implements StateStore {
 
     @Override
     public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
-        Optional<String> optionalTaskName = Optional.empty();
+        Optional<Protos.TaskInfo> taskInfoOptional = Optional.empty();
 
         for (Protos.TaskInfo taskInfo : fetchTasks()) {
             if (taskInfo.getTaskId().getValue().equals(status.getTaskId().getValue())) {
-                if (optionalTaskName.isPresent()) {
+                if (taskInfoOptional.isPresent()) {
                     logger.error("Found duplicate taskIDs in Task {} and  Task {}",
-                            optionalTaskName.get(), taskInfo.getName());
+                            taskInfoOptional.get(), taskInfo.getName());
                     throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
                             "There are more than one tasks with TaskID: %s", status));
                 }
-                optionalTaskName = Optional.of(taskInfo.getName());
+                taskInfoOptional = Optional.of(taskInfo);
             }
         }
 
-        if (!optionalTaskName.isPresent()) {
+        if (!taskInfoOptional.isPresent()) {
             throw new StateStoreException(Reason.NOT_FOUND, String.format(
                     "Failed to find a task with TaskID: %s", status));
         }
-        putStatus(optionalTaskName.get(), status);
-    }
 
-    /* storeStatus(taskName, status):
-                            is the only way to update status if there are duplicate taskIDs.
-     Protected method to be called directly using taskName (i.e. by a config *upgrade*). It does error checking before
-     writing status to TaskPath
-     */
-    protected void storeStatus(String taskName, Protos.TaskStatus status) throws StateStoreException {
-        Optional<Protos.TaskInfo> optionalTaskInfo;
-
-        try {
-            optionalTaskInfo = fetchTask(taskName);
-        } catch (Exception e) {
-            throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
-                    "Unable to retrieve matching TaskInfo for the provided TaskStatus name %s.", taskName), e);
-        }
-
-        if (!optionalTaskInfo.isPresent()) {
-            throw new StateStoreException(Reason.LOGIC_ERROR,
-                    String.format("The following TaskInfo is not present in the StateStore: %s. " +
-                            "TaskInfo must be present in order to store a TaskStatus.", taskName));
-        }
-
-        if (!optionalTaskInfo.get().getTaskId().getValue().equals(status.getTaskId().getValue())) {
-            throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
-                    "Task ID '%s' of updated status doesn't match Task ID '%s' of current TaskInfo."
-                            + " Task IDs must exactly match before status may be updated."
-                            + " NewTaskStatus[%s] CurrentTaskInfo[%s]",
-                    status.getTaskId().getValue(), optionalTaskInfo.get().getTaskId().getValue(),
-                    status, optionalTaskInfo));
-        }
-        putStatus(taskName, status);
-    }
-
-    private void putStatus(String taskName, Protos.TaskStatus status) throws StateStoreException {
-
+        String taskName = taskInfoOptional.get().getName();
         Optional<Protos.TaskStatus> currentStatusOptional = fetchStatus(taskName);
 
         if (currentStatusOptional.isPresent()
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
@@ -3,7 +3,6 @@ package com.mesosphere.sdk.curator;
 import com.google.common.annotations.VisibleForTesting;
 import com.mesosphere.sdk.dcos.DcosConstants;
 import com.mesosphere.sdk.offer.CommonTaskUtils;
-import com.mesosphere.sdk.offer.TaskException;
 import com.mesosphere.sdk.state.*;
 import com.mesosphere.sdk.storage.Persister;
 import com.mesosphere.sdk.storage.StorageError.Reason;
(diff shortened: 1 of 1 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,17 +192,27 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
+        Optional<String> optionalTaskName = Optional.empty();
+        for (Protos.TaskInfo taskInfo : fetchTasks()) {
+            if (taskInfo.getTaskId().getValue().equals(status.getTaskId().getValue())) {
+                if (optionalTaskName.isPresent()) {
+                    logger.error("Found duplicate taskIDs in Task {} and  Task {}",
+                            optionalTaskName.get(), taskInfo.getName());
+                    throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
+                            "There are more than one tasks with TaskID: %s", status));
+                }
+                    optionalTaskName = Optional.of(taskInfo.getName());
```
gabrielhartmann: Indentation is incorrect.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,17 +192,27 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
+            }
         }
+        if (!optionalTaskName.isPresent()) {
+            throw new StateStoreException(Reason.NOT_FOUND, String.format(
```
gabrielhartmann: This is an undesirable behavior change.  See the comment from the old code.  It is a pretty common occurrence to receive stale TaskStatus messages.  Logging a warning would be fine, but throwing an exception isn't a good idea.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -193,17 +192,27 @@ public void storeTasks(Collection<Protos.TaskInfo> tasks) throws StateStoreExcep
...
+    }
 
-        // Validate that a TaskInfo with the exact same UUID is currently present. We intentionally
-        // ignore TaskStatuses whose TaskID doesn't (exactly) match the current TaskInfo: We will
-        // occasionally get these for stale tasks that have since been changed (with new UUIDs).
+    protected void putStatus(String taskName, Protos.TaskStatus status) throws StateStoreException {
+        //putStatus(taskName, status) is the only way to update status if there are duplicate taskIDs.
         Optional<Protos.TaskInfo> optionalTaskInfo;
         try {
             optionalTaskInfo = fetchTask(taskName);
```
gabrielhartmann: We're fetching the TaskInfo twice now.  In the loop above we already have the TaskInfo we want.

gabrielhartmann:
@mbdcos:  What's wrong with (this)[https://github.com/mesosphere/dcos-commons/compare/stateStore3]?

The `protected` `storeStatus` implementation you introduce here wasn't called by anything.  Once that is removed the `putStatus` method is used only by `storeStatus` so didn't need to be separate.

If you need custom `StateStore` logic for the Kafka migration, it should be done in `dcos-kafka-service`.  This includes extending / overriding methods in the `CuratorStateStore` if necessary.

mbdcos:
Re: _What's wrong with this?_

The diff  you sent is below. It removes protected new **storeStatus(name,status)** method. 

 Good, since TaskName generation from TaskID will be dropped since it  is a major requirement. So this PR is helpful.

I want to add _stateStore(name, status)_ too.  See my next messages for my explanation.

Lets discuss about the proposal of adding stateStatus(name,status) interface under the original PR #621, not in this one.

     @Override
     public void storeStatus(Protos.TaskStatus status) throws StateStoreException {
-        Optional<String> optionalTaskName = Optional.empty();
+        Optional<Protos.TaskInfo> taskInfoOptional = Optional.empty();
 
         for (Protos.TaskInfo taskInfo : fetchTasks()) {
             if (taskInfo.getTaskId().getValue().equals(status.getTaskId().getValue())) {
-                if (optionalTaskName.isPresent()) {
+                if (taskInfoOptional.isPresent()) {
                     logger.error("Found duplicate taskIDs in Task {} and  Task {}",
-                            optionalTaskName.get(), taskInfo.getName());
+                            taskInfoOptional.get(), taskInfo.getName());
                     throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
                             "There are more than one tasks with TaskID: %s", status));
                 }
-                optionalTaskName = Optional.of(taskInfo.getName());
+                taskInfoOptional = Optional.of(taskInfo);
             }
         }
 
-        if (!optionalTaskName.isPresent()) {
+        if (!taskInfoOptional.isPresent()) {
             throw new StateStoreException(Reason.NOT_FOUND, String.format(
                     "Failed to find a task with TaskID: %s", status));
         }
-        putStatus(optionalTaskName.get(), status);
-    }
-
-    /* storeStatus(taskName, status):
-                            is the only way to update status if there are duplicate taskIDs.
-     Protected method to be called directly using taskName (i.e. by a config *upgrade*). It does error checking before
-     writing status to TaskPath
-     */
-    protected void storeStatus(String taskName, Protos.TaskStatus status) throws StateStoreException {
-        Optional<Protos.TaskInfo> optionalTaskInfo;
-
-        try {
-            optionalTaskInfo = fetchTask(taskName);
-        } catch (Exception e) {
-            throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
-                    "Unable to retrieve matching TaskInfo for the provided TaskStatus name %s.", taskName), e);
-        }
-
-        if (!optionalTaskInfo.isPresent()) {
-            throw new StateStoreException(Reason.LOGIC_ERROR,
-                    String.format("The following TaskInfo is not present in the StateStore: %s. " +
-                            "TaskInfo must be present in order to store a TaskStatus.", taskName));
-        }
-
-        if (!optionalTaskInfo.get().getTaskId().getValue().equals(status.getTaskId().getValue())) {
-            throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
-                    "Task ID '%s' of updated status doesn't match Task ID '%s' of current TaskInfo."
-                            + " Task IDs must exactly match before status may be updated."
-                            + " NewTaskStatus[%s] CurrentTaskInfo[%s]",
-                    status.getTaskId().getValue(), optionalTaskInfo.get().getTaskId().getValue(),
-                    status, optionalTaskInfo));
-        }
-        putStatus(taskName, status);
-    }
-
-    private void putStatus(String taskName, Protos.TaskStatus status) throws StateStoreException {
 
+        String taskName = taskInfoOptional.get().getName();
         Optional<Protos.TaskStatus> currentStatusOptional = fetchStatus(taskName);
 
         if (currentStatusOptional.isPresent()

mbdcos:
* Re: _“The protected storeStatus implementation you introduce here wasn't called by anything. Once that is removed the putStatus method is used only by storeStatus so didn't need to be separate.”_

Original PR#621 introduced stateStore(name, status) method. Your review to PR#621 was to remove it from StateStore interface. When it was removed from interface definition, I left it as a protected method in CuratorStateStore. It will be overridden and called inside KafkaConfigUpgrade. See below:

```
     public static class NewCuratorStateStore extends CuratorStateStore {    
                public NewCuratorStateStoreUpdate(String name, String zk) {        
                           super(name, zk);     
                }     
                public void storeStatus(String taskName, Protos.TaskStatus status) 
                            throws StateStoreException {
                          super.storeStatus(taskName, status);     
                }
         }
```

If we drop the protected method, _NewCuratorStateStore_ in this example, must implement everything for storeStatus(name, status). It can only extend other methods. And I do not like this solution. See below:

  ```
  void storeStatus(String taskName, Protos.TaskStatus status) throws StateStoreException {
        Optional<Protos.TaskInfo> optionalTaskInfo;
        try {
            optionalTaskInfo = fetchTask(taskName);
        } catch (Exception e) {
            throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
                    "Unable to retrieve matching TaskInfo for the provided TaskStatus name %s.", taskName), e);
        }
        if (!optionalTaskInfo.isPresent()) {
            throw new StateStoreException(Reason.LOGIC_ERROR,
                    String.format("The following TaskInfo is not present in the StateStore: %s. " +
                            "TaskInfo must be present in order to store a TaskStatus.", taskName));
        }
        if (!optionalTaskInfo.get().getTaskId().getValue().equals(status.getTaskId().getValue())) {
            throw new StateStoreException(Reason.LOGIC_ERROR, String.format(
                    "Task ID '%s' of updated status doesn't match Task ID '%s' of current TaskInfo."
                            + " Task IDs must exactly match before status may be updated."
                            + " NewTaskStatus[%s] CurrentTaskInfo[%s]",
                    status.getTaskId().getValue(), optionalTaskInfo.get().getTaskId().getValue(),
                    status, optionalTaskInfo));
        } 
        Optional<Protos.TaskStatus> currentStatusOptional = fetchStatus(taskName);
        if (currentStatusOptional.isPresent()
                && status.getState().equals(Protos.TaskState.TASK_LOST)
                && CommonTaskUtils.isTerminal(currentStatusOptional.get())) {
            throw new StateStoreException(Reason.LOGIC_ERROR,
                    String.format("Ignoring TASK_LOST for Task already in a terminal state %s: %s",
                            currentStatusOptional.get().getState(), taskName));
        }
        String path = taskPathMapper.getTaskStatusPath(taskName);
        logger.info("Storing status '{}' for '{}' in '{}'", status.getState(), taskName, path);
        try {
            curator.set(path, status.toByteArray());
        } catch (Exception e) {
            throw new StateStoreException(Reason.STORAGE_ERROR, e);
        }
    }
```

Is there another way to not rewrite storeStatus(name, status) code entirely?

mbdcos:
* Re: “_If you need custom StateStore logic for the Kafka migration, it should be done in dcos-kafka-service. This includes extending / overriding methods in the CuratorStateStore if necessary._”

Yes Kafka changes are in Kafka.  And major SDK changes such as this one should be in SDK.

StateStore is a core functionality and I am not comfortable extending it outside of sdk.curator. 

There is of course always other ways. If introducing a new method in StateStore interface will break some other design requirements, we could ignore. 

* What is the reason to not have storeStatus(name,status) method in StateStore interface? What is it breaking?

* Why I was proposing to introduce storeStatus(name, status) in StateStore interface:

- It complements StateStore. If you do not know taskName, storeStatus(status) will search and find for you. If you know taskName, no need to search, you can use storeStatus(name, status) directly. 
Other than KafkaConfigUpgrade, there might be other use cases in the future.
    
- Reimplementing stateStore functionality of CuratorStateStore externally, outside of sdk.curator, breaks the design.
   - You get lock-in to CuratorStateStore, you can not use any other StateStore
   - If CuratorStateStore is dropped, or changed, you need to update your NewCuratorStateStore class
   - Since NewCuratorStateStore is outside, any maintainer will not know that you extend CuratorStateStore and you depend on it.

- This is our SDK and we should be able to update as necessary.

mbdcos:
retest this please

PR #642, opened 2017-03-02
[SDK][INFINITY-1058] Fix statestore inconsistencies on startup

gabrielhartmann:
It is possible in some failure scenarios for TaskInfo and TaskStatus
objects regarding the same Task to have inconsistent TaskIDs.  This
breaks dependent logic that assumes TaskInfo and TaskStatus for the
same Task will share a TaskID.

TaskState is set to TASK_FAILED in the case of any detected inconsistency to
cause a restart of the Task and arrive at a categorically consistent state with Mesos.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
@@ -6,8 +6,12 @@ import com.mesosphere.sdk.offer.CommonTaskUtils;
 import com.mesosphere.sdk.state.*;
 import com.mesosphere.sdk.storage.Persister;
 import com.mesosphere.sdk.storage.StorageError.Reason;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.curator.RetryPolicy;
 import org.apache.mesos.Protos;
+import org.apache.mesos.Protos.TaskInfo;
+import org.apache.mesos.Protos.TaskStatus;
+import org.apache.mesos.Protos.TaskState;
 import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


gabrielhartmann:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -7,9 +7,12 @@
 import com.mesosphere.sdk.state.*;
 import com.mesosphere.sdk.storage.Persister;
 import com.mesosphere.sdk.storage.StorageError.Reason;
-
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.curator.RetryPolicy;
 import org.apache.mesos.Protos;
+import org.apache.mesos.Protos.TaskInfo;
```
nickbp: Duplicate include vs `Protos`

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorStateStore.java
```diff
@@ -448,6 +452,38 @@ private String getTasksRootPath() {
         }
     }
 
+    @SuppressFBWarnings("UC_USELESS_OBJECT")
```
nickbp: Should add a comment explaining the scenario and what needs to be done to fix it (copy/paste from PR desc)

gabrielhartmann: It's commented two lines below.

nickbp: I mean describe the repair operation itself (bad choice of lines to comment on)

PR #645, opened 2017-03-03
[INFINITY-1074] Authorize marathon to launch tasks as root in strict mode tests

mgummelt:
I've also refactored `setup_permissions.sh` to be organized by permission, and I've changed the comments in that file, which were incorrect.

### DIFF
--- a/tools/setup_permissions.sh
+++ b/tools/setup_permissions.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-SERVICE_ACCOUNT_NAME=service-acct
-SECRET_NAME=secret
-LINUX_USER=$1
-ROLE=$2
+SERVICE_ACCOUNT_NAME=$1
+LINUX_USER=$2
+ROLE=$3
+
 if [ -n "$CLUSTER_URL" ]
 then
     # CI
@@ -11,6 +11,7 @@ then
 else
     DCOS_URL=$(dcos config show core.dcos_url)
 fi
+
 if [ -n "$CLUSTER_AUTH_TOKEN" ]
 then
     # CI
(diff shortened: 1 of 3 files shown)


mgummelt:
retest this please

susanxhuynh:
@mgummelt dcos-cassandra-service/test.sh and dcos-kafka-service/test.sh both call setup_permissions.sh. They would have to be updated, too.

mgummelt:
@susanxhuynh good catch.  I'll update those.

mgummelt:
That shouldn't block this PR, though, since this has to go in first.

mgummelt:
I just removed the interface change to `setup_permissions.sh`

mgummelt:
cc @susanxhuynh for approval.

tools/setup_permissions.sh
```diff
@@ -19,81 +20,94 @@ else
...
+         -d '{"description":"Execute Mesos tasks as user=${LINUX_USER}"}' \
+         -H 'Content-Type: application/json'
+    curl -k -L -X PUT \
+         -H "Authorization: token=${ACS_TOKEN}" \
+         "${DCOS_URL}/acs/api/v1/acls/dcos:mesos:master:task:user:${LINUX_USER}/users/${SERVICE_ACCOUNT_NAME}/create"
+
+    # In order for the Spark Dispatcher to register with Mesos as
+    # root, we must launch the dispatcher task as root.  The other
+    # frameworks are launched as nobody, but then register as
+    # service.user, which defaults to root
```
susanxhuynh: I'm not following this. Why can't Spark Dispatcher do the same as the other frameworks? Launch as nobody and then register as root?

mgummelt: The dispatcher registers as the current linux user.  It's not configurable.  We could make it configurable, but there's really no reason to.

susanxhuynh: OK

PR #647, opened 2017-03-03
[SDK:ELASTIC-81] Include "native" endpoints by default, pretty-print all returned JSON

nickbp:
- Deprecates the `format=native` argument in EndpointsResource, which was found to be unintuitive in practice. Now the native endpoints are just included in the default response under a `native` type. The CLI has been updated to temporarily return an error when the `--native` argument is used.
- Updates all JSON API responses (both `JSONArray`/`JSONObject` and Jackson beans) to produce nicely-formatted output.


nickbp:
PTAL

gabrielhartmann:
@nickbp:  What's the example output supposed to look like?  HDFS has custom endpoints, but I don't see "native" endpoints.  Does it need to be updated like Kafka?

nickbp:
- The 'native' output is only applied to *autogenerated* endpoint listings (created by presence of VIPs). Example for elastic:
```
$ dcos elastic endpoints
[
  "coordinator",
  "data",
  "kibana",
  "master",
  "ingest"
]
$ dcos elastic endpoints master
{
  "native": [
    "10.0.2.50:1025",
    "10.0.1.236:1025",
    "10.0.3.170:1025"
  ],
  "direct": [
    "master-1-server.elastic.mesos:1025",
    "master-0-server.elastic.mesos:1025",
    "master-2-server.elastic.mesos:1025"
  ],
  "vip": "master.elastic.l4lb.thisdcos.directory:9200"
}
```
- The kafka updates were just to have it pretty-print its output like the other in-scheduler resources

nickbp:
Do the integration tests test the kafka CLI?

mbdcos:
There is no integration test yet. They are going in in the next PR, waiting all clis and APIs to be approved first bec integration tests need to adapt based on what is available.

PR #648, opened 2017-03-03
[ELASTIC/HDFS][INFINITY-1019] Standardize strategy naming

gabrielhartmann:


gabrielhartmann:
@keithchambers:  Everything is updated.  The template doesn't have any strategies / custom deploy plan at all.

PR #651, opened 2017-03-04
[ELASTIC-45] Fix Elastic CLI help errors

loren:
- Changed boilerplate text
- Removed unused example section

### DIFF
--- a/frameworks/elastic/cli/dcos-elastic/main.go
+++ b/frameworks/elastic/cli/dcos-elastic/main.go
@@ -14,17 +14,16 @@ func main() {
 		log.Fatalf(err.Error())
 	}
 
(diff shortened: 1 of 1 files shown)


loren:
retest this please

keithchambers:
@triclambert I suspect most developers that use our SDK will miss this. Instead of solving this for Elastic, how about we eliminate the command description (e.g., "Provides an example for DC/OS service developers") altogether? Same thing for removing (or commenting out) the Example commands.

nickbp:
FWIW I did some of that so far, specifically removing the 'Example' sections, in #589 .

While I didn't remove the description strings in that PR, I agree that it'd also make sense to remove them as they're not really helpful to users and are yet another thing for a new developer to worry about updating

nickbp:
I'll do the remaining cleanup in a separte PR (tracking: INFINITY-1093)

PR #663, opened 2017-03-07
[SDK] Fix Darwin CLI build, add uninstall instructions to build.sh footer

nickbp:
- Darwin binary must not be UPX-encoded, or else it immediately produces `Killed: 9`
- Include all steps to reinstall a service in build output

### DIFF
--- a/tools/publish_aws.py
+++ b/tools/publish_aws.py
@@ -152,9 +152,6 @@ class AWSPublisher(object):
         # print universe url early
         universe_url = self._upload_artifact(universe_path)
         logger.info('---')
-        logger.info('Built and uploaded stub universe:')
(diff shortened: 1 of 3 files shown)


tools/publish_aws.py
```diff
@@ -152,9 +152,6 @@ def upload(self):
         # print universe url early
         universe_url = self._upload_artifact(universe_path)
         logger.info('---')
-        logger.info('Built and uploaded stub universe:')
-        logger.info(universe_url)
-        logger.info('---')
```
jsrodman: Unrelated?

nickbp: Yes, it's redundant given the install steps provided at the end

PR #664, opened 2017-03-07
[CCM] Don't abort launch if CCM returns status '8'

nickbp:


tools/launch_ccm_cluster.py
```diff
@@ -173,7 +177,7 @@ def wait_for_status(self, cluster_id, pending_status_label, complete_status_labe
                     else:
                         logger.error('Cluster {} has entered state {}, but lacks cluster_info...'.format(
                             cluster_id, status_label))
-                elif status_code != pending_state_code:
+                elif status_code != pending_state_code and not status_code in self._CCM_STATUS_IGNORED:
```
colin-msphere: Will this perform the same as `status_code not in self._CCM_STATUS_IGNORED`?

This could also be considered a pending state - could `pending_state_code` be modified to be a list/tuple and check `status_code in pending_state_codes`?

nickbp: Switched to multiple pending states

PR #666, opened 2017-03-08
Support specifying port ranges in task requirements

nickbp:
A shorthand for reserving a block of ports:

```
ports:
  foo:
    port-begin: 1000
    port-end: 1500
```

Results in default envvars of:

```
PORT_FOO_BEGIN = 1000
PORT_FOO_END = 1500
```

Note: blocks of dynamic ports are not supported -- the validator checks for this


keithchambers:
Let's discuss before merging this.

PR #671, opened 2017-03-08
[INFINITY-1100] Default strategy for phases/steps

benclarkwood:


nickbp:
Take a look at `DefaultPlanGenerator.generate()` which handles converting the `RawPlan` (raw/direct representation of the yaml tree) into a `Plan`

benclarkwood:
Right, that's basically one layer above StrategyFactory.

It seemed more cluttered to set defaults in `generate` or `from` than to do it in `StrategyFactory`.

PR #673, opened 2017-03-09
Scaffolding support for self hosted projects

mohitsoni:
SDK user can use following command to generate a standalone project that can be independently hosted:
```bash
./new-standalone-service.sh service-name /fs/parent/location
```

The script does following:
1. Generate the project structure from the template project
2. Switch the project to use binary dependencies
3. Initialize gradlew for the project

TODO:
~* Add documentation~

Request:
Please try this PR out on your local as part of your review before merging.


frameworks/template/build.self.hosted.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+
+# capture anonymous metrics for reporting
+curl https://mesosphere.com/wp-content/themes/mesosphere/library/images/assets/sdk/build-sh-start.png >/dev/null 2>&1
+
+FRAMEWORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+BUILD_DIR=$FRAMEWORK_DIR/build/distributions
+PUBLISH_STEP=${1-none}
+export REPO_NAME=template
```
nickbp: Should `REPO_NAME` just be the name of the current directory? `basename $FRAMEWORK_DIR`

I don't know the answer to this, to be clear

frameworks/template/build.self.hosted.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
nickbp: How about writing this file out from within `new-standalone-service.sh`? It's closely tied to the logic in there, could make sense to just emit it directly. See examples in `tools/vagrant/build-dcos-docker.sh`

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
+
+cleanup() {
+    # TODO(mohit): Maybe add a trap for SIGINT.
```
nickbp: trap example: https://github.com/mesosphere/dcos-commons/blob/master/tools/vagrant/build-dcos-docker.sh#L18

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+    exit 1
+fi
+
+UPPER_CASE_PROJECT_NAME=$(echo $PROJECT_NAME | awk '{print toupper($0)}')
+
+find $PROJECT_PATH/$PROJECT_NAME -type f -exec sed -i.bak "s/template/$PROJECT_NAME/g; s/TEMPLATE/$UPPER_CASE_PROJECT_NAME/g; s/template/$PROJECT_NAME/g" {} \;
+sed -i.bak -e '21,$ d' $PROJECT_PATH/$PROJECT_NAME/src/main/dist/svc.yml
+find $PROJECT_PATH/$PROJECT_NAME -type f -name *.bak -exec rm {} \;
+
+sed -i.bak "s/compile project(\":scheduler\")/compile \"mesosphere:scheduler:0.12.0\"/g" $PROJECT_PATH/$PROJECT_NAME/build.gradle
```
nickbp: This `0.12.0` version could be a separate variable at the top of the file

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+    rm -rf template.zip
+}
+
+info() {
+    MESSAGE=$1
+    echo $MESSAGE
+}
+
+debug() {
+    if [[ -z "${DEBUG// }" ]]; then
```
nickbp: Out of curiosity, what do the extra slashes/space do? I've just been using e.g. `-z "${DEBUG}"`

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+export REPO_NAME="$(basename $FRAMEWORK_DIR)"
+export BUILD_BOOTSTRAP=no
+export TOOLS_DIR=${FRAMEWORK_DIR}/tools
+export CLI_DIR=${FRAMEWORK_DIR}/cli
+export ORG_PATH=github.com/$REPO_NAME
+${FRAMEWORK_DIR}/tools/build_framework.sh $PUBLISH_STEP $REPO_NAME $FRAMEWORK_DIR $BUILD_DIR/$REPO_NAME-scheduler.zip
+
+# capture anonymous metrics for reporting
+curl https://mesosphere.com/wp-content/themes/mesosphere/library/images/assets/sdk/build-sh-finish.png >/dev/null 2>&1
+
```
nickbp: mega nitpick: extra newline

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+EOF
+
+chmod +x $PROJECT_PATH/$PROJECT_NAME/build.sh
+mv $PROJECT_PATH/$PROJECT_NAME/cli/dcos-template $PROJECT_PATH/$PROJECT_NAME/cli/dcos-$PROJECT_NAME
+mv $PROJECT_PATH/$PROJECT_NAME/src/main/java/com/mesosphere/sdk/template/ $PROJECT_PATH/$PROJECT_NAME/src/main/java/com/mesosphere/sdk/$PROJECT_NAME/
+mv $PROJECT_PATH/$PROJECT_NAME/src/test/java/com/mesosphere/sdk/template/ $PROJECT_PATH/$PROJECT_NAME/src/test/java/com/mesosphere/sdk/$PROJECT_NAME/
+
+if [ $? -ne 0 ]; then
```
nickbp: this only checks the outcome of the last `mv` operation. Maybe do a `set -e` and configure a trap so that any errors are caught/logged automatically? Similar to this: https://github.com/mesosphere/dcos-commons/blob/master/tools/vagrant/build-dcos-docker.sh#L18

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+GRADLE_EXISTS="yes"
+command -v gradle >/dev/null 2>&1 || { echo >&2 "I require gradle but it's not installed. Please install gradle and then run 'gradle wrapper --gradle-version 3.4' inside $PROJECT_PATH/$PROJECT_NAME"; GRADLE_EXISTS="no"; }
+
+if [ "$GRADLE_EXISTS" == "yes" ]; then
+    pushd $PROJECT_PATH/$PROJECT_NAME
+    gradle wrapper --gradle-version 3.4
+    popd
+fi
+
+if [ $? -eq 0 ]; then
```
nickbp: this likely just checks the outcome of `popd`, see above comment about `set -e` + trap

new-framework.sh
```diff
@@ -20,6 +20,7 @@ rm -rf $1/build
 rm -rf $1/cli/dcos-*/*.whl
 rm -rf $1/cli/dcos-*/dcos-*
 rm -rf $1/cli/python/{build,dist}
+rm -rf $1/build.self.hosted.sh
```
nickbp: file is no longer present

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
+
+cleanup() {
+    # TODO(mohit): Maybe add a trap for SIGINT.
+    rm -rf template.zip
```
nickbp: If this is added as part of a trap, you'll probably need to give an absolute path to delete. The script may be in some random directory at the time of an error...

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+}
+
+info() {
+    MESSAGE=$1
+    echo $MESSAGE
```
nickbp: just `echo $1` imo

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+    echo "Failed to create new project"
+    cleanup
+    exit 1
+fi
+
+UPPER_CASE_PROJECT_NAME=$(echo $PROJECT_NAME | awk '{print toupper($0)}')
+
+find $PROJECT_PATH/$PROJECT_NAME -type f -exec sed -i.bak "s/template/$PROJECT_NAME/g; s/TEMPLATE/$UPPER_CASE_PROJECT_NAME/g; s/template/$PROJECT_NAME/g" {} \;
+sed -i.bak -e '21,$ d' $PROJECT_PATH/$PROJECT_NAME/src/main/dist/svc.yml
+find $PROJECT_PATH/$PROJECT_NAME -type f -name *.bak -exec rm {} \;
```
nickbp: I hit this locally:
```
rm: remove write-protected regular file '/home/nick/code/dcos-commons/dse/tools/shakedown_env/src/shakedown/.git/objects/a2/59f08996c35899ce8205aae51f7f0052682ae7.bak'? y
rm: remove write-protected regular file '/home/nick/code/dcos-commons/dse/tools/shakedown_env/src/shakedown/.git/objects/a2/9e4d25d57dc360038d81ec0175a5eb46509958.bak'? y
[...]
```

To get around that, this command could be written with `rm -f`: `find $PROJECT_PATH/$PROJECT_NAME -type f -name *.bak -exec rm -f {} \;`

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+
+sed -i.bak "s/compile project(\":scheduler\")/compile \"mesosphere:scheduler:0.12.0\"/g" $PROJECT_PATH/$PROJECT_NAME/build.gradle
+sed -i.bak "s/compile project(\":executor\")/compile \"mesosphere:executor:0.12.0\"/g" $PROJECT_PATH/$PROJECT_NAME/build.gradle
+sed -i.bak "s/testCompile project(\":testing\")/testCompile \"mesosphere:testing:0.12.0\"/g" $PROJECT_PATH/$PROJECT_NAME/build.gradle
+sed -i.bak '/distZip.dependsOn ":executor:distZip"/d' $PROJECT_PATH/$PROJECT_NAME/build.gradle
+sed -i.bak '/distZip.finalizedBy copyExecutor/d' $PROJECT_PATH/$PROJECT_NAME/build.gradle
+
+GRADLE_EXISTS="yes"
+command -v gradle >/dev/null 2>&1 || { echo >&2 "I require gradle but it's not installed. Please install gradle and then run 'gradle wrapper --gradle-version 3.4' inside $PROJECT_PATH/$PROJECT_NAME"; GRADLE_EXISTS="no"; }
```
nickbp: This could just use the `gradlew` command provided at the root of `dcos-commons`. I was able to build a wrapper like this just fine:

`myproject$ path/to/dcos-commons/gradlew wrapper --gradle-version 3.4`

nickbp:
Other comments:
- New project needs to have `.gitignore` copied from `dcos-commons`
- There was an extra `build.gradle.bak` in the new project
- `gradlew.bat` could be removed from the new project as well (after calling `gradlew wrapper`)

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+
+if [ "$GRADLE_EXISTS" == "yes" ]; then
+    pushd $PROJECT_PATH/$PROJECT_NAME
+    gradle wrapper --gradle-version 3.4
```
nickbp: looks like there's a version `3.4.1` now.

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+curl https://mesosphere.com/wp-content/themes/mesosphere/library/images/assets/sdk/build-sh-start.png >/dev/null 2>&1
+
+FRAMEWORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+BUILD_DIR=$FRAMEWORK_DIR/build/distributions
+PUBLISH_STEP=${1-none}
+export REPO_NAME="$(basename $FRAMEWORK_DIR)"
+export BUILD_BOOTSTRAP=no
+export TOOLS_DIR=${FRAMEWORK_DIR}/tools
+export CLI_DIR=${FRAMEWORK_DIR}/cli
+export ORG_PATH=github.com/$REPO_NAME
```
nickbp: wouldn't this result in e.g. `ORG_PATH=github.com/dcos-commons` instead of `ORG_PATH=github.com/mesosphere`?

also it looks like this is overridden in `tools/build_go_exe.sh` so it might not actually be doing anything

nickbp: Confirmed: looks like omitting this didn't matter. BUT that was in an example where the organization was `mesosphere` anyway, which is what `tools/build_go_exe.sh` assumes.

Should probably just update `tools/build_go_exe.sh` to detect the repo path from `git remote -v`.

new-standalone-service.sh
mohitsoni: Those double slashes ensure that non-empty strings containing only spaces `     ` are also treated as empty.

Without those magical `//`, [[ -z "    " ]] will evaluate to `true`.

new-standalone-service.sh
```diff
@@ -0,0 +1,108 @@
...
+set -e
+
+# capture anonymous metrics for reporting
+curl https://mesosphere.com/wp-content/themes/mesosphere/library/images/assets/sdk/build-sh-start.png >/dev/null 2>&1
+
+FRAMEWORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+BUILD_DIR=$FRAMEWORK_DIR/build/distributions
+PUBLISH_STEP=${1-none}
+export REPO_NAME="$(basename $FRAMEWORK_DIR)"
+export BUILD_BOOTSTRAP=no
```
nickbp: How should the bootstrap executable be included in projects that need it? Should probably copy just bootstrap into the new repo so that it's available, or have the bootstrap build do a `go get` of `dcos-commons` to get it.

nickbp: In my local build I was able to do the following for now:
- Copy `dcos-commons/sdk/bootstrap/` into `myrepo/sdk/bootstrap/`
- Set this value to `true`

nickbp:
Currently missing some pieces for testing:
- A copy of `/testing/`. Should probably just copy that over like what's being done with `/tools/`.
- In addition, the project's `/tests/__init__.py` should be updated to point to `/testing/` wherever it goes.
- A basic `test.sh` could be added that just runs `tools/run_tests.py` against a required DC/OS URL (i.e. omit all the cluster launch login in `dcos-commons/test.sh`).

nickbp:
This could work for `test.sh`:

```
#!/usr/bin/env bash

set -e

    if [ -z "$1" ]; then
        echo "Syntax: $0 <cluster-url>, or CLUSTER_URL=<cluster-url> $0"
        exit 1
    fi
    export CLUSTER_URL=$1
fi

REPO_ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
${REPO_ROOT_DIR}/tools/run_tests.py shakedown ${FRAMEWORK_DIR}/tests/
```

nickbp:
Another thing I hit was an error when launching the service: `Error: Could not find or load main class com.mesosphere.sdk.myservice.scheduler.Main`

The solution was to create a `settings.gradle` in the root project dir with the following: `rootProject.name = 'myservice'`

Should probably a `settings.gradle` by default?

nickbp:
The external project also lacks an `executor.zip` in the build/upload, so the executors fail to launch.

new-standalone-service.sh
mohitsoni: We need to either pass the version to `wrapper` command OR put it inside the build.gradle.

new-standalone-service.sh
mohitsoni: Valid suggestions. I'm also a bit undecided about the scope of this. So, I decided of using $REPO_NAME as both org name and project name.

Another thing to consider is that a developer using this project might be on a private github instance OR using some other git hosting service.

To reduce the scope I went with this assumption.

new-standalone-service.sh
mohitsoni: I was thinking of publishing bootstrap executables as part of our build process. And then projects which need to use them, will just reference them inside `resource.json`.

How does that sound ?

tools/build_cli.sh
```diff
@@ -11,7 +11,8 @@ fi
 CLI_EXE_NAME=$1
 CLI_DIR=$2
 REPO_CLI_RELATIVE_PATH=$3 # eg 'frameworks/helloworld/cli/'
-TOOLS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+source $TOOLS_DIR/init_paths.sh
```
nickbp: The $TOOLS_DIR requirement on calling scripts can be removed by just doing the following:
```
TOOLS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
source $TOOLS_DIR/init_paths.sh
```

tools/build_framework.sh
```diff
@@ -25,9 +25,9 @@ echo ARTIFACT_FILES=$ARTIFACT_FILES
 CLI_DIR=${CLI_DIR:=${FRAMEWORK_DIR}/cli}
 UNIVERSE_DIR=${UNIVERSE_DIR:=${FRAMEWORK_DIR}/universe}
 CLI_EXE_NAME=${CLI_EXE_NAME:=dcos-${FRAMEWORK_NAME}}
+BUILD_BOOTSTRAP=${BUILD_BOOTSTRAP:=yes}
 
-TOOLS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-REPO_ROOT_DIR=$(dirname $TOOLS_DIR) # note: need an absolute path for REPO_CLI_RELATIVE_PATH below
+source $TOOLS_DIR/init_paths.sh
```
nickbp: The $TOOLS_DIR requirement on calling scripts can be removed by just doing the following:
```
TOOLS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
source $TOOLS_DIR/init_paths.sh
```

tools/build_go_exe.sh
```diff
@@ -8,17 +8,16 @@ if [ $# -lt 2 ]; then
     exit 1
 fi
 
-TOOLS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-REPO_ROOT_DIR=$(dirname $TOOLS_DIR)
+source $TOOLS_DIR/init_paths.sh
```
nickbp: The $TOOLS_DIR requirement on calling scripts can be removed by just doing the following:
```
TOOLS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
source $TOOLS_DIR/init_paths.sh
```

tools/init_paths.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+# exit immediately on failure
+set -e
+
+if [ -z "$TOOLS_DIR" ]; then
```
nickbp: It looks like callers are already effectively required to set `TOOLS_DIR` to something

tools/build_cli.sh
mohitsoni: `TOOLS_DIR` is required when working in context of standalone repos. In context of mono-repos it works fine. Also, because we had to use `TOOLS_DIR` in one of the non-tools dir scripts, I standardized on using `TOOLS_DIR` on all other scripts for consistency.

gabrielhartmann:
@mohitsoni:  Is this PR still relevant?  Looks like last comment was 13 days ago.

gabrielhartmann:
Open too long, re-open if still relevant.

benclarkwood:
@mohitsoni @gabrielhartmann This merge broke ongoing master:

```
BUILD SUCCESSFUL

Total time: 39.005 secs
+ ./tools/release.sh
./tools/release.sh: line 6: init_paths.sh: No such file or directory
Build step 'Execute shell' marked build as failure
Taking single-use slave mesos-jenkins-b5d8cb11b92c4a5f9aa91584e823ceae-infinity offline.
[ssh-agent] Stopped.
Sending e-mails to: infinity-ci@mesosphere.io
Finished: FAILURE
```

PR #681, opened 2017-03-09
[ELASTIC-43] Rename task names from *-server" to *-node" like other I…

benclarkwood:
…nfinity services

@nickbp Not sure if you're best reviewer or not. If not, tag them in.

### DIFF
--- a/frameworks/elastic/src/main/dist/svc.yml
+++ b/frameworks/elastic/src/main/dist/svc.yml
@@ -1,5 +1,5 @@
 name: {{FRAMEWORK_NAME}}
-web-url: http://proxylite-0-server.{{FRAMEWORK_NAME}}.mesos:{{PROXYLITE_PORT}}
+web-url: http://proxylite-0-node.{{FRAMEWORK_NAME}}.mesos:{{PROXYLITE_PORT}}
 scheduler:
   principal: {{FRAMEWORK_PRINCIPAL}}
   user: {{FRAMEWORK_USER}}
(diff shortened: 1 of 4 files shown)


frameworks/elastic/src/main/dist/svc.yml
```diff
@@ -289,9 +289,9 @@ pods:
       - {{SCHEDULER_URI}}
       - {{BOOTSTRAP_URI}}
     tasks:
-      server:
+      node:
```
nickbp: It may make sense to leave `proxylite` as-is, as it isn't really an ES component but instead something we're providing

benclarkwood: Yeah, I debated on that. It certainly makes more sense as a server (and it's a singleton). I'll change it back.

frameworks/elastic/src/main/dist/svc.yml
```diff
@@ -1,5 +1,5 @@
 name: {{FRAMEWORK_NAME}}
-web-url: http://proxylite-0-server.{{FRAMEWORK_NAME}}.mesos:{{PROXYLITE_PORT}}
+web-url: http://proxylite-0-node.{{FRAMEWORK_NAME}}.mesos:{{PROXYLITE_PORT}}
```
nickbp: revert to server

benclarkwood: Argh, I checked but didn't check good enough. Good catch.

PR #634, opened 2017-03-01
[INFINITY-1054] Fix suppression by only considering running plans whe…

mgummelt:
…n deciding when to suppress

Blocked on:
#751 
#752 
#753 
#754

PR #687, opened 2017-03-10
Reduce health check spam

nickbp:


mgummelt:
I'm fine with this if you want to merge, but I've found that simply printing the entire protobuf, which is what `shortDebugString` does, is too verbose.  I like to opt for explicitly printing the relevant information.  For example: https://github.com/mesosphere/dcos-commons/pull/634/files#diff-c33540a008b1ab3565a8cf8204535cb8R591

nickbp:
Makes sense. In this case the CommandInfo has the commandline + env and not much else so it's difficult to pick a subset of the data.

PR #692, opened 2017-03-12
Test runs should not by default create clusters and not tear them down.

jsrodman:
Possibly this might be okay as a build parameter.

It's not okay to, on developer machines, leave the clusters running by default.  How are they expected to clean it up when the CLUSTER_URL can be easily lost?

cc: @colin-msphere @gabrielhartmann @loren @susanxhuynh


jsrodman:
retest this please

jsrodman:
retest this please

PR #693, opened 2017-03-13
[INTINITY-1106] CNI Port Mapping, basic implementation and hello-world

ArtRand:
This PR allows `pods` to specify container-to-host port mappings using the mesos CNI and DC/OS overlay network interface.


ArtRand:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -540,32 +540,45 @@ private static boolean isPortResource(Protos.Resource resource) {
...
+        if (containerSpec != null && !containerSpec.getRLimits().isEmpty()) {
             containerInfo.setRlimitInfo(getRLimitInfo(containerSpec.getRLimits()));
         }
 
         return containerInfo.build();
     }
 
     private static Protos.NetworkInfo getNetworkInfo(NetworkSpec networkSpec) {
-        return Protos.NetworkInfo.newBuilder().setName(networkSpec.getName()).build();
+        LOGGER.info("Loading NetworkInfo for network named \"" + networkSpec.getNetworkName() + "\"");
```
mrbrowning: slf4j's `Logger`, at least, supports bracket-style formatting like so: 
`LOGGER.info("Loading NetworkInfo for network named \"{}\"", networkSpec.getNetworkName())`
which is a little easier to read.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultNetworkSpec.java
```diff
@@ -2,31 +2,55 @@
...
+        builder.portMap = copy.getPortMappings();
+
+        return builder;
+    }
+
+    @Override
+    public String getNetworkName() { return networkName; }
+
+    @Override
+    public Map<Integer, Integer> getPortMappings() { return portMappings; }
```
mrbrowning: I know it's a one-liner, but since IDEs auto-fold anyway house style seems to recommend a fully-indented block for a method definition like this.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultNetworkSpec.java
```diff
@@ -2,31 +2,55 @@
...
+    public static Builder newBuilder(NetworkSpec copy) {
+        Builder builder = new Builder();
+        builder.networkName = copy.getNetworkName();
+        builder.portMap = copy.getPortMappings();
+
+        return builder;
+    }
+
+    @Override
+    public String getNetworkName() { return networkName; }
```
mrbrowning: See below.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultNetworkSpec.java
```diff
@@ -2,31 +2,55 @@
...
+            @JsonProperty("port-mappings") Map<Integer, Integer> portMapings) {
+        this.networkName  = networkName;
+        this.portMappings = portMapings;
     }
 
-    @Override
-    public String getName() {
-        return name;
+    private DefaultNetworkSpec(Builder builder) {
+        this(
```
mrbrowning: Might as well be one line, it fits and doesn't sacrifice readability IMO.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
```diff
@@ -111,16 +123,17 @@ public Integer getCount() {
...
-    }
+    public List<TaskSpec> getTasks() { return tasks; }
 
     @Override
     public Optional<PlacementRule> getPlacementRule() {
         return Optional.ofNullable(placementRule);
     }
 
     @Override
+    public Optional<Collection<NetworkSpec>> getNetworks() { return Optional.ofNullable(networkSpecs); }
```
mrbrowning: With collections, we tend not to wrap in `Optional` and instead let something like `Collection.emptyList` represent the null case. That way, you don't need to branch on non-emptiness, and instead can just always iterate.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -540,32 +540,45 @@ private static boolean isPortResource(Protos.Resource resource) {
...
     }
 
-    private static Protos.ContainerInfo getContainerInfo(ContainerSpec containerSpec) {
+    private static Protos.ContainerInfo getContainerInfo(ContainerSpec containerSpec,
+                                                         Collection<NetworkSpec> networks) {
         Protos.ContainerInfo.Builder containerInfo = Protos.ContainerInfo.newBuilder()
                 .setType(Protos.ContainerInfo.Type.MESOS);
 
-        if (containerSpec.getImageName().isPresent()) {
+        if (containerSpec != null && containerSpec.getImageName().isPresent()) {
```
mrbrowning: Since optionality in the `containerSpec` parameter is already implicitly a part of this method's contract, why not just have it be of type `Optional<ContainerSpec>`? Saves an unwrapping in the caller that would need to happen here anyway in a different form via the `null` check.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawNetwork.java
```diff
@@ -1,6 +1,28 @@
...
+        this.hostPorts = hostPorts;
+        this.containerPorts = containerPorts;
+    }
+
+    public List<Integer> getHostPorts() { return hostPorts; }
+
+    public List<Integer> getContainerPorts() { return containerPorts; }
+
+    public Integer numberOfPortMappings() {
+        return Math.min(hostPorts.size(), containerPorts.size());
```
mrbrowning: Will this fail silently if the lengths of the two are different? As a framework writer I want to be yelled at if I mess this up as early as possible in the process.

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultServiceSpecTest.java
```diff
@@ -146,6 +148,59 @@ public void validReadinessCheck() throws Exception {
...
+        WriteOnceLinkedHashMap<String, RawNetwork> rawNetworkMap = rawServiceSpec
+                .getPods()
+                .get("meta-data")
+                .getNetworks();
+        Assert.assertTrue(MapUtils.isNotEmpty(rawNetworkMap));
+        RawNetwork rawNetwork = rawNetworkMap.get("dcos");
+        ArrayList<Integer> expectedHostPorts = new ArrayList<>();
+        expectedHostPorts.add(4040);
+        ArrayList<Integer> expectedContainerPorts = new ArrayList<>();
+        expectedContainerPorts.add(8080);
```
mrbrowning: Can you split this into a few paragraphs, maybe one per logical assertion?

sdk/common/src/main/java/com/mesosphere/sdk/testutils/TestConstants.java
```diff
@@ -39,6 +39,11 @@
     public static final Integer PORT_API_VALUE = 8080;
     public static final String HAS_DYNAMIC_PORT_ASSIGNMENT_LABEL = "has-dynamic-port-assignment";
     public static final String HAS_VIP_LABEL = "hasvip";
+    // CNI port mapping constants
+    public static final String EXPECTED_NETWORK_NAME = "dcos";
```
mrbrowning: Since it's understood that we'll both be creating fake objects with these values and then testing that the protobufs still contain them on the other end, probably no need to prefix them with `EXPECTED_`.

PR #697, opened 2017-03-14
Adds @PositiveScalarProtoValue annotation

mohitsoni:
When used on a Scalar Protos.Value this annotation will validate to true
if the value of the scalar is > 0. This is useful for validation of
resources like CPUs, Memory and Disk. We don't want to accidently
reserve a zero valued CPU, Memory and Disk resource.


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultResourceSpec.java
```diff
@@ -23,6 +24,7 @@
     @Size(min = 1)
     private final String name;
     @NotNull
+    @NonZeroScalarProtoValue
```
gabrielhartmann: I'd change this to `PositiveScalar` or something similar.  `NonZero` could mean negative, and the annotation as a whole is a bit verbose.

mohitsoni:
retest this please

gabrielhartmann:
retest this please

PR #708, opened 2017-03-14
[CASSANDRA-554] Remove JMX and Remote Statsd from new SDK based Cassandra Service

benclarkwood:
Setup Cassandra to report metrics to DC/OS StatsD.

### DIFF
--- a/frameworks/cassandra/src/main/dist/metrics-reporter-config.yaml
+++ b/frameworks/cassandra/src/main/dist/metrics-reporter-config.yaml
@@ -1,0 +1,6 @@
+statsd:
+  - period: {{STATSD_PERIOD}}
+    timeunit: 'SECONDS'
+    hosts:
+      - host: {{STATSD_UDP_HOST}}
+        port: {{STATSD_UDP_PORT}}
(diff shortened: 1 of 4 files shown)


benclarkwood:
@keithchambers The only configurable statsd thing is the reporting period (defaults to 60 seconds).

Deprecation notice is a separate PR against dcos-cassandra-service.

keithchambers:
Based on my understanding of how DC/OS Metrics works, changing the sample rate will have little effect since the DC/OS Metrics Agent caches the latest value and an http client needs to pool the metrics API (likely at some interval) to fetch the latest value.

Can you check with Jeff to understand if changing the sample rate will deliver enough customer value or if it would be acceptable to KISS and hard code to a value like 15 seconds?

benclarkwood:
Checking with Jeff

benclarkwood:
@keithchambers Jeff says:

Jeff Malnick [9:38 AM] 
The metrics service only exposes the latest values, we don't cache (yet).
So send as often as you want to scrap the HTTP API where it's exposed

Which makes me think it is a useful knob to have exposed.

benclarkwood:
Talked with Keith. We're going to go with a hardcoded 10 second default.

benclarkwood:
retest this please

gabrielhartmann:
@benclarkwood:  Where is the 10s hard-coded?

PR #713, opened 2017-03-15
[HDFS-446] Colocate name node with zkfc

benclarkwood:


build.gradle
```diff
@@ -26,7 +26,7 @@ allprojects {
     // at ANY TIME as changes are merged into master. They represent a fully unstable API.
     // ---
     group = "mesosphere"
-    version = "0.12.0"
+    version = "0.13.0-SNAPSHOT"
```
gabrielhartmann: Should be rebased on `master`, as this change is already on `master`.

PR #717, opened 2017-03-16
[Infinity-1012] parallelize pr tests

jsrodman:
Goals:
 * run tests faster in parallel
 * Better logging (timestamps!)
 * Render failures unambiguously via junit xml (not done yet)
 * Try to reduce the number of environment variables floating around.

### DIFF
--- a/test.py
+++ b/test.py
@@ -13,6 +13,15 @@ import time
 def get_repo_root():
     return os.path.dirname(sys.argv[0])
 
+
+logger = logging.getLogger("dcos-commons-test")
+logging.basicConfig(level=logging.INFO, format="%(asctime)s %(name)s %(message)s blob")
+
+sys.path.append(os.path.join(get_repo_root(), 'tools'))
+import clustinfo
+import fwinfo
+import launch_ccm_cluster
+
 work_dir = None
 def get_work_dir():
     global work_dir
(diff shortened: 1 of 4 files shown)


susanxhuynh:
Looks good to me.

test.py
```diff
@@ -0,0 +1,467 @@
...
+            help="requires a prior build")
+    parser.add_argument("--build-only", action='store_false', dest='run_tests')
+    parser.add_argument("--parallel", action='store_true',
+            help="Use more than one cluster to try to speed up the tests")
+    parser.add_argument("--cluster-count", type=int, default=2,
+            help="Ignored unless using --parallel.  "
+                 "Number of clusters to run tests on. "
+                 "Large values are likely to break CCM.")
+    parser.add_argument("--order", choices=('random', 'ordered'), default='random',
+            help="Run tests in randomr order, or the order given on the command line.  "
```
loren: typo nit: randomr

tools/fwinfo.py
```diff
@@ -0,0 +1,74 @@
...
+def init_repo_root(repo_root):
+    global _repo_root
+    _repo_root=repo_root
+
+def add_framework(framework_name, repo_root=None):
+    if not repo_root:
+        repo_root=_repo_root
+    logger.debug("add_framework(%s, repo_root=%s)", framework_name, repo_root)
+    if framework_name in _framework_infos:
+        raise Exceptoin("Tried to add framework %s when it already exists")
```
loren: Exception

jsrodman:
BTW @loren , definitely fixing those things you pointed out; just merged test.py so arthur could use it.

test.py
jsrodman: the ordering isn't implemented here, the full args were more or less trying to plan what I needed to support, and the pull request was kind of a minimum "this runs at all, and please stop me if this is awful"

PR #718, opened 2017-03-16
[HDFS] Use Cli, it already handles 503 status when querying for plans

mbdcos:
Better than HTTP exceptions. You can see spin time, and also view plan details (in console)


mbdcos:
+1 @gabrielhartmann

PR #722, opened 2017-03-16
[INFINITY-1114] Hello-world does not restart tasks when upgrading to new version

benclarkwood:
Begin considering the executor as a part of a pods configuration (spoiler, it already was)


benclarkwood:
retest this please

benclarkwood:
I did it the more better way.

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -275,7 +275,6 @@ def fn():
     zk_config_new = shakedown.get_zk_node_data(zk_path)
     assert zk_config_old == zk_config_new
 
-
```
gabrielhartmann: Should be 2 new lines.

PR #726, opened 2017-03-17
Use 'address'/'dns' in endpoint responses

nickbp:
Discards current confusing 'direct' / 'native' and returns to prior responses from 'connections' calls.


mbdcos:
didnt replace this in tests?

nickbp:
good point, grepping them now...

mbdcos:
You can merge this PR as it is if you do not want to wait for another Jenkins run. 
PR # 728 will update endpoints in tests.

nickbp:
I'll just close this one as #728 has this fix in it. No reason to break the build in the interim :)

PR #727, opened 2017-03-17
Infinity 1130 launchccm permissions templates

jsrodman:
quick fix; not sure why this shows all changes in branch instead of actually diffing against master; change is just https://github.com/mesosphere/dcos-commons/commit/aefc174664875e21102bea48befcb7725665e7b6 at this point

### DIFF
--- a/tools/launch_ccm_cluster.py
+++ b/tools/launch_ccm_cluster.py
@@ -204,6 +204,7 @@ class CCMLauncher(object):
 
     def _start(self, config):
         is_17_cluster = config.ccm_channel in self._DCOS_17_CHANNELS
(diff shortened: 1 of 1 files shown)


jsrodman:
@susanxhuynh : I don't know; I called that out in my original comment.

I reused the branch, which could have contributed (we reused the jira after all), but the branch has *every change* in it from master, so... not sure how we're diffing.

susanxhuynh:
@jsrodman OK, hopefully, it will figure this out when merging.

jsrodman:
Git merging the same changes will be fine, but it's disturbing for sure.

PR #730, opened 2017-03-18
[INFINITY-1128] Reset github statuses on retest, and calculate/emit operation duration

nickbp:
- Status reset: Github doesn't allow removing old statuses, only updating statuses. Therefore we just set all non-`velocity` statuses as pending in github if a build is retriggered (via 'retest this please'). Note that this doesn't fix things in all situations: If the old build is still running in jenkins then it can still override the reset-to-pending with additional updates.
- Timers: Stores a timestamp against the operation name when a step calls `pending`, then compares that timestamp when `failed` or `success` is called for that same operation name. End result is we can log duration for each operation.

### DIFF
--- a/tools/github_update.py
+++ b/tools/github_update.py
@@ -174,20 +174,23 @@ class GithubStatusUpdater(object):
 
     def _pretty_duration(self, seconds):
         """ Returns a user-friendly representation of the provided duration in seconds.
-        For example: 62.8 => "1m2.8s", or 129837.8 => "2d12h4m57.8s"
+        For example: 62.8 => "1m3s", or 129837.8 => "1d12h"
         """
-        ret = ''
         if seconds >= 86400:
-            ret += '{:.0f}d'.format(int(seconds / 86400))
-            seconds = seconds % 86400
+            # >= 1day: just do XdYh without minutes/seconds. int cast on days to avoid rounding up.
+            return '{:.0f}d{:.0f}h'.format(int(seconds / 86400), (seconds % 86400) / 3600)
         if seconds >= 3600:
-            ret += '{:.0f}h'.format(int(seconds / 3600))
-            seconds = seconds % 3600
+            # >= 1hr: just do XhYm without seconds. int cast on hours to avoid rounding up.
+            return '{:.0f}h{:.0f}m'.format(int(seconds / 3600), (seconds % 3600) / 60)
+
+        ret = ''
         if seconds >= 60:
+            # int cast to avoid rounding up:
             ret += '{:.0f}m'.format(int(seconds / 60))
             seconds = seconds % 60
         if seconds > 0:
-            ret += '{:.1f}s'.format(seconds)
+            # round to nearest second:
+            ret += '{:.0f}s'.format(seconds)
         return ret
 
 
--- a/tools/github_update.py
+++ b/tools/github_update.py
@@ -16,6 +16,8 @@ import pprint
 import re
 import sys
 import subprocess
+import tempfile
+import time
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
@@ -26,11 +28,17 @@ except ImportError:
     # Python 2
     from httplib import HTTPSConnection
 
-class GithubStatusUpdater(object):
+'''reserved contexts which may not be modified by this script'''
+BLACKLISTED_CONTEXT_LABELS = ['velocity'] # not set by us, do not update
 
-    def __init__(self, context_label):
-        self._context_label = context_label
+'''the pending/incomplete state'''
+PENDING_STATE = 'pending'
 
+'''list of all valid states'''
+VALID_STATES = [PENDING_STATE, 'success', 'error', 'failure']
+
+
+class RepoInfo(object):
 
     def _get_dotgit_path(self):
         '''returns the path to the .git directory for the repo'''
@@ -48,7 +56,7 @@ class GithubStatusUpdater(object):
                         'Run this command within the git repo, or provide GIT_REPOSITORY_ROOT')
 
 
-    def _get_commit_sha(self):
+    def commit_sha(self):
         '''returns the sha1 of the commit being reported upon'''
         # 1. try 'ghprbActualCommit', 'GIT_COMMIT', and 'sha1' envvars:
         commit_sha = os.environ.get('ghprbActualCommit', '')
@@ -73,7 +81,7 @@ class GithubStatusUpdater(object):
         return commit_sha
 
 
-    def _get_repo_path(self):
+    def repo_orgname(self):
         '''returns the repository path, in the form "mesosphere/some-repo"'''
         repo_path = os.environ.get('GITHUB_REPO_PATH', '')
         if repo_path:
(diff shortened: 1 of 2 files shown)


nickbp:
retest this please (check reset works)

nickbp:
retest this please

mgummelt:
> If the old build is still running in jenkins then it can still override the reset-to-pending with additional updates.

Can we check the latest commit in the Pull Request, and only update the status if it matches the commit in the build?

nickbp:
In the 'retest this' scenario, multiple builds may be running against the latest commit

nickbp:
To be clear, the 'PR status' is actually assigned on a per-commit basis rather than a per-PR basis. So if there's a build running against an older commit in a PR it won't collide with the status of the current commit. The main issues come up with re-running tests against the same commit an additional time.

mgummelt:
Oh I see, that should be fine then.

tools/github_update.py
```diff
@@ -96,118 +104,235 @@ def _get_repo_path(self):
...
         print_help(argv)
         return 1
+
+    updater = GithubStatusUpdater()
     state = argv[1]
-    if state != 'pending' \
-            and state != 'success' \
-            and state != 'error' \
-            and state != 'failure':
+    if state == 'reset':
```
mgummelt: "reset" isn't a `state`.  Maybe this variable should be renamed.

mgummelt: For readability, can you factor each of these two stanzas into its own function?

tools/github_update.py
```diff
@@ -96,118 +104,235 @@ def _get_repo_path(self):
...
+        if not _should_access_github():
+            return
+
+        message = ' '.join(argv[2:])
```
mgummelt: What is this message for?  It seems to be never used?

tools/github_update.py
```diff
@@ -96,118 +104,235 @@ def _get_repo_path(self):
...
+        logger.error('Request:')
+        logger.error(pprint.pformat(request))
+        logger.error('Response:')
+        logger.error(pprint.pformat(response.read().decode('utf-8')))
+        return False
 
 
-        request = self._build_request(state, message, details_url)
+    def list_contexts(self):
+        '''returns a list of context labels of all statuses in a given commit'''
```
mgummelt: s/list/set

tools/github_update.py
```diff
@@ -96,118 +104,235 @@ def _get_repo_path(self):
...
+        request = self._build_get_list_request()
```
mgummelt: The following four lines should probably be abstracted into a `GithubAPI.get_status_list()` method.  It's unnecessarily low-level as it is.  Clients shouldn't have to deal with request construction and error checking.  See the `dcos` package for an example of abstracting out a remote API.

tools/github_update.py
```diff
@@ -96,118 +104,235 @@ def _get_repo_path(self):
...
 
-    def update(self, state, message, details_url = ''):
-        '''sends an update to github.
-        returns True on success or False otherwise.
-        state should be one of 'pending', 'success', 'error', or 'failure'.'''
-        logger.info('[STATUS] {} {}: {}'.format(self._context_label, state, message))
-        if details_url:
-            logger.info('[STATUS] URL: {}'.format(details_url))
+    def _pretty_duration(self, seconds):
+        """ Returns a user-friendly representation of the provided duration in seconds.
```
mgummelt: Can we just use str(datetime.timedelta(seconds=seconds))?

tools/github_update.py
```diff
@@ -96,118 +104,235 @@ def _get_repo_path(self):
...
-            logger.error(pprint.pformat(request))
-            logger.error('Response:')
-            logger.error(pprint.pformat(response.read()))
+        if not self._check_success(response, request, 'status list'):
+            return []
+        return set([status['context'] for status in json.loads(response.read().decode('utf-8'))])
+
+
+    def update(self, state, message='', details_url='', context_label=''):
+        '''sends a commit status update to github.
```
mgummelt: Same here.  Can we factor out all Github API calls into a `GithubAPI` class?

nickbp: Done

tools/github_update.py
nickbp: Sure. Done

tools/github_update.py
nickbp: Broke out the lower-level half into a separate GithubAPI class

tools/github_update.py
nickbp: Done

tools/github_update.py
nickbp: It's a message to be stored against the status. The message variable is passed to updater.update()

tools/github_update.py
nickbp: Done

nickbp: Done, switched to 'command'

tools/github_update.py
```diff
@@ -96,118 +105,238 @@ def _get_repo_path(self):
...
-        return encoded_tok.decode('utf-8').rstrip('\n')
+        return github_token
+
 
+class GithubAPI(object):
 
-    def _build_request(self, state, message, details_url = ''):
-        '''returns everything needed for the HTTP request, except the auth token'''
+    def __init__(self, repo_orgname, commit_sha, github_token):
+        info = RepoInfo()
```
mgummelt: unused

nickbp: Whoops, fixed

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
+    logger.info('- Reset pending: {} reset [a replacement message ...]'.format(argv[0]))
+
+
+def reset_states(updater, message):
+    if not _should_access_github():
+        # reset disabled due to local build, exit silently
+        return 0
+
+    contexts = [context for context in updater.list_contexts() if context not in BLACKLISTED_CONTEXT_LABELS]
+    if not contexts:
```
mgummelt: This check should be unnecessary.

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
+
+    if _should_access_github():
+        logger.info('[GH-STATUS] {}'.format(logmsg))
+        updater.update(state, message=message, context_label=context_label, details_url=details_url)
+    else:
+        logger.info('[STATUS] {}'.format(logmsg))
+    return 0
 
 
 def main(argv):
```
mgummelt: Can't we just do a blanket return 0 here if `!_should_access_github()`?

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
+    if not _should_access_github():
+        # reset disabled due to local build, exit silently
+        return 0
+
+    contexts = [context for context in updater.list_contexts() if context not in BLACKLISTED_CONTEXT_LABELS]
+    if not contexts:
+        # nothing to reset, exit silently
+        return 0
+
+    contexts.sort()
```
mgummelt: To be functional, use `sorted` instead of `sort()`.

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
-    state = argv[1]
-    if state != 'pending' \
-            and state != 'success' \
-            and state != 'error' \
-            and state != 'failure':
+
+    updater = GithubStatusUpdater()
+    command = argv[1]
+    if command == 'reset':
+        return reset_states(updater, ' '.join(argv[2:]))
```
mgummelt: I get that argv[2:] is a message, but we never use it, right?  So why have it?

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
-            logger.error('Response:')
-            logger.error(pprint.pformat(response.read()))
-            return
-        logger.info('Updated GitHub PR with status: {}'.format(request['path']))
+        self._check_success(response, request, 'status update')
+
+
+class GithubStatusUpdater(object):
+
+    def __init__(self, default_context_label = ''):
```
mgummelt: `default_context_label` doesn't seem to be used anymore.

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
-        if not details_url:
-            raise Exception(
-                'Failed to determine URL for details link. ' +
-                'Provide either GITHUB_COMMIT_STATUS_URL or BUILD_URL in env.')
-        return details_url
-
-
-    def _get_auth_token(self):
+    def github_auth_token(self):
         github_token = os.environ.get('GITHUB_TOKEN_REPO_STATUS', '')
```
mgummelt: github_token = os.environ.get('...') or os.environ.get('...')

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
 
-    def _build_request(self, state, message, details_url = ''):
-        '''returns everything needed for the HTTP request, except the auth token'''
+    def __init__(self, repo_orgname, commit_sha, github_token):
+        self._repo_orgname = repo_orgname
+        self._commit_sha = commit_sha
+        self._github_token = github_token
+
+
+    def _build_post_update_request(self, context_label, state, message, details_url):
```
mgummelt: s/update/status

This method is posting statuses.

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
+
+    def _build_get_list_request(self):
+        '''returns everything needed for an HTTP request to retrieve repo statuses, except the auth token'''
+        return {
+            'method': 'GET',
+            'path': '/repos/{}/commits/{}/statuses'.format(self._repo_orgname, self._commit_sha),
+            'headers': {
+                'User-Agent': 'github_update.py',
+                'Content-Type': 'application/json',
+                'Authorization': 'Basic HIDDENTOKEN'} # replaced within update_query
```
mgummelt: Why even set this if _send_request is going to override it?

tools/github_update.py
```diff
@@ -96,118 +105,237 @@ def _get_repo_path(self):
...
                 'User-Agent': 'github_update.py',
                 'Content-Type': 'application/json',
                 'Authorization': 'Basic HIDDENTOKEN'}, # replaced within update_query
-            'payload': {
-                'context': self._context_label,
-                'state': state,
-                'description': message,
-                'target_url': self._get_details_link_url(details_url)
-            }
+            'payload': payload
```
mgummelt: This return value is a very unusual and indirect way of creating a request.  A traditional remote API client would have a method like `_post_status(state, message, ...)`, which constructs and sends the request.  Why have this intermediate stage where you're creating this dict that gets interpreted and decomposed in _send_request?

tools/github_update.py
nickbp: Would prefer continuing to emit status during local builds

tools/github_update.py
nickbp: This is returning an empty string if the envvar is empty/unset

tools/github_update.py
nickbp: it's building post update requests

tools/github_update.py
nickbp: Being explicit that a token would have been provided in error logs, without exposing the token in the logs

tools/github_update.py
mgummelt: Yea, and if it's set to an empty string, the following conditional will execute as true.  The above one liner has the same effect.

tools/github_update.py
mgummelt: Sorry, I meant it's "buliding commit status" requests, but the method name doesn't say anything about commit statuses.

tools/github_update.py
mgummelt: not a blocker, though

tools/github_update.py
mgummelt: not a blocker

tools/github_update.py
mgummelt: not a blocker

tools/github_update.py
mgummelt: I don't think it communicates anything to the local user that they don't already know, but not a big deal. not a blocker.

tools/github_update.py
nickbp:
```
$ grep -R GithubStatusUpdater *
tools/publish_http.py:        self._github_updater = github_update.GithubStatusUpdater('upload:{}'.format(package_name))
tools/github_update.py:class GithubStatusUpdater(object):
tools/github_update.py:    updater = GithubStatusUpdater()
tools/publish_aws.py:        self._github_updater = github_update.GithubStatusUpdater('upload:{}'.format(package_name))
tools/run_tests.py:        self._github_updater = github_update.GithubStatusUpdater('test:{}'.format(github_label))
tools/launch_ccm_cluster.py:        self._github_updater = github_update.GithubStatusUpdater('cluster:{}'.format(github_label))
```

tools/github_update.py
nickbp: Oh I see, I didn't see the following lines. Done here and elsewhere.

tools/github_update.py
nickbp: I *think* the original theory was to build a request object that could be logged if the request failed. But looking at it again now, this capability adds a lot of complexity.

Dropped that 'feature' in favor of much simpler structure.

tools/github_update.py
nickbp: These functions have gone away via the structural cleanup

tools/github_update.py
nickbp: Felt like a nice to have, in case we wanted to say 'status reset' or similar in the cleared status. Didn't add much complexity.

tools/github_update.py
nickbp: Done

tools/github_update.py
nickbp: The check is just to avoid the following message saying that we're resetting statuses, if there are no statuses to be reset

mgummelt:
Thanks!

nickbp:
retest this please

nickbp:
retest this please

nickbp:
retest this please

PR #732, opened 2017-03-20
[SDK][KAF-285] Fix port envvars

gabrielhartmann:
Unclear why the style fixes were needed in this PR and missed in other PRs.  The functional change is related to `CommandUtils.java`.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/CommandUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/CommandUtils.java
@@ -1,7 +1,9 @@
 package com.mesosphere.sdk.offer;
 
-import org.apache.mesos.Protos;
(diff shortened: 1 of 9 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/CommandUtils.java
```diff
@@ -1,20 +1,23 @@
...
 
 /**
  * This class provides utility methods for the construction of {@link org.apache.mesos.Protos.CommandInfo} protobufs
  * from {@link CommandSpec}s.
  */
 public class CommandUtils {
 
-    public static Protos.CommandInfo addEnvVar(
-            Protos.CommandInfo command,
+    public static Protos.CommandInfo setEnvVar(
```
benclarkwood: Why return a CommandInfo rather than a CommandInfo.Builder when you take the latter?

PR #733, opened 2017-03-20
[INFINITY-1110] Minor cleanup of port evaluation

nickbp:
- Evaluate port envvar name at offer time, rather than at spec conversion time. Gets port envvar handling all in one place.
- Move YAML -> spec conversion of ports from `DefaultResourceSet` to `YAMLToInternalMappers` to align with other handling
- Rename `PortsEvaluationStage` to `MultiEvaluationStage` as it's not inherently port-specific
- Minor formatting fixes to offer eval logs
- In placement constraint handling, refrain from passing `null` collection, instead pass an empty collection.


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
```diff
@@ -150,9 +151,10 @@ public EvaluationOutcome evaluate(MesosResourcePool mesosResourcePool, PodInfoBu
 
         return pass(
                 this,
-                offerRecommendation == null ? null : Arrays.asList(offerRecommendation),
-                "Offer contains sufficient '%s'",
-                resourceRequirement.getName());
+                offerRecommendation == null ? Collections.emptyList() : Arrays.asList(offerRecommendation),
+                "Offer contains sufficient '%s': requirement=%s",
+                resourceRequirement.getName(),
+                TextFormat.shortDebugString(resourceRequirement.getValue()));
```
mrbrowning: I wonder if there's a better way to include these -- obviously you want some record of the real protos being compared in the logs, but when they're mixed up in logging about what _happened_ it gets annoying to have to read around them. I wonder if we could structure things so that they're presented first, with some ad hoc ID (the way that offers are) and then referenced by that ID in actual evaluation logging.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/CommandUtils.java
```diff
@@ -11,16 +11,22 @@
...
         Map<String, String> envMap = CommonTaskUtils.fromEnvironmentToMap(builder.getEnvironment());
         envMap.put(key, value);
         builder.setEnvironment(CommonTaskUtils.fromMapToEnvironment(envMap));
-        return builder.build();
     }
 
+    /**
+     * Returns the value of the provided environment variable, or {@code null} if no matching environment variable was
+     * found.
+     */
```
nickbp: Good point, switched to Optional

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/PortRequirement.java
```diff
@@ -2,31 +2,41 @@
...
+import java.util.Optional;
+
 import org.apache.mesos.Protos;
 
 /**
  * A {@link PortRequirement} encapsulates a needed {@link MesosResource} representing a port.
  */
 public class PortRequirement extends ResourceRequirement {
-    private final String envKey;
+
```
nickbp: Reverted to fit format elsewhere

mgummelt:
Can you explain how this fixes the endpoint problem, which is the subject of the JIRA?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -359,12 +361,70 @@ private static DefaultResourceSet from(
...
+        return new PortsSpec(
+                Constants.PORTS_RESOURCE_TYPE, portsValueBuilder.build(), role, principal, envKey, portSpecs);
+    }
+
+    /**
+     * This visibility information isn't currently used by DC/OS Service Discovery. At the moment it's only enforced in
+     * our own {@link com.mesosphere.sdk.api.EndpointsResource}.
+     */
+    private static DiscoveryInfo.Visibility toVisibility(Boolean rawIsVisible) {
+        if (rawIsVisible == null) {
```
mgummelt: Other `Raw*` classes, such as `RawPod` do null conversions in the constructor, so we can get rid of null values early.  Maybe we should do that here.  Then `PUBLIC_VIP_VISIBILITY` can go in `RawVIP`.  And it should probably be called `DEFAULT_VIP_VISIBILITY`.  That's my understanding of what it is.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -359,12 +361,70 @@ private static DefaultResourceSet from(
...
+                final String vipName = StringUtils.isEmpty(rawVip.getPrefix()) ? name : rawVip.getPrefix();
+                portSpecs.add(new NamedVIPSpec(
+                        Constants.PORTS_RESOURCE_TYPE,
+                        portValueBuilder.build(),
+                        role,
+                        principal,
+                        rawPort.getEnvKey(),
+                        name,
+                        protocol,
+                        toVisibility(rawVip.isAdvertised()),
```
mgummelt: Is the `advertised` YAML field documented?  It doesn't seem to be documented in the dev guide.  What does it mean to "advertise" a VIP?

nickbp:
Re relevancy: this precedes https://github.com/mesosphere/dcos-commons/commits/endpoints-without-vips

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
nickbp: It's intentionally undocumented right now for that reason. At the moment all it does is hide the VIP in the endpoints endpoint when set to false (default is true). It'd probably make sense to migrate it to the port level in the next PR, when non-VIPed ports will also be listed in that endpoint. Or just remove it until something actually needs it..

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceEvaluationStage.java
nickbp: Alternately the evaluation could store the logged info to a buffer instead of logging directly, then logged as a block at the end of the evaluation cycle. Somewhat exceeding the scope of this PR though.

PR #734, opened 2017-03-20
[SDK] Be explicitly non-greedy when consuming sha256 tags

nickbp:
A user encountered this interpreting `{{sha256:myfile}} ... {{someothertag}}` as `myfile }} ... {{someothertag` instead of `myfile`


jsrodman:
Probably not important enough to redo, but I'd really suggest the general strategy of something like

`for shafilename in re.findall('{{sha256:([^}]+)', orig_content):`

PR #735, opened 2017-03-20
[SDK] Make VIP visible in endpoint. Set protocol and visibility if VI…

mbdcos:
…P already exists

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
@@ -320,6 +320,7 @@ public class ResourceUtils {
                 .getLabelsBuilder()
                 .addLabels(getVIPLabel(vipName, vipPort));
 
+        builder.setVisibility(DiscoveryInfo.Visibility.CLUSTER);
         return builder;
     }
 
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -320,6 +320,7 @@ public static Resource getExpectedRanges(
                 .getLabelsBuilder()
                 .addLabels(getVIPLabel(vipName, vipPort));
 
+        builder.setVisibility(DiscoveryInfo.Visibility.CLUSTER);
```
nickbp: this is overriding the visibility setting at L319

nickbp: Maybe add a comment saying that this is meant to update visibility from prior service versions (it looks like that's what this is for?)

mbdcos: NamedVIPEvaluationStage.java:L95 
Task visibility should be enough, this is to ensure discovery visibility is always CLUSTER.

mbdcos: PR takes so much time to pass all tests. Will add the comment in the next one.

PR #736, opened 2017-03-20
[INFINITY-1152] Move PlanUtils methods.

mgummelt:


mgummelt:
@nickbp PTAL

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/plan/strategy/ParallelStrategyTest.java
```diff
@@ -55,45 +60,16 @@ public void testParallelExecution() {
...
-        Assert.assertEquals(1, getCandidates().size());
-        Assert.assertEquals(el2, getCandidates().iterator().next());
-
-        when(el2.isComplete()).thenReturn(true);
-        when(el2.isPending()).thenReturn(false);
-        Assert.assertTrue(getCandidates().isEmpty());
-    }
-
-    @Test
-    public void testDirtyAssetAvoidance() {
```
nickbp: Any reason for this test being dropped? (And a similar one in SerialStrategyTest)

mgummelt: The dirtiness logic is in the Step now, which is mocked out.  So it doesn't belong here, but I suppose I should add the test back into StepTest.

mgummelt: It's now added to `DeploymentStepTest`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/ParentElement.java
```diff
@@ -42,4 +51,106 @@ default void updateParameters(Map<String, String> parameters) {
...
+        errors.addAll(parentErrors);
+        Collection<? extends Element> children = getChildren();
+        children.forEach(element -> errors.addAll(element.getErrors()));
+        return errors;
+    }
+
+    @Override
+    default Status getStatus() {
+        // Ordering matters throughout this method.  Modify with care.
+        // Also note that this function MUST NOT call parent.getStatus() as that creates a circular call.
```
nickbp: This second note (on L102) can go away. The equivalent under the new system would be `getStatus()` calling `this.getStatus()` -- the result there shouldn't be surprising for anyone.

PR #742, opened 2017-03-21
[INFINITY-1164] strict mode tests fail when dcos bin missing

jsrodman:
The framework tests manually added the dcos binary before invoking test.sh, but this is just ridiculous.  If test.sh purports to support SECURITY=strict, then it should work.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -58,6 +58,23 @@ cd $REPO_ROOT_DIR
 
 # Get a CCM cluster if not already configured (see available settings in dcos-commons/tools/README.md):
 if [ -z "$CLUSTER_URL" ]; then
+    if [ "$SECURITY" = "strict" ]; then
+        # in launch_ccm_cluster, for strict, we run
+        # tools/create_service_account.sh which depends on the dcos binary.
+        if ! which dcos >/dev/null 2>&1; then
+            echo "Will fetch dcos binary for use by strict mode cluster setup"
+            dcos_cli_bindir="$REPO_ROOT_DIR"/tmp/dcos_bin
+            if [ -d $dcos_cli_bindir ]; then
+                echo "Wiping prior dcos bin dir $dcos_cli_bindir"
+                rm -rf "$dcos_cli_bindir"
+            fi
+            mkdir -p "$dcos_cli_bindir"
+            echo "curl https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.8/dcos --output $dcos_cli_bindir/dcos"
+            curl https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.8/dcos --output "$dcos_cli_bindir/dcos"
+            export PATH="$PATH":"$dcos_cli_bindir"
+        fi
+    fi
+
     echo "CLUSTER_URL is empty/unset, launching new cluster."
     export CCM_AGENTS=5
     CLUSTER_INFO=$(${REPO_ROOT_DIR}/tools/launch_ccm_cluster.py)


jsrodman:
@susanxhuynh @mgummelt

test.sh
```diff
@@ -58,6 +58,23 @@ cd $REPO_ROOT_DIR
 
 # Get a CCM cluster if not already configured (see available settings in dcos-commons/tools/README.md):
 if [ -z "$CLUSTER_URL" ]; then
+    if [ "$SECURITY" = "strict" ]; then
```
mgummelt: The CLI is also installed in `run_tests.py`.  That functionality needs to be factored out, not duplicated.

mgummelt: Where do the strict mode tests fail if this code isn't included?

jsrodman: I don't agree.

run_tests.py accepts a running cluster, and needs to be able to talk to it, and people use run_tests.py directly already.

test.sh believes it needs to create a cluster, and launch_ccm_cluster.py unfortunately just  launch a cluster, it launches it and possibly adds a service account via use of the dcos command line.

These needs are distinct, and while restructuring the scripts so there's a tool that adds a dcos command line to the path for general use is worthwhile, there are opportunities for race conditions with a stable name, and fixing one problem can't require fixing all problems at once.

In short, yeah definitely I'll be doing that but not now to make smoke test work at all.

jsrodman: Strict mode tests fail in launch_ccm_cluster.py's use of create_service_account.sh which depends on the dcos binary.

gabrielhartmann: @jsrodman:  Can you please file a JIRA capturing what refactors you think are necessary for better installation of the DC/OS CLI, so we don't lose track of it.  I agree however that getting smoke tests passing again is the higher priority at the moment as other teams depend on it.  We can refactor from a working state.

mgummelt: This is a 10 line refactor.  It's trivial.  It's not going to block the smoke tests, which have already been failing for a week.

But I won't block this further, so long as you fix it right after.

jsrodman:
The fundamental claim here: we should not ever introduce duplication, is unfortunately false in the case of the test runners.  There is *no* way to test the tests before introducing change to the live tests.  Yes, I can run them locally and do, and in practice each time I make a change I learn about more undocumented behavior.  Additionally, in the effort to move the test driver to python I will *not* be changing over every type of test at once (even if I could, it would take hours to click around in Jenkins).  Therefore I cannot agree to be bound to this standard.

I agree wholeheartedly that I need to work towards reducing duplication, and in this case one of the things I'm doing is killing off strange inline logic in shellscripts no one ever looks at in test definitions.  So the duplication is actually staying neutral here, though it's now more visible.

jsrodman:
I should add the other reason I wasn't excited about doing the cli download cleanup now is that the way we're doing it currently is pretty clearly not right (brittle in many ways) and I have no idea who to ask what the right way is.  I'll file a jira to solve both of those problems.

test.sh
```diff
@@ -58,6 +58,23 @@ cd $REPO_ROOT_DIR
...
+        if ! which dcos >/dev/null 2>&1; then
+            echo "Will fetch dcos binary for use by strict mode cluster setup"
+            dcos_cli_bindir="$REPO_ROOT_DIR"/tmp/dcos_bin
+            if [ -d $dcos_cli_bindir ]; then
+                echo "Wiping prior dcos bin dir $dcos_cli_bindir"
+                rm -rf "$dcos_cli_bindir"
+            fi
+            mkdir -p "$dcos_cli_bindir"
+            echo "curl https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.8/dcos --output $dcos_cli_bindir/dcos"
+            curl https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.8/dcos --output "$dcos_cli_bindir/dcos"
```
susanxhuynh: This should be the dcos-1.9 version.

PR #743, opened 2017-03-21
Versioned artifacts

mohitsoni:
Adds additional scripts to upload versioned bootstrap.zip and executor.zip as part of the release process. These artifacts are meant to be consumed by frameworks, especially, the ones that reside outside the monorepo.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -9,6 +9,8 @@ cd $REPO_ROOT_DIR
 # Build and upload Java library to Maven repo:
 ./build.sh
 ./gradlew publish
+
+# Scripts below expects S3_BUCKET and SDK_VERSION env vars. To be supplied by Jenkins.
 ./tools/build_publishable.sh
 ./tools/release_artifacts.sh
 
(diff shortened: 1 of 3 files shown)


mohitsoni:
retest this please

release.sh
```diff
@@ -10,5 +10,9 @@ cd $REPO_ROOT_DIR
 ./build.sh
 ./gradlew publish
 
+# Scripts below expects S3_BUCKET and SDK_VERSION env vars. To be supplied by Jenkins.
```
benclarkwood: These are existing envvars, right?

benclarkwood: Looks like it, and they have sane defaults 👍

mohitsoni: Yes. They are passed in from the Build job. And, like you noticed they have sane defaults.

nickbp:
retest this please (looks like host resolv flake)

tools/build_publishable.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
mgummelt: Let's get in the habit of documenting what a script does at the top of the script.  The name alone doesn't tell me.

PR #751, opened 2017-03-22
[INFINITY-1140] Add steps to CanaryStrategy constructor

mgummelt:


mgummelt:
cc @mbdcos since you commented on the JIRA.

mbdcos:
LGTM after all tests are green.  Instead of waiting getCandidates() to setup the object, you do that in constructor. So, isInterrupted() returns true before getCandidates() call.

PR #753, opened 2017-03-22
[INFINITY-1185] Rename DefaultPlanFactory to DeployPlanFactory

mgummelt:


mrbrowning:
I'm not feeling dogmatic either way, but by this logic, should `DefaultPlan` also be moved to `DeployPlan`? The downside is that there would no longer be an obvious target to override for developers who want to reuse that logic in their custom plan definitions, which probably isn't worth the naming consistency between the factory and its product.

mgummelt:
DefaultPlan isn't only used for deploys.  The `HDFSRecoveryPlanManager` uses it.

mgummelt:
retest this please

mgummelt:
The tests failed with some error during kafka install, but the logs are too sparse to diagnose.  Running again to repro.

mgummelt:
network error

mgummelt:
retest this please

PR #754, opened 2017-03-22
[INFINITY-1188] fix ParentElement.getStatus

mgummelt:


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/ParentElement.java
```diff
@@ -100,47 +98,48 @@ default Status getStatus() {
...
         // Also note that this function MUST NOT call parent.getStatus() as that creates a circular call.
 
-        final Collection<? extends Element> children = getChildren();
+        final List<C> children = getChildren();
         if (children == null) {
             LOGGER.error("Parent element returned null list of children: {}", getName());
             return Status.ERROR;
         }
 
+        final Collection<? extends Element> candidateChildren =
```
mgummelt: Parent Status depends on its strategy.  A parent with "PENDING, WAITING" children should be PENDING, whereas a parent with "WAITING, PENDING" children should be WAITING.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/ParentElement.java
```diff
@@ -100,47 +98,48 @@ default Status getStatus() {
...
+                getStrategy().getCandidates(children, Collections.emptyList());
+
         Status result;
-        if (!getErrors().isEmpty()) {
+        if (!getErrors().isEmpty() || anyHaveStatus(Status.ERROR, children)) {
             result = Status.ERROR;
-            LOGGER.debug("({} status={}) Elements contains errors", getName(), result);
-        } else if (CollectionUtils.isEmpty(children)) {
```
mgummelt: This check is redundant.  It's covered by the following check.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/ParentElement.java
```diff
@@ -100,47 +98,48 @@ default Status getStatus() {
...
-        } else if (allHaveStatus(Status.PENDING, children)) {
-            result = Status.PENDING;
-            LOGGER.debug("({} status={}) All elements have status: {}",
-                    getName(), result, Status.PENDING);
         } else if (anyHaveStatus(Status.PREPARED, children)) {
             result = Status.IN_PROGRESS;
             LOGGER.debug("({} status={}) At least one phase has status: {}",
                     getName(), result, Status.PREPARED);
-        } else if (anyHaveStatus(Status.IN_PROGRESS, children)) {
+        }  else if (anyHaveStatus(Status.WAITING, candidateChildren)) {
```
mgummelt: A Parent is waiting only if its candidate children are waiting.

mgummelt:
cc @mbdcos for review

mbdcos:
Much better than previous PlanUtils.getStatus()

PR #756, opened 2017-03-22
Update test script to create two mount volumes

nickbp:
- Prior config: 1 12GB volume => 1 12GB mount
- New config: 1 24GB volume => 2 12GB mounts

Multiple partitions on a single volume keeps management simple, while also ensuring the script doesn't take twice as long (most time is spent waiting on the EBS volume).

### DIFF
--- a/tools/enable_mount_volumes.py
+++ b/tools/enable_mount_volumes.py
@@ -15,6 +15,7 @@ import boto3
 import botocore
 import logging
 import os
+import os.path
 import pprint
 import sys
 import time
@@ -70,7 +71,7 @@ def filter_instances_private(instances):
 
 def create_volume(client, zone):
     response = client.create_volume(
-        Size=12,
+        Size=24,
         AvailabilityZone=zone,
         VolumeType='gp2',
         Encrypted=False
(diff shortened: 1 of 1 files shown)


tools/enable_mount_volumes.py
```diff
@@ -136,23 +137,32 @@ def detach_volume(client, volume_id, instance_id, device='/dev/xvdm'):
...
     """
-    Format the attached EBS volume and adds an entry into fstab
+    Format the attached EBS volume as two MOUNT volumes and adds entries into fstab.
+    DC/OS will autodetect the '/dcos/volume#' volumes.
     """
+    device_name = os.path.basename(device)
+    run('until [[ "$(lsblk -o NAME -r | grep {} | wc -l)" -gt "0" ]]; do echo "Waiting for {}"; sleep 2; done'.format(device_name, device_name))
+    run('sudo parted -s {} mklabel gpt'.format(device))
+
+    run('sudo parted -s {} mkpart primary ext4 0% 50%'.format(device))
```
mgummelt: Can you factor this out into a function to avoid the redundancy?

nickbp: Good point. Done

PR #760, opened 2017-03-23
add suppressed endpoint for Kafka

sascala:


frameworks/kafka/docs/api-reference.md
```diff
@@ -400,7 +400,16 @@ Note the warning in the output from the commands above. You can change the indic
     {
         "message": ""
     }
-    
+
+# Service Status
```
mgummelt: Shouldn't we factor out docs for dcos-commons APIs into a central location?  This documentation isn't specific to Kafka.

sascala: I'm adding this to all infinity services with a suppressed endpoint -- it doesn't look at the moment like the APIs for the various DC/OS commons services are identical. If that's the plan, let's talk about how and where to put central API docs.

frameworks/kafka/docs/api-reference.md
```diff
@@ -9,34 +9,34 @@ enterprise: 'no'
...
 If you are using Enterprise DC/OS, follow these instructions to [create a service account and an authentication token](https://docs.mesosphere.com/1.9/administration/id-and-access-mgt/service-auth/custom-service-auth/). You can then configure your service to automatically refresh the authentication token when it expires. To get started more quickly, you can also [get the authentication token without a service account](https://docs.mesosphere.com/1.9/administration/id-and-access-mgt/iam-api/), but you will need to manually refresh the token.
 
-If you are using open source DC/OS, follow these instructions to [pass your authentication token to the DC/OS endpoint](https://dcos.io/docs/1.9/administration/id-and-access-mgt/iam-api/). 
+If you are using open source DC/OS, follow these instructions to [pass your authentication token to the DC/OS endpoint](https://dcos.io/docs/1.9/administration/id-and-access-mgt/iam-api/).
 
 Once you have the authentication token, you can store it in an environment variable and reference it in your REST API calls:
 
 ```
-$ export AUTH_TOKEN=uSeR_t0k3n
+$ export auth_token=uSeR_t0k3n
```
susanxhuynh: Why did you change this to lower case? Environment variables are usually all caps.

frameworks/kafka/docs/api-reference.md
susanxhuynh: @sascala There are a bunch of endpoints that will be common across frameworks, within dcos-commons. The suppress endpoint is one example. We may want to consolidate the docs for those endpoints, so we don't have duplication. We may also want to point this fact out to users, so that they know they can use the same endpoint for different SDK frameworks. If it's easier, you could do this consolidation in a separate PR.

frameworks/kafka/docs/api-reference.md
sascala: to be consistent with what we do elsewhere. I can change all of them to CAPS too

frameworks/kafka/docs/api-reference.md
sascala: Yes, that would be easier. This way this information doesn't get held up while I/we figure out how best to do this.

frameworks/kafka/docs/api-reference.md
susanxhuynh: I see. If you're already doing lower-case in other DC/OS docs, it's fine to be consistent.

PR #764, opened 2017-03-23
add surpressed endpoint for HDFS

sascala:


frameworks/hdfs/docs/api-reference.md
```diff
@@ -2046,8 +2046,15 @@ $ dcos hdfs config show 9a8d4308-ab9d-4121-b460-696ec3368ad6
...
-$ curl -H "Authorization:token=<auth_token>" <dcos_url>/service/hdfs/v1/configurations/9a8d4308-ab9d-4121-b460-696ec3368ad6
+$ curl -H "Authorization:token=$auth_token" <dcos_url>/service/hdfs/v1/configurations/9a8d4308-ab9d-4121-b460-696ec3368ad6
 {
     ... same format as target config above ...
 }
 ```
+
+# Service Status Info
+Send a GET request to the `/v1/state/properties/suppressed` endpoint to learn if HDFS is in a `suppressed` state and not receiving offers. If a service does not need offers, Mesos can "suppress" it so that other services are not starved for resources.
+You can use this request to troubleshoot: if you think Cassandra should be receiving resource offers, but is not, you can use this API call to see if Cassandra is suppressed.
```
susanxhuynh: Should be 'HDFS' not 'Cassandra' (two places).

PR #766, opened 2017-03-23
[KAFKA] Remove Configuration Upgrade code

mbdcos:
After dcos-kafka-service is absolute. See comments below.


mbdcos:
retest this please

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/scheduler/KafkaService.java
```diff
@@ -29,13 +30,16 @@ public KafkaService(File pathToYamlSpecification) throws Exception {
...
-        /* Upgrade */
-        new KafkaConfigUpgrade(schedulerBuilder.getServiceSpec());
-        CuratorStateStoreFilter stateStore = new CuratorStateStoreFilter(schedulerBuilder.getServiceSpec().getName(),
+        /* Post Upgrade for version 1.1.21 */
+        CuratorStateStore stateStore = new CuratorStateStore(schedulerBuilder.getServiceSpec().getName(),
                 DcosConstants.MESOS_MASTER_ZK_CONNECTION_STRING);
-        stateStore.setIgnoreFilter(RegexMatcher.create("broker-[0-9]*"));
-        schedulerBuilder.setStateStore(stateStore);
-        /* Upgrade */
+        StringMatcher oldTaskNameFilter = RegexMatcher.create("broker-[0-9]*");
```
susanxhuynh: Not sure I understand this: which versions would this upgrade from? 1.1.20 -> 1.1.21? Version 1.1.20 does not have 'broker-[0-9]' tasks, does it? Wouldn't it already have the new task names?

mbdcos:
retest this please

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/scheduler/KafkaService.java
mbdcos: Current master has code to autodetect if Configuration upgrade needed or not. If upgrade needed, it does and while doing that keeps taskInfo backups. If new Kafka fails, you can revert back to old Kafka.

It is optional, whether cleanup backups or not after upgrade.  Default is keep backups. Backups are not useful after new Kafka is running correctly. So, we delete them. 

PS: if we do not delete those backup taskInfos, should use CuratorFilter, otherwise those will confuse recovery process, and placement rule calculations.

mbdcos:
I mark this as WIP.  

If we delete Configuration format upgrade code, we automatically enforce subsequent framework upgrades. In order to upgrade to v 1.1.21, you should have upgraded to v 1.1.20. 
This does not have to be a necessary requirement. If we keep the Configuration upgrade code, you can switch from v 1.1.19 to 1.1.21, or 1.1.18 to 1.1.21. 

We can merge this PR later, but I do not see any reason to rush, and merge it now.  After a couple of versions, after Kafka binary is also upgraded, we can merge this PR.

gabrielhartmann:
Too old.  Re-open if still relevant

PR #767, opened 2017-03-24
[INFINITY-1012] Make tests go faster ( parallelize)

jsrodman:
Goals:
 * run tests faster in parallel
 * Better logging (timestamps!)
 * Render failures unambiguously via junit xml (not done yet)
 * Try to reduce the number of environment variables floating around.
 * New test runner, parallelizes, junit for test actions, verifies requirements.

### DIFF
--- a/test.py
+++ b/test.py
@@ -21,7 +21,6 @@ import clustinfo
 import fwinfo
 import launch_ccm_cluster
 import junit_xml
-
 work_dir = None
 def get_work_dir():
     global work_dir
--- a/test.py
+++ b/test.py
@@ -4,23 +4,32 @@ import sys
 
 import argparse
 import logging
+import random
 import shutil
 import subprocess
 import tempfile
 import time
 
-
 def get_repo_root():
     return os.path.dirname(sys.argv[0])
-
 
 logger = logging.getLogger("dcos-commons-test")
 logging.basicConfig(level=logging.INFO, format="%(asctime)s %(name)s %(message)s")
 
 sys.path.append(os.path.join(get_repo_root(), 'tools'))
+sys.path.append(os.path.join(get_repo_root(), 'tools', 'contrib'))
 import clustinfo
 import fwinfo
 import launch_ccm_cluster
+import junit_xml
+
+work_dir = None
+def get_work_dir():
+    global work_dir
+    if not work_dir:
+        work_dir = tempfile.mkdtemp(prefix='test_workdir_', dir=get_repo_root())
+        logger.info("Using %s for test run files", work_dir)
+    return work_dir
 
 work_dir = None
 def get_work_dir():
@@ -42,7 +51,7 @@ def parse_args(args=sys.argv):
                  "Number of clusters to run tests on. "
                  "Large values are likely to break CCM.")
     parser.add_argument("--order", choices=('random', 'ordered'), default='random',
-            help="Run tests in randomr order, or the order given on the command line.  "
+            help="Run tests in random order, or the order given on the command line.  "
             "In the case of no tests listed, ordered means alpha order.")
     parser.add_argument("--cluster-url", help="Use this already existing cluster, "
             "don't bring up new ones.")
@@ -172,6 +181,16 @@ def setup_frameworks(framework_list):
     logger.info("Frameworks initialized: %s", ", ".join(fw_names))
 
 
+def _action_wrapper(action_name, framework, function, *args):
+    framework.start_action(action_name)
+    try:
+        val = function(*args)
+        framework.finish_action_ok(action_name)
+        return val
+    except:
+        framework.finish_action_fail(action_name)
+        raise
+
 def build_and_upload(run_attrs=parse_args([])):
     """
     Build a list of framework scheduler and put them at URLs so a cluster can use it.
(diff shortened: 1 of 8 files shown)


jsrodman:
So this *seems* like it will reduce test runs by about an hour (33% reduction); but I can't get a test run to pass for reasons unrelated to this change.  Maybe by tomorrow.

That's not a very exciting improvement, sorry.  However, most of the work was figuring out where all the problems were running overlapping tests.  We could get a larger improvement in a variety of ways, and I'll continue to work on that mixed in with other tasks.

jsrodman:
Reviewers, I welcome critique.
I haven't addressed Susan's existing criticism that we should fetch the dcos command line from 1.9, but I plan to do that in the next few working days, by fetching the cluster version and fetching the dcos command line that matches.

mgummelt:
How are the junit changes related to the parallelization?  If it's a separate task, can you put that in a different PR?

test.py
```diff
@@ -9,18 +9,26 @@
 import tempfile
 import time
 
-
```
mgummelt: pep8 standard is to have two empty lines

test.py
```diff
@@ -9,18 +9,26 @@
 import tempfile
```
mgummelt: Where is test.py used?

mgummelt:
This PR description seems to be the same as that of the last commit: https://github.com/mesosphere/dcos-commons/commit/bf496380964403d33bd3f9b4c290eef9ef87fe2c

What's different about this?  Does it resolve INFINITY-1012?  If not, can you create a JIRA for what this resolves?

jsrodman:
The last commit was intended as wip but art was.having significant trouble with build dependency that were addressed in the wip. Probably I should have pointed him to the branch to copy the script out of.

tools/contrib/junit_xml/__init__.py
```diff
@@ -0,0 +1,322 @@
+#!/usr/bin/env python
```
jsrodman: It cannot be done in the manner of run_tests.py and achieve the goals. 
Run_tests.py pays the cost of a subshell and secondary python process. This code is intended to catch meaningful exceptions so that failures can be accurately represented in Jenkins without reading many megabytes of logs.  activating a virtualenv within a python process is somewhat black magic, but installing modules via pip inside a python process and then using them will require buckets of it. 
There is no intention to add any modules beyond junit_xml in the future, and it may be worth it at some point to generate the xml myself. 

As a further point if this design is a real problem, someone could have replied when I explicitly raised the question as a concern on slack.

test.py
jsrodman: If you mean tempfile, it can now be removed. I had initially implemented the output for in temp but later changed to store it in the working dir probably for developer obviousness.

test.py
jsrodman: Pep8 is a set of guifelines, not rules. We can and will have many deviations and toys is fine.

test.py
mgummelt: I mean where is test.py invoked?

tools/contrib/junit_xml/__init__.py
mgummelt: Can we have a thin shell script wrapper around test.py that sets up the virtualenv, then?  Seems preferable to having third party code in the repo.

test.py
mgummelt: Pep8 is a fairly well established standard throughout the company.  The DC/OS CLI as well as all the DC/OS tooling itself also adheres to pep8.  I'm fine with deviations when they make sense, but let's not add regressions.

test.py
jsrodman: Currently nowhere.  The intent is to replace test.sh

test.py
jsrodman: This is not a regression. Regression is a serious word that should not be diluted by nitpicks over a blank linke.

tools/contrib/junit_xml/__init__.py
jsrodman: That would work directly against one of the goals here, to reduce the number of confusing little subprocesses and files that serve only to indirect and make it harder to see how information is moving through the scripts.  So I don't want to do that.  Perhaps some others may have opinions or suggestions.

test.py
mrbrowning: +1 to two line separation. Consistent style pays dividends, and while there can be cases where the tradeoffs make a style violation worth it, there's really no upshot to doing so here. Ideally we should just be running `flake8` against our Python code as a PR check, since enforcing style is mostly a job computers can do, saving reviewers' time for other concerns.

tools/contrib/junit_xml/__init__.py
mrbrowning: I'm also pretty skeptical of vendoring code, but I appreciate the desire to simplify the logic here, since it makes maintaining these test scripts more burdensome. I think, though, that significant disambiguation could be achieved just with better naming and organization: modularize the scripts so that environment set up and test logic are as orthogonal as possible, instead of the confusing nesting doll topology that we have now. The actual process graph doesn't seem to me like a significant impediment to understanding so much as which script is doing what.

mgummelt: Also, it looks like `test.py` shells out to `run_test.py`?  Why do we need both scripts?  If they are different in function, why not just call `run_test.py` instead of shelling out to it, so we don't have the complexity of multiple python environments?

jsrodman: When in --parallel, running the test process in the background under control by run_tests.py would require significant change to run_tests.py.  Of course simplifying this is possible and I already planned to do it.  I don't see why that improvement is required now.

PR #771, opened 2017-03-24
[INFINITY-1107] merge pull request for build & test

jsrodman:
Simple goal; test the merged state.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -34,6 +34,7 @@ if [ x$PULLREQUEST = "xtrue" ]; then
   if ! git pull origin master; then
     _notify_github failure "Merge from master branch failed"
     exit 1
+  else
     _notify_github pending "Merge from master branch done"
   fi
 fi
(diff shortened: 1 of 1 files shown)


jsrodman:
Seems to work; the idea is to add to build.sh to mean like "this is a *p*ull request" to ask it to merge before building.

When we get Jenkinsfile logic going, maybe this should leave build.sh and go in there entirely.

jsrodman:
Note that the pull-requests won't test this logic, but I hacked in a forcing change to test it anyway, later removed.

build.sh
```diff
@@ -15,6 +29,15 @@ _notify_github() {
     $REPO_ROOT_DIR/tools/github_update.py $1 build:sdk $2
 }
 
+if [ x$PULLREQUEST = "xtrue" ]; then
+  echo "Merging master into pull request branch."
+  if ! git pull origin master; then
+    _notify_github failure "Merge from master branch failed"
+    exit 1
+    _notify_github pending "Merge from master branch done"
```
susanxhuynh: add else here

PR #774, opened 2017-03-28
[INFINITY-1171]Centralize cli downloading in our scripts.

jsrodman:
@susanxhuynh & @nickbp I'd appreciate your comments.

* Extract cli download to separate module.
* Expose as script in case shell needs it (I don't need that now) -- also enables easy testing
* Update cli atomically in case of goofy redownload while running
* Automatically get the right version for the cluster
* launch_ccm_cluster fetches dcos cli itself if required.


susanxhuynh:
LGTM

PR #775, opened 2017-03-28
Allows optional values to be cleared

dannypurcell:
It looks like the intention here was to allow these values to be set or "empty" as they are initially. However, `Optional.of` does a null check which prevents a consumer from clearing these values once they are set.
The use case for clearing is a class using `DefaultScheduler` that wants to remove the default config validators.
Current workaround is
```
DefaultScheduler.Builder builder =
                DefaultScheduler.newBuilder(serviceSpecWithCustomizedPods(rawServiceSpec))
                        .setPlansFrom(rawServiceSpec)
                .setConfigValidators(Collections.emptyList());
```


gabrielhartmann:
retest this please

PR #777, opened 2017-03-28
Fix local builds (INFINITY-1216)

nickbp:
Github updater was only checking for CI env when run directly as a python script.
Refactor to also apply the CI env check when using the updater as an API.

### DIFF
--- a/tools/github_update.py
+++ b/tools/github_update.py
@@ -172,13 +172,32 @@ class GithubAPI(object):
 class GithubStatusUpdater(object):
 
     def __init__(self, default_context_label=''):
-        info = RepoInfo()
-        self._api = GithubAPI(info.repo_orgname(), info.commit_sha(), info.github_auth_token())
+        if self._should_access_github():
+            info = RepoInfo()
+            self._api = GithubAPI(info.repo_orgname(), info.commit_sha(), info.github_auth_token())
+        else:
+            self._api = None
         self._default_context_label = default_context_label
+
+
+    def _should_access_github(self):
+        '''returns whether the local environment appears to be running in CI'''
+        if 'WORKSPACE' not in os.environ:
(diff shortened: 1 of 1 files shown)


tools/github_update.py
```diff
@@ -172,20 +172,49 @@ def set_commit_status(self, context_label, state, message, details_url):
...
+            self._api = None
         self._default_context_label = default_context_label
 
 
+    def _should_access_github(self):
+        '''returns whether the local environment appears to be running in CI'''
+        if 'WORKSPACE' not in os.environ:
+            # not running in CI. skip actually sending anything to GitHub
+            return False
+        if os.environ.get('GITHUB_DISABLE', ''):
```
jsrodman: Random nitpick, this `get('key', '')` pattern is pretty weird.  You set the default to the empty string a lot but never use it.
get('key') returns None if it isn't there, so this looks like you explicitly needed an empty string but don't.  

I prefer the `'WORKSPACE' in os.environ` pattern above for testing if present.

Just food for the future.

tools/github_update.py
```diff
@@ -172,20 +172,49 @@ def set_commit_status(self, context_label, state, message, details_url):
...
 
 
     def update(self, state, message='', details_url='', context_label=''):
         '''sends a commit status update to github.
         returns True on success or False otherwise.
         state should be one of the values in 'VALID_STATES'.'''
+        if details_url:
+            logmsg = '{} {}: {} ({})'.format(context_label, state, message, details_url)
+        else:
+            logmsg = '{} {}: {}'.format(context_label, state, message)
```
jsrodman: These logs are the reason we're bothering with this object handling the no-github at all?

nickbp: Yes, it just falls back to logging the status if it doesn't have anywhere to write it.

PR #783, opened 2017-03-30
Remove testpy

jsrodman:
Revert test.py for later resubmission.


tools/run_tests.py
```diff
@@ -148,9 +148,7 @@ def run_shakedown(self, test_dirs, requirements_txt=None, pytest_types='sanity')
 #!/bin/bash
 set -e
 echo "VIRTUALENV CREATE/UPDATE: {venv_path}"
-if  [ ! -f {venv_path}/bin/python3 ] ; then
-    virtualenv -p $(which python3) --always-copy {venv_path}
-fi
+virtualenv -p $(which python3) --always-copy {venv_path}
```
mrbrowning: Is the assumption behind removing the check here that this script is meant to be run in CI primarily, so we can always depend on a fresh workspace?

tools/launch_ccm_cluster.py
```diff
@@ -409,32 +409,15 @@ def _write_jenkins_config(github_label, cluster_info, error = None):
...
-    return cluster_info
-
 def main(argv):
     ccm_token = os.environ.get('CCM_AUTH_TOKEN', '')
     if not ccm_token:
         raise Exception('CCM_AUTH_TOKEN is required')
 
     # used for status and for jenkins .properties file:
-    github_label = determine_github_label()
+    github_label = os.environ.get('CCM_GITHUB_LABEL', '')
```
mrbrowning: A matter of taste, but when a chunk of logic cleanly encapsulates some need (as here, where the need is "get the right github label"), it should have its own function. The inlining below makes sense to me, since `start_cluster` was not really a complete representation of what's needed to start a cluster, but this less so.

jsrodman: This is a revert.  So it's uncleaning this.  I could change it to retain this cleanup.

tools/run_tests.py
jsrodman: the if was a failed idea as to how to safely parallelize.  In this revert it's going back to how it was, but this idea didn't work anyway as pip would uninstall packages out from other another test run.

jsrodman: For now, I kept the idea that actually worked; use different virtualenvs for different frameworks.

PR #784, opened 2017-03-30
Marathon-style placement constraints are not pod-specific

loren:
Fixes INFINITY-1209

- Plumbed podName from YAMLToInternalMappers to MarathonConstraintParser in order to apply task-filter logic to `RoundRobinByHostnameRule`, `RoundRobinByAttributeRule`, `MaxPerHostnameRule`, and `MaxPerAttributeRule`
- Hooked up RoundRobinByHostnameRule and RoundRobinByAttributeRule as placement rules for Jackson
- Added Shakedown tests around CLUSTER, GROUP_BY, UNIQUE, and MAX_PER
- Updated Javadocs to reflect new task-filter behavior

Unrelated cleanup:
- Renamed test_discovery.yml -> test_discovery.py
- DRY up get_marathon_host() in elastic shakedown


frameworks/helloworld/src/main/dist/examples/marathon_constraint.yml
```diff
@@ -0,0 +1,43 @@
+name: "hello-world"
+pods:
+  hello:
+    count: {{HELLO_COUNT}}
+    placement: {{HELLO_PLACEMENT}}
```
nickbp: Could this use the default svc.yml for hello-world? Looks like that has placement constraint fields defined

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/MarathonConstraintParser.java
```diff
@@ -44,58 +44,45 @@ private MarathonConstraintParser() {
     }
 
     /**
```
nickbp: This should probably be kept around in some form, it's expected to be used in frameworks that want to combine some hardcoded rules (in a service's Main.java) with user-entered rules.

For example HDFS could be exposing a user-visible setting to place all of the HDFS service in a particular location, and then using this to AND that setting against the hardcoded per-type rules.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/MarathonConstraintParser.java
```diff
@@ -44,58 +44,45 @@ private MarathonConstraintParser() {
...
-        return new AndRule(rule, marathonRule);
-    }
-
-    /**
      * Creates and returns a new {@link PlacementRule} against the provided marathon-style
      * constraint string. Returns a {@link PassthroughRule} if the provided constraint string is
      * {@code null} or empty.
      *
+     *
+     * @param podName
```
nickbp: Needs javadoc content. Can't hurt to specify what to do if a given constraint string is to be exposed across the whole service and not just a single pod, or if that's even supported..

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/MarathonConstraintParser.java
```diff
@@ -44,58 +44,45 @@ private MarathonConstraintParser() {
...
+            rowRules.add(parseRow(podName, row));
         }
         return new AndRule(rowRules);
     }
 
     /**
      * Converts the provided marathon constraint entry to a PlacementRule.
      *
+     *
+     * @param podName
```
nickbp: javadoc (also pending above suggestion to switch to passing in a StringMatcher)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/MarathonConstraintParser.java
```diff
@@ -44,58 +44,45 @@ private MarathonConstraintParser() {
...
      *     content isn't valid or supported
      */
-    public static PlacementRule parse(String marathonConstraints) throws IOException {
+    public static PlacementRule parse(String podName, String marathonConstraints) throws IOException {
         if (marathonConstraints == null || marathonConstraints.isEmpty()) {
             // nothing to enforce
             return new PassthroughRule();
         }
         List<List<String>> rows = splitConstraints(marathonConstraints);
         if (rows.size() == 1) {
```
nickbp: Looks like downstream either turns podName into a StringMatcher, or ignores it, depending on the operation. How about just creating the StringMatcher here, and passing that to parseRow (and beyond) instead?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -234,6 +234,8 @@ public boolean equals(ServiceSpec first, ServiceSpec second) {
                 OrRule.class,
                 PassthroughRule.class,
                 RegexMatcher.class,
+                RoundRobinByHostnameRule.class,
+                RoundRobinByAttributeRule.class,
```
nickbp: Reverse to fit alphabetical order

frameworks/helloworld/src/main/dist/examples/marathon_constraint.yml
loren: `svc.yml` doesn't have a plan strategy exposed, and I want the tasks to deploy in parallel while keeping the phases serial. Otherwise the tests will take over an hour. Also, I'd prefer to guard against any future changes to `svc.yml` and have something that acts more like a test fixture tied to this suite of tests.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/MarathonConstraintParser.java
loren: I put it back in and updated its method signature, but I'm a little uncomfortable modifying code that currently is not touched by either the codebase or the test suite. I see how it can be used in that HDFS scenario (HDFS-412 ?) so I'll offer to work on that.

loren:
@nickbp I addressed your review comments, please have another look.

frameworks/helloworld/src/main/dist/examples/marathon_constraint.yml
nickbp: Got it. :+1:

loren:
Thank you @nickbp for reviewing!

PR #788, opened 2017-03-31
Quick and dirty injection of modifying master envvars before running …

benclarkwood:
…tests

### DIFF
--- a/tools/modify_master.py
+++ b/tools/modify_master.py
@@ -1,0 +1,93 @@
+#!/usr/bin/python
+
+import shakedown
+import logging
+
+
+# Methods to modify the envvar settings of a mesos master and restart
+# the master process after they are modified
+
+logger = logging.getLogger(__name__)
+
+
+def set_master_envvar(envvar, value):
+    as_dict = {envvar: value}
+    modify_envvars(lambda envvars: envvars.update(as_dict))
+
+
+def set_master_envvars(new_envvars_as_dict):
+    modify_envvars(lambda envvars: envvars.update(new_envvars_as_dict))
+
+
+def remove_master_envvar(envvar):
+    modify_envvars(lambda envvars: envvars.pop(envvar, None))
+
+
+def modify_envvars(modifier):
+    logger.info("Modifying envvars on master node...")
+    # Get the current set of master envvars
+    success, out = shakedown.run_command_on_master('cat /opt/mesosphere/etc/mesos-master')
+    if success is not True:
+        logger.info('Unable to get current envvars from master: {}'.format(out))
+        raise RuntimeError("Unable to get current envvars")
+
+    logger.info("Current envvars:\n{}".format(out))
+    envvars, commented = process_envvars(out)
+    modifier(envvars)
+    write_envvars(envvars, commented)
+    restart_master()
+    logger.info("Modification complete.")
+
+
+def process_envvars(input):
+    logger.info("Processing envvars...")
+    lines = input.split('\n')
+
+    envvars = {}
+    commented = []
+    for line in lines:
+        if '#' in line:
+            commented.append(line)
+            continue
+        if '=' not in line:
+            continue
+        bits = line.split('=')
+        if len(bits) != 2:
+            raise ValueError('Envvar file had badly formatted line: {}\nfile: {}'.format(
+                line, input))
+        
+        envvars[bits[0]] = bits[1]
+
+    return envvars, commented
+
+
+def write_envvars(envvars, commented_envvars):
+    logger.info("Writing envvars...")
+    content = []
+    for evar in envvars:
+        content.append('{}={}'.format(evar, envvars[evar]))
+
+    for comment in commented_envvars:
+        content.append(comment)
+
+    new_file_content = '\n'.join(content)
+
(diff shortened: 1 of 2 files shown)


benclarkwood:
Testing that the envvar override works correctly.

jsrodman:
Going to make idiomatic kibitzes here I encourage you to heartily ignore unless you find them useful.

tools/modify_master.py
```diff
@@ -0,0 +1,93 @@
...
+
+def remove_master_envvar(envvar):
+    modify_envvars(lambda envvars: envvars.pop(envvar, None))
+
+
+def modify_envvars(modifier):
+    logger.info("Modifying envvars on master node...")
+    # Get the current set of master envvars
+    success, out = shakedown.run_command_on_master('cat /opt/mesosphere/etc/mesos-master')
+    if success is not True:
```
jsrodman: This idiom may not be yours but typical would be `if not success:`

tools/modify_master.py
```diff
@@ -0,0 +1,93 @@
...
+        logger.info('Unable to get current envvars from master: {}'.format(out))
+        raise RuntimeError("Unable to get current envvars")
```
jsrodman: Maybe not worth it this time, but ideally some info about the failure (stderr?) should go in here. Probably too much work to figure out the right things to get for now.

tools/modify_master.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def process_envvars(input):
+    logger.info("Processing envvars...")
+    lines = input.split('\n')
+
+    envvars = {}
+    commented = []
+    for line in lines:
+        if '#' in line:
```
jsrodman: I think this wants to be
`if line.strip().startswith('#'):`
or similar (though id' probably strip on an independent line)

tools/modify_master.py
```diff
@@ -0,0 +1,93 @@
...
+    envvars, commented = process_envvars(out)
+    modifier(envvars)
+    write_envvars(envvars, commented)
+    restart_master()
+    logger.info("Modification complete.")
+
+
+def process_envvars(input):
+    logger.info("Processing envvars...")
+    lines = input.split('\n')
```
jsrodman: I think 'input' is a file-like object, in which case this can be spelled

`for line in input:`
By default, files when used as iterators return lines, which neatly solves the 'data' too big' problem.

tools/modify_master.py
```diff
@@ -0,0 +1,93 @@
...
+    for line in lines:
+        if '#' in line:
+            commented.append(line)
+            continue
+        if '=' not in line:
+            continue
+        bits = line.split('=')
+        if len(bits) != 2:
+            raise ValueError('Envvar file had badly formatted line: {}\nfile: {}'.format(
+                line, input))
```
jsrodman: The slightly ugly 
`var, value = line.slit('=', 1)`
handles this.  (Ugly, becauase wtf is that 1?  The split count.  maxsplit=1 is a syntax error because the C implementation is lazy.)
I'd bet a value with an equals in it is technically valid, so you can probably drop the error handling.

tools/modify_master.py
```diff
@@ -0,0 +1,93 @@
...
+        
+        envvars[bits[0]] = bits[1]
+
+    return envvars, commented
+
+
+def write_envvars(envvars, commented_envvars):
+    logger.info("Writing envvars...")
+    content = []
+    for evar in envvars:
```
jsrodman: Optionally could be:
`for evar, value in envvars.items():`

tools/modify_master.py
```diff
@@ -0,0 +1,93 @@
...
+        content.append('{}={}'.format(evar, envvars[evar]))
+
+    for comment in commented_envvars:
+        content.append(comment)
+
+    new_file_content = '\n'.join(content)
```
jsrodman: I'm not sure if this loses the terminal newline (which I assume won't break anything).  I'd usually construct the lines with embedded newlines (this is what for line in file gives you)  But then again I have no real idea what's going to happen with that echo anyway.

tools/modify_master.py
```diff
@@ -0,0 +1,93 @@
...
+    logger.info("Restarting master process...")
+
+    success, out = shakedown.run_command_on_master('sudo systemctl condrestart dcos-mesos-master')
+    if success is not True:
+        print("wtf...")
+        raise RuntimeError("Unable to restart master")
+
+
+# Note: This is a hack for now. Talk to bwood about it.
+set_master_envvar('MESOS_SLAVE_REMOVAL_RATE_LIMIT', '100/20mins')
```
jsrodman: This incantation makes this only happen if run directly:
`if __name__ == "__main__":`
which makes it an externally usable module.

tools/modify_master.py
benclarkwood: Input is a string, it's the stdout of executing cat remotely

tools/modify_master.py
benclarkwood: That is better. #firsttimepythoner

tools/modify_master.py
benclarkwood: Right. This is my super hacky "Hardcode it to do this one thing when called from that bash script.

tools/modify_master.py
benclarkwood: Yeah, that would be better and less false-positivey

PR #791, opened 2017-04-03
[INFINITY-1252]split cluster create and config

jsrodman:
First pass at INFINITY-1252; configure our clusters separately from bringing them up

Unfortunately this required fixing some bugs in enable_mount_voluemes, and
bugs made it clear launch_ccm_cluster can't catch every exception and treat it
as a retry (this leads to silently abandoning perfectly good clusters)

----

Perfectly happy to be told to do this another way.  This seems far larger than I'd like but as seems to handle the SECURITY=strict & SECURITY=permissive cases from a naive test.sh invocation.


mgummelt:
You shouldn't need to test python2.  We (should) be using python3 everywhere.

jsrodman:
@mgummelt I presumed python2 was being used because by default on most systems test.sh will use python2 for the launch_ccm_cluster etc.  I was going to make a separate jira / PR to change all our shebangs to use 3 explicitly, unless you want that here.

mgummelt:
Changing to python3 explicitly is fine, but python 3 is the only python we support.  If a user tries to run with python 2 and it fails, that's on the user.

tools/configure_test_cluster.py
```diff
@@ -0,0 +1,93 @@
...
+    return os.path.dirname(os.path.realpath(__file__))
+
+class ClusterInitializer(object):
+    def __init__(self, cluster_id, stack_id, auth_token, dns_address,
+            enterprise=False, security=None):
+        self.cluster_id = cluster_id
+        self.stack_id = stack_id
+        self.auth_token = auth_token
+        self.dns_address = dns_address
+        self.entperise = enterprise
```
nickbp: typo, also it may make sense to name this `is_enterprise`

tools/configure_test_cluster.py
```diff
@@ -0,0 +1,93 @@
...
+
+import logging
+import os
+import os.path
+import subprocess
+import sys
+
+import venvutil
+
+# Things this needs: cluster_id, stack_id (???), auth_token, dns_address,
```
nickbp: The `cluster_id` is a sequential int made up by CCM, while the `stack_id` is the AWS cloudformation stack (colon-delimited string iirc).
The variable names below could be updated to `ccm_cluster_id` and `aws_stack_id` to clarify the difference a bit.

tools/configure_test_cluster.py
```diff
@@ -0,0 +1,93 @@
...
+            self.dcos_url = 'http://%s' % dns_address
+        else:
+            self.dcos_url = 'https://%s' % dns_address
+
+
+    def create_mount_volumes(self):
+        fmt = 'Enabling mount volumes for cluster {} (stack id {})'
+        logger.info(fmt.format(self.cluster_id, self.stack_id))
+        import enable_mount_volumes
+        # fabric spams to stdout, which causes problems with launch_ccm_cluster.
```
nickbp: looks like this comment can go away now

tools/enable_mount_volumes.py
```diff
@@ -16,7 +16,7 @@
 import logging
 import os
 import os.path
-import pprint
+#import pprint
```
nickbp: remove this?

tools/launch_ccm_cluster.py
```diff
@@ -258,60 +261,43 @@ def _start(self, config):
...
         logger.info('Fetching auth token')
         dcos_url = 'https://' + dns_address
         auth_token = dcos_login.DCOSLogin(dcos_url).get_acs_token()
 
-        if config.permissions:
-            logger.info('Setting up permissions for cluster {} (stack id {})'.format(cluster_id, stack_id))
+        is_enterprise = config.cf_template.startswith('ee.')
+        clustinit = configure_test_cluster.ClusterInitializer(cluster_id,
+                stack_id, auth_token, dns_address, is_enterprise,
+                os.environ.get('SECURITY'))
```
nickbp: This env fetch could be moved upstream into the StartConfig object, which currently has all the other env fetches

tools/configure_test_cluster.py
jsrodman: I'll kill it, I was not entirely sure of this change yet.

tools/enable_mount_volumes.py
jsrodman: I could do. We have a bunch of commented pprint lines, which is why i commented this too (not currently used), but didn't evaluate whether those lines are useless.

tools/launch_ccm_cluster.py
jsrodman: It's sort of there already; as self.permissions.
Options to move this there:
1 - make redundant item
2 - change self.permissions to self.security and adjust all current users.

jsrodman: Took option 2 in update.

tools/requirements.txt
```diff
@@ -0,0 +1 @@
+-e git+https://github.com/dcos/shakedown.git@master#egg=shakedown
```
jsrodman: This is really here to say that modify_master.py needs this.  I could just put it in the python that calls modify_master, or I could potentially put it in the modify_master module itself, though that would require it to delay imports (maybe fine).

Making the modules themselves able to declare what modules they need pip installed probably will require more magic than I'm super comfortable wedging into this pull request, but maybe theres a clear better state for right now.

PR #794, opened 2017-04-03
[SDK][HDFS-437] Wait for API Server startup before accepting offers

gabrielhartmann:


mbdcos:
Delay accepting offers till API server is started. Here is the order:
       initializeGlobals(driver);
        initializeDeploymentPlanManager();
        initializeRecoveryPlanManager();
        initializePlanCoordinator();
        initializeResources();
        initializeApiServer();
initializeApiServer() is starting a thread. How likely is it that APIserver will not initialize immediately and scheduler will decline first bunch of offers? In terms of Mesos scheduling, how bad is it to decline offers?

gabrielhartmann:
@mbdcos:  Declining offers isn't a big deal.  We do it all the time and is a normal part of being a framework.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -632,8 +634,32 @@ private void initializeResources() throws InterruptedException {
         }
         resources.add(new StateResource(stateStore, new StringPropertyDeserializer()));
         resources.add(new TaskResource(stateStore, taskKiller, serviceSpec.getName()));
-        // use add() instead of put(): throw exception instead of waiting indefinitely
-        resourcesQueue.add(resources);
+    }
+
+    private void initializeApiServer() {
+        if (!apiServerDisabled.get()) {
```
nickbp: could just do `if (apiServerDisabled) { return; }`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -693,6 +719,12 @@ public void reregistered(SchedulerDriver driver, Protos.MasterInfo masterInfo) {
 
     @Override
     public void resourceOffers(SchedulerDriver driver, List<Protos.Offer> offersToProcess) {
+        if (!apiServerDisabled.get() && (apiServer == null || !apiServer.isStarted())) {
+            LOGGER.info("Waiting for API Server to start ...");
```
nickbp: How about `Declining all offers: Waiting for API Server to start...`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -693,6 +719,12 @@ public void reregistered(SchedulerDriver driver, Protos.MasterInfo masterInfo) {
...
+            declineOffers(driver, Collections.emptyList(), offersToProcess);
+            return;
```
nickbp: There should probably be a check here: exit the scheduler process if the API server is still down after some duration (60s?). The previous code in getResources() looks like it was effectively doing this.

loren: I'd be interested to see how that affects Elastic, since it takes 1.5-2 minutes for its API server to come up. Would that make a scheduler restart part of the normal process of installing the package? https://jira.mesosphere.com/browse/INFINITY-996

gabrielhartmann: I wasn't aware Elastic took so long.  Why is that?  It's take 1.5-2 minutes for the first Task to get launched?  I don't remember that being the case.

loren: I'm not really sure what's causing that delay for the API server. Some of the tasks actually come up and are COMPLETE before that API server starts responding. Anyway, didn't mean to hijack this PR with it, but I thought it might be relevant.

nickbp:
CI test failure is likely relevant:

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: It is relevant.  I'm going to leave a timeout out of this PR.  @nickbp. that 60s timeout was for the Scheduler to provide in memory Resource objects.  That timeout was NEVER hit.  It was an artifact of the weird interdependent design with that weird queue.  So not relevant anymore.  I am hesitant to put a timeout now given ES behavior.  I don't want the Scheduler crash looping because it's on a slow machine.  Happy to revisit if we see this screwing things up.

gabrielhartmann:
@nickbp:  Added timeout for ApiServer and Flags file.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -691,10 +724,33 @@ public void reregistered(SchedulerDriver driver, Protos.MasterInfo masterInfo) {
...
     }
 
+    public boolean apiServerReady() {
+        boolean serverStarted = apiServer != null && apiServer.isStarted();
+
+        if (serverStarted || apiServerDisabled.get()) {
+            apiServerStopwatch.reset();
+            serverStarted = true;
+        } else if (
+                apiServerStopwatch.elapsed(TimeUnit.MILLISECONDS) > SchedulerFlags.getApiServerTimeout().toMillis()) {
```
nickbp: indent is off

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerFlags.java
```diff
@@ -0,0 +1,25 @@
...
+    public static final String DEFAULT_API_SERVER_TIMEOUT_S = "600";
+
+    public static Duration getApiServerTimeout() {
+        String apiServerTimeoutSecs = getEnvDefault(API_SERVER_TIMEOUT_S, DEFAULT_API_SERVER_TIMEOUT_S);
+        return Duration.ofSeconds(Integer.parseInt(apiServerTimeoutSecs));
+    }
+
+    private static String getEnvDefault(String envKey, String dflt) {
+        String value = System.getenv(envKey);
+        if (value == null) {
```
nickbp: could just do `return (value == null) ? dflt : value;`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -94,6 +96,10 @@
     private final Optional<RecoveryPlanManagerFactory> recoveryPlanManagerFactoryOptional;
     private final Optional<ReplacementFailurePolicy> failurePolicyOptional;
 
+    private JettyApiServer apiServer;
+    private AtomicBoolean apiServerDisabled = new AtomicBoolean(false);
```
nickbp: maybe switch this to `apiServerEnabled`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -875,4 +931,9 @@ public void update(Observable observable) {
             suppressOrRevive();
         }
     }
+
+    @VisibleForTesting
+    void setApiServerDisabled(boolean disabled) {
```
nickbp: (very optional/subjective) alternate option: a `protected` method that returns whether the API server is enabled, and then a `TestDefaultScheduler extends DefaultScheduler` which overrides that protected method

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: switched

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: No it's not.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: too much `DefaultSchedulerTest` infrastructure to rewrite.  You should take up scope expansion as a hobby.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerFlags.java
gabrielhartmann: k

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: Actually maybe not so bad.  Giving it a try anyway.

loren:
retest this please because https://github.com/mesosphere/dcos-commons/commit/7c0d3213111f11f4c63d3a95ea330e1b72e23a2d

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
nickbp: Oh I see thats a `(` not a `{`

PR #795, opened 2017-04-03
[SDK] Fix footprint reservation calculation

gabrielhartmann:
The reverted commit was not recording Task's filling out the pod footprint was causing those resources to be `UNRESERVE`d.  They must be recorded, however the `TaskStatus`es with empty TaskIDs should not be recorded.  The original bug of failing StateStore validation on Scheduler restart is fixed.


gabrielhartmann:
retest this please

PR #799, opened 2017-04-04
[INFINITY-1268] Be permissive of network flakes in github_updater

nickbp:
Was still an issue when the updater was being called directly from python scripts.

### DIFF
--- a/tools/github_update.py
+++ b/tools/github_update.py
@@ -19,6 +19,7 @@ import sys
 import subprocess
 import tempfile
 import time
+import traceback
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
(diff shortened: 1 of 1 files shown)


tools/github_update.py
```diff
@@ -198,12 +199,16 @@ def list_contexts(self):
         '''returns a set of context labels of all statuses in a given commit'''
         if not self._api:
             return set([])
-        return set([status['context'] for status in self._api.get_commit_statuses()])
+        try:
+            return set([status['context'] for status in self._api.get_commit_statuses()])
+        except:
+            # in the event of network error, just log the exception and continue
+            traceback.print_exc()
```
jsrodman: I'm slightly obsessive about format_exc() and reporting it via standard channels.  Take or leave.

nickbp: Yeah I'd checked the docs just in case. It looks like it at least defaults to stderr as opposed to stdout.

PR #800, opened 2017-04-04
Add upload plan for putting snapshots in configurable bucket

mrbrowning:


frameworks/cassandra/universe/resource.json
```diff
@@ -5,7 +5,8 @@
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
       "scheduler-zip": "{{artifact-dir}}/cassandra-scheduler.zip",
       "bootstrap-zip": "{{artifact-dir}}/bootstrap.zip",
-      "executor-zip": "{{artifact-dir}}/executor.zip"
+      "executor-zip": "{{artifact-dir}}/executor.zip",
+      "awscli-zip": "https://s3cli-artifacts.s3.amazonaws.com/s3cli-0.0.55-linux-amd64"
```
nickbp: who owns this bucket? should it be copied into downloads.mesosphere.com? (which is confusingly backed by a bucket named downloads.mesosphere.io)

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -76,6 +80,13 @@ pods:
         goal: FINISHED
         cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool snapshot -p {{TASKCFG_ALL_JMX_PORT}} -t $CASSANDRA_SNAPSHOT_NAME -- $CASSANDRA_KEYSPACES
         resource-set: sidecar-resources
+      upload-aws:
+        goal: FINISHED
+        cmd: >
+                chmod a+x {{S3CLI_VERSION}} && eval "echo \"$(sed 's/\"/\\\"/g' < s3config.json)\"" > config.json
```
nickbp: what's the `sed` doing?

could it make sense to just write config.json from scratch with an `echo` here?

frameworks/cassandra/universe/resource.json
mrbrowning: It should be, yeah.

frameworks/cassandra/src/main/dist/svc.yml
mrbrowning: `eval` would otherwise nuke the quote characters in the JSON content of the file, so they need to be escaped again. I could go either way with having the JSON being part of the command, but lean towards it being easier to read/edit JSON in a separate file.

frameworks/cassandra/universe/resource.json
```diff
@@ -5,7 +5,8 @@
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
       "scheduler-zip": "{{artifact-dir}}/cassandra-scheduler.zip",
       "bootstrap-zip": "{{artifact-dir}}/bootstrap.zip",
-      "executor-zip": "{{artifact-dir}}/executor.zip"
+      "executor-zip": "{{artifact-dir}}/executor.zip",
+      "awscli-zip": "https://s3.amazonaws.com/downloads.mesosphere.io/dcos-commons/artifacts/s3cli-0.0.55-linux-amd64.zip"
```
nickbp: `https://downloads.mesosphere.com/dcos-commons/...`?

frameworks/cassandra/src/main/dist/svc.yml
gabrielhartmann: I like the JSON in a separate file.  The command isn't too beautiful though.  I'm trying to say something constructive, but I don't have an elegant solution handy, just think we could find a better way if we thought about it a bit.

gabrielhartmann:
@mrbrowning:  Is the intent to have an endpoint / command at some point that executes both plans?  i.e. snapshot and then upload?  I would have though we'd get a backup plan, that organizes the whole routine.

gabrielhartmann:
@mrbrowning:  I think this isn't relevant anymore.  Correct me if I'm wrong, by re-opening the PR.

PR #804, opened 2017-04-05
Populate template README with starter content and identifiers stubbed out

nickbp:
Updates packaging with generic bits to match.


mgummelt:
I'd rather have a single document for all the shared functionality that the service-specific docs can link to, rather than duplicating the information in each service.

nickbp:
INFINITY-1266

PR #805, opened 2017-04-05
[INFINITY-1278]Change the default cluster name to better hint at responsible parties.

jsrodman:


jsrodman:
Well, at least the test failure is entirely unrelated.

PR #806, opened 2017-04-05
[CASSANDRA] Add replace test that shutdowns a node and replaces a pod

mbdcos:


mbdcos:
retest this please

frameworks/cassandra/tests/test_recovery_shutdown.py
```diff
@@ -0,0 +1,75 @@
...
+    print('marathon ip = {}'.format(service_ip))
+
+    node_ip = 0
+    for pod_id in range(0, DEFAULT_TASK_COUNT):
+        node_ip = get_pod_host(pod_id)
+        if node_ip != service_ip:
+            break
+
+    if node_ip is None:
+        return
```
benclarkwood: Will this fail the test? Because the test _should_ fail if it can't determine an agent to terminate right?

frameworks/cassandra/tests/test_recovery_shutdown.py
```diff
@@ -0,0 +1,75 @@
...
+
+    old_agent = get_pod_agent(pod_id)
+    print('pod id = {},  node_ip = {}, agent = {}'.format(pod_id, node_ip, old_agent))
+
+    task_ids = tasks.get_task_ids(PACKAGE_NAME, 'node-{}'.format(pod_id))
+
+    shakedown.run_command_on_agent(node_ip, 'sudo shutdown now')
```
benclarkwood: run_command_on_agent returns a boolean, as well as the output of the command.

Can you please verify that boolean comes back as True, and print the output of the command?

I don't expect the command to ever fail, but it'd be nice to have the logging there if it does.

frameworks/cassandra/tests/test_recovery_shutdown.py
```diff
@@ -0,0 +1,75 @@
...
+
+    cmd.run_cli('cassandra pods replace node-{}'.format(pod_id))
+
+    tasks.check_tasks_updated(PACKAGE_NAME, 'node', task_ids)
+    tasks.check_running(PACKAGE_NAME, DEFAULT_TASK_COUNT)
```
benclarkwood: The replacement happens pretty quickly right? If not, the default timeout of check_running may not be long enough.

frameworks/cassandra/tests/test_recovery_shutdown.py
mbdcos: There is no guarantee that tasks will be in separate nodes. But for Cassandra, that is the case. Correct.

frameworks/cassandra/tests/test_recovery_shutdown.py
mbdcos: k

frameworks/cassandra/tests/test_recovery_shutdown.py
mbdcos: check_task_updated waits 15mins, and task lost comes after 10min.

PR #807, opened 2017-04-05
Update _config.yml

sascala:


nickbp:
The CI image has been updated with the jekyll redirect plugin, so this should be all set, modulo any caching of CI images in jenkins

nickbp:
Looks like the build was successful, but a bug in generate.sh resulted in the redirect file not being created in the gh-pages branch (specifically: a test is being done to check if files were modified, but that check apparently doesn't detect added files). I'll fix that check in a separate commit...

sascala:
@nickbp Thanks for getting this over the finish line.

PR #808, opened 2017-04-05
Add preview notice to services

ryadav88:


gabrielhartmann:
@mrbrowning and @ryadav88:  Since the release of Cassandra from `dcos-commons` is coming up fast, might want to update that one as well.

PR #811, opened 2017-04-05
[SDK] Skip placement constraints in transient failure, even if placem…

mbdcos:
* Skip placement constraints in transient failure, even if placement rule are set/modified after first task launch.

Current code will restart all tasks since changing placement rule is a config update. For example:
If I give a wrong placement rule (there is no way to satisfy the rule), all tasks will be restarted and will be tried to find a matching offer. Since no such offer exists, no tasks will be relaunched.

Here we want to make sure that if a reservation is already made, do not consider placement rules in offer evaluation. If you want to move your task and apply a new placement rule, you should manually do that with replace.

We are delaying evaluation of placement rules (till replace or new deployment). 

"You start a scheduler with a placement rule (or no placement rule). All tasks are running. You update to a new config with a new placement rule. Maybe your placement rule is invalid. Nothing happens after update. No restart, no replace. Old config is deleted. New config is target of all tasks. Till you run replace, or till you fail, you have no idea whether your placement rule will work or not. Placement rule will be applied for new deployments or for permanent failures (including replace)."


benclarkwood:
Talked with @gabrielhartmann. To fully fix this we want to do:

1. Remove placement rules from config change detection (See branch `constraints`, for sufficient code)

2. Never consider placement rules in transient (existing) offer evaluation. The current code changes do this (with some changes in review comments). 

Gabriel's fixes for #1 should get added into this branch.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -225,7 +225,9 @@ public OfferRequirement getExistingOfferRequirement(PodInstanceRequirement podIn
                 podInstance.getIndex(),
                 taskRequirements,
                 ExecutorRequirement.create(getExecutor(podInstance, serviceName, targetConfigurationId)),
-                podInstance.getPod().getPlacementRule());
+                //TODO(MB): do not add placement rules to getExistingOfferRequirement
```
benclarkwood: What does this TODO mean? It seems like the Optional.empty() satisfies this?

Also, can we write a unit test that exercises this?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -39,8 +39,9 @@ public OfferEvaluator(StateStore stateStore, OfferRequirementProvider offerRequi
 
     public List<OfferRecommendation> evaluate(OfferRequirement offerRequirement, List<Offer> offers)
             throws StateStoreException, InvalidRequirementException {
+        List<OfferEvaluationStage> evaluationStages = getEvaluationPipeline(offerRequirement);
+
         for (int i = 0; i < offers.size(); ++i) {
-            List<OfferEvaluationStage> evaluationStages = getEvaluationPipeline(offerRequirement);
```
benclarkwood: The evaluation stages should be recreated for every offer as they mutate their state.

mbdcos:
Adding others since this changes behavior of scheduler!

This PR ignores placement rule if it is a transient failure (restarting and using already reserved resource). 

A scenario that bothers me:  You start a scheduler with a placement rule (or no placement rule). All tasks are running. You update to a new config with a new placement rule. Maybe your placement rule is invalid. Nothing happens after update. No restart, no replace. Old config is deleted. New config is target of all tasks. Till you run replace, or till you fail, you have no idea whether your placement rule will work or not. 

The following does not fit into our design. We need to discuss further if we would like to implement with two steps.
 1) hold offers that satisfy placement rule, 2) stop task, 3) reserve new offers and start task, 4a) release old offers, 4b) if can not find new offers or task can not start, start task with old offers and release new offers.

gabrielhartmann:
@mbdcos:  Two quick things.  

1. We should merge this PR with #817, which suppresses config change detection based on placement rules.  As written this PR would allow for a scenario where all tasks are restarted and cannot be relaunched because the constraint is not satisfiable.
1. It is core to our design that we role out configuration changes as soon as they're detected, this is true.  However it is also core the current design that we never automatically perform a destructive operation.  As a migration of a service from one set of placement rules to another could potentially move every Task, we could lose all data.  We have always defer this decision to the user.  This might not be the policy for all time, but it is a line we so far choose not to cross.

mbdcos:
@gabrielhartmann Yes this PR is doing as you mention in .2  I want to get opinions, also it would be great if we get placement rule behavior document.

mbdcos:
Could you please explain #817 here little bit. How is it different than checking config updater check?  

private static boolean areDifferent(PodSpec podSpec1, PodSpec podSpec2) {
        if (podSpec1.equals(podSpec2)) {
            return false;
        }
       podSpec1 = DefaultPodSpec.newBuilder(podSpec1).count(0).placementRule(null).build();
       podSpec2 = DefaultPodSpec.newBuilder(podSpec2).count(0).placementRule(null).build();
       return !podSpec1.equals(podSpec2);
    }

gabrielhartmann:
@mbdcos:  The podSpec comparison change looks identical in result.

gabrielhartmann:
@mbdcos:  Effect looks the same to me.  Your way is probably better.  I didn't see that change before.

mbdcos:
This PR is modifying "areDifferent" method, following the same pattern as it is done for "count".

gabrielhartmann:
Yes, looks like a good way of doing it to me @mbdcos.

gabrielhartmann:
@mbdcos:  What sort of placement rule behavior document would you like?  Design doc? Scope doc?  The current behavior is broken so I don't think documenting that it doesn't work is worth doing.  Hopefully we can get this PR in pretty soon.

mbdcos:
Thank you.

frameworks/helloworld/tests/test_marathon_constraint.py
```diff
@@ -127,3 +130,119 @@ def ensure_multiple_per_agent(hello, world):
             assert False, "Unknown task: " + task['name']
     assert len(hello_agents) == len(set(hello_agents)) * hello
     assert len(world_agents) == len(set(world_agents)) * world
+
+
+@pytest.mark.sanity
+@pytest.mark.recovery
+@pytest.mark.placement_update
```
gabrielhartmann: I would remove this mark.

frameworks/helloworld/tests/test_marathon_constraint.py
```diff
@@ -127,3 +130,119 @@ def ensure_multiple_per_agent(hello, world):
...
+def test_change_constraint():
```
gabrielhartmann: I thought we were going to have 3 different tests.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -225,7 +225,9 @@ public OfferRequirement getExistingOfferRequirement(PodInstanceRequirement podIn
                 podInstance.getIndex(),
                 taskRequirements,
                 ExecutorRequirement.create(getExecutor(podInstance, serviceName, targetConfigurationId)),
-                podInstance.getPod().getPlacementRule());
+                // do not add placement rules to getExistingOfferRequirement
```
gabrielhartmann: Please capitalize comments.  Not a big deal.

mbdcos:
retest this please

mbdcos:
retest this please

mbdcos:
retest this please

mbdcos:
retest this please

mbdcos:
retest this please

benclarkwood:
retest this please

.gitignore
```diff
@@ -46,3 +46,5 @@ frameworks/*/dcos-commons-tools/
 # OS X
 .DS_Store
 
+# ignore shakedown_env
+tools/*/shakedown_env
```
nickbp: maybe put next to existing `tools/shakedown_env/` ignore

mbdcos:
@gabrielhartman This PR has all tests passed already.  Committing this PR, so any changes on top of it can be done in the next PR. No need to wait another couple of hours for all tests to complete (assuming this PR is urgently needed since marked as a blocker today).

Commit d428fba 6  days ago passed tests. Commit f948645 (which only divides test function into three separate tests according to PR review) is rebased with master and updated, so it also passed all tests: infinity-dcos-commons-pr #3082 

https://jenkins.mesosphere.com/service/jenkins/job/dcos-commons/job/infinity-dcos-commons-pr/3082/console

@benclarkwood About your comment on DefaultConfigurationUpdaterTest: Need to change tester classes to facilitate placement rules in comparison. It requires some work to modify test helpers. Skipping it for now. This can be addressed in the next PR. 

@gabrielhartman integration test in d428fba is divided into three separate tests. Those cover more broad restart/replace test cases as requested in review, and all those three tests uninstalls/installs separately in f948645. 

Other review comments (including the one from @nickbp) are also addressed. Test from @loren that describes expected behavior has been included.

mbdcos:
retest this please

gabrielhartmann:
@mbdcos: This should not have been merged.  This was closed for a reason.  I have been avoiding having a very strict policy around reviews and who is allowed to merge PRs.  This kind of thing starts to make a stricter policy necessary.

mbdcos:
What is that reason?

I do not see any note saying this PR has a problem. Nothing in any other channel. In fact, this PR is reverted and exact duplicate with a subset of tests were merged. 

What kind of things? It is an approved and all tests cleared PR. 

Anyways, important is that this bug is fixed at least.

frameworks/helloworld/tests/test_marathon_constraint.py
```diff
@@ -127,3 +131,215 @@ def ensure_multiple_per_agent(hello, world):
...
+        "hello": {
+            "count": hello_count - 1,
+            "placement": "hostname:UNIQUE"
+        },
+        "world": {
+            "count": 0
+        }
+    }
+
+    # hostname:CLUSTER:{} will produce:
```
mbdcos: placements are processed

PR #812, opened 2017-04-06
[INFINITY-1012]once again introduce test startup script which can: run tests in parallel, report accurate failures, isolate outoutput from test runs, provide a clean interface at invocation instead of a million env vars, etc etc

jsrodman:
Reintroduce test.py; moved junit out for a separate pull request.


jsrodman:
retest this please

jsrodman:
heh, so test failure was because:
1 - branch still had the bad master manipulation
2 - the merge-master-on-pullrequests wasn't turned on in the jenkins job (build.sh -p)

jsrodman:
retest this please

jsrodman:
@mrbrowning :
At a high level, I don't know if this set of makework scripts is worth trying to convolve into something that  closely resembles "software", especially when we're discussing creating tooling that is largely redundant possibly in another language.  

That said, a large number of essentially unrelated scripts with unrelated styles isn't even the local minimum.

If we want to go down the road of adopting a strict style checker (I'd rather start with an error checker), then I'd rather model the style document that senthil and others are working on than a literalist reading of PEP8.  Still, greater adherence to PEP8 is worthwhile IMO.

test.py
```diff
@@ -0,0 +1,595 @@
...
+import tempfile
+import time
+
+def get_repo_root():
+    return os.path.dirname(sys.argv[0])
+
+logger = logging.getLogger("dcos-commons-test")
+logging.basicConfig(level=logging.INFO, format="%(asctime)s %(name)s %(message)s")
+
+sys.path.append(os.path.join(get_repo_root(), 'tools'))
```
jsrodman: the bulk of test.py probably doesn't belong in the root.  I'm not racing to move the location of every .py file in tools though, and that's why i did the path monkeying.

test.py
```diff
@@ -0,0 +1,595 @@
...
+
+work_dir = None
+def get_work_dir():
+    global work_dir
+    if not work_dir:
+        work_dir = tempfile.mkdtemp(prefix='test_workdir_', dir=get_repo_root())
+        logger.info("Using %s for test run files", work_dir)
+    return work_dir
+
+work_dir = None
```
jsrodman: Bad merge / rebase, I'm sure.  I'll kill it.

test.py
```diff
@@ -0,0 +1,595 @@
...
+            fwinfo.add_framework(framework)
+
+    if run_attrs.order == "random":
+        fwinfo.shuffle_order()
+
+    fw_names = fwinfo.get_framework_names()
+    logger.info("Frameworks initialized: %s", ", ".join(fw_names))
+
+
+def _action_wrapper(action_name, framework, function, *args):
```
jsrodman: Maybe you could explain this one in person.  The scope is very small and the chain is one deep, so I'm not sure what you mean.

This is
1 remember we started doing this thing
2 try to do the thing
3 remember whether the thing worked or failed

Yeah, this could be inlined, or we could have 4 nearly identical functions, but neither seems better to me.

test.py
```diff
@@ -0,0 +1,595 @@
...
+    logger.info("SECURITY set to: '%s'", security)
+    if security == "strict":
+        logger.info("running %s role strict setup script(s)", framework.name)
+        perm_setup_script = os.path.join(repo_root, 'tools', 'setup_permissions.sh')
+
+        custom_env = os.environ.copy()
+        custom_env['CLUSTER_URL'] = cluster.url
+        custom_env['CLUSTER_AUTH_TOKEN'] = cluster.auth_token
+
+        for script_num in (1, 2):
```
jsrodman: pre-existing magic numbers.  I wasn't changing any of the "external interfaces" of this while I wrote it because there was more than enough to figure out, and finding the exceptions usually only happens after you've merged the change.

I suspect what this should be a is a declarative file in the framework test definitions saying what roles they want, where this parses and creates them.

test.py
```diff
@@ -0,0 +1,595 @@
...
+
+    logger.info("Launching shakedown for %s", framework.name)
+
+    custom_env = os.environ.copy()
+    custom_env['TEST_GITHUB_LABEL'] = framework.name
+    custom_env['STUB_UNIVERSE_URL'] = framework.stub_universe_url
+    custom_env['CLUSTER_URL'] = cluster.url
+    custom_env['CLUSTER_AUTH_TOKEN'] = cluster.auth_token
+
+    runtests_script = os.path.join(repo_root, 'tools', 'run_tests.py')
```
jsrodman: Calling directly into run_tests.py makes a lot of sense.  However, it wants a couple things.

First it needs to be able to run pytest in a detached fashion.  
Next there's a LOT of overlap in concepts at this point between run_tests.py and test.py + venvutil + cli_install 
I'm currently stuck on the idea of a workspace for a given framework test run.  run_tests.py certainly a fairly clean one but insists on wiping it exit, which makes debugging very hard.  Also I'd like the output to be where I can conveniently refer to it from a jenkins run to collect artifacts.  However, developers are usually not going to want to keep a lot of test run temp files accruing.  

These aren't hard problems, I just haven't planned it out.

tools/clustinfo.py
```diff
@@ -0,0 +1,159 @@
...
+    return _launch_recorder.get_list()
+
+def add_running_cluster(url, auth_token):
+    cluster = ClusterInfo(url, auth_token, external=True)
+    _clusters.append(cluster)
+    return cluster
+
+def get_cluster_by_url(cluster_url):
+    matches = [cluster for cluster in _clusters if cluster.url == cluster_url]
+    assert len(matches) in (0, 1)
```
jsrodman: I chose this to explicitly emphasize that 0 was valid, partly because I felt the list comprehension was misleading.  It might be better off just returning the first match or None.

tools/clustinfo.py
```diff
@@ -0,0 +1,159 @@
...
+    if matches:
+        return matches[0]
+    return None
```
jsrodman: Yes, but explicit is better than implicit.

tools/fwinfo.py
```diff
@@ -0,0 +1,166 @@
...
+        # failures
+        action['error_message'] = args.get('error_message')
+        action['error_output'] = args.get('error_output')
+
+if __name__ == "__main__":
+    import unittest
+    rel_root_path = os.path.dirname(os.path.dirname(__file__))
+    abs_repo_root = os.path.abspath(rel_root_path)
+
+    class tests(unittest.TestCase):
```
jsrodman: I think inline tests for a relatively small module are virtuous.  But I'm not sure I got your point.

mrbrowning:
I think you're right that, in general, scripts like these need not be treated as "software" in the sense that they be constructed with maintainability in mind, but there's a complexity cutoff beyond which that's no longer true that I think we've crossed here. I stress readability primarily because I don't believe it's orthogonal to error incidence. Error checkers are a treatment, linters are a prophylactic, and stylistic consistency matters more than any particular maxim, so honestly the choice of style guide is arbitrary as long as it's statically enforceable, since really it's a computer's job to do that.

PR #814, opened 2017-04-06
install Elastic with check_suppression=False

loren:
[INFINITY-1296]


loren:
retest this please now that jenkins is back up.

PR #815, opened 2017-04-06
[Infinity-1107] merge pullrequests for tests

jsrodman:
https://github.com/mesosphere/dcos-commons/pull/771 was not sufficient.  It worked in one test, but probably there was no real work to do.  When there was a real merge to record, git would balk saying "hey i don't have a name and email address for this commit".  

One approach would be to configure git with a name and address, since the whole local repo is not currently intended to be kept.  However, this feel dangerous to me, if we change workflow later.

The other approach is to avoid ever creating a commit, which I'm taking here.


jsrodman:
I guess I can only test this properly with a custom jenkins job using this branch, pointing to other git branches that I arrange to be in different states.

build.sh
```diff
@@ -31,9 +36,32 @@ _notify_github() {
     $REPO_ROOT_DIR/tools/github_update.py $1 build:sdk $2
 }
 
+merge_master() {
+    echo "attmpting to merge changes from master"
```
nickbp: typo

build.sh
```diff
@@ -31,9 +36,32 @@ _notify_github() {
...
+    # git sucks and won't let you update files without knowing a name
+    echo "Creating fake user."
+    command="git config --global user.email pullrequestbot@mesospherebot.com"
+    echo $command
+    if ! $command; then
+        return 1 # fail
+    fi
+    command="git config --global user.name Infinity-tools-fake-user"
+    FAKE_EMAIL="pullrequestbot@mesospherebot.com"
```
nickbp: this FAKE_EMAIL line looks unused

build.sh
```diff
@@ -31,9 +36,32 @@ _notify_github() {
     $REPO_ROOT_DIR/tools/github_update.py $1 build:sdk $2
 }
 
+merge_master() {
+    echo "attmpting to merge changes from master"
+    # git sucks and won't let you update files without knowing a name
+    echo "Creating fake user."
+    command="git config --global user.email pullrequestbot@mesospherebot.com"
```
nickbp: FWIW the `--global` may not be necessary, see:
- docs generator script which pushes to gh-pages: https://github.com/mesosphere/dcos-commons/blob/master/docs/generate.sh#L111
- sample result: https://github.com/mesosphere/dcos-commons/commit/0ee0f3a22a731a4ae3a035b7a56e3b873f0b381c

build.sh
jsrodman: Done.

build.sh
jsrodman: Done.

build.sh
jsrodman: Removed.

jsrodman:
retest this please

jsrodman:
It would be amazing if CCM could give anything a working cluster.

PR #817, opened 2017-04-06
[SDK] Suppress config change detection based on placement rules

gabrielhartmann:


mbdcos:
@gabrielhartmann Could you please explain what you are doing here? May be a few lines in description.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -175,11 +177,24 @@ private Comparator() {
         }
 
         @Override
-        public boolean equals(ServiceSpec first, ServiceSpec second) {
-            return EqualsBuilder.reflectionEquals(first, second);
+        public boolean equals(ServiceSpec current, ServiceSpec proposed) {
```
mbdcos: See: DefaultConfigurationUpdater.java:L117 
You can not change placement rule at all if this PR is in. Am I missing something here?

benclarkwood: Yep, you are right. Good catch.

benclarkwood:
This PR would have the bad consequence of not letting you change placement constraints ever.

https://github.com/mesosphere/dcos-commons/pull/811 has the proper solution.

PR #820, opened 2017-04-07
[SDK] Fix consumption of multiple mount volumes in a Task

gabrielhartmann:


sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorTest.java
```diff
@@ -330,22 +330,56 @@ public void testReserveTaskDynamicVIPPort() throws Exception {
...
+        Assert.assertEquals(5, recommendations.size());
+        Assert.assertEquals(Operation.Type.RESERVE, recommendations.get(0).getOperation().getType());
+        Assert.assertEquals(Operation.Type.CREATE, recommendations.get(1).getOperation().getType());
+        Assert.assertEquals(Operation.Type.RESERVE, recommendations.get(2).getOperation().getType());
+        Assert.assertEquals(Operation.Type.CREATE, recommendations.get(3).getOperation().getType());
+        Assert.assertEquals(Operation.Type.LAUNCH, recommendations.get(4).getOperation().getType());
+    }
+
+    @Test
+    public void testConsumeMultipleMountVolumesFailure() throws Exception {
```
mohitsoni: Maybe we should add another failure test (if not already present) where an offerResource volume is not sufficient in size.

gabrielhartmann: See: `testFailToCreateVolumeWithWrongResource`

nickbp:
Verified via testing of 0.13.2.1 hotfix which contained this patch

PR #826, opened 2017-04-08
Implement backup and restore plans and tasks for Cassandra

mrbrowning:


frameworks/cassandra/universe/resource.json
```diff
@@ -5,7 +5,8 @@
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
       "scheduler-zip": "{{artifact-dir}}/cassandra-scheduler.zip",
       "bootstrap-zip": "{{artifact-dir}}/bootstrap.zip",
-      "executor-zip": "{{artifact-dir}}/executor.zip"
+      "executor-zip": "{{artifact-dir}}/executor.zip",
+      "awscli-zip": "https://downloads.mesosphere.io/dcos-commons/artifacts/s3cli-0.0.55-linux-amd64.zip"
```
nickbp: IIRC this should be `downloads.mesosphere.com`

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -77,9 +82,84 @@ pods:
         goal: FINISHED
         cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool cleanup -p {{TASKCFG_ALL_JMX_PORT}} -- $CASSANDRA_KEYSPACE_TABLES
         resource-set: sidecar-resources
+      backup-schema:
+        goal: FINISHED
+        cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/cqlsh --cqlversion="3.4.0" -e "DESC SCHEMA" node-0-server.{{SERVICE_NAME}}.mesos > container-path/schema.cql
```
nickbp: Can the `--cqlversion` here and elsewhere be removed or is it needed for this to work? (I don't know either way, just bringing it up)

frameworks/cassandra/universe/resource.json
nickbp: The binary inside could be `strip`ped, would cut the zip file by a third

frameworks/cassandra/universe/marathon.json.mustache
```diff
@@ -129,6 +130,7 @@
     "TASKCFG_ALL_REMOTE_SEEDS": "{{service.remote_seeds}}",
     "EXECUTOR_URI": "{{resource.assets.uris.executor-zip}}",
     "BOOTSTRAP_URI": "{{resource.assets.uris.bootstrap-zip}}",
+    "AWSCLI_URI": "{{resource.assets.uris.awscli-zip}}",
```
nickbp: Should probably just refer to it as the S3CLI throughout, as the aws cli is something else.

frameworks/cassandra/src/main/dist/svc.yml
mohitsoni: Also, this one relies on `node-0-server`, I'm just wondering if this will ever leave somebody in a state where they cannot backup data, because `node-0-server` cannot be brought up ?

mohitsoni: FWIW, I also don't have a solution for this problem for the moment.

mrbrowning: @nickbp it's not necessary, given that `cqlsh` will always be the version that shipped with the version of C* we're running against. Going to remove.

@mohitsoni it could use the VIP instead? Would that work?

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -77,9 +82,84 @@ pods:
...
         cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool cleanup -p {{TASKCFG_ALL_JMX_PORT}} -- $CASSANDRA_KEYSPACE_TABLES
         resource-set: sidecar-resources
+      backup-schema:
+        goal: FINISHED
+        cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/cqlsh -e "DESC SCHEMA" "node-${POD_INSTANCE_INDEX}-server.{{SERVICE_NAME}}.mesos" > container-path/schema.cql
+        resource-set: sidecar-resources
       snapshot:
         goal: FINISHED
-        cmd: ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool snapshot -p {{TASKCFG_ALL_JMX_PORT}} -t $CASSANDRA_SNAPSHOT_NAME -- $CASSANDRA_KEYSPACES
+        cmd: >
```
gabrielhartmann: For the large `cmd`s here and below, I think it might be a bit nicer to separate these out into separate files.

frameworks/cassandra/tests/test_backup.py
```diff
@@ -0,0 +1,108 @@
...
+def launch_and_verify_job(job_name, expected_successes=1):
+    cmd.run_cli('job run {}'.format(qualified_job_name(job_name)))
+
+    spin.time_wait_noisy(lambda: (
+        'Successful runs: {}'.format(expected_successes) in
+        cmd.run_cli('job history {}'.format(qualified_job_name(job_name)))
+    ))
+
+
+def test_backup_and_restore_flow():
```
gabrielhartmann: No `pytest` mark?

mrbrowning:
@mohitsoni `cqlsh` doesn't like the VIP. The problem remains, anyway, since we're forced to choose some pod index to run the sidecar task in. This could be resolved with a step specifier similar to `default`, but that means choose any pod. For now, this seems to be the way to go.

PR #828, opened 2017-04-10
[INFINITY-1305] Automate SDK releases against tags

nickbp:
This automates the process for creating a new release of the SDK (jars for now, other bits soon)
- `release.sh -r snapshot`: Publish build of current master (what current release.sh does)
- `release.sh -r 1.2.3`: Create/push 1.2.3 tag (with build.gradle updated), then publish build of 1.2.3
- Escape hatch if a tag already exists (e.g. build flaked after creating the tag on prior attempt): '-f' to force override

Policy changes:
- Master only has SNAPSHOT versions, never release versions.
- Release versions are created against release tags.

TODO:
- Once script is available, create manually-invoked jenkins job with parameters: version string and force checkbox

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,16 +16,9 @@ allprojects {
 
     apply from: "$rootDir/gradle/quality.gradle"
 
-    // ---
-    // To create a release release:
-    // 1. Merge a PR with "-SNAPSHOT" suffix removed
-    // 2. Merge a second PR which bumps the version with "-SNAPSHOT" suffix added back
-    // For example, given a version of "1.2.3-SNAPSHOT", merge "1.2.3" then "1.2.4-SNAPSHOT".
-    // ---
-    // In general, users of this SDK should NOT depend on "-SNAPSHOT" builds, as they may break
-    // at ANY TIME as changes are merged into master. They represent a fully unstable API.
-    // ---
     group = "mesosphere"
+    // In master, this is ALWAYS a -SNAPSHOT version.
+    // Releases are made FROM TAGS. To create a release, see release.sh.
     version = "0.14.0-SNAPSHOT"
 
     sourceCompatibility = '1.8'
--- a/build.gradle
+++ b/build.gradle
@@ -14,11 +14,25 @@ allprojects {
     apply plugin: 'maven-publish'
     apply plugin: 'com.github.ksoichiro.console.reporter'
 
-    apply from: "$rootDir/gradle/quality.gradle"
+    apply from: '$rootDir/gradle/quality.gradle'
 
-    group = "mesosphere"
-    // In master, this is ALWAYS a -SNAPSHOT version.
-    // Releases are made FROM TAGS. To create a release, see release.sh.
+    group = 'mesosphere'
+    // ---
+    // INFO:
+    //
+    // In general, users of this SDK may use '-SNAPSHOT' builds for testing or development, but
+    // they MUST NOT be used for releases. A given '-SNAPSHOT' build will vary on a per-PR basis
+    // and APIs WILL break at any time. See the tags in this repo for a list of stable releases.
+    //
+    // In repo master, this version MUST ALWAYS be a '-SNAPSHOT' version.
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -16,16 +16,9 @@ allprojects {
...
 
-    // ---
-    // To create a release release:
-    // 1. Merge a PR with "-SNAPSHOT" suffix removed
-    // 2. Merge a second PR which bumps the version with "-SNAPSHOT" suffix added back
-    // For example, given a version of "1.2.3-SNAPSHOT", merge "1.2.3" then "1.2.4-SNAPSHOT".
-    // ---
-    // In general, users of this SDK should NOT depend on "-SNAPSHOT" builds, as they may break
-    // at ANY TIME as changes are merged into master. They represent a fully unstable API.
-    // ---
```
jsrodman: Where is the equivalent hinting moving?

release.sh
```diff
@@ -1,16 +1,95 @@
 #!/bin/bash
 
+# Builds and uploads a release, optionally creating a tag for that release.
```
jsrodman: Perhaps we can say what we're releaseing explicitly.  I think it's the SDK, but we have services in the repo.

release.sh
```diff
@@ -1,16 +1,95 @@
...
+        exit 1
+    fi
+else
+    # Uploading a release: Create and publish the release tag, which contains a commit updating build.gradle.
+    # Sanity check: Tagged releases of snapshots is disallowed (case insensitive comparison)
+    if [[ "${FLAG_RELEASE^^}" = *"SNAPSHOT"* ]]; then
+        echo "Requested tag release version may not contain 'SNAPSHOT'."
+        exit 1
+    fi
+    sed -i "s/version = \\\"${SDK_VERSION}\\\"/version = \\\"${FLAG_RELEASE}\\\"/" $GRADLE_BUILD_FILE
```
jsrodman: Shell :-(

release.sh
```diff
@@ -1,16 +1,95 @@
...
+    export SDK_VERSION=$FLAG_RELEASE
+    git add $GRADLE_BUILD_FILE
+    git ci -m "Automated cut of $FLAG_RELEASE"
+    if [ "x$FLAG_FORCE" = "xtrue" ]; then
+        git tag -f "$FLAG_RELEASE"
+        git push origin -f "$FLAG_RELEASE"
+    else
+        FAILED=""
+        git tag "$FLAG_RELEASE" || FAILED="true"
+        # Jump through hoops to avoid conflicts with 'set -e':
```
jsrodman: I'm pretty sure this is equivalent to
```
if ! git tag "$FLAG_RELEASE"; then 
    echo ...
    exit 1
fi
```

Though set -e is vastly underdocumented, so I couldn't say 100%.

release.sh
```diff
@@ -1,16 +1,95 @@
...
+        fi
+    fi
+fi
+
 # Build and upload Java library to Maven repo:
 ./build.sh
 ./gradlew publish
 
-# Scripts below expects S3_BUCKET and SDK_VERSION env vars. To be supplied by Jenkins.
+# Scripts below expect an S3_BUCKET env var. To be supplied by Jenkins.
```
jsrodman: Are the commented lines still meaningful?

release.sh
nickbp: Added some more info. This just does the SDK itself (currently just jars, soon also prebuilt versions of `bootstrap`/executor/CLI).

build.gradle
nickbp: Good point, updated this section to retain background info and step-by-step under the new system

release.sh
nickbp: Agreed. This script is doing a lot of cmdline interaction with all the `git` calls though

release.sh
nickbp: Yes they'll be reenabled in a couple days.

nickbp:
retest this please

nickbp:
retest this please

PR #835, opened 2017-04-11
[INFINITY-1318] Fix UNRESERVE of MOUNT volume disk resource

gabrielhartmann:
1. Deployed `hello-world` with MOUNT volumes with just enough machines for it to fit. 
1. Replaced a node and verified that it couldn't get rescheduled because it was generating an invalid UNRESERVE operation
1. Upgraded to this fix.
1. Success

Also matches code from Marathon now: https://github.com/mesosphere/marathon/blob/9dab2704f27243825cbd252588d3476ce980d9bb/src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala#L76

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/UnreserveOfferRecommendation.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/UnreserveOfferRecommendation.java
@@ -16,13 +16,21 @@ public class UnreserveOfferRecommendation implements OfferRecommendation {
 
     public UnreserveOfferRecommendation(Offer offer, Resource resource) {
         this.offer = offer;
+        Resource.Builder resourceBuilder = resource.toBuilder();
+
+        if (resource.hasDisk() && resource.getDisk().hasSource()) {
+            resource = resourceBuilder.setDisk(
+                    Resource.DiskInfo.newBuilder()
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/UnreserveOfferRecommendation.java
```diff
@@ -16,13 +16,21 @@
 
     public UnreserveOfferRecommendation(Offer offer, Resource resource) {
         this.offer = offer;
+        Resource.Builder resourceBuilder = resource.toBuilder();
+
+        if (resource.hasDisk() && resource.getDisk().hasSource()) {
```
benclarkwood: Can you include a comment similar to what they have in Marathon to indicate _why_ this is being done?

benclarkwood: https://github.com/mesosphere/marathon/blob/9dab2704f27243825cbd252588d3476ce980d9bb/src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala#L76

gabrielhartmann:
@nickbp @benclarkwood.  Comments addressed.

PR #836, opened 2017-04-11
Improve the master settings modification code

benclarkwood:
Add an intelligent wait that waits for the mesos master to be back up before continuing

Set agent_removal_rate_limit and max_agent_ping_timeouts.

### DIFF
--- a/tools/modify_master.py
+++ b/tools/modify_master.py
@@ -8,8 +8,8 @@ import time
 
 
 # Methods to modify the envvar settings of a mesos master and restart
-# the master process after they are modified
-
+# the master process after they are modified. Ideally, we would just get clusters
+# that are preconfigured like this. But, this allows us to modify an existent cluster.
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


tools/modify_master.py
```diff
@@ -82,20 +82,29 @@ def write_envvars(envvars, commented_envvars):
...
+        # During our testing, we remove agents. This requires us to 
+        # increase the rate at which we can remove agents. The new value
+        # of the rate limit indicates we can remove 100 agents every 20 minutes.
+        'MESOS_AGENT_REMOVAL_RATE_LIMIT': '100/20mins',
+        # During our testing, we remove agents. This requires us to lower the number of checks
+        # that are done before removal when an agent is not responding so that removal
+        # is quicker. The ping timeout (MESOS_AGENT_PING_TIMEOUT) is 15 seconds, and
+        # thus we will wait 1 minute for agent death.
+        'MESOS_MAX_AGENT_PING_TIMEOUTS': '4'
+    }
```
jsrodman: Shouldn't modify take care of overwrite? (In an ideal world anyway)
Technically it's possible to have multiple entries by the same name in an environment table, but nothing should permit it.

benclarkwood: I'm swapping envvar names, out of spite.

PR #837, opened 2017-04-11
[Infinity-1311] test output stderr

jsrodman:
After a few failed experiments, the working pattern is to just ask python not to buffer output.  Since test.py is already running run_tests.py with the same file as fd 1 and 2, this should provide a reliable linear stream of events.   That's a very small change.

The much larger linecount change is the move from print to a dedicated "this is test output" function.  I'm not married this.  But I wanted to move away from print inline everywhere since that's a growing problem.  

I can also decouple the test.py fix, though given the pace of my reviews and merges I'm not excited about it.

### DIFF
--- a/test.py
+++ b/test.py
@@ -511,7 +511,8 @@ def start_test_background(framework, cluster, repo_root):
     _setup_strict(framework, cluster, repo_root)
 
     logger.info("Launching shakedown in background for %s", framework.name)
-    logger.debug("stub_universe:{framework.stub_universe_url} cluster_url:{cluster.url} authtoken:{cluster.auth_token}".format(locals()))
+    logger.debug("stub_universe:%s cluster_url:%s authtoken:%s",
+            framework.stub_universe_url, cluster.url, cluster.auth_token))
 
     custom_env = os.environ.copy()
     custom_env['TEST_GITHUB_LABEL'] = framework.name
@@ -543,7 +544,8 @@ def run_test(framework, cluster, repo_root):
     _setup_strict(framework, cluster, repo_root)
 
     logger.info("Launching shakedown for %s", framework.name)
-    logger.debug("stub_universe:%s cluster_url:%s authtoken:%s" % (framework.stub_universe_url, cluster.url, cluster.auth_token))
+    logger.debug("stub_universe:%s cluster_url:%s authtoken:%s",
+            framework.stub_universe_url, cluster.url, cluster.auth_token))
     custom_env = os.environ.copy()
     custom_env['STUB_UNIVERSE_URL'] = framework.stub_universe_url
     custom_env['TEST_GITHUB_LABEL'] = framework.name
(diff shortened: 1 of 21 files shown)


nickbp:
Maybe shorten usage via either or both:
- name the function `p`?
- import in tests against an alias like `p`?

jsrodman:
@nickbp I don't like copying references out of module namespaces; I feel it leads to less clarity as to what code is being called and I've experience the pain of multiple copies of module implementations in memory at times.

The name certainly could be shorter.  out() ?

testing/sdk_utils.py
```diff
@@ -1,8 +1,35 @@
...
+    #    del parent
+    #name = inspect.getmodulename(parent_filename)
+    #out = "{current_time} {name} {msg}\n".format(current_time=current_time,
+    #                                             name=name,
+    #                                             msg=msg)
+    #sys.stderr.write(out)
+
+
 def gc_frameworks():
+    '''Reeclaims private agent disk space consumed by Mesos but not yet garbage collected'''
```
loren: reclaims

test.py
```diff
@@ -505,7 +510,8 @@ def start_test_background(framework, cluster, repo_root):
                 framework.name)
     _setup_strict(framework, cluster, repo_root)
 
-    logger.info("Launching shakedown for %s", framework.name)
+    logger.info("Launching shakedown in background for %s", framework.name)
+    logger.debug("stub_universe:{framework.stub_universe_url} cluster_url:{cluster.url} authtoken:{cluster.auth_token}".format(locals()))
```
loren: If you are keeping this debug and the one on line 546, please pick one style of formatting that string

jsrodman: Sure, i'll keep the one that actually works.

jsrodman:
btw, the failures are of course me cancelling redundant jobs.

benclarkwood:
I like `out()`

benclarkwood:
And also don't like removing module namespacing. Namespacing is one of my favorite parts of Go

jsrodman:
retest this please

benclarkwood:
@jsrodman has this fixed what you wanted it to fix?

jsrodman:
@benclarkwood it's "fixed enough" when being run by test.py with output redirection.  

Without redirection, there are still open questions about whatever receiver system is doing with the streams.  How does jenkins pretend to be a terminal? I couldn't say.

Initially i'd wanted everything going to stderr with timestamps, but experimentation showed that is not achievable yet; shakedown would need significant change.

jsrodman:
Unrelatedly,  frameworks.cassandra.tests.test_recovery_shutdown.test_shutdown_host_test failed.

jsrodman:
retest this please

jsrodman:
retest this please

jsrodman:
retest this please

jsrodman:
retest this please

jsrodman:
well, kafka hit an SSL error.  Okay whatever.

jsrodman:
retest this please

PR #838, opened 2017-04-11
[INFINITY-1295] followup - 1295 declare python3

jsrodman:
although we used to explicitly reference /usr/bin/python; some mac setups have that, and have an installer-provided python3, but no python3 at /usr/bin/python3


jsrodman:
retest this please

jsrodman:
fwiw, @gabrielhartmann verified this works in his env.

PR #842, opened 2017-04-12
Make default user `nobody`

loren:
- Update upgrade/downgrade to accept additional options, so we can keep username consistent between upgrade/downgrade

Completes ELASTIC-67


loren:
retest this please

gabrielhartmann:
retest this please

PR #847, opened 2017-04-12
[INFINITY-1330] Only template json files; only copy in non-dotfiles.

jsrodman:
@nickbp @mrbrowning Comments and/or approval please.

### DIFF
--- a/tools/universe_builder.py
+++ b/tools/universe_builder.py
@@ -92,6 +92,9 @@ class UniversePackageBuilder(object):
         pkgdir = os.path.join(treedir, 'repo', 'packages', self._pkg_name[0].upper(), self._pkg_name, '0')
         os.makedirs(pkgdir)
         for pkgfile in os.listdir(self._input_dir_path):
+            # Don't copy in .dotfiles; editor temps etc
+            if pkgfile.startswith('.'):
+                continue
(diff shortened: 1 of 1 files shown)


tools/universe_builder.py
```diff
@@ -92,6 +92,9 @@ def _create_tree(self, scratchdir):
         pkgdir = os.path.join(treedir, 'repo', 'packages', self._pkg_name[0].upper(), self._pkg_name, '0')
         os.makedirs(pkgdir)
         for pkgfile in os.listdir(self._input_dir_path):
+            # Don't copy in .dotfiles; editor temps etc
+            if pkgfile.startswith('.'):
```
nickbp: is `.stub_dir` omitted with this check?

mrbrowning:
LGTM otherwise but I defer to Nick's concern.

tools/universe_builder.py
jsrodman: Yes this does not affect .stub_dir.  The .stub_dir is created in the target location higher in the function.  This is simply copying framework/*/universe into the target.  copyfile would fail anyway on a directory.

nickbp: Ah got it, ok great

PR #861, opened 2017-04-13
[DONOTREVIEW] Fix scheduler handling of removed envvars: Ensure task env is correctly updated

nickbp:


nickbp:
To be continued once other prereq work is done

PR #867, opened 2017-04-14
[INFINITY-1162] Fix plan errors

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -289,7 +289,7 @@ public Builder setCapabilities(Capabilities capabilities) {
          * @throws IllegalStateException if config validation failed when updating the target config for a default
          *     {@link OfferRequirementProvider}, or if creating a default {@link ConfigStore} failed
          */
-        public DefaultScheduler build() {
+        public DefaultScheduler build() throws ConfigStoreException {
```
mrbrowning: If it must be so, but I think I'd be a little surprised to catch a ConfigStoreException from this method in general.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -923,10 +931,13 @@ private void postRegister() {
         revive();
     }
 
-    @Override
-    public void update(Observable observable) {
-        if (observable == planCoordinator) {
-            suppressOrRevive();
-        }
+    private Plan getDeployPlan() {
+        return getDeployPlan(plans).get();
```
mrbrowning: What's the point of this?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -689,6 +690,13 @@ private void declineOffers(SchedulerDriver driver, List<Protos.OfferID> accepted
     }
 
     @Override
+    public void update(Observable observable) {
```
mrbrowning: I know it just got moved, but this could probably use a javadoc.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -923,10 +931,13 @@ private void postRegister() {
...
-        if (observable == planCoordinator) {
-            suppressOrRevive();
-        }
+    private Plan getDeployPlan() {
+        return getDeployPlan(plans).get();
+    }
+
+    private static Optional<Plan> getDeployPlan(Collection<Plan> plans) {
+        return plans.stream()
+                .filter(plan -> plan.isDeployPlan())
```
mrbrowning: I'm sure there's not a way for it to happen currently, but for safety's sake we ought to assert somewhere that there's no more than one deploy plan defined. Doesn't have to be in this PR.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: Oh, right.  I'll fix this.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: So I don't have to duplicate code for finding the plan named `deploy` in a list of plans.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: Agreed.  I thought that while I was looking at it.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
mrbrowning: Ok. Maybe still return `Optional<Plan>` for semantics' sake?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Plan.java
```diff
@@ -21,6 +22,8 @@
  * list of one or more error messages to be shown to the user,
  */
 public interface Plan extends ParentElement<Phase> {
+    void setErrors(List<String> errors);
```
nickbp: Feels weird adding a setter to this interface. How about having DefaultScheduler just create a new plan with errors pre-populated in that plan? The plan could omit the other pieces.

Something like:
```
final List<String> errors = new ArrayList<>();
errors.add("err: bad config");
return new Plan() {
  public List<String> getErrors() { return errors; }
};
```

gabrielhartmann: I agree to a degree.  I didn't really like adding the `setter` method.  However the point of the implementation is that we have rejected the proposed target and kept the previous good target.  By the lights of the previous target the plan remains unchanged and is correct.  In fact, if an attempt were made to change from a "good" config to a "bad" one, that original good plan would still be executed, or should be.

If anything we should consider putting errors outside the plan interface altogether.  However the implementation here is a return to the original behavior which had regressed.  Because configuration validation and plan generation are so decoupled (as they should be), I didn't see a good solution that avoided the setter.

nickbp: How about something that `extends Plan`, wraps the original Plan object, and just overrides the `getErrors()` call (and probably also the `getStatus()` call)? Then it'd be a bit more explicit that we're just adding errors onto the existing plan.

```
public class ErrorWrapper {
  public ErrorWrapper(Plan wrapMe, List<String> errors) {
    this.wrapMe = wrapMe;
    this.errors = errors;
  }

  @Override
  public List<String> getErrors() {
    List<String> tmp = new ArrayList<>();
    tmp.addAll(wrapMe.getErrors())
    tmp.addAll(errors);
    return tmp;
  }

  @Override
  public Status getStatus() {
    return ERROR;
  }

  [.. probably some boilerplate here ..]
}
```

gabrielhartmann: We actually already have an appropriate constructor in `DefaultPlan` for this.  Updating and removing the setter.  Thanks for pushing on this @nickbp.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -555,21 +574,8 @@ private void initializeGlobals(SchedulerDriver driver) {
...
-            deployPlan =
-                    new DeployPlanFactory(
-                            new DefaultPhaseFactory(
-                                    new DefaultStepFactory(configStore, stateStore))).getPlan(serviceSpec);
-        } else {
-            deployPlan = deploy.get();
-        }
-        deploymentPlanManager = new DefaultPlanManager(deployPlan);
+        LOGGER.info("Initializing deployment plan manager...");
+        deploymentPlanManager = new DefaultPlanManager(getDeployPlan());
```
nickbp: nitpick: could just call `getDeployPlan(plans).get()` directly, looks like this is the only place using `getDeployPlan()`

PR #869, opened 2017-04-14
[INFINITY-1340] Add a Marathon label to identify SDK based services

gabrielhartmann:


gabrielhartmann:
This is for Cosmos and the UI

PR #870, opened 2017-04-14
Fix race condition in test_unchanged_scheduler_restarts_without_restarting_tasks()

loren:


gabrielhartmann:
@loren: Great.  Thanks for fixing this.

PR #879, opened 2017-04-17
[INFINITY-1363]Incomplete refactor or similar while editing this in e8df7d4db88809bb43145e1a394aa1b44726873e

jsrodman:
@nickbp Just notifying you of typo.


jsrodman:
I'm merging this without tests because the PR-tests don't exercise this file anyway.  cassandra nightly etc will show it works.

PR #880, opened 2017-04-17
[SDK] Refactor all task label access behind a hierarchy of access classes, shrink CommonTaskUtils

nickbp:
This large PR is effectively a no-op, consolidating existing label setter/getter logic behind some thick boilerplate, making it very explicit as to which parts of the code need access to which embedded task metadata, and cordoning that off into a new `[...].offer.taskdata` package. Separate PRs will be created to do the following:
- Similar boilerplate handling for writing/reading task envvars (should be smaller than this one).
- Actual logic changes centered around writing labels and envvars to tasks (mainly in offer eval).

Additionally, a few other non-common parts of `CommonTaskUtils` were broken out into executor-specific and scheduler-specific code. For example, in addition to moving out all the label-related functions, yaml template rendering, sending task status from executors, and checking task status at schedulers are now broken out. As a result the remainder of `CommonTaskUtils` is reduced to what's actually shared logic.

The new Label access structure can be summarized as:
- common: `LabelWriter`/`LabelReader` (underlying read/write access, with one common call to read an unpacked readiness check object)
- executor: `ExecutorLabel[Writer/Reader]` (executor-specific labels. extends common)
- scheduler: `SchedulerLabel[Writer/Reader]` (scheduler-specific labels. extends common)

Bonus fix: Discovered via refactoring that `isMustacheFullyRendered()` wasn't really working (always returned `true`), as the regex search used was broken by newlines. Fixed now, with some unit tests.


sdk/common/src/main/java/com/mesosphere/sdk/offer/taskdata/LabelReader.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    private LabelReader(String taskName, Labels labels) {
+        this.taskName = taskName;
+        this.labels = LabelUtils.toMap(labels);
+    }
+
+    /**
+     * Returns the readiness check to be run by the Executor on task startup.
+     */
+    public Optional<HealthCheck> getReadinessCheck() throws TaskException {
```
benclarkwood: Does it make more sense for this to be in ExecutorLabelReader?

sdk/common/src/main/java/com/mesosphere/sdk/offer/taskdata/LabelReader.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.commons.codec.binary.Base64;
+import org.apache.mesos.Protos.HealthCheck;
+import org.apache.mesos.Protos.Labels;
+import org.apache.mesos.Protos.TaskInfo;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.mesosphere.sdk.offer.TaskException;
+
+/**
+ * Implements common logic for read access to a task's Labels. Any access to component-specific values is provided by
```
benclarkwood: mega-nitpick: Would be spiffier if these were @see's

nickbp:
retest this please (gradle net flake)

sdk/common/src/main/java/com/mesosphere/sdk/offer/taskdata/LabelReader.java
nickbp: Yes. It had been here since both the executor and scheduler had been reading it. The scheduler cases were somewhat special and I was able to move them into the Scheduler-specific wrappers. This has now been moved into ExecutorLabelReader.

sdk/common/src/main/java/com/mesosphere/sdk/offer/taskdata/LabelReader.java
nickbp: Done, with @links

nickbp:
retest this please

nickbp:
retest this please

sdk/common/src/main/java/com/mesosphere/sdk/offer/taskdata/EnvUtils.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.apache.mesos.Protos.Environment;
+
+/**
+ * Utilities relating to environment variable manipulation
+ * TODO(nickbp): Remove in favor of boilerplated env access
```
benclarkwood: What do you mean by this? e.g. an example?

nickbp: This was just intended as a temporary note. There's a separate branch based on this one which does similar work for envvar handling, to be put up as a separate PR once this one's merged.

sdk/scheduler/src/test/java/com/mesosphere/sdk/testutils/OfferRequirementTestUtils.java
```diff
@@ -147,6 +151,15 @@ public static OfferRequirement getOfferRequirement(
         return resourceRequirements;
     }
 
+    public static Optional<HealthCheck> getReadinessCheck(TaskInfo taskInfo) throws TaskException {
+        return new SchedulerLabelWriter(taskInfo) {
+            @Override
+            public Optional<HealthCheck> getReadinessCheck() throws TaskException {
```
mrbrowning: This is overridden just to expand visibility to `public`, right?

sdk/common/src/main/java/com/mesosphere/sdk/offer/taskdata/TaskPacking.java
```diff
@@ -0,0 +1,100 @@
...
+ *   - command
+ *
+ * Packed:
+ * - taskInfo
+ *   - executor
+ *   - data: serialized executorinfo
+ *     - data: custom
+ *     - command
+ */
+public class TaskPacking {
```
mrbrowning: Maybe have this class name match other similar classes (`TaskPackingUtils` or similar) for guessability.

sdk/scheduler/src/test/java/com/mesosphere/sdk/testutils/OfferRequirementTestUtils.java
nickbp: Yes

sdk/common/src/main/java/com/mesosphere/sdk/offer/taskdata/TaskPacking.java
nickbp: Good point. Done!

PR #883, opened 2017-04-17
Ensure that cassandra host shutdown test doesn't kill scheduler host

mrbrowning:


benclarkwood:
retest this please

frameworks/cassandra/tests/test_recovery_shutdown.py
```diff
@@ -29,35 +29,34 @@ def teardown_module(module):
 @pytest.mark.recovery
 @pytest.mark.shutdown_node
 def test_shutdown_host_test():
+    scheduler_ip = shakedown.get_service_ips('marathon', PACKAGE_NAME).pop()
```
benclarkwood: At first this didn't make sense and I was going to say "add a comment, you wizard", but then I read better.

PR #886, opened 2017-04-18
[SPARK-448] Shotgun approach to spark venvutil problem; logging + attempted workaround.

jsrodman:
I have a theory, mentioned in the jira, that this is a known problem with ensurepip, that it can shortcircuit with "I already have pip!"   Indeed, the spark job is different in that it explicitly downloads pip before launch_ccm_cluster.py ever runs.

This suggests a few possible workarounds.

1 - update the spark code that explicitly downloads a fairly old version of pip for python2 to use python3 and/or venvutil (not certain to work)
2 - drop the downloaded pip off the path before invoking launch_ccm_cluster
3 - have venvutil sniff for pip on the PATH and editing it out if present (seems wrong)
4 - explicitly tell pip to install the packages into the venvutil (what I'm trying here).

Also in this change I'm including a bunch of logging to try to find out if I'm wrong.

-----

Update: what i'm trying is: don't do an unnecessary step (create virtualenv, launch shakedown) in the middle of launching the cluster.

### DIFF
--- a/tools/configure_test_cluster.py
+++ b/tools/configure_test_cluster.py
@@ -105,7 +105,7 @@ class ClusterInitializer(object):
         import modify_master
         modify_master.set_local_infinity_defaults()
 
-    def apply_default_config(self):
+    def apply_default_config(self, initmaster=True):
         saved_env = os.environ.copy()
         try:
             # TODO; track a cluster-specific working dir, and keep this in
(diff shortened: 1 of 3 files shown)


jsrodman:
@nickbp 

* in order to solve a failure in cassandra (all the cluster config spew is going to standard out anyway , which makes me think something is making stdout==stderr), I added a --output flag to write the cluster info to a file, for use by shell.
* in order to avoid a failure on spark (creating the shakedown venv is failing due to a combination of factors I can't directly troubleshoot), I'm making the mesos master config step optional; and in fact all cluster config optional; I'd like to change consumers to request cluster configuration independently (users should be able to configure their own clusters from CCM direct-use after all).

Thus I brought in argparse to slightly improve arg handling.

mgummelt:
Can we just setup the virtualenv before calling into tools, rather than setting it up dynamically via python code?  That would be more standard, and I think would avoid this problem.

jsrodman:
the failure that is occuring now is a problem of venv creation, rather than a problem of venv use, so I don't agree it would solve this problem.

mgummelt:
Well the standard approach would be for client code to create the virtualenv as well:

```
virtualenv env
source env/bin/activate
./launch_ccm_cluster.py
```

But not a blocker.

PR #888, opened 2017-04-18
 [SDK] Change marathon memory limit to 1024 and manually set Xmx to 512M

mbdcos:


mbdcos:
I am hoping that our scheduler app will often stay within the Xms size; and starting with a smaller heap, GC will happen may be more often and earlier.   I agree Xmx==Xms means less GC,  allocating all heap space at the beginning. We may want to reduce container memory limit + change heap settings later.

PR #896, opened 2017-04-20
Add test for running cassandra backup and restore in soak

mrbrowning:


susanxhuynh:
@mrbrowning In the soak cluster, there are multiple instances of cassandra running. In fact, sdk-cassandra runs on non-default ports. You'll have to update the jobs/*.json cqlsh queries to use the non-default port. (https://github.com/mesosphere/soak-cluster-configurations/blob/master/common/packages/sdk-cassandra.json#L13)

PR #897, opened 2017-04-20
[INFINITY-1335] API for endpoints should emit plural `vips`, not `vip`.

paegun:
* Changed `endpoints` API resource to emit vips (plural) instead of the now deprecated vip field.
  * Fixed previously closed-to-extension shakedown test w/i kafka framework test_endpoints_address.


keithchambers:
There should not be more than 1 VIP per type of endpoint. Is there a situation where you believe this is not true?

paegun:
@keithchambers an example of multiple vips is elastic's node which has an "http" and "transport" port exposed.

keithchambers:
In this case, the entire response appears flawed. It should be something like:
``` json
{
  "http": {
    "vip": [
      "coordinator.elastic.l4lb.thisdcos.directory:9200"
    ],
    "address": [
      "10.0.2.88:1026"
    ],
    "dns": [
      "coordinator-0-node.elastic.mesos:1026"
    ]
  },
  "transport": {
    "vip": [
      "coordinator.elastic.l4lb.thisdcos.directory:9300"
    ],
    "address": [
      "10.0.2.88:1027"
    ],
    "dns": [
      "coordinator-0-node.elastic.mesos:1027"
    ]
  }
}
```

benclarkwood:
@paegun There is further work to be done before this is ready, correct? Can we close this PR in the meantime while that work is done?

paegun:
w/o the suggested change by Keith to emit names for the vip's ports, the consumer would not know which port speaks which protocol

gabrielhartmann:
@paegun:  [Consumers do know what protocols to speak to on which endpoint](https://www.elastic.co/guide/en/elasticsearch/guide/current/_talking_to_elasticsearch.html).  VIPs by design are defined against well known ports adhering to the underlying base technology's default scheme.

Additionally, introduction of field `transport` doesn't give much more of a clue, although it would make some sense to someone familiar with ElasticSearch, just as the VIPs with ports 9200 and 9300 do.

As it is today, multiple VIPs are defined per endpoint and information is being lost / overwritten, and this is a **bug**.  This fix is the minimal implementation of expected behavior for the current design.

@triclambert and @keithchambers:  The formatting change requested by @keithchambers is a feature request.  This PR addresses a bug within the context of the old design which already went through the scoping and requirements process.  @keithchambers, please submit your proposal as a feature request.

triclambert:
Agree with this bug fix, and a feature request for the format change.

nickbp:
FYI to manually retrigger a build you can post a comment of "retest this please" (please is required, lol)

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointsResource.java
```diff
@@ -211,9 +215,27 @@ private static void addPortToEndpoints(
...
-            vipEndpoint.put(RESPONSE_KEY_VIP, String.format("%s.%s.%s:%d",
-                    vipInfo.name, serviceName, ResourceUtils.VIP_HOST_TLD, vipInfo.port));
+
+            // (distinctly) append entry to 'vips' for this task:
+            String vipHostPort = String.format("%s.%s.%s:%d",
+                    vipInfo.name, serviceName, ResourceUtils.VIP_HOST_TLD, vipInfo.port);
+            Boolean foundVip = false;
+            if (vipEndpoint.has(RESPONSE_KEY_VIPS)) {
+                JSONArray vips = vipEndpoint.getJSONArray(RESPONSE_KEY_VIPS);
+                for (Object vipPresent : vips) {
```
nickbp: Nitpick: Could alternately do this and forego the typecast. Looks like JSONArray would have some additional internal checks for string types as well:
```
for (int i = 0; i < vips.length(); ++i) {
  if (vipHostPort.equals(vips.getString(i)) {
    foundVip = true;
    break;
  }
}
```

PR #898, opened 2017-04-20
[INFINITY-1370] Download cli once and only once in test.sh & test.py

jsrodman:
This is essentially a step towards cleaning up how temp dirs and the dcos cli are managed across test runs.

Here we have one working dir for the whole run, and download the cli once (though it's copied a few times).  

It's not exciting alone, but it felt like a necessary preamble to gathering diagnostics on test failure; since we have to do that in test.py  (or test.sh) to make it make sense, which means we need the dcos cli ensured available at that level (the alternative is to just increase the undesirable action pile in CI).

### DIFF
--- a/test.py
+++ b/test.py
@@ -414,7 +414,7 @@ def _multicluster_linear_per_cluster(run_attrs, repo_root):
                     start_config = launch_ccm_cluster.StartConfig(private_agents=6)
                     avail_cluster = clustinfo.start_cluster(start_config,
                             reporting_name="Cluster %s" % human_count)
-                    avail_cluster.ensure_cli_downloaded(cluster.url, get_work_dir())
+                    cli_install.ensure_cli_downloaded(avial_cluster.url, get_work_dir())
                 elif not avail_cluster:
                     # We're not supposed to start more clusters, so wait, and
                     # check for test completion.
--- a/test.py
+++ b/test.py
@@ -16,25 +16,26 @@ logger = logging.getLogger("dcos-commons-test")
 logging.basicConfig(level=logging.INFO, format="%(asctime)s %(name)s %(message)s")
 
 sys.path.append(os.path.join(get_repo_root(), 'tools'))
+import cli_install
 import clustinfo
 import fwinfo
 import launch_ccm_cluster
 
 
 work_dir = None
-def get_work_dir():
+def set_work_dir(location):
+    """Create and remember a directory for storing changing files modified by
+    the test run"""
     global work_dir
-    if not work_dir:
-        work_dir = tempfile.mkdtemp(prefix='test_workdir_', dir=get_repo_root())
-        logger.info("Using %s for test run files", work_dir)
-    return work_dir
-
-work_dir = None
-def get_work_dir():
-    global work_dir
-    if not work_dir:
+    assert not work_dir
+    assert location in ('tmp', 'repo_root')
+    if location == 'tmp':
+        work_dir = tempfile.mkdtemp(prefix='test_workdir')
+    else:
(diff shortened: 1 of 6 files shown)


jsrodman:
Totally unrelated build failure; merged master and retesting.

tools/cli_install.py
```diff
@@ -85,6 +85,10 @@ def _get_tempfilename(a_dir):
 def _mark_executable(path):
     os.chmod(path, 0o755)
 
+def install_cli_from_dir(src_dir, write_dir):
+    src_file = os.path.join(src_dir, get_cli_filename())
+    install_cli(src_file, write_dir)
+
```
benclarkwood: Single new line should be double new line.

If you're doing single new line to match the existing style, I can accept that.

tools/cli_install.py
```diff
@@ -144,21 +158,32 @@ def usage(f=sys.stderr):
...
+    source = args[1]
+    target_dir = args[2]
 
     # convenience for command line
     if not os.path.exists(target_dir):
         os.makedirs(target_dir)
 
     if source.startswith('http://') or source.startswith('https://'):
-        download_cli(source, target_dir)
+        if maybe_install:
```
benclarkwood: Correct me if I'm wrong, but download_cli will overwrite an already installed CLI, correct?

Will we actually ever want to do that?

jsrodman:
cli_install is quite new, so no real traditions there.  we could readily require 'python cli_install' for direct invocation.  I felt that the idea of "a script that fetches the right cli" had legitimate utility standalone, which is why i added the invocation interface.

tools/cli_install.py
jsrodman: I don't really think this style suggestion matters at all.  However, I'm happy to update the file to do that overall if you care.

benclarkwood: /shrug. Python style guides sure seem to care about it.

tools/cli_install.py
jsrodman: Sure, you test against 1.8 and then against 1.9, and don't care to bother to worry about unique filenames or directories.

tools/cli_install.py
jsrodman: The weird thing about python is that the development community has confused PEP8, the style guide for python core, with good python style.  So there's a bunch of stuff in there that's super sensible for all projects, and then a bunch of other stuff that's completely arbitrary. 

It's fine that a particular project has arbitrary style enforcement, but this is an example of one of the style choices that doesn't make universal sense.  For some code that's primarily well-factored libraries with clearly defined responsibility, the vertical density may not matter as much, because you consider things in an isolated way typically.  In a script it just seems pretty counterproductive to me.

test.py
```diff
@@ -404,6 +414,7 @@ def _multicluster_linear_per_cluster(run_attrs, repo_root):
                     start_config = launch_ccm_cluster.StartConfig(private_agents=6)
                     avail_cluster = clustinfo.start_cluster(start_config,
                             reporting_name="Cluster %s" % human_count)
+                    cli_install.ensure_cli_downloaded(avial_cluster.url, get_work_dir())
```
benclarkwood: Did you see this?

test.py
```diff
@@ -553,6 +565,8 @@ def run_test(framework, cluster, repo_root):
     custom_env['TEST_GITHUB_LABEL'] = framework.name
     custom_env['CLUSTER_URL'] = cluster.url
     custom_env['CLUSTER_AUTH_TOKEN'] = cluster.auth_token
+    custom_env['TESTRUN_TEMPDIR'] = get_work_dir()
+    custom_env['TESTRUN_TEMPDIR'] = get_work_dir()
```
jsrodman: Whoops.

PR #899, opened 2017-04-20
Fix handling of MOUNT volumes (allow leading slash in spec)

shirkevich:
To specify MOUNT volume for framework consumption it is needed to allow leading slash in validation.


mgummelt:
cc @nickbp

shirkevich:
Sorry I've mistaken ROOT with MOUNT volumes, actually we are using attached managed disks on azure, and they are available at /dcos/volume0 and can be consumed with following config:
```
        volume:
          path: /dcos/volume0
          type: MOUNT
```
What's wrong with mounting underlying storage outside of container? I'd be happy to add integration test, but not sure how to reproduce MOUNT volumes in test env...

gabrielhartmann:
@shirkevich:  Thank you for the PR.  We will be integrating with external storage through CSI, not this out of band solution.  Resources must be handled through Mesos.

PR #900, opened 2017-04-20
Add boto to test deps

mgummelt:
cc @jsrodman 

Spark tests require boto

### DIFF
--- a/tools/run_tests.py
+++ b/tools/run_tests.py
@@ -118,7 +118,7 @@ class CITester(object):
             requirements_file = open(requirements_txt, 'w')
             requirements_file.write('''
 requests==2.10.0
-
+boto==2.46.1
 -e git+https://github.com/dcos/shakedown.git@master#egg=shakedown
 ''')
             requirements_file.flush()


jsrodman:
Do you prefer this over providing a requirements file?  That's sort of why the requirements file interface exists, but I don't think this causes any harm.

mgummelt:
I'd prefer to provide a requirements file so I can keep Spark's requirements separate.  Is there a way to do that currently?

jsrodman:
run_tests.py takes a requirements file as an argument, which overrides the inline generated requirements file.
It's good to have an internal user too, we weren't really validating the codepath by any means, and it's pretty much intended for partners.

mgummelt:
OK, great, I'll do that.

BTW, even after I make Spark use `run_tests.py`, there's still opportunity for further consolidation with `dcos-commons`.  Ideally, I would just build the Spark package, then run some script like `test.py STUB_UNIVERSE_URL /path/to/tests`, which would start a cluster and run the tests.

The problem is that, currently, starting a cluster and building the framework are coupled together in `dcos-commons/test.sh`.  If you could factor that out, it would go a long way towards reducing duplication.

jsrodman:
I'm not sure why framework builds happened in test.sh in general.  However, dcos-commons/test.py --test-only already skips builds, but the interface for finding out the stub universe url is still a hack bound fairly tightly to the test.py build step.  In other words, it assumes a temp file will still exist somewhere with the url in it.

PR #901, opened 2017-04-20
[INFINITY-1381] Change incomplete plan status code from 503 to 202

gabrielhartmann:


jsrodman:
I'm confused.  This is changing the tests.  where's the change that caused 503s to become 202?  Off in dcos somewhere?

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PlansResource.java
```diff
@@ -57,7 +57,7 @@ public Response getPlanInfo(@PathParam("planName") String planName) {
             Plan plan = planManagerOptional.get().getPlan();
             return jsonResponseBean(
                     PlanInfo.forPlan(plan),
-                    plan.isComplete() ? Response.Status.OK : Response.Status.SERVICE_UNAVAILABLE);
```
benclarkwood: @jsrodman This is the place

testing/sdk_tasks.py
```diff
@@ -67,7 +67,7 @@ def fn():
 
 
 def check_tasks_not_updated(service_name, prefix, old_task_ids):
-    sdk_plan.get_deployment_plan(service_name)
+    sdk_plan.wait_for_completed_deployment(service_name)
```
mbdcos: @gabrielhartmann You really do not need to wait deployment to complete here. get_deploymeny_plan will wait for plan API, ensuring that scheduler started. 
PS: canary strategy will not automatically complete.

gabrielhartmann: You do need to wait for deployment to complete.  The point of this check is to validate that the Scheduler thinks no more work should be accomplished and so isn't planning on updating any tasks.  The "change" here is exactly equivalent behavior as in the past, as `get_deployment_plan` would never return until a `200` was returned, which only occurred on plan completion.

gabrielhartmann:
@mbdcos:  Thanks for the CLI catch.  Removed special handling of 503.

PR #904, opened 2017-04-20
update doc links

sascala:


sascala:
@mrbrowning I default to sending users to the OSS site since the services are open source. For Enterprise-specific information (like security) I send them to docs.mesosphere. Having written that, I can see I should default to enterprise so users aren't taken from one site to the other.

PR #906, opened 2017-04-21
Add volumes to executor

mrbrowning:
After moving to task groups, volumes defined on individual task groups will not be visible between other task groups in the same pod. This change allows volumes to be defined at the pod level so that they are attached to the `ExecutorInfo` on launch, which is the mandated method for sharing volumes between task groups in the same pod.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
@@ -187,15 +187,15 @@ public class DefaultOfferRequirementProvider implements OfferRequirementProvider
             }
         }
 
-        Map<String, Protos.Resource> volumeMap = new HashMap<>();
         if (executorInfo == null) {
             executorInfo = getNewExecutorInfo(
                     podInstance.getPod(), serviceName, targetConfigurationId, schedulerFlags);
-        } else {
-            volumeMap.putAll(executorInfo.getResourcesList().stream()
-                    .filter(r -> r.hasDisk() && r.getDisk().hasVolume())
-                    .collect(Collectors.toMap(r -> r.getDisk().getVolume().getContainerPath(), Function.identity())));
         }
+
+        Map<String, Protos.Resource> volumeMap = new HashMap<>();
+        volumeMap.putAll(executorInfo.getResourcesList().stream()
+                .filter(r -> r.hasDisk() && r.getDisk().hasVolume())
+                .collect(Collectors.toMap(r -> r.getDisk().getVolume().getContainerPath(), Function.identity())));
 
         List<ResourceRequirement> resourceRequirements = new ArrayList<>();
         for (VolumeSpec v : podInstance.getPod().getVolumes()) {
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
@@ -173,22 +173,38 @@ public class DefaultOfferRequirementProvider implements OfferRequirementProvider
             String serviceName,
             UUID targetConfigurationId) throws InvalidRequirementException {
         List<Protos.TaskInfo> podTasks = TaskUtils.getPodTasks(podInstance, stateStore);
+        Protos.ExecutorInfo executorInfo = null;
 
         for (Protos.TaskInfo taskInfo : podTasks) {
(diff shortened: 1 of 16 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ExecutorRequirement.java
```diff
@@ -55,20 +63,23 @@ private static ExecutorRequirement getExistingExecutorRequirement(ExecutorInfo e
...
     }
 
-    private ExecutorRequirement(ExecutorInfo executorInfo)
-            throws InvalidRequirementException {
+    private ExecutorRequirement(
+            ExecutorInfo executorInfo,
+            Collection<ResourceRequirement> resourceRequirements) throws InvalidRequirementException {
         validateExecutorInfo(executorInfo);
         this.executorInfo = executorInfo;
+        this.resourceRequirements = resourceRequirements != null ? resourceRequirements :
```
gabrielhartmann: This seems a little strange.  So sometimes you're supposed to pass in all `ResourceRequirement`s.  Other times we derive them from the `ExecutorInfo`.  This seems to be a side effect of passing in volume related `ResourceRequirement`s sometimes.  It's a little funny that when we pass in requirements its always ONLY volume requirements.  Why don't we do one or th other.  Either all the `ResourceRequirement`s are derived from the `ExecutorInfo`, or none are.  If none are, why pass in an `ExecutorInfo`? Why couldn't the ExecutorInfo have all the appropriate volume information on it already and have it extracted here?

gabrielhartmann:
@mrbrowning:  Can we also add an integration test or two.  One sharing a volume between two tasks launched together, one with a sidecar consuming a shared volume, used by some other main task (a proxy for C*).  Have you tried C*'s sidecar plans with this approach?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ExecutorRequirement.java
mrbrowning: It's not really possible to move it over to exclusively instantiating resource requirements from the resources themselves -- since subtypes of `ResourceRequirement` (e.g. `VolumeRequirement`) are polymorphic in the `ResourceEvaluationStage` they yield, we need to keep a handle to whichever `ResourceRequirement` we got from the specification. The ability to not pass in resource requirements explicitly was useful in the case of an existing offer requirement, which seemed safe since disk resources can't change after creation (so we don't need the volume-create logic in `VolumeEvaluationStage` because we intend to just claim the volume that we already reserved as-is), but of course I realize now that a user might want to add a new volume entirely to the pod. So probably the best bet is to always pass in the `ResourceRequirement` collection.

mrbrowning:

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java
```diff
@@ -173,22 +173,38 @@ private ExecutorRequirement getExecutorRequirement(
...
                         "Reusing executor from task '{}': {}",
                         taskInfo.getName(),
-                        TextFormat.shortDebugString(taskInfo.getExecutor()));
-                return ExecutorRequirement.create(taskInfo.getExecutor());
+                        TextFormat.shortDebugString(executorInfo));
             }
         }
 
+        Map<String, Protos.Resource> volumeMap = new HashMap<>();
+        if (executorInfo == null) {
```
mrbrowning: The logic here is basically the same as in [DefaultOfferRequirementProvider#149](https://github.com/mesosphere/dcos-commons/blob/824ee8146ecf56fbf50fb364c2268a3affa0617e/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/DefaultOfferRequirementProvider.java#L149): to do the right bookkeeping for resources on an "existing" requirement, we need to pass the existing `Resource` protobuf around so that the resource ID ends up on the `TaskInfo` that we launch. I suppose, thinking about this a little more, that that's unnecessary -- executor resources can't change after launch (and offers aren't evaluated for executor resources in the case that we're placing a task on an already-running executor), so we may as well not even bother.

mrbrowning: I do think that `ExecutorInfo` should continue to be passed to `ExecutorRequirement` -- `DefaultOfferRequirementProvider` has to know a lot of things to create an `ExecutorInfo`, like the pod name, how to construct the `ContainerInfo`, the pod user, URIs...right now I don't think it makes sense to pass all of that stuff to an `ExecutorRequirement` just so it can know how to create the protobuf. All of the data dependencies here could probably stand some refactoring, but it's a larger process than just moving where a proto is instantiated.

PR #907, opened 2017-04-21
[INFINITY-1085] (also INFINITY-1088) junit failures on failure

jsrodman:
Reintroduce work from a long while ago, this time without any vendoring.
Explicitly itemize things like: cluster bringup, framework build failure, etc.


jsrodman:
Don't read the full changelog; everything up to 9276356 is already merged.

test.py
```diff
@@ -563,16 +563,65 @@ def run_test(framework, cluster, repo_root):
         logger.info(msg, runtests_script, framework.name)
         raise CommandFailure(cmd_args)
 
+def setup_junit_xml():
+    """Try to download and import junit_xml"""
```
benclarkwood: Can we just add this to the infinity container ahead of time? Or is it having to suck this into the venv? or?

test.py
```diff
@@ -563,16 +563,65 @@ def run_test(framework, cluster, repo_root):
...
+    import junit_xml
+
+def emit_junit_xml(cluster_launch_attempts, frameworks):
+    """Write out all the test actions failures to a junit file for jenkins or
+    similar"""
+    import junit_xml
+    launch_fake_testcases = []
+    for launch_attempt in cluster_launch_attempts:
+        attempt_duration = launch_attempt.end_time - launch_attempt.start_time
+        fake_test = junit_xml.TestCase(launch_attempt.name,
```
benclarkwood: What... is a fake test?

benclarkwood: Is the goal to have mock results for _all_ tests, even those that weren't able to run due to a failure?

test.py
```diff
@@ -563,16 +563,65 @@ def run_test(framework, cluster, repo_root):
         logger.info(msg, runtests_script, framework.name)
         raise CommandFailure(cmd_args)
 
+def setup_junit_xml():
+    """Try to download and import junit_xml"""
+    workdir = get_work_dir()
+    lib_dir = os.path.join(workdir, 'lib', 'python')
+    pip_cmd = ['pip', 'install', 'junit-xml', '-t', lib_dir]
```
mrbrowning: It's slightly confusing that this script modifies its own execution environment in a subprocess and then imports on the basis of that change. Could we group all this environment mutation into one script and the actual test logic (`emit_junit_xml`) in another?

test.py
```diff
@@ -563,16 +563,65 @@ def run_test(framework, cluster, repo_root):
...
+                framework_testcases)
+        fake_suites.append(framework_suite)
+
+    with open("junit_testpy.xml", "w") as f:
+        junit_xml.TestSuite.to_file(f, fake_suites)
+
+
 def report_failed_actions():
-    "Do useful things with the recorded successful and failed actions"
+    """Do useful things with the recorded successful and failed actions"""
```
mrbrowning: https://www.python.org/dev/peps/pep-0257/#id16

test.py
jsrodman: It's possible to make this a container dependency, but I don't know how to update the container, and we don't have the ability to quickly update the container.  In short, I don't believe that's likely to be a better outcome.

test.py
jsrodman: what i meant is, it's not a test, but junit_xml only can describe tests, so I'm pretending "create a cluster" is a test.

test.py
jsrodman: mega?  I wouldn't do this either but browning seemed to care.

test.py
jsrodman: Are you objecting to the pattern of a process downloading and loading a module itself, or just looking for different module organization?

I have no problem with moving this single function into another module, but there wont' be any meaningful encapsulation or interface, partly because junit_xml is pretty awful and doesn't offer interfaces for sensibly building out its state iteratively.   (Also solvable but there's some definite diminishing returns here.) 

If it's a goal always using subprocesses for loading modules dynamically, that's definitely not something I want.  I want exception outputs that give clear traces of failures.  I have many years of experience of being able to produce fixes for very specific edge-case problem from a single error log by following good-stewardship patterns allowing quality exception output.  We are currently not in that position, and many failures are question marks for precisely this reason.

mrbrowning: > Are you objecting to the pattern of a process downloading and loading a module itself, or just looking for different module organization?

Kind of, yeah. I meant this in the most simple sense possible: setting up the environment (here, installing Python modules) is an orthogonal concern to something like emitting junit.xml output. Keeping these two concerns grouped together in different files/modules makes them easier to find and modify.

test.py
benclarkwood: "Triple quotes are used even though the string fits on one line. This makes it easy to later expand it."

Reasonable, I suppose.

I use mega as my qualifier for: "Don't change this, but I'd like to know why"

test.py
benclarkwood: 👍

test.py
jsrodman: I'm still not clear. I said A or B and you said "yes".  Do you mean A and B?

It would be nice to have all the module downloading in one place.  However,  pip download is a common enough case of test failure, that I'd like to use junit to report it, which is why moved this to the very start of the run.

It could still make sense to do all the module fetching in one module with two different cases, but that would require further surgery to run_tests.py to readily achieve.  That said, I could readily introduce a utility module used both here and in configure_test_cluster.py and make run_tests.py use it later.  

If that covers your concerns then I'll do it.

benclarkwood:
@mrbrowning @jsrodman: I want these changes, do we want to do the described refactor? Or is this monkeypatch okay?

test.py
```diff
@@ -582,6 +631,12 @@ def main():
         logger.error("Aborting run.")
         return False
 
+    try:
+        setup_junit_xml()
+    except Exception as e:
+        logging.exception("Failed to load junit_xml; no templated success/fail"
```
nickbp: Able to log embedded exception?

test.py
mrbrowning: That sounds reasonable.

jsrodman:
@mrbrowning @benclarkwood : The existing code isn't really a monkeypatch, it's just inline in a file browning feels should have a very small responsibility.  That said, I'll move the code here in this pull.

test.py
jsrodman: I don't understand the question.
logging.exception() is special magic that takes an informative string and records the string and the current exception into the logger channel.  I'd have preferred `logger.exception(e, msg)`, but I didn't write it.

nickbp: OK cool, didn't know it was included already.

test.py
jsrodman: On looking to implement this; it's messy.  venvutil knows about virtualenvs which I've always found overcomplicated and now discovering are more fragile than I realized; so one of my goals is to only have one for most runs, and none for this failsafe purpose.

Having a module that knows about virtualenv specific things for downloading seems weird; venvutil already covers that.  So I just put in a lame convenience method to module_downloader that calls through.

jsrodman:
retest this please

PR #913, opened 2017-04-24
[INFINITY-1396] Remove unused MIGRATION labels

gabrielhartmann:


jasongilanfarr:
I couldn't find evidence that marathon actually reads these endpoints. It probably should... @aquamatthias ?

aquamatthias:
@jasongilanfarr the readiness check defines the contract (incl. path etc). Marathon still has checks for the tags: https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/state/AppDefinition.scala#L484

jasongilanfarr:
I know that we have the validation checks, but i didn't see us actually _use_ the fields.

aquamatthias:
It is used by the UI to query the plan and do the proceed operations.

gabrielhartmann:
The DC/OS UI team doesn't use this.

Issue #916, opened 2017-04-25
Support for external/docker volumes in pods


disrani-px:
The currently supported volumes types (ROOT and MOUNT) don't allow for external volumes to be used for services.

External volumes can be supported by allowing docker volumes to be used in the pods.

The proposal is to allow a DOCKER volume type to be specified along with docker_volume_driver, docker_volume_name and docker_driver_options in the service yaml for the pod specification.

The changes for this proposal can be viewed here:
https://github.com/portworx/dcos-commons/commit/1f1fbbc4c1cc85ce1a4b7efab009ef1d7bc47747

This has been tested with the Cassandra and HDFS service.
Example of changes required in Cassandra service to support this:
https://github.com/portworx/dcos-commons/commit/7f5432b2f00ba30c2d0c527f02a795283b988fe6

gabrielhartmann:
@disrani-px:  These changes look pretty good to me in general.  Out of curiosity why the extensive effort to support upgrade from the old [`dcos-kafka-service`](https://github.com/mesosphere/dcos-kafka-service/) to the new `kafka` implementation?

disrani-px:
@gabrielhartmann: Since these services are in beta users might still choose to use the old service (with changes to run on Docker volumes) and upgrade to this later.
The Cassandra service uses a Docker image, so that should be running UCR.
Your comments actually reflected on the PR I had sent out: https://github.com/mesosphere/dcos-commons/pull/918

keithchambers:
Plan is to support CSI once available in DC/OS.

PR #918, opened 2017-04-25
Framework changes to be able to use DOCKER volumes in pods

disrani-px:
This is related to https://github.com/mesosphere/dcos-commons/issues/916

Changes also include updates for HDFS and Cassandra service to add DOCKER volume options. Default is still set to ROOT


gabrielhartmann:
Do the use of docker volumes require use of the Docker engine?

disrani-px:
It would use dvdi for the docker volumes, not sure if that has a dependency on Docker engine.

disrani-px:
Updated with feedback from the review and added some tests for the DockerVolumeSpec.

keithchambers:
Hi Dinesh.

Thank you very much for contributing this PR. We are focusing on CSI in partnership
with the boarder container ecosystem: https://mesosphere.com/blog/2017/03/30/csi-towards-universal-storage-interface-for-containers. We are confident CSI as the path to supporting external volumes in this project. Is the Pirtworx team working on CSI?

Thanks,
Keith

gourao:
Hi @keithchambers Yes, we are a coe part of the CSI effort, working with Ben and others on the core specs.  This PR is sort of unrelated to that in that, it only solves a current need for the DCOS ecosystem.  It is not incompatible with CSI, and to clarify that, when CSI gets ironed out and implemented, existing frameworks can and will migrate to that without impacting existing deployments.  

This PR is simply to allow existing DCOS users to deploy stateful applications while the internal plumbing gets sorted into a CNCF endorsed standard.

Hope that makes sense

gourao:
@gabrielhartmann @disrani-px DVDI has no requirement of the Docker engine.

keithchambers:
I understand the reasoning, but DVDI is not something we will support in our SDK. We will support CSI once available.

gourao:
@keithchambers I understand your stance on closing the PR... we do have joint customers that need to deploy applications like cassandra, hdfs in production, today, with external volumes.  Would like to hear your recommendation for that.

PR #921, opened 2017-04-26
[INFINITY-1406] Change modify_master to act as a context manager, to get write-on-descope

jsrodman:
@benclarkwood I haven't yet done the bugfixing round, but I'm showing this now so you can critique if desired.

### DIFF
--- a/tools/modify_master.py
+++ b/tools/modify_master.py
@@ -1,5 +1,20 @@
 #!/usr/bin/env python3
 
+""" Interface to reconfigure files on a mesos master as referenced by the
+current dcos command line.
+
+Usage:
+
+with MesosMasterModifier() as mmm:
+    mmm.set_master_envvar('ABC')
+    mmm.remove_master_envvar('XYZ')
+
+The settings take effect at the conclusion of the block, when mmm leaves
+scope.
+"""
+
+
+
 # shakedown requires python3
 
 import logging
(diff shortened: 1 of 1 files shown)


benclarkwood:
Am cool with this approach.

benclarkwood:
I am fine with how this works. Deferring to @mrbrowning for any Python quibbles.

tools/modify_master.py
```diff
@@ -13,98 +28,171 @@
...
+            raise RuntimeError("Unable to get current envvars")
+
+        logger.info("Current envvars:\n{}".format(out))
+        envvars, commented_lines = self._decode_master_config(out)
+        self.master_envvars = envvars
+        self.commented_lines = commented_lines
+
+    def _modify_envvars(self, modifier_f):
+        """Performs a transformation on the mesos master environment
+        variables.  Accepts a funciton which accepts the mesos environment
```
mrbrowning: Typo, `funciton`

tools/modify_master.py
```diff
@@ -13,98 +28,171 @@
...
+        before the transformation, and returns the new environment.  The
+        function should accept and return dicts of str:str
+
+        :param modifier_f: The function to apply the transformation to the envvar state
+        :type modifier_f: func
+        """
```
mrbrowning: What's the value of this pseudo-immutable treatment of the `master_envvars` member? Couldn't we just modify the dict in place?

jsrodman:
Currently the intent was to change some settings on the master, not to replace the settings entirely.  If we wanted to wipe what was there, accepting a default dict might work, but I'd probably want the consumer to explicitly document that the prior state is being wiped with an explicit action.

tools/modify_master.py
jsrodman: We could modify the dict in place, but it felt gross to me to have the functor dustructively modifying an object it doesn't really know the lifetime of.
I could move the protection against this to _modify_envvars(), or i could remove, but this felt more right to me than a destructive change that could break on refactor in weird ways.

PR #925, opened 2017-04-26
[INFINITY-1410] decomposed venv construction needs setup_scripts.

jsrodman:
handle args correctly in cli_install


jsrodman:
Pre-emptive merge to remove regressions, since test.sh will start breaking for people, even if not in CI

PR #926, opened 2017-04-26
[INFINITY-1319] Replace pods immediately.

benclarkwood:
Right now, we wait for Mesos to kill all of a pod's tasks before we revive and try to replace it.

Instead, immediately "kill" the task as far as the scheduler is concerned (via TaskStatusWriter), and then tell Mesos to kill it as well. 

This insulates us from agent failure delaying replaces (and agent failure is one of the few times you'll really want to do replace).


benclarkwood:
retest this please

benclarkwood:
Looks like this may have caused a regression in HDFS recovery. Investigating.

benclarkwood:
retest this please

PR #935, opened 2017-04-29
Use repeated -p parameter for specifying plan envvars

mrbrowning:
Manually verified that this works as expected with single and multi-parameter plans.

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -31,15 +31,15 @@ func GetArguments() []string {
 	return os.Args[2:]
 }
 
-func GetPlanParameterPayload(parameters string) (string, error) {
+func GetPlanParameterPayload(parameters []string) (string, error) {
 	envPairs := make(map[string]string)
-	for _, pair := range strings.Split(parameters, ",") {
+	for _, pair := range parameters {
 		elements := strings.Split(pair, "=")
-		if len(elements) != 2 {
+		if len(elements) < 2 {
 			return "", errors.New(fmt.Sprintf(
 				"Must have one variable name and one variable value per definition"))
 		}
-		envPairs[elements[0]] = elements[1]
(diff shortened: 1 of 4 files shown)


mrbrowning:
@benclarkwood I think supplying these optionally as JSON is a great idea too, but this interface will continue to be useful for simpler plans, and I needed it quickly because of the existing way's stupid handling of values that have `"="` in them, as many base64-encoded values do.

mrbrowning:

gabrielhartmann:
@mrbrowning I agree that this change is valuable on its own.  It will be nice to have a JSON option in the future, but this need not be blocked by that.  Do you have a JIRA for that work?

mrbrowning:
I agree that some tests of this particular function are useful, but in general I think we should be judicious in spending time on a unit test suite for the CLI package. It's mostly stable, it's very simple, and although (as a user-facing component) it not breaking is a high priority, a CLI bug is not show-stopping given the availability of the HTTP API. These are all factors that diminish the likely value of a rigorous test suite, so I don't think anyone need spend hours chasing coverage here.

mrbrowning:
@jsrodman do you think it would be appropriate for CLI unit tests to be run in `build.sh`? Currently we only build the CLI package on a per-framework basis, so I'm not sure if the assumptions about the environment `build.sh` is running in will check out.

jsrodman:
@mrbrowning Sounds good to me, and all builds are running on the 'infinity' image, so it shouidln't break.

mrbrowning:

benclarkwood:
@mrbrowning Agreed that the CLI need not be exhaustively tested. I think a good rule of thumb is tests around user input to ensure good UX. WDYT?

Thank you for setting this up!

PR #938, opened 2017-04-30
Added uninstall functionality

loren:
- Change `DefaultService` to instantiate new `UninstallScheduler` when
  `SDK_UNINSTALL` flag is set.
- Added UninstallScheduler
- Added UninstallResourceCleaner to destroy/unreserve all resources
- Fixed some Javadoc @link issues
- Use `Scheduler` instead of a `DefaultScheduler.Builder` in `DefaultService`
- Convert some code to use Java 8 idioms around lambdas and `Optional`


gabrielhartmann:
@loren:  I know this is WIP.  Something that's important to me in the final version is a good view of the progress of work in the `v1/plans/deploy` endpoint, so we can very easily see where things get stuck in the unlikely event we are not receiving offers for something we need to `UNRESERVE` or `DESTROY`.

loren:
Thank you for the feedback @gabrielhartmann . I will definitely expose some sort of plan via `/deploy`. This first pass was just to see if I was getting the resource cleaning right, and to make sure I didn't totally break every other framework with my changes to `DefaultService`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceCleaner.java
```diff
@@ -1,44 +1,25 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
 /**
  * The Resource Cleaner provides recommended operations for cleaning up
  * unexpected Reserved resources and persistent volumes.
  */
-public class ResourceCleaner {
+public abstract class ResourceCleaner {
```
nickbp: Given how little code is left here, it feels like ResourceCleaner could just be an interface to be implemented (independently) by DefaultResourceCleaner / UninstallResourceCleaner. The remaining piece in `evaluate()`, below, could be a separate static function to be called by DefaultResourceCleaner / UninstallResourceCleaner. I'm leaning towards this structure as it'll make the two a little more isolated and reduce the chance of unintentional changes from one side affecting the other, as ideally they don't really have much to do with each other.

TLDR:
- ResourceCleaner could be an interface
- DefaultResourceCleaner and UninstallResourceCleaner could implement that interface (but be independent otherwise)
- The common `evaluate()` code currently in ResourceCleaner could be moved into a static method to be invoked by DefaultResourceCleaner/UninstallResourceCleaner.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -237,9 +228,7 @@ private static void registerAndRunFramework(
 
         // The framework ID is not available when we're being started for the first time.
         Optional<Protos.FrameworkID> optionalFrameworkId = stateStore.fetchFrameworkId();
-        if (optionalFrameworkId.isPresent()) {
-            fwkInfoBuilder.setId(optionalFrameworkId.get());
-        }
+        optionalFrameworkId.ifPresent(fwkInfoBuilder::setId);
```
nickbp: In the uninstall case, we should behave differently when optionalFrameworkId is missing: If it's missing then it implies ZK is missing, likely due to a previously completed uninstall.

In this scenario, the scheduler would ideally start in a no-op mode where it's not doing anything other than advertising a completed plan state.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/UninstallScheduler.java
```diff
@@ -0,0 +1,446 @@
...
+        LOGGER.info("Initializing...");
+        // NOTE: We wait until this point to perform any work using configStore/stateStore.
+        // We specifically avoid writing any data to ZK before registered() has been called.
+        initializeGlobals(driver);
+//        initializeApiServer();
+        LOGGER.info("Done initializing.");
+    }
+
+
+    private void initializeGlobals(SchedulerDriver driver) {
```
nickbp: I'd expect many of these to go away or be replaced once this is finished

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/UninstallScheduler.java
```diff
@@ -0,0 +1,446 @@
...
+        LOGGER.info("Initializing globals...");
+        taskKiller = new DefaultTaskKiller(taskFailureListener, driver);
+        reconciler = new DefaultReconciler(stateStore);
+        offerAccepter = new OfferAccepter(Arrays.asList(new PersistentLaunchRecorder(stateStore, serviceSpec)));
+        OfferEvaluator offerEvaluator = new OfferEvaluator(stateStore, offerRequirementProvider);
+        planScheduler = new DefaultPlanScheduler(offerAccepter, offerEvaluator, stateStore, taskKiller);
+    }
+
+
+    private void initializeResources() throws InterruptedException {
```
nickbp: same

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/DeregisterStep.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param schedulerDriver Must be set before call to {@link #start()}
+     */
+    public void setSchedulerDriver(SchedulerDriver schedulerDriver) {
+        this.schedulerDriver = schedulerDriver;
+    }
+
+    @Override
+    public Optional<PodInstanceRequirement> start() {
+        logger.info("Stopping SchedulerDriver...");
+        // Unregisters the framework in addition to stopping the SchedulerDriver thread:
```
nickbp: remove the framework id before the unregister

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/DeregisterStep.java
```diff
@@ -0,0 +1,71 @@
...
+        schedulerDriver.stop(true);
+        StateStoreUtils.setUninstalling(stateStore, false);
```
nickbp: skip setting the value false -- should be done when ZK is wiped

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreUtils.java
```diff
@@ -210,35 +211,45 @@ public static void validateValue(byte[] value) throws StateStoreException {
...
+     * Returns the current value of the 'uninstall' property in the provided {@link StateStore}.
+     */
+    public static boolean isUninstalling(StateStore stateStore) throws StateStoreException {
+        return fetchBooleanProperty(stateStore, UNINSTALLING_PROPERTY_KEY);
+    }
 
+    /**
+     * Sets an 'uninstall' property in the provided {@link StateStore} to the provided value.
+     */
+    public static void setUninstalling(StateStore stateStore, boolean isUninstalling) {
```
nickbp: with the `false` case removed above, the boolean param here could be removed as well (always `true`). don't forget to update the javadoc

nickbp:
Feel free to address above final wrap-up behavior in a separate PR -- assuming things are green it'd be good to merge and iterate IMO

loren:
Thanks @nick for reviewing. I agree that the wrap-up behavior would best be addressed in a separate JIRA/PR. I will create that now.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/UninstallRecorder.java
```diff
@@ -0,0 +1,97 @@
+package com.mesosphere.sdk.state;
```
nickbp: This could go into the `uninstall` package, looks like it's only used there

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/UninstallResourceCleaner.java
```diff
@@ -0,0 +1,48 @@
+package com.mesosphere.sdk.offer;
```
nickbp: This could be in the `uninstall` package as it looks like it's only used there.

loren: Good idea. OK if I move both of these over as part of the separate wrap-up work? I hate to waste a green CI.

nickbp: sure :+1:

PR #941, opened 2017-05-01
[INFINITY-1439] Disable modify_master in strict mode.

jsrodman:
modify_master is failing in test runs.  Disable it until we can solve why it
is not working and/or reimplement via a mechanism that doesn't break.


jsrodman:
For what it's worth, this change is not required by any automated tests runs.  I've temporarily changed the framework-test-master nightly to explicitly ask for this code to not run, but we'd like most of the tests to work in strict mode, so this is still needed just not urgent.

PR #947, opened 2017-05-02
Update tooling to use drop-in stub universe json format

nickbp:
Code for building stub universes is switched entirely to the json format. The release builder, meanwhile, still supports both old zips and new jsons for now. The json file now being produced is identical in usage to the prior zip file (`dcos package repo add http://arrl.org/foo.json`, etc...)

Other notable changes:
- Fixes a nasty bug in `release_builder` which would allow always overwriting AWS data (bool('false') => True)
- Adds support in `publish_aws` and `publish_http` for the bizarre universe content type that Cosmos requires for these json files.
- Breaks out the universe version in `universe_builder` to allow easier switching to v4 later.
- Misc formatting fixes that pylint complained about

### DIFF
--- a/tools/publish_aws.py
+++ b/tools/publish_aws.py
@@ -90,14 +90,15 @@ class AWSPublisher(object):
             self._artifact_paths.append(artifact_path)
 
 
-    def _upload_artifact(self, filepath):
+    def _upload_artifact(self, filepath, content_type=None):
         filename = os.path.basename(filepath)
+        cmd = 'aws s3'
         if self._aws_region:
-            cmd = 'aws s3 --region={} cp --acl public-read {} {}/{} 1>&2'.format(
-                self._aws_region, filepath, self._s3_directory, filename)
-        else:
-            cmd = 'aws s3 cp --acl public-read {} {}/{} 1>&2'.format(
-                filepath, self._s3_directory, filename)
+            cmd += ' --region={}'.format(self._aws_region)
+        cmd += ' cp --acl public-read'
+        if content_type:
+            cmd += ' --content-type "{}"'.format(content_type)
+        cmd += ' {} {}/{} 1>&2'.format(filepath, self._s3_directory, filename)
         if self._dry_run:
             logger.info('[DRY RUN] {}'.format(cmd))
             ret = 0
@@ -140,17 +141,18 @@ class AWSPublisher(object):
 
     def upload(self):
         '''generates a unique directory, then uploads artifacts and a new stub universe to that directory'''
+        builder = universe_builder.UniversePackageBuilder(
+            self._pkg_name, self._pkg_version,
+            self._input_dir_path, self._http_directory, self._artifact_paths)
         try:
-            universe_path = universe_builder.UniversePackageBuilder(
-                self._pkg_name, self._pkg_version,
-                self._input_dir_path, self._http_directory, self._artifact_paths).build_zip()
+            universe_path = builder.build_package()
         except Exception as e:
             err = 'Failed to create stub universe: {}'.format(str(e))
             self._github_updater.update('error', err)
             raise
 
         # print universe url early
-        universe_url = self._upload_artifact(universe_path)
+        universe_url = self._upload_artifact(universe_path, content_type=builder.content_type())
         logger.info('---')
         logger.info('Uploading {} artifacts:'.format(len(self._artifact_paths)))
 
--- a/tools/publish_aws.py
+++ b/tools/publish_aws.py
@@ -92,13 +92,14 @@ class AWSPublisher(object):
 
     def _upload_artifact(self, filepath, content_type=None):
         filename = os.path.basename(filepath)
-        cmd = 'aws s3'
+        cmdlist = ['aws s3']
         if self._aws_region:
-            cmd += ' --region={}'.format(self._aws_region)
-        cmd += ' cp --acl public-read'
+            cmdlist.append('--region={}'.format(self._aws_region))
+        cmdlist.append('cp --acl public-read')
         if content_type:
-            cmd += ' --content-type "{}"'.format(content_type)
-        cmd += ' {} {}/{} 1>&2'.format(filepath, self._s3_directory, filename)
+            cmdlist.append('--content-type "{}"'.format(content_type))
+        cmdlist.append('{} {}/{} 1>&2'.format(filepath, self._s3_directory, filename))
+        cmd = ' '.join(cmdlist)
         if self._dry_run:
             logger.info('[DRY RUN] {}'.format(cmd))
             ret = 0
--- a/tools/release_builder.py
+++ b/tools/release_builder.py
@@ -2,7 +2,6 @@
 
 import base64
 import difflib
-import http.client
 import json
 import logging
 import os
@@ -15,6 +14,7 @@ import tempfile
 import urllib.request
 import zipfile
 from collections import OrderedDict
+import http.client
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


nickbp:
Looks like there's a bug in the CLI's handling of binary CLI hashes. Debugging... In the meantime feel free to review.

```
E               dcos.errors.DCOSException: The hash for the downloaded subcommand [d9fa98944bfa1ded2b0b401d2365c25db94e14e94df3a1aeb9273f9b5c77c352] does not match the expected value [0131180f265dc6e206c67ec2344a616a57d95401ba09de8b1df9df386b974c7e]. Aborting...
```

(spoiler: when downloaded manually, the executable matches the expected hash)

tools/release_builder.py
```diff
@@ -74,33 +74,81 @@ def _run_cmd(self, cmd, exit_on_fail=True, dry_run_return = 0):
...
-        if not os.path.isdir(pkgdir_path):
-            raise Exception('Didn\'t find expected path {} after unzipping {}'.format(
-                pkgdir_path, local_zip_path))
-        os.unlink(local_zip_path)
-        return pkgdir_path
+        '''Returns the path to the package directory in the stub universe.'''
+        stub_universe_file = urllib.request.urlopen(self._stub_universe_url)
+
+        stub_universe_extension = os.path.splitext(self._stub_universe_url)[1]
+        if stub_universe_extension == '.zip':
```
benclarkwood: Nitpick: I'd rather see these cases split into functions. Don't have a strong argument for it other than stylistic.

tools/publish_aws.py
```diff
@@ -90,14 +90,15 @@ def __init__(
...
-            cmd = 'aws s3 --region={} cp --acl public-read {} {}/{} 1>&2'.format(
-                self._aws_region, filepath, self._s3_directory, filename)
-        else:
-            cmd = 'aws s3 cp --acl public-read {} {}/{} 1>&2'.format(
-                filepath, self._s3_directory, filename)
+            cmd += ' --region={}'.format(self._aws_region)
+        cmd += ' cp --acl public-read'
+        if content_type:
+            cmd += ' --content-type "{}"'.format(content_type)
+        cmd += ' {} {}/{} 1>&2'.format(filepath, self._s3_directory, filename)
```
jsrodman: Not really specific to this code, but I strongly prefer command-as-vector over command-as-big-string  Programs that break when a dir introduces a space is so 1980.

tools/release_builder.py
```diff
@@ -2,7 +2,6 @@
 
 import base64
 import difflib
-import http.client
```
jsrodman: Why de-sort this import?

nickbp:
retest this please

tools/release_builder.py
```diff
@@ -74,33 +74,81 @@ def _run_cmd(self, cmd, exit_on_fail=True, dry_run_return = 0):
...
-            '0')
-        if not os.path.isdir(pkgdir_path):
-            raise Exception('Didn\'t find expected path {} after unzipping {}'.format(
-                pkgdir_path, local_zip_path))
-        os.unlink(local_zip_path)
-        return pkgdir_path
+        '''Returns the path to the package directory in the stub universe.'''
+        stub_universe_file = urllib.request.urlopen(self._stub_universe_url)
+
+        stub_universe_extension = os.path.splitext(self._stub_universe_url)[1]
```
jsrodman: I suggest `_, stub_unverse_extention = os.path.splitext(self._stub_universe_url)`
tuple unpack is generally clearer than tuple indexing.

tools/release_builder.py
```diff
@@ -74,33 +74,81 @@ def _run_cmd(self, cmd, exit_on_fail=True, dry_run_return = 0):
...
+                raise Exception('Expected "packages" key in root of stub universe JSON: {}'.format(
+                    self._stub_universe_url))
+            if len(stub_universe_json['packages']) != 1:
+                raise Exception('Expected single "packages" entry in stub universe JSON: {}'.format(
+                    self._stub_universe_url))
+
+            package_json = stub_universe_json['packages'][0]
+
+            def extract_json_file(package_dict, name):
+                file_dict = package_dict.get(name, None)
```
jsrodman: Odd;  `dict.get(key)` is the same as `dict.get(key, None)`

I'm not sure yet whey you're processing package_dict destructively, but I'd probably have just written

```python
if name in package_dict:
    file_dict = package_dict.pop(name)
    ....
```

tools/release_builder.py
```diff
@@ -74,33 +74,81 @@ def _run_cmd(self, cmd, exit_on_fail=True, dry_run_return = 0):
...
+                del package_json['marathon']
+                marathon_file = open(os.path.join(pkgdir, 'marathon.json.mustache'), 'w')
+                marathon_file.write(base64.standard_b64decode(marathon_json).decode())
+                marathon_file.flush()
+                marathon_file.close()
+
+            if 'releaseVersion' in package_json:
+                del package_json['releaseVersion']
+
+            json.dump(package_json, open(os.path.join(pkgdir, 'package.json'), 'w'), indent=2)
```
jsrodman: And now the destructive handling becomes clear.  Maybe worth a brief note at the top?

tools/universe_builder.py
```diff
@@ -1,102 +1,71 @@
...
 import logging
 import os
 import os.path
 import re
-import shutil
 import sys
 import tempfile
-import time
-import zipfile
+from collections import OrderedDict
```
jsrodman: opinion: My preference is avoiding `from` unless it's truly onerous.  Iteratively it costs very little, but when you find yourself debugging a scenario where there are multiple copies of a module in memory, you become very sad.

tools/universe_builder.py
```diff
@@ -120,103 +89,96 @@ def _get_file_template_mapping(self, filepath):
...
-                    'Please provide the full path to this artifact (known artifacts: {})'.format(self._artifact_files))
+                    'Please provide the full path to this artifact (known artifacts: {})'.format(self._artifact_file_paths))
             template_mapping['sha256:{}'.format(shafilename)] = self._calculate_sha256(shafilepath)
 
         # import any custom TEMPLATE_SOME_PARAM environment variables:
         for env_key, env_val in os.environ.items():
             if env_key.startswith('TEMPLATE_'):
                 # 'TEMPLATE_SOME_KEY' => 'some-key'
-                template_mapping[env_key[9:].lower().replace('_','-')] = env_val
+                template_mapping[env_key[9:].lower().replace('_', '-')] = env_val
```
jsrodman: 9?  len('TEMPLATE_')?
Maybe not worth cleaning up here, but this line is doing way too much for my taste.

tools/release_builder.py
```diff
@@ -15,6 +14,7 @@
 import urllib.request
 import zipfile
 from collections import OrderedDict
+import http.client
```
jsrodman: There's an import sort tool for python isort which is reasonable if you're actually developing in python.  I'm hesitant to spend time on that level of tooling for our use patterns.

tools/release_builder.py
```diff
@@ -24,19 +24,19 @@
...
+                 min_dcos_release_version=os.environ.get('MIN_DCOS_RELEASE_VERSION', '1.8'),
+                 http_release_server=os.environ.get('HTTP_RELEASE_SERVER', 'https://downloads.mesosphere.com'),
+                 s3_release_bucket=os.environ.get('S3_RELEASE_BUCKET', 'downloads.mesosphere.io'),
+                 release_docker_image=os.environ.get('RELEASE_DOCKER_IMAGE'),
+                 release_dir_path=os.environ.get('RELEASE_DIR_PATH', ''),
+                 beta_release=os.environ.get('BETA', 'False')):
         self._dry_run = os.environ.get('DRY_RUN', '')
-        self._force_upload = bool(os.environ.get('FORCE_ARTIFACT_UPLOAD', 'false'))
-        name_match = re.match('.+/stub-universe-(.+).zip$', stub_universe_url)
+        self._force_upload = bool(os.environ.get('FORCE_ARTIFACT_UPLOAD', ''))
```
jsrodman: I really don't like this pattern of using empty string as false.  
None, False are good.

Here though, 
`self._force_upload = 'FORCE_ARTIFACT_UPLOAD' in os.environ`
or if you prefer
`self._force_upload = os.environ.has_key('FORCE_ARTIFACT_UPLOAD')`

tools/release_builder.py
nickbp: Pylint is broken per below

tools/release_builder.py
nickbp: The prior version was very broken, as:
```
>>> bool('false')
True
```

Switched to an explicit string comparison for 'true' so that non-empty envvars dont get parsed as True either.

tools/release_builder.py
nickbp: Done

jsrodman:
I only feel really strongly about a few hints about what things are doing (e.g. extract_json_file() is probably obvious to you but not me), and beginning to move away from empty string as a false sentinel.

I also, in *general* feel strongly about commands as vectors/sequences, and tuple unpacking vs thing[0] and thing[1], but not in this pull request.

tools/release_builder.py
nickbp: Removed explicit `None` fallback here and elsewhere

tools/release_builder.py
nickbp: Added a comment

tools/universe_builder.py
nickbp: Switched to `collections.OrderedDict` here and elsewhere

tools/universe_builder.py
nickbp: Now using len('TEMPLATE_')

tools/release_builder.py
jsrodman: I guess part of me thinks that bool() led to the error.  Thanks.

nickbp:
Manually tested the release script as of `61f3459`. Successfully reuploaded to prod and created the branch/PR. A side effect of moving to this new `stub-universe.json` is that it normalizes the JSON formatting in package files (WITHOUT reordering keys) so the release commit shows a bunch of whitespace fixes: https://github.com/mesosphere/universe/compare/version-3.x...automated/release_kafka_0.0.0-test_+qEvAw?expand=1

PR #951, opened 2017-05-03
[INFINITY-1448] Vendorise CLIs dependency on dcos-commons/cli

ssk2:
This allows us to make changes of the master branch of dcos-commons/cli without breaking the frameworks CLIs that are contained within this monorepo. 

I have tested that these each build correctly from a clean start and that the builds use the correct (i.e. vendored) dependency.

PR #946, opened 2017-05-02
Small tweaks to build instructions.

ssk2:
Specifically:
* Tidy up .gitignore to put the `native-*` line at the right level in the hierarchy.
* Update the build instructions to use the latest version of `build.sh` which scripts away many of the manual steps required before.
* Small grammatical tweak to one of the output lines.

Note that I tried using just build_cli.sh and refactoring it to require just one argument but it has a number of dependencies on environment variables that made it more effort than it was worth. Using `build.sh` incurs some overhead (of framework compilation) that we can probably make optional at some point later.

PR #953, opened 2017-05-04
[INFINITY-1449] CLI code refactoring

ssk2:
Merge after https://github.com/mesosphere/dcos-commons/pull/946 and https://github.com/mesosphere/dcos-commons/pull/951.

This PR refactors the flat directory structure of the CLI into three packages:
+ `commands`: containing handlers and methods for each subcommand.
+ `config`: containing a handful of constants and global variables that encapsulate configuration
+ `utils`: containing helper methods that facilitate HTTP communication, response parsing and integration with the DC/OS CLI.

I also updated the helloworld example to use this new layout. Will probably want to update this again the vendoring work is merged as well.


benclarkwood:
@ssk2 Please resolve conflicts.

cli/commands/connection.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+	"github.com/mesosphere/dcos-commons/cli/utils"
+	"gopkg.in/alecthomas/kingpin.v2"
+	"strings"
+)
+
+// Connection section, manually implemented by some services (DEPRECATED, use common Endpoints)
+
+type ConnectionHandler struct {
```
benclarkwood: Let's deprecate this now.

benclarkwood: (everything is using HandleEndpoints AFAIK, and if they aren't: it's vendored.

cli/utils/http.go
```diff
@@ -85,40 +67,40 @@ func HTTPPutJSON(urlPath, jsonPayload string) *http.Response {
 }
 
 func HTTPQuery(request *http.Request) *http.Response {
```
benclarkwood: Unless this is used outside of the package, HTTPQuery + CreateHTTPRequest and anything else only used in the package should be package private (lower case first letters).

benclarkwood: Looks like everything south of HTTPPutJSON should be private.

cli/utils/response.go
```diff
@@ -1,4 +1,4 @@
-package cli
+package utils
```
benclarkwood: `utils` is IMO the wrong name.

Everything in this package is for a client to the scheduler.

cli/config/config.go
```diff
@@ -0,0 +1,13 @@
+package config
```
benclarkwood: Can config be made non-globals?

e.g. can it be created at the beginning and then passed around? (Makes DI easier in tests... ish)

cli/utils/response.go
benclarkwood: Additionally, if the "client" can just be created once and re-used, then injected into commands (rather than globally referenced) writing tests will be much, much easier. (Not saying: make this change now, but it will eventually need to happen)

cli/config/config.go
ssk2: I think so, but this will require more refactoring. I'd like to do that in a separate PR if that is acceptable?

benclarkwood: Yes, this can get bucketed into the same category as the comment above it (Will probably need to happen eventually but can be deferred for now).

cli/utils/response.go
nickbp: As of this writing there are three files in utils/:
- dcoscli.go: access config data in the main CLI, run main CLI commands
- http.go: build requests to query the scheduler (and soon likely other services in the cluster)
- response.go: handle http responses

IMO two changes could be made here:
- response.go could be folded into http.go as they're pretty closely related (I think http.go even invokes response.go)
- I'm a bit hesitant to have a `utils` package in general. I think @benclarkwood 's point of instead calling it `client` or `clients` would work, as the functionality here is basically client code for querying 1) schedulers over HTTP and 2) the main DC/OS CLI over the commandline.

frameworks/helloworld/cli/dcos-hello-world/main.go
```diff
@@ -2,17 +2,18 @@ package main
 
 import (
 	"github.com/mesosphere/dcos-commons/cli"
+    "github.com/mesosphere/dcos-commons/cli/commands"
```
nickbp: indent

tools/build_framework.sh
```diff
@@ -79,7 +79,7 @@ case "$PUBLISH_STEP" in
         echo "---"
         echo "Build complete, skipping publish step."
         echo "Use one of the following additional arguments to get something that runs on a cluster:"
-        echo "- 'local': Host the build in a local HTTP server for use by a vagrant cluster."
+        echo "- 'local': Host the build in a local HTTP server for use by a DC/OS Vagrant cluster."
```
nickbp: strictly speaking this was written for DC/OS Docker clusters but within a vagrant container. We found DC/OS Vagrant to not really be useful for our purposes (required huge machines when we're targeting e.g. laptops with 4-8GB)

see also: https://github.com/mesosphere/dcos-commons#quick-start

frameworks/helloworld/cli/dcos-hello-world/main.go
```diff
@@ -2,17 +2,18 @@ package main
...
 
 func main() {
 	app := cli.New()
 
-	cli.HandleConfigSection(app)
-	cli.HandleEndpointsSection(app)
-	cli.HandlePlanSection(app)
-	cli.HandlePodsSection(app)
-	cli.HandleStateSection(app)
+	commands.HandleConfigSection(app)
```
nickbp: I assume other `main.go`s need the same update?

nickbp: Something to think about: These explicit section calls were added mainly to get around the downsides of unvendored dependencies. Now that vendoring is being introduced, we could bring back a single `HandleDefaultSections()` call that functions as a shortcut to all of these other calls.

cli/commands/connection.go
ssk2: Done.

cli/utils/http.go
ssk2: Done.

cli/utils/response.go
ssk2: I will rename this to `client` for now - we can fold response.go into http.go at a later point.

ssk2: Done.

frameworks/helloworld/cli/dcos-hello-world/main.go
ssk2: Good point.

ssk2: I was only doing this for the example `helloworld` CLI - what's our policy about updating CLIs?

ssk2: Done (for helloworld).

tools/build_framework.sh
ssk2: Ah, okay. This might be confusing with both just `Vagrant` and `DC/OS Vagrant`. Would something like `local DC/OS cluster` work instead.

ssk2:
@benclarkwood, @nickbp, rebased, conflicts resolved and most feedback addressed.

ssk2:
retest this please

benclarkwood:
Not sure what that merge issue is.

@jsrodman any thoughts?

jsrodman:
This merge is odd.  Why is anything at the time of the go build trying to do a merge? is build.sh being invoked  recursively?

ssk2:
retest this please

ssk2:
retest this please

PR #960, opened 2017-05-05
Work around cli incompatibilities for now.

jsrodman:


tools/cli_install.py
```diff
@@ -113,6 +113,11 @@ def download_cli(dcos_url, write_dir):
         cluster_version, _ = cluster_version.split('-', 1) # "1.9-dev" -> 1.9
     major, minor = cluster_version.split('.')[:2] # 1.8.8 -> 1.8
     cluster_version = '%s.%s' % (major, minor)
+
+    # XXX 1.10 has compatability issues; so until we hae the time to modify
+    # all our tests
+    cluster_version="1.9"
```
benclarkwood: I was thinking something like:
```
if int(minor) >= 9:
        minor = 9

    cluster_version = '%s.%s' % (major, minor)
```

But this works too.

jsrodman: I just felt like there's no correct behavior for now, so make it as clearly wrong as possible.

benclarkwood: Fair!

benclarkwood:
I vote, that if this manages to run any tests, we go ahead and merge it so we can keep CI moving.

benclarkwood:
retest this please

jsrodman:
I've now proven that the problem is the new cli code but that this approach doesn't avoid the new logic, so killing this.

PR #961, opened 2017-05-05
Force v 0.4.16 for dcos python until we fix our test code & shakedown.

jsrodman:
I haven't finished testing this locally, but i have shown pip does what I want.


benclarkwood:
👍

PR #963, opened 2017-05-05
[INFINITY-1380] Fix cleaning of resources for permanently failed tasks.

gabrielhartmann:
If a task is permanently failed, it's resources should not be considered "expected," so they can be cleaned up.


benclarkwood:
retest this please

gabrielhartmann:
retest this please

mrbrowning:
Looks good. I suppose we'll need to remember to revisit this logic if we ever add support for shared volumes.

PR #968, opened 2017-05-09
yaml: Add json parsing to env vars with prefix "JSONPARSE_"

nlsun:


nlsun:
@benclarkwood thoughts?

benclarkwood:
1. Tests?
2. I don't see a need to have a special prefix. Just try to parse every envvar to JSON, and if it fails, then take it as a string (this is a common pattern in Java parsing libraries).

It is possible this could break things for some users, as I'm not super familiar with Mustache.

@gabrielhartmann @nickbp WDYT?

nlsun:
I'll write some tests if people agree that this is a good idea.

I was thinking that we've likely been relying on the fact that env vars are passed through verbatim

benclarkwood:
Right. 

If Mustache does the right thing with parsed JSON (that the user isn't doing anything with), then I think the simplicity is worth it.

nickbp:
What's the use-case? So far we've effectively been doing fixed hierarchies via prefixing env names. Not the most elegant but it's easy to understand/isn't surprising:
```
{
  "foo": 5,
  "prefix2": {
    "bar": 8
  }
}
```
vs env equivalent:
```
PREFIX1_FOO = 5
PREFIX1_PREFIX2_BAR = 8
```

nlsun:
@nickbp Use case is as follows:
```
pods:
  edgelb-pool:
    count: {{POOL_COUNT}}
    placement: {{POOL_CONSTRAINTS}}
    image: {{POOL_IMAGE}}
    resource-sets:
      server-resources:
        cpus: {{POOL_CPUS}}
        memory: {{POOL_MEM}}
        ports:
          {{#POOL_PORTS}}
          p{{v}}:
            port: {{v}}
          {{/POOL_PORTS}}
```
```
POOL_PORTS=[{ "v": 80 }, { "v": 443 }]
```

nlsun:
I'm not quite sure how this is related to hierarchies

gabrielhartmann:
@nlsun:  Let's put some kind of a doc together describing a little more clearly what your needs are, then design a solution before we jump straight to the PR.  We're happy to help in person if you'd like or in a hangout.  I have some time today.  Want to schedule something?

nickbp:
FWIW, from the example, I'd recommend adding custom logic to your `Main.java` that dynamically adds `PortSpec`s to your `ResourceSet`. Basically the `ServiceSpec` Java object model of your svc.yaml can be manually edited before it's passed to `DefaultScheduler`.

nickbp:
As an example, HDFS does some manual placement rule configuration here: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java#L84

PR #975, opened 2017-05-12
[INFINITY-1484] strict 18 19

jsrodman:
Some hacks to increase reliability in the face of trouble with virtualenv & jenkins, for use in the CI job running permissive & strict against dcos 1.8 and 1.9


jsrodman:
Merging this, without the JIRA being complete, because:
1 - It's correct, even if the jenkins job isn't finished yet.
2 - I want the version sensitivity in other work.

jsrodman:
Realized I need to disable fail-fast to get any clear picture of what the set of failures are in these test runs.

PR #977, opened 2017-05-12
Refactor all ZK access (StateStore, ConfigStore, ...)

nickbp:
- Gets rid of `StateStoreCache` in favor of caching at the `Persister` level: `PersisterCache`
- Implements a `MemPersister` for use by `PersisterCache`.
- Switches most tests (that aren't specifically testing Curator behavior) to use `MemPersister` as well. `:scheduler:test` went from 66s to 40s here.
- Moves all Curator-specific logic into the `CuratorPersister`, allowing renaming of `Curator[State|Config]Store` to `Default[State|Config]Store` as they no longer contain any Curator-specific logic.
- Breaks out `DefaultService`'s locking logic into a separate `CuratorLocker` class in the `curator` package.


nickbp:
Should be in good shape now, PTAL

nickbp:
retest this please (ccm failure)

nickbp:
Broke out most of the non-functional changes into #993. Will rebase this against that. Go look at that one for now.

nickbp:
OK, now we're down to about a third of the original size. PTAL

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterCache.java
```diff
@@ -0,0 +1,121 @@
...
+        } finally {
+            rlock.unlock();
+        }
+    }
+
+    @Override
+    public void set(String path, byte[] bytes) throws PersisterException {
+        rwlock.lock();
+        try {
+            persister.set(path, bytes);
```
benclarkwood: What happens if the persister.set succeeds but the cache.set fails?

sdk/scheduler/src/test/java/com/mesosphere/sdk/api/StateResourceTest.java
```diff
@@ -119,22 +123,26 @@ public void testGetPropertyFails() {
...
         assertEquals(409, response.getStatus());
     }
 
     @Test
-    public void testRefreshCacheFailure() {
-        doThrow(new StateStoreException(Reason.UNKNOWN, "hi")).when(mockStateStoreCache).refresh();
-        Response response = new StateResource(mockStateStoreCache).refreshCache();
+    public void testRefreshCacheFailure() throws PersisterException {
+        when(mockStateStore.getPersister()).thenReturn(mockPersisterCache);
+        doThrow(new PersisterException(Reason.UNKNOWN, "hi")).when(mockPersisterCache).refresh();
```
benclarkwood: hello!

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterCache.java
nickbp: The cache.set() call is unlikely to fail as it's seeking/setting a value in a local tree structure. Failure there would imply OOM

nickbp:
retest this please (looks like the cluster went away while elastic was deploying?)

nickbp:
DC/OS broke again. Not going to wait for it to be fixed to verify that my javadoc fix passes integration.

PR #982, opened 2017-05-12
Template framework needs better templating

drewkerrigan:
* `new-framework.sh` and `new-standalone-service.sh` both did a replace of the word "template" in `frameworks/template`, but there are legitimate uses of that word in `frameworks/template/vendor`. The proposed solution is to instead use a more template friendly term like `{{template}}` for replacements.
* Also updated the template's cli and vendor dirs with a more recent version of the dcos-commons cli package 

Altenative solution to this problem?
* Remove `frameworks/template/vendor` and instead run `govendor fetch github.com/mesosphere/dcos-commons/cli` once the new framework has been generated

### DIFF
--- a/frameworks/template/build.gradle
+++ b/frameworks/template/build.gradle
@@ -31,9 +31,9 @@ distZip.finalizedBy copyExecutor
 
 distributions {
     main {
-        baseName = 'template-scheduler'
+        baseName = '{{template}}-scheduler'
         version = ''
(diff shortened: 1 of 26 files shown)


benclarkwood:
@drewkerrigan We have an open ticket to address these templating issues. I'm going to close this (since it's drifted) in favor of that future work.

PR #984, opened 2017-05-15
Modify the upgrade integration test to work with beta packages

susanxhuynh:
Modified the upgrade/downgrade integration test to work with services that have beta packages in the Universe. Now the integration test will install the beta- package, if specified, rather than the non-beta package.
- The main change was to add additional parameters to the upgrade_downgrade() method so that the caller can specify different values for (1) Universe package name, (2) test package name, and (3) service name.
- For example, for Elastic, (1) Universe package name = `beta-elastic`, (2) test package name = `elastic`, and (3) service name = `elastic`.
- Before this change, all three names were always the same value, so there was no need to distinguish.

### DIFF
--- a/frameworks/elastic/tests/test_shakedown.py
+++ b/frameworks/elastic/tests/test_shakedown.py
@@ -136,13 +136,13 @@ def test_kibana_proxylite_adminrouter_integration():
 @pytest.mark.upgrade
 @pytest.mark.sanity
 def test_upgrade_downgrade():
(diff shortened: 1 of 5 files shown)


susanxhuynh:
@paegun FYI, the way this is implemented, there will be a small code change when a framework moves off of beta: the `universe_package_name` will have to be updated.

paegun:
@susanxhuynh please see https://github.com/mesosphere/confluent/pull/13 which was updated to specify the beta bits as required by this change.

PR #985, opened 2017-05-15
[WIP] U2 commands for the CLI

ssk2:
Still a work in progress.

To Do:
+ Unit tests:
    + Testing command line arguments
    + Testing file parsing
    + Testing JSON parsing
    + Testing HTTP calls
+ Better / more usable output

Done: 
+ Initial implementation of the CLI commands
+ Modification of `http.py` to talk to Cosmos.
+ Tested with Cosmos on DC/OS cluster
+ Different output for non Enterprise clusters.


ssk2:
@nickbp and @benclarkwood: pinging you here for early review. Still a WIP.

cli/client/http.go
```diff
@@ -5,68 +5,85 @@ import (
...
 	"net/url"
 	"path"
 	"strings"
+
+	"github.com/mesosphere/dcos-commons/cli/config"
 )
 
-func HTTPGet(urlPath string) *http.Response {
-	return CheckHTTPResponse(HTTPQuery(CreateHTTPRequest("GET", urlPath)))
+func HTTPServiceGet(urlPath string) *http.Response {
```
nickbp: IMO I think it'd be fine to leave the current Service ones with the original "default" names. The Cosmos queries in comparison are still a bit of a special case.

cli/commands/service.go
```diff
@@ -0,0 +1,138 @@
...
+
+type DescribeHandler struct {
+	DescribeName string
+}
+
+type DescribeRequest struct {
+	AppID string `json:"appId"`
+}
+
+func unmarshallJSON(jsonBytes []byte) (map[string]interface{}, error) {
```
nickbp: unmarshal

cli/commands/service.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (cmd *DescribeHandler) DescribeConfiguration(c *kingpin.ParseContext) error {
+	// TODO: add error handling
+	requestContent, _ := json.Marshal(DescribeRequest{config.ServiceName})
+	response := client.HTTPCosmosPostJSON("describe", string(requestContent))
+	responseBytes := client.GetResponseBytes(response)
+	resolvedOptions, err := parseDescribeResponse(responseBytes)
+	if err != nil {
+		log.Printf("Failed to unmarshall response.")
```
nickbp: - unmarshal
- Maybe just put both of these messages on a single line?
- print `err`?

cli/commands/service.go
```diff
@@ -0,0 +1,138 @@
...
+			log.Fatalf("Failed to load specified options file %s: %s", cmd.OptionsFile, err)
+		}
+		request.OptionsJSON = optionsJSON
+	}
+	requestContent, _ := json.Marshal(request)
+	response := client.HTTPCosmosPostJSON("update", string(requestContent))
+	responseBytes := client.GetResponseBytes(response)
+	deploymentID, err := parseUpdateResponse(responseBytes)
+	if err != nil {
+		log.Printf("Failed to unmarshall response.")
```
nickbp: same comments as above

cli/client/http.go
```diff
@@ -182,16 +208,42 @@ func createURL(urlPath, urlQuery string) *url.URL {
...
+}
+
+func createServiceURL(urlPath, urlQuery string) *url.URL {
+	getDCOSURL()
+	joinedURLPath := path.Join("service", config.ServiceName, urlPath)
+	return createURL(config.DcosUrl, joinedURLPath, urlQuery)
+}
+
+func createCosmosURL(urlPath string) *url.URL {
+	joinedURLPath := ""
```
nickbp: Doesn't look like this declaration is needed here

cli/commands/service.go
benclarkwood: Definitely print err

cli/commands/service.go
```diff
@@ -0,0 +1,138 @@
...
+	requestContent, _ := json.Marshal(DescribeRequest{config.ServiceName})
+	response := client.HTTPCosmosPostJSON("describe", string(requestContent))
+	responseBytes := client.GetResponseBytes(response)
+	resolvedOptions, err := parseDescribeResponse(responseBytes)
+	if err != nil {
+		log.Printf("Failed to unmarshall response.")
+		log.Printf("Original data follows:")
+		outBuf := *bytes.NewBuffer(responseBytes)
+		outBuf.WriteTo(os.Stdout)
+		os.Exit(1)
```
benclarkwood: Is this our typical pattern? This seems... bad to me?

Should just return an error here and have it handled higher.

benclarkwood: Also, if you're doing this multiple times, DRY and have a function for it.

cli/commands/config.go
```diff
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+
```
benclarkwood: Is gofmt inserting this newline?

cli/client/http.go
ssk2: I'd like to be explicit - it may be the case in the future that we add support for querying other services on DC/OS directly (e.g. Marathon). What do you think?

cli/client/http.go
ssk2: Yup - good catch! (Leftover variable.)

cli/commands/service.go
ssk2: Fixed.

cli/commands/service.go
ssk2: Fixed.

cli/commands/service.go
ssk2: Fixed.

cli/commands/service.go
ssk2: I copied what response.go was doing. Do you have a better recommendation? Happy to change. For now, I'll move this out into a function.

cli/commands/config.go
ssk2: My editor uses `goimports` (which is identical to `gofmt` except it organises imports too). I quite like it (it separates third party libraries from official libraries and removes unused imports) - but let me know if you'd like me  to stick with just `gofmt`.

benclarkwood: No, that's fine. 

Just wanted to make sure it was some automated tool doing it.

cli/client/cosmos_test.go
```diff
@@ -0,0 +1,146 @@
+package client
+
+import (
+	"bytes"
+	"net/http"
```
benclarkwood: FYI: The `httptest` package has useful utilities for testing HTTP requests

cli/client/dcoscli.go
```diff
@@ -1,11 +1,12 @@
...
-	"github.com/mesosphere/dcos-commons/cli/config"
 	"log"
 	"os"
 	"os/exec"
 	"strings"
+
+	"github.com/mesosphere/dcos-commons/cli/config"
 )
 
 // TODO(nick): Consider breaking this config retrieval out into a separate independent library?
```
benclarkwood: Seems like this TODO has been done

cli/client/http.go
```diff
@@ -119,60 +131,56 @@ func HTTPQuery(request *http.Request) *http.Response {
 		switch err.(type) {
 		case x509.UnknownAuthorityError:
 			// custom suggestions for a certificate error:
-			log.Printf("HTTP %s Query for %s failed: %s", request.Method, request.URL, err)
-			log.Printf("- Is someone intercepting the connection to steal your credentials?")
-			log.Printf("- Is the cluster CA certificate configured correctly? Check 'dcos config show core.ssl_verify'.")
-			log.Fatalf("- To ignore the unvalidated certificate and force your command (INSECURE), use --force-insecure")
+			logMessage("HTTP %s Query for %s failed: %s", request.Method, request.URL, err)
+			logMessage("- Is someone intercepting the connection to steal your credentials?")
```
benclarkwood: Let's not log this. It's terrifying and will lead to users filing pointless support tickets.

cli/commands/service.go
```diff
@@ -0,0 +1,142 @@
...
+		resolvedOptionsBytes, err := json.Marshal(resolvedOptions)
+		if err != nil {
+			return nil, err
+		}
+		return resolvedOptionsBytes, nil
+	}
+	return nil, nil
+}
+
+func (cmd *DescribeHandler) DescribeConfiguration(c *kingpin.ParseContext) error {
```
benclarkwood: If the parse context isn't being used, I would omit it as a parameter.

cli/commands/service.go
```diff
@@ -0,0 +1,142 @@
...
+func (cmd *UpdateHandler) UpdateConfiguration(c *kingpin.ParseContext) error {
+	if cmd.Status {
+		printStatus()
+		return nil
+	}
+	doUpdate(cmd.OptionsFile, cmd.PackageVersion)
+	return nil
+}
+
+func HandleServiceSection(app *kingpin.Application) {
```
benclarkwood: As discussed in the sprint review, we'll move this directly under the "root" of the CLI.

cli/client/cosmos_test.go
ssk2: Will check it out.

cli/client/dcoscli.go
ssk2: I imagine this was meant specifically for the bits that interact with the dcoscli, so I reckon this is still a "todo".

cli/commands/service.go
ssk2: Unfortunately needed by the Kingpin interface.

cli/client/cosmos.go
```diff
@@ -0,0 +1,83 @@
...
+				logMessage("Cosmos error: %s: %s", errorType, message)
+			}
+			printErrorAndExit(response)
+		}
+	}
+}
+
+func checkCosmosHTTPResponse(response *http.Response) *http.Response {
+	switch {
+	case response.StatusCode == 404:
```
ssk2: Done.

cli/client/http.go
ssk2: Removed.

cli/commands/service.go
ssk2: Done! Note that this breaks the one "section" per file convention that we've been going with. I kept them in the same file so that the code for these commands and their related code are grouped logically.

PR #987, opened 2017-05-16
Improve handling of framework deregistration

loren:
- Moved uninstall-specific classes into install package
- Test for unexpected OfferRecommendation types in UninstallRecorder
- Folded DeleteServiceRootPathStep into DeregisterStep so the phase has just one Step now.
- If there is no framework ID, wipe ZK and return a COMPLETE plan
- If there is no framework ID, don't register the FW with a new schedule driver
- Renamed "misc" to "deregister" in Plan steps/phases
- Removed boolean from setting state store's "installing" property key
- Removed unneeded continue/return's from CuratorStateStore


sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreUtils.java
```diff
@@ -231,10 +231,10 @@ public static boolean isUninstalling(StateStore stateStore) throws StateStoreExc
     }
 
     /**
-     * Sets an 'uninstall' property in the provided {@link StateStore} to the provided value.
+     * Sets an 'uninstall' property in the provided {@link StateStore} to true
```
nickbp: Mega-nitpick: `{@code true}`

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallRecorderTest.java
```diff
@@ -0,0 +1,24 @@
...
+    public void testHandlingOfUnexpectedOfferRecommendation() throws Exception {
+        Protos.Resource resource = ResourceUtils.getUnreservedScalar("cpus", 1.0);
+        OfferRecommendation unsupportedOfferRecommendation = new CreateOfferRecommendation(null, resource);
+        StateStore mockStateStore = mock(StateStore.class);
+        OperationRecorder operationRecorder = new UninstallRecorder(mockStateStore, null);
+        // should just return without error
+        operationRecorder.record(unsupportedOfferRecommendation);
+    }
+
+}
```
nickbp: Missing trailing newline

loren: addressed newline and code javadoc

PR #991, opened 2017-05-16
De-duplicate ProcessBuilder construction

nickbp:


sdk/executor/src/main/java/com/mesosphere/sdk/executor/ProcessTask.java
```diff
@@ -39,18 +37,8 @@ public static ProcessTask create(
...
             Protos.TaskInfo taskInfo,
             boolean exitOnTermination) {
-        return create(executorDriver, taskInfo, getProcess(taskInfo), exitOnTermination);
-    }
-
-    public static ProcessBuilder getProcess(TaskInfo taskInfo) {
-        CommandInfo commandInfo = taskInfo.getCommand();
-        String cmd = commandInfo.getValue();
-
-        ProcessBuilder builder = new ProcessBuilder("/bin/sh", "-c", cmd);
```
mohitsoni: Thinking out loud: This one used to use `/bin/sh` and now we are going to use `/bin/bash`. It seems like a safe change to me.

The only thing that I would like to bring up is that usually in linux distributions, `/bin/sh` symlinks to `/bin/bash` or the default system shell. 
Now, I'm wondering if we should make it explicit in our utility to use `/bin/bash` explicitly OR fallback to the default distro preferred behavior ?

nickbp: I think it'd make more sense to specifically go with `/bin/bash` as done here if only because of that issue -- to make the expected environment more consistent.

mohitsoni: 👍

PR #992, opened 2017-05-16
Fixed JettyApiServerTest use of bind any address as loopback

paegun:
* fixed jetty api server test use of bind any ip address (0.0.0.0) as the address that the client performs the http get against, but should be localhost (127.0.0.1 or ::1 or ...)
  * 0.0.0.0 is not universally hacked to be treated equivalent to proper loopback addresses, but it may be, i.e. on various linux distros that prefer UX over universally correct implementation.
* made the test ~properly threaded where properly is not using a busy wait.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/JettyApiServer.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/JettyApiServer.java
@@ -12,6 +12,10 @@ import java.util.Collection;
 
 /**
  * A JettyApiServer takes a list of POJO JAX-RS Resources and serves them on the indicated port of 0.0.0.0.
+ *
+ * When constructed with a port of zero, an available ephemeral port will be
+ * determined and bound by the JettyApiServer. In this case, call getLocalPort
+ * to obtain the port number.
  */
 public class JettyApiServer {
     private final Server server;
@@ -55,6 +59,9 @@ public class JettyApiServer {
         if (this.port != 0) {
             return this.port;
         }
+        // When constructed w/ port 0, an available ephemeral port will be
+        // bound and this will be on the first and only network interface
+        // (connector) for the Jetty Server.
         NetworkConnector networkConnector = (NetworkConnector) server.getConnectors()[0];
         this.port = networkConnector.getLocalPort();
         return this.port;
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/JettyApiServer.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/JettyApiServer.java
@@ -1,6 +1,8 @@
 package com.mesosphere.sdk.api;
 
+import org.eclipse.jetty.server.NetworkConnector;
 import org.eclipse.jetty.server.Server;
+
 import org.glassfish.jersey.jetty.JettyHttpContainerFactory;
 import org.glassfish.jersey.server.ResourceConfig;
 
@@ -13,6 +15,7 @@ import java.util.Collection;
  */
 public class JettyApiServer {
     private final Server server;
+    private int port;
 
     public JettyApiServer(int port, Collection<Object> resources) {
         ResourceConfig resourceConfig = new ResourceConfig();
@@ -25,6 +28,9 @@ public class JettyApiServer {
                 .port(port).build();
 
         this.server = JettyHttpContainerFactory.createServer(baseUri, resourceConfig);
+        if (port != 0) {
+            this.port = port;
+        }
     }
 
     public void start() throws Exception {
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/test/java/com/mesosphere/sdk/api/JettyApiServerTest.java
```diff
@@ -16,22 +16,44 @@
...
  * This class tests the JettyApiServer.
  */
 public class JettyApiServerTest {
     private JettyApiServer jettyApiServer;
-    private int port;
     private static final String TEST = "test";
 
     @Before
     public void beforeEach() throws IOException {
+        final int jettyStartRetries = 10;
```
benclarkwood: We weren't previously retrying the starts, correct? Why start now?

sdk/scheduler/src/test/java/com/mesosphere/sdk/api/JettyApiServerTest.java
```diff
@@ -16,22 +16,44 @@
...
+                    } catch (Exception e) {
+                        e.printStackTrace();
+                    }
+
+                    sleep(jettyStartRetryDelay);
+                }
+            }
+        }).start();
+
+        for(int i = 0; i < jettyStartRetries; i++) {
```
benclarkwood: You could use LifeCycle.Listener to do this more fancily.

This "looks" a little funky, but if it works it works.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/JettyApiServer.java
```diff
@@ -44,4 +50,13 @@ public void stop() throws Exception {
     public boolean isStarted() {
         return server.isStarted();
     }
+
+    public int getLocalPort() {
+        if (this.port != 0) {
+            return this.port;
+        }
+        NetworkConnector networkConnector = (NetworkConnector) server.getConnectors()[0];
```
benclarkwood: Comment describing why this works please :)

paegun:
Thank you @benclarkwood . You identified a few areas of the code that were unclear w/o further comment, so I've commented.

Also thank you for the suggestion to look into using LifeCycle.Listener. I felt that the general application of busy-wait + retry patterns was a quick and dirty solution to something that is causing `build.sh` to locally fail for me (and likely others) consistently. I did not feel that this is something that is worth the time to investigate a specific solution that is more elegant. BUT your suggestion is now captured in code commentary, and I hope to loop back to implement. Thanks for the education/edification.

sdk/scheduler/src/test/java/com/mesosphere/sdk/api/JettyApiServerTest.java
paegun: comments \/ b/c these points should live beyond the review.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/JettyApiServer.java
```diff
@@ -1,6 +1,8 @@
 package com.mesosphere.sdk.api;
 
+import org.eclipse.jetty.server.NetworkConnector;
 import org.eclipse.jetty.server.Server;
+
```
benclarkwood: Is this the result of an import format? If it is that's fine

sdk/scheduler/src/test/java/com/mesosphere/sdk/api/JettyApiServerTest.java
```diff
@@ -75,4 +83,22 @@ public Response getFrameworkId() {
...
+    private void sleep(int ms) {
+        // an interruptable sleep b/c if interrupted, i.e. by SIGINT, np,
+        // continue to crash, don't worry about sleep
+        try {
+            Thread.sleep(ms);
+        } catch (InterruptedException e) { }
+    }
+
+    private int getJettyApiServerPort() {
+        if (jettyApiServer == null) {
```
benclarkwood: This seems a little weird. I'd just throw a runtime exception here.

Or, just call `jettyApiServer.getLocalPort()` directly at https://github.com/mesosphere/dcos-commons/pull/992/files#diff-c9f51af3b06d6f233439266a232d6d3cR72

I think that would fail in a logical both cases?

Right now, you have to see what request failed to parse, then see what that number means in this method.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/JettyApiServer.java
paegun: yes, the line break on change in org is the same imposed magical formatting rule as the other line breaks in the import section.

sdk/scheduler/src/test/java/com/mesosphere/sdk/api/JettyApiServerTest.java
paegun: Calling directly does not match the exposure and call of isStarted() . And yes, returning an error code rather than throwing an exception at the point of the exception is not correct Java. I don't think a spelunking expedition to re-find the existing code that lead me to follow it's pattern, but IMHO this is relatively unimportant as an attempt to request a resource over HTTP on port -1 or -2 will fail miserably, throw an exception and lead the developer to this line of code quickly.

PR #993, opened 2017-05-17
Prelude to ZK refactoring

nickbp:
Gets most of the big nonfunctional changes out of the way before #977

- Renames Curator[State|Config|SchemaVersion]Store to Default[*]Store and removes MOST curator-specific code therein (remainder will be cleaned out in a following PR)
- Adds a builder to CuratorPersister
- Moves curator locking logic from DefaultService into a separate CuratorLocker class
- Moves some non-Curator-specific calls from CuratorUtils into a new PersisterUtils for use by upstream Persister users now and by other Persisters later.
- Moves ACL handling function into the aforementioned CuratorPersister builder.
- Simplifies handling of ConfigStore/StateStore creation in DefaultScheduler, to be further reduced in a following PR


sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/Persister.java
```diff
@@ -4,41 +4,65 @@
...
+ * A low-level interface for key/value storage in a tree structure.
+ *
+ * <p>Individual nodes may be the parent of other nodes. Some of these parent nodes may lack any data of their own. The
+ * root-level node (with path "", or "/") is considered to always be present.
+ *
+ * <p>This interface should be implemented in order to store and fetch data, with paths delimited by
+ * {@link com.mesosphere.sdk.storage.PersisterUtils#PATH_DELIM}.
  */
 public interface Persister {
+
```
nickbp: Reviewer note: This is just doing the following:
- renaming `delete()` to `deleteAll()` to be clearer on what the expected behavior is
- reordering the existing methods to be a bit easier to understand
- adding more thorough javadocs on behavior

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterUtils.java
```diff
@@ -0,0 +1,57 @@
+package com.mesosphere.sdk.storage;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utilities relating to usage of {@link Persister}s.
+ */
+public class PersisterUtils {
```
nickbp: reviewer note: moved from CuratorUtils

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorLocker.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Gets an exclusive lock on service-specific ZK node to ensure two schedulers aren't running simultaneously for the
+     * same service.
+     */
+    public void lock() {
+        if (curatorClient != null) {
+            throw new IllegalStateException("Already locked");
+        }
+        curatorClient = CuratorFrameworkFactory.newClient(zookeeperConnection, CuratorUtils.getDefaultRetry());
```
nickbp: From what the unit test has, I think the ACLs are effectively applied to newly written data, so in theory the ACLs written via CuratorPersister shouldn't intersect with the lock.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorLocker.java
```diff
@@ -0,0 +1,104 @@
...
+    @VisibleForTesting
+    protected TimeUnit getWaitTimeUnit() {
+        return TimeUnit.SECONDS;
+    }
+
+    /**
+     * Broken out into a separate function to allow overrides in tests.
+     */
+    @VisibleForTesting
+    protected void exit() {
```
nickbp: Looks like I forgot to include `CuratorLockerTest` when copying `CuratorLocker` from #977: https://github.com/mesosphere/dcos-commons/pull/977/files#diff-8e1335388f184b40f695532d408127eb

This is a strategy to allow the test to override the exit behavior. In the unit test, calling `exit()` results in setting a boolean which the test can then validate was set.

I'm going to update this PR with CuratorLockerTest.java and then merge it.

nickbp:
Build is green after (only) adding CuratorLockerTest, gonna call it mergeable (red X is from me cancelling it before it went and created clusters etc)

PR #995, opened 2017-05-17
Second prelude to PR #977

nickbp:
Removes remaining pieces of Curator-specific logic from the Default[Config|State|SchemaVersion]Stores.

- Moves `/dcos-service-foo` prefixing from Default[*]Stores into `CuratorPersister`. As a result the service name is now passed directly to the `CuratorPersister` instead of to the Default[*]Stores (updates callers to do so).
- Switches `Persister` interface over to `PersisterException`s. This replaces Curator-specific `KeeperException` type handling in `Persister` callers.
- Moves the new root-delete functionality in `DefaultStateStore` into `CuratorPersister` as a special case of `deleteAll()`. This localizes knowledge of the lock file as well as how the root node cannot be deleted by the Scheduler.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
@@ -3,7 +3,9 @@ package com.mesosphere.sdk.curator;
 import com.google.common.annotations.VisibleForTesting;
 import com.mesosphere.sdk.specification.ServiceSpec;
 import com.mesosphere.sdk.storage.Persister;
+import com.mesosphere.sdk.storage.PersisterException;
 import com.mesosphere.sdk.storage.PersisterUtils;
+import com.mesosphere.sdk.storage.StorageError.Reason;
 
 import org.apache.curator.RetryPolicy;
 import org.apache.curator.framework.CuratorFramework;
@@ -11,7 +13,6 @@ import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.api.ACLProvider;
 import org.apache.curator.framework.api.transaction.CuratorTransaction;
 import org.apache.curator.framework.api.transaction.CuratorTransactionFinal;
-import org.apache.curator.framework.api.transaction.CuratorTransactionResult;
 import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.ZooDefs;
 import org.apache.zookeeper.data.ACL;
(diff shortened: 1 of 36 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -126,95 +130,198 @@ public CuratorPersister build() {
...
 
     @VisibleForTesting
-    CuratorPersister(CuratorFramework client) {
+    CuratorPersister(String serviceName, CuratorFramework client) {
+        this.serviceRootPath = CuratorUtils.getServiceRootPath(serviceName);
         this.client = client;
         this.client.start();
     }
 
     @Override
```
nickbp: Reviewer note: this file has a bit more churn than strictly necessary as its functions were reordered to match the order of the `Persister` interface. The effective changes are:
- Addition of custom root-delete handling in `deleteAll()` (ported from `DefaultStateStore`)
- Catching `Exception`/`KeeperExceptions`s and converting them into the appropriate `PersisterException`s
- Prefixing received paths with `/dcos-service-foo` before accessing Curator itself

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/DefaultConfigStore.java
```diff
@@ -108,64 +102,68 @@ public void clear(UUID id) throws ConfigStoreException {
...
     }
 
     @Override
     public void setTargetConfig(UUID id) throws ConfigStoreException {
         try {
-            persister.set(targetPath, id.toString().getBytes(StandardCharsets.UTF_8));
-        } catch (Exception e) {
-            throw new ConfigStoreException(Reason.STORAGE_ERROR, String.format(
+            persister.set(TARGET_PATH_NAME, id.toString().getBytes(StandardCharsets.UTF_8));
+        } catch (PersisterException e) {
```
benclarkwood: Were only KeeperExceptions possible here before?

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/DefaultStateStore.java
```diff
@@ -322,22 +274,26 @@ private CuratorTransactionFinal deleteChildren(String root, CuratorTransactionFi
...
             try {
-                byte[] bytes = persister.get(taskPathMapper.getTaskStatusPath(taskName));
+                byte[] bytes = persister.get(getTaskStatusPath(taskName));
                 taskStatuses.add(Protos.TaskStatus.parseFrom(bytes));
-            } catch (KeeperException.NoNodeException e) {
-                // The task node exists, but it doesn't contain a TaskStatus node. This may occur if
-                // the only contents are a TaskInfo.
-            } catch (Exception e) {
-                throw new StateStoreException(Reason.STORAGE_ERROR, e);
+            } catch (PersisterException e) {
```
benclarkwood: Same Q here: Are we losing anything by not catching Exception?

Where do these exceptions propagate to? Can they crash the scheduler?

nickbp: These catches were effectively moved upstream into `CuratorPersister` (including `} catch (Exception e) {`). Any exceptions thrown by Curator are translated to `PersisterException`s before they're passed upstream to `Default[State|Config]Store`.

nickbp:
retest this please (ccm failure)

PR #997, opened 2017-05-18
[INFINITY-1539] create service account on dcos18

jsrodman:


jsrodman:
retest this please

jsrodman:
Yeah, maybe we're at a point where I'll need to maintain a table of dependencies.  Awkward but possibly needed.

PR #998, opened 2017-05-19
Update preInstallNotes

triclambert:


ssk2:
We should make sure this is consistent with the messaging we're using for the other apps. I've attempted to improve them slightly but just noticed they're different to what is proposed here for HDFS: https://github.com/mesosphere/dcos-commons/pull/1058.

nickbp:
retest this please

ssk2:
I've cherry-picked this onto my changes to the preInstallNotes field over in #1058.

PR #1000, opened 2017-05-19
Upgrade HDFS JRE path from 8u112 to 8u131

paegun:
* set hdfs (duplicate specification of) JRE_VERSION and TASKCFG_ALL_JRE_VERSION to account for jre path change due to jre upgrade from 8u112 to 8u131.


paegun:
retest this please

PR #1001, opened 2017-05-19
[INFINITY-1588] Correct created role name.

jsrodman:
Seems hdfs-role was never validated by the tests, only hdfs2-role,
so this error wasn't caught.


jsrodman:
retest this please

jsrodman:
IMO the right way would be for the framework to explicitly declare the roles it uses and the script to only create those.

PR #1004, opened 2017-05-19
Replace ResourceUtils

nickbp:
Hold on to your butts...

- Replaces `ResourceUtils`-based resource construction with a `ResourceBuilder`. Some code (particularly some evaluation stages) is still manually editing `Resource` objects but this cleans things up a ton regardless.
- Where applicable, some `ResourceUtils` functs were moved into `ResourceTestUtils` for use by tests, or into other `*Utils` classes.


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceCollectUtils.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.mesos.Protos;
+import org.apache.mesos.Protos.*;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * This class encapsulates common methods for scanning collections of Resources.
+ */
+public class ResourceCollectUtils {
```
benclarkwood: ResourceCollectionUtils is the best name for this class. Please?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallResourceCleaner.java
```diff
@@ -42,8 +44,9 @@
      */
     @Override
     public Collection<Resource> getPersistentVolumesToBeDestroyed(Offer offer) {
-        List<Resource> resources = offer.getResourcesList().stream().
-                filter(resource -> ResourceUtils.getPersistenceId(resource) != null).collect(Collectors.toList());
+        List<Resource> resources = offer.getResourcesList().stream()
+                .filter(resource -> resource.hasDisk() && resource.getDisk().hasPersistence())
```
benclarkwood: nitpick: I'm a little worried about us now having this pattern of checking persistence getting copy pasted around.

Can it just become a public method of Resource?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceBuilder.java
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * Constructs Mesos {@link Resource} protobufs.
+ */
+public class ResourceBuilder {
+
+    /** The name used for reserved network port resources. */
+    public static final String PORTS_RESOURCE_TYPE = "ports";
+    /** The name used for reserved storage/disk resources. */
+    public static final String DISK_RESOURCE_TYPE = "disk";
```
ArtRand: given that these types have associated types (SCALAR, RANGE, etc). Why not make these proper objects so we can validate input easier?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallResourceCleaner.java
nickbp: Resource is a mesos protobuf. We should be able to resolve this once we've got our own classes for these types.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceBuilder.java
nickbp: I think that'll happen as we start moving away from passing around mesos protobufs internally

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallResourceCleaner.java
benclarkwood: Ah, yeah.

Argh.

PR #1011, opened 2017-05-23
Add replacement logic for Cassandra

mrbrowning:


frameworks/cassandra/src/main/java/com/mesosphere/sdk/cassandra/scheduler/CassandraRecoveryPlanOverrider.java
```diff
@@ -0,0 +1,118 @@
...
+
+        // Dig all the way down into the command, so we can append the replace_address option to it.
+        PodInstance podInstance = inputLaunchStep.start().get().getPodInstance();
+        PodSpec podSpec = podInstance.getPod();
+        TaskSpec taskSpec = podSpec.getTasks().stream().filter(t -> t.getName().equals("server")).findFirst().get();
+        CommandSpec command = taskSpec.getCommand().get();
+
+        DefaultCommandSpec.Builder builder = DefaultCommandSpec.newBuilder(command);
+        builder.value(String.format(
+                "%s -Dcassandra.replace_address=%s -Dcassandra.consistent.rangemovement=false%n",
```
mohitsoni: Do you know if `-Dcassandra.consistent.rangemovement=false` needs to be turned back on at some point ?

mohitsoni: Also, we should used `-Dcassandra. replace_address_first_boot` vs `-Dcassandra.replace_address`. As this only takes effect on the first boot after the replacement, and gets ignored by Cassandra on subsequent restarts.

PR #1013, opened 2017-05-24
[INFINITY-1606][INFINITY-1616] strict mode failures in helloworld

jsrodman:
1 - We were not allowing the install-time selection of service-account/principal to work in the examples/*.yml test configurations.
2 - We were creating the wrong role name due to problems with the helloworld/hello-world service name.


jsrodman:
CI failed     frameworks.helloworld.tests.test_marathon_constraint.test_rr_by_hostname, which I think will flap, but I've never seen it flap outside of this work.  That makes me think that use of mesos V0 protocol  caused this instability.  Buhh.

PR #1017, opened 2017-05-24
Pass all URIs via resource.json

loren:


loren:
Retest this please. Stuck on cluster launch.

PR #1020, opened 2017-05-24
[INFINITY-1620] Drop proxylite.

jsrodman:
We're replacing it with edgeLB.
It doesn't work in strict mode.


triclambert:
Bye proxylite, thanks for everything.

PR #1025, opened 2017-05-26
CLI tests

ssk2:
This PR includes a few things:
* Tests (and related fixtures) for the new describe and update commands
* Moved the Cosmos tests to use testify/suite.
* Fixed the plan tests (that were in the wrong place).

PR #1026, opened 2017-05-26
Update the CLIs to work with the latest version of the dcos-commons/cli library

ssk2:
I also removed the vendored dependency on dcos-commons/cli so that now framework CLI builds/tests will always run against the version of the dcos-commons/cli in the same commit. This will allow our tests to catch any issues with the CLI code proactively (at the expense of extra pain when making changes).

Merge after https://github.com/mesosphere/dcos-commons/pull/1025.


nickbp:
Will probably want to merge #1025 first anyway but the changes in here look good

benclarkwood:
The beauty of the monorepo, is you get to do all the pain at once! LGTM.

PR #1028, opened 2017-05-26
Added ability for hello-world framework installation to specify the pod-level user

paegun:
* added ability for hello-world framework installation to specify the pod-level user.
  * as it stands, the same user is used for the scheduler and both pods: hello and world.
  * this fix is required to be able to install in strict mode since the user "nobody" should be used, but w/o this change "root" is used for the pods.
  * we may expand this functionality to allow for distinct users for each, but there is currently very little value in such a thing.


paegun:
duplicates: #1013

PR #1029, opened 2017-05-27
Support foldered services

nickbp:
1. ZK changes only really affect `CuratorPersister` as that's the only place where service name mapping occurs:
- Maps slashes in service names to periods. For example `/path/to/myservice` => `/dcos-service-path.to.myservice` in ZK
- Automatically stores and validates the service name in a zk entry at `/dcos-service-<myservice>/servicename`. This protects against collisions if someone uses literal periods in their service names, and covers any future situation where someone wants to deterministically tell the difference between a service with literal periods and a service with folders when looking at data in zk.

2. In addition, envvar and hostname handling have been cleaned up a bit in order to make it straightforward to use folder-safe service names in hostnames:
- Extremely boilerplatey/generic `TaskConfigRouter`/`ConfigNamespace`/`ConfigValue` have been replaced with a much more succinct `TaskEnvRouter`. The prior had a lot of additional functionality that hadn't ended up being needed.
- Replaced `YAMLServiceSpecFactory`'s static methods with a `DefaultServiceSpecBuilder` and `RawServiceSpecBuilder` (the prior for `ServiceSpec` and the latter for `RawServiceSpec`). Switched 
- Added `EndpointUtils` for building folder-safe endpoint URLs (whether mesos-dns or VIP)
- Fixed folder-unsafe strings in common SDK code, hello-world, and kafka to use `CuratorUtils`/`EndpointUtils` instead (start off with some simple cases to validate the above interface changes). For example the SDK's config template URLs are now folder safe, and Kafka's `KAFKA_ZOOKEEPER_URI` envvar is now constructed to be folder-safe.
- Hello-world and kafka now have integration tests that the service comes up when given a foldered service name (i.e. one containing slashes). Other services will similarly be converted in a separate PR.

3. Finally, one more thing that needed to be fixed up was role/principal handling, as roles cannot start with a slash. We now strip out any leading slash from the service name when constructing the role.


nickbp:
TODO (please don't merge yet): A quick integration test to at least check that things start as expected, and clean up some old principal/role handling which is blocking that integration test..

nickbp:
All set. The new integration test will likely fail until the upcoming DC/OS update which fixes admin router handling of foldered services (should be fixed in a day or two), works fine in a manually created test cluster.

nickbp:
TODO: need to fix l4lb hostnames, e.g.: `Failed to fetch 'http://api./group1/group2/group3/group4/group5/kafka.marathon.l4lb.thisdcos.directory/v1/artifacts/template/c0981484-452a-4acd-aa21-64cfa5405bf2/kafka/broker/server-properties': Error downloading resource: Couldn't resolve host name`

nickbp:
The test failure is from the test I added. Going to fix that one while also adding a test for kafka. Should still be reviewable in the meantime; manual testing has been fine.

nickbp:
retest this please (marathon flake about redeploying an app)

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultTaskEnvRouter.java
```diff
@@ -0,0 +1,127 @@
...
+ * Default implementation of a {@link TaskEnvRouter} which handles forwarding environment variables from two sources:
+ * <ul>
+ * <li>Scheduler's system environment: searches for environment variables that start with {@code "TASKCFG_<TASKTYPE>_"}
+ * and/or {@code "TASKCFG_ALL_"}, and returns appropriate {@link ConfigNamespace}s for task types based on that
+ * environment.</li>
+ * <li>Manually-provided values: The service developer may provide custom values which will then be forwarded to tasks
+ * as needed.</li>
+ * </ul>
+ */
+public class DefaultTaskEnvRouter implements TaskEnvRouter {
```
benclarkwood: I don't see a compelling reason to have an Interface in front of this?

Is it just to limit what we see elsewhere to just the output of the router, and not expose the inputs?

benclarkwood: (This is really just stylistic: We can mock classes, and I loath the Default* soup we have all over the place).

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorUtils.java
```diff
@@ -1,40 +1,121 @@
...
+    static RetryPolicy getDefaultRetry() {
         return new ExponentialBackoffRetry(
                 CuratorUtils.DEFAULT_CURATOR_POLL_DELAY_MS, CuratorUtils.DEFAULT_CURATOR_MAX_RETRIES);
     }
-
     /**
-     * Returns the root node to store all scheduler ZK data inside.
+     * Compares the service name to the previously stored name in zookeeper, or creates a new node containing this data
+     * if it isn't already present. This is useful for two situations where foldered service names may be confused with
+     * literal dot-delimited names:
```
benclarkwood: I am confused, aren't we using `__`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorUtils.java
```diff
@@ -1,40 +1,121 @@
...
-            // "/dcos-service-" + "/foo"
-            return DcosConstants.SERVICE_ROOT_PATH_PREFIX + frameworkName.substring(1);
-        } else {
-            // "/dcos-service-" + "foo"
-            return DcosConstants.SERVICE_ROOT_PATH_PREFIX + frameworkName;
+    static void initServiceName(Persister persister, String serviceName) {
+        try {
+            byte[] bytes = persister.get(SERVICE_NAME_NODE);
+            if (bytes.length == 0) {
+                throw new IllegalArgumentException(String.format(
```
benclarkwood: This is the case that the node exists but has no data, right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/DefaultServiceSpecBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+        this.rawServiceSpec = rawServiceSpec;
+        this.schedulerFlags = schedulerFlags;
+        this.taskEnvRouter = taskEnvRouter;
+        this.fileReader = new FileReader();
+    }
+
+    /**
+     * Assigns an environment variable to be included in all service tasks. Note that this may be overridden via
+     * {@code TASKCFG_*} scheduler environment variables at runtime, and by pod-specific settings provided via
+     * {@link #setPodTaskEnv(String, String)}.
```
benclarkwood: Javadoc is out of date (missing a String)

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/DefaultServiceSpecBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+public class DefaultServiceSpecBuilder {
+
+    /**
+     * Implementation for reading files from disk. Meant to be overridden by a mock in tests.
+     */
+    @VisibleForTesting
+    public static class FileReader {
+        public String read(String path) throws IOException {
+            return FileUtils.readFileToString(new File(path), StandardCharsets.UTF_8);
```
benclarkwood: I was curious if there was a Java 8 native method for this. There isn't precisely (doesn't let you specify a CHARSET). Boo 👎

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawServiceSpecBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+    static {
+        // If the user provides duplicate fields (e.g. 'count' twice), throw an error instead of silently dropping data:
+        YAML_MAPPER.enable(JsonParser.Feature.STRICT_DUPLICATE_DETECTION);
+    }
+
+    private final String yamlTemplate;
+    private Map<String, String> env;
+
+    public RawServiceSpecBuilder(File pathToYamlTemplate) throws IOException {
+        this(FileUtils.readFileToString(pathToYamlTemplate));
```
benclarkwood: Specify UTF_8?

sdk/scheduler/src/test/java/com/mesosphere/sdk/api/EndpointUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Test
+    public void testToEndpoint() {
+        assertEquals("foo:5", EndpointUtils.toEndpoint("foo", 5));
+    }
+
+    @Test
+    public void testToMesosDnsEndpoint() {
+        assertEquals("task.svc.mesos:5", EndpointUtils.toMesosDnsEndpoint("svc", "task", 5));
+        assertEquals("task.svc-to-path.mesos:5", EndpointUtils.toMesosDnsEndpoint("/path/to/svc", "task", 5));
```
benclarkwood: Aside: What is the behavior if the service name they provide is `a.special.service`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * "/group1/group2/group3/group4/group5/kafka" => "group1group2group3group4group5kafka".
+     */
+    private static String removeSlashes(String serviceName) {
+        return serviceName.replace("/", "");
+    }
+
+    /**
+     * "/group1/group2/group3/group4/group5/kafka" => "kafka-group5-group4-group3-group2-group1.marathon.mesos".
```
benclarkwood: Why is the behavior between Navstar and Mesos-DNS different?

Also, what's the maximum number of characters allowed in a DNS record for either?

```Mesos-DNS follows RFC 952 for name formatting. All fields used to construct hostnames for A records and service names for SRV records must be up to 24 characters and drawn from the alphabet (A-Z), digits (0-9) and minus sign (-). No distinction is made between upper and lower case. If the task name does not comply with these constraints, Mesos-DNS will trim it, remove all invalid characters, and replace period (.) with sign (-) for task names. For framework names, we allow period (.) but all other constraints apply. For example, a task named apiserver.myservice launch by framework marathon.prod, will have A records associated with the name apiserver-myservice.marathon.prod.mesos and SRV records associated with name _apiserver-myservice._tcp.marathon.prod.mesos.```

(from: https://mesosphere.github.io/mesos-dns/docs/naming.html)

That looks like 24 characters to me... Should we construct and then hash the service name? Maybe, hash the group and append it?

They say that they will truncate, but that would still lead to conflicts for us, I believe?

Comment (benclarkwood) [inline sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointUtils.java]: Ugh, but then we lose the predictability in the task naming... Were you already planning that Flo would switch to a templating variable?

Comment (nickbp) [inline sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointUtils.java]: This is effectively reproducing what navstar/mesos-dns currently produce for a given service name. The mesos-dns hostnames will likely go away via #1047

**Comment by nickbp** [comment]
retest this please (CI apparently took too long)

**Comment by nickbp** [comment]
retest this please (after passing all the hard stuff, it flakes out on a marathon deployment in hello-world tests!!!)

File: frameworks/helloworld/tests/test_folders.py [hunk 9]
```diff
@@ -0,0 +1,40 @@
...
+    uninstall_foldered()
+
+
+def teardown_module(module):
+    uninstall_foldered()
+
+
+@pytest.mark.sanity
+@pytest.mark.smoke
+def test_install_foldered():
```
mgummelt: We should probably verify that the adminrouter endpoint is reachable

nickbp: Done, added a CLI command (which effectively hits adminrouter) for both helloworld/kafka (the two that support it at the moment)

PR #1031, opened 2017-05-28
Services fail to start in strict mode

loren:
- Per Mohit's suggestion, migrated the initService() piece inside run() so that we do the initialization on thread run after the certs are installed vs on object creation. Now only DefaultService calls initService(), not in subclasses like KafkaService/CassandraService.
- Removed try-catch around Throwable in initService(). If something throws an exception in there, we should catch the most specific one we can.
- Removed unneeded ElasticService class


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -58,47 +45,36 @@
...
 /**
  * This scheduler when provided with a ServiceSpec will deploy the service and recover from encountered faults
  * when possible.  Changes to the ServiceSpec will result in rolling configuration updates, or the creation of
  * new Tasks where applicable.
  */
-public class DefaultScheduler implements Scheduler, Observer {
+public class DefaultScheduler extends AbstractScheduler implements Observer {
 
     /**
      * Time to wait for the executor thread to terminate. Only used by unit tests.
```
loren: sorry, what is private now?

benclarkwood: Ah. I should have appended my comment one line lower. The method went from protected -> private.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -58,47 +45,36 @@
...
-    // master re-election. Avoid performing initialization multiple times, which would cause resourcesQueue to be stuck.
-    protected final AtomicBoolean isAlreadyRegistered = new AtomicBoolean(false);
-
     protected final ServiceSpec serviceSpec;
     protected final SchedulerFlags schedulerFlags;
     protected final Collection<Plan> plans;
-    protected final StateStore stateStore;
     protected final ConfigStore<ServiceSpec> configStore;
-    protected final Optional<RecoveryPlanOverriderFactory> recoveryPlanOverriderFactory;
+    final Optional<RecoveryPlanOverriderFactory> recoveryPlanOverriderFactory;
```
mohitsoni: Why does this have `default` access specifier ?

PR #1035, opened 2017-05-30
[INFINITY-1366] Changed services to run as 'nobody' by default instead of 'root'

NimaVaziri:
Confirmed UID of files in tasks is 'nobody'


benclarkwood:
Once the DC/OS bug is fixed, I will be curious to see if our tests pass running as `nobody` or not

benclarkwood:
retest this please

paegun:
@NimaVaziri, @benclarkwood this change only sets the default, but several of the `frameworks/$framework/src/main/dist/svc.yml` are not setting user at the scheduler-, let alone pod-level.

The following is a list of frameworks and how they differ from what I believe is the expectation, that the service allows for override of the default via install options as well as configuration edit:

| framework | scheduler-user | pod-user |
|------------|----------------|-----------|
| cassandra | ✅  | ❌ |
| dse | ❌ | ❌ |
| elastic | ✅  | ❌ |
| hdfs | ❌ | ❌ |
| helloworld | ✅ | ✅ |
| kafka | ✅ | ❌ |
| confluent | ✅ | ❌ |
|template | ✅ | ✅ |


*) Additionally, some of the frameworks use `{{SERVICE_USER}}` and `{{SERVICE_PRINCIPAL}}` while others use `{{FRAMEWORK_USER}}` and `{{FRAMEWORK_PRINCIPAL}}` . I believe while touching all of the same files that would be required to make a change to standardize on either SERVICE_ or FRAMEWORK_ , barring bike shedding, we should make the change to standardize on one, which afaik is SERVICE_ . TBH I can't determine if we should use FRAMEWORK_ or SERVICE_ as the count of each is near even and helloworld uses SERVICE_ while template uses FRAMEWORK_ .

NimaVaziri:
@paegun yeah you're correct in that the user needs to be set at several more places to match expected behavior. The first commit was just to initialize the PR :)

Re: consistency issue - IMO it'd make more sense to standardize on `SERVICE_` since the docs/guide refer to them as services (in which case we should probably also rename the `frameworks` folder to `services`)

NimaVaziri:
All the scheduler and pod users have been set now, meaning executors and schedulers running permissive and strict mode, with the exception of the scheduler in permissive mode: in this case the scheduler seems to be not able to be ran as `nobody` despite setting the service's user in the config and updating the scheduler code to provide this info in the right protobuf.

AFAICT [this protobuf](https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L218) indicates that only the executors can run as the given `user` so strict mode probably has the default scheduler user hardwired to `nobody`.

NimaVaziri:
retest please

benclarkwood:
@NimaVaziri Can you please clean up conflicts? CI should be more stable now.

NimaVaziri:
Yup merge resolution coming up

benclarkwood:
retest this please

NimaVaziri:
retest this please

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -1,9 +1,11 @@
 name: {{FRAMEWORK_NAME}}
 scheduler:
   principal: {{SERVICE_PRINCIPAL}}
+  user: {{TASKCFG_ALL_TASK_USER}}
```
NimaVaziri: Ah alright, in that case I'll remove the wiring I setup in this PR to inject this into the FrameworkInfo protobuf then.

benclarkwood: Oh... I didn't notice that. Before you do that, @nickbp @gabrielhartmann is that required to make strict work?

NimaVaziri: AFAIK it's not required

NimaVaziri:
Closed in favor of splitting this on other service-specific branches.

PR #1055, opened 2017-06-01
Fix compile error from double-merge

loren:


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -141,29 +155,21 @@ private boolean allButStateStoreUninstalled() {
...
 
-    protected ServiceSpec getServiceSpec() {
-        return this.schedulerBuilder.getServiceSpec();
+    private boolean tasksNeedClearing() {
+        return ResourceCollectionUtils.getResourceIds(
+                ResourceCollectionUtils.getAllResources(stateStore.fetchTasks())).stream()
+                .allMatch(resourceId -> resourceId.startsWith(Constants.TOMBSTONE_MARKER));
     }
 
-    public static Boolean serviceSpecRequestsGpuResources(ServiceSpec serviceSpec) {
```
ArtRand: Did you remove this function? We need to keep it for the GPU opt-in policy.

PR #1057, opened 2017-06-01
Improvements to CLI messaging

ssk2:
These are some improvements to CLI messaging that were suggested by @keithchambers last week.

### DIFF
--- a/cli/commands/update.go
+++ b/cli/commands/update.go
@@ -11,7 +11,6 @@ import (
 )
 
 type DescribeHandler struct {
-	DescribeName string
 }
 
 type DescribeRequest struct {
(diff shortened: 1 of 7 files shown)


cli/commands/update.go
```diff
@@ -97,6 +98,7 @@ func parseUpdateResponse(responseBytes []byte) (string, error) {
 }
 
 func doUpdate(optionsFile, packageVersion string) {
+	config.Command = "update"
```
benclarkwood: This seems like it has weird side effects. Config is thrown away after the command completes?

ssk2: @benclarkwood yup - it's in memory. I'll move this to the actual handler method instead so it's only ever set upon invocations of that command.

ssk2:
@benclarkwood: amended. I removed a couple of fields that weren't being used, and used the actual command being passed through by Kingpin. `config.Command` is now set in the callback methods for `describe` and `handle` respectively.

ssk2:
retest this please

PR #1060, opened 2017-06-01
[INFINITY-1658] Add newly-required 1.10 ACLs

jsrodman:


jsrodman:
retest this please

tools/setup_permissions.sh
```diff
@@ -46,6 +46,15 @@ function grant_task_execution() {
          -H "Authorization: token=${ACS_TOKEN}" \
          "${DCOS_URL}/acs/api/v1/acls/dcos:mesos:master:task:user:${LINUX_USER}/users/${SERVICE_ACCOUNT_NAME}/create"
 
+    # XXX 1.10 curerrently requires this mesos:agent permission as well as
+    # mesos:task permission.  unclear if this will be ongoing requirement.
+    # See DCOS-15682
+    curl -k -L -X PUT \
+         -H "Authorization: token=${ACS_TOKEN}" \
+         "${DCOS_URL}/acs/api/v1/acls/dcos:mesos:agent:task:user:${LINUX_USER}" \
```
paegun: +1 on this being a new 1.10 requirement.

I believe we are missing a similar curl for
"${DCOS_URL}/acs/api/v1/acls/dcos:mesos:agent:task:user:${LINUX_USER}/users/${framework_name}-principal"
where ${framework_name} is the framework part of the name, i.e. 'kafka' in 'kafka-principal'.

tools/setup_permissions.sh
```diff
@@ -56,6 +65,12 @@ function grant_task_execution() {
          "${DCOS_URL}/acs/api/v1/acls/dcos:mesos:master:task:user:root" \
          -d '{"description":"Execute Mesos tasks as user=root"}' \
          -H 'Content-Type: application/json'
+    # XXX see above
+    curl -k -L -X PUT \
+         -H "Authorization: token=${ACS_TOKEN}" \
+         "${DCOS_URL}/acs/api/v1/acls/dcos:mesos:agent:task:user:root" \
```
paegun: I believe this is wrong. We should only need to grant the SERVICE_USER, FRAMEWORK_USER, or LINUX_USER (all the same thing semantically, but named different things throughout). We should not need to grant to 'root' unless the scheduler or pods are running as 'root'.

jsrodman: I think you're suggesting lines 49 - 58 were already wrong.

paegun: Yes, previous lines 49-58 were wrong. wrt mesos task ace's, the frameworks need the following:

| ace | purpose |
|-----|---------|
| dcos:mesos:master:task:user:${LINUX_USER} | may launch the scheduler on a master |
| dcos:mesos:agent:task:user:${LINUX_USER} | may launch the scheduler or pod member on an agent |

For each of ^^, there also needs to be a ".../users/${FRAMEWORK_NAME}-principal" entry so the (dc/os concept) service account can perform the action.

tools/setup_permissions.sh
jsrodman: We use ${SERVICE_ACCOUNT_NAME} here, not ${framework_name}-principal.  I'm not excited about shifting to ${framework_name}-principal because it's more work, and using a non-default name has found several bugs.

I've gotten some tests to pretty much work with the acl diffs here.  Can you explain what your additional acl change represents and why we need it, or reference something that does?

paegun: I didn't mean to explicitly use ${framework_name}, the point I was making is that this script needs to also add the "similar curl" to grant the access control action to the service account (...-principal).

jsrodman: Reading through the implementation of the code behind the ACL crap, it seems there is the possibility to create an ACL in abstract that applies to *nothing*.  (Perhaps it applies to everything?  Who knows.)

Thus the call I added creates this new ACL (if it doesn't already exist).  An ACL that says that tasks can be run as the user in question.  However, this ACL isn't applying to the service account specifically, as was done in the prior lines.  Empirically, this gave a sitaution where frameworks worked.  However, I have no idea how the permission system actually works, since there seem to be no real docs of any kind so I should at least keep in parallel with what went before.

PR #1061, opened 2017-06-01
Add --package-versions command

ssk2:
I still need to verify this against actual Cosmos output but there's a bug that's blocking testing. Would like to get your feedback on the implementation in any case.


ssk2:
retest this please

cli/commands/update.go
```diff
@@ -138,6 +158,7 @@ func HandleUpdate(app *kingpin.Application) {
 	updateCmd := &UpdateHandler{}
 	update := app.Command("update", "Update the package version or configuration for this DC/OS service").Action(updateCmd.UpdateConfiguration)
 	update.Flag("options", "Path to a JSON file that contains customized package installation options").StringVar(&updateCmd.OptionsFile)
-	update.Flag("package-version", "The desired package version").StringVar(&updateCmd.PackageVersion)
-	update.Flag("status", "View status of this update").BoolVar(&updateCmd.Status)
+	update.Flag("package-version", "The desired package version to update to").StringVar(&updateCmd.PackageVersion)
+	update.Flag("package-versions", "View a list of available package versions to downgrade or upgrade to").BoolVar(&updateCmd.ViewPackageVersions)
```
nickbp: this list operation feels like it should be its own command rather than an argument against `update`

ssk2:
Cherry picked these commits onto #1084.

PR #1065, opened 2017-06-02
Revert buggy url parsing patch

nlsun:
My previous patch actually broke it.


nlsun:
@ssk2 we just discovered that this actually breaks it because the expected url is minus one `/`. Right after I pinged to merge too... Sorry about this.

nlsun:
hold on, still buggy

nlsun:
ok, thinking about it more, the original behavior was probably the more correct one, as a url with both "/" and "/#/" will get trimmed.

It may actually be better to just revert my previous patch.

Also, an unsolicited suggestion, is to not be modifying global state in `"github.com/mesosphere/dcos-commons/cli/config"`

benclarkwood:
@nlsun Yes please submit a reversion patch.

And yes, I agree we do weird things to the config state. CLI code is in a state of flux currently.

nlsun:
Changed to a git revert

benclarkwood:
retest this please

PR #1068, opened 2017-06-02
[INFINITY-1683] [INFINITY-1692] Fix Uninstall Issues

benclarkwood:
Executor Resource Uninstall Issues:
  - When resources are at executor level, they are shared across tasks.
  - Only create one step per resource, despite the resource showing up in multiple tasks.
  - Test this case.

Frameworks left behind as inactive:
  - DeregisterStep was calling `shutdown` with the wrong value for `failover`

Enable uninstall for frameworks
  - Enable uninstall for the frameworks by adding the capability label

Update Test Code
  - Remove sdk_spin.py that was left behind in an earlier refactor.
  - Update our `uninstall` convenience function to understand new uninstall.


benclarkwood:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
```diff
@@ -73,16 +73,13 @@ private Plan getPlan() {
...
-            if (!resourceId.isPresent()) {
-                continue;
-            }
-            Status status = resourceId.get().startsWith(TOMBSTONE_MARKER) ? Status.COMPLETE : Status.PENDING;
-            taskSteps.add(new UninstallStep(resourceId.get(), status));
-        }
-
+        // We filter to unique Resource Id's, because Executor level resources are tracked
+        // on multiple Tasks.
+        List<Protos.Resource> allResources = ResourceCollectionUtils.getAllResources(stateStore.fetchTasks());
```
nickbp: Maybe add an example of what exactly it's going for here, like:

```
Given this scenario:
- Task 1: resource A, resource B
- Task 2: resource A, resource C
<Description of what we want to do...>
```

loren:
retest this please

elastic's `test_bump_node_counts` failed and cluster was somehow torn down.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
loren: @benclarkwood added a unit test `testInitialPlanTaskResourceOverlap` around that scenario. I'll add in your comments so the intent is clearer in the code.

loren:
Retest this please

CI failed [here](https://github.com/mesosphere/dcos-commons/blob/15640d45be4b59413ea3e636df098c6aeb91d51f/frameworks/cassandra/tests/test_maintenance_plans.py#L20-L20) in cassandra setup in `test_maintenance_plans.py`:
```
>       launch_and_verify_job(WRITE_DATA_JOB)
```
These tests were written weeks ago, so it may just be another unrelated failure.

loren:
Retest this please

Hello-world failed on uninstall in test_marathon_constraint

nickbp:
Test failure against the added metronome cleanup. Apparently it throws if the job in question doesn't currently exist. remove_cassandra_jobs() could catch that?

loren:
@nickbp Good point. I handle that case now.

nickbp:
retest this please (ccm)

loren:
retest this please

kafka failed in `test_canary_init`

loren:
retest this please

accidentally killed wrong CI job

benclarkwood:
@loren @nickbp Turns out I didn't fix spit. I did find the real bug, however!

We were pulling task + executor resources when building the plan (and creating some dupes) but we weren't pulling task + executor when we marked things as finished.

benclarkwood:
Apparently my hubris was unfounded.

benclarkwood:
Given that I am fixing a bug in Cosmos' handling of the uninstall, it is totally possible that bug is causing us pain here.

nickbp:
the `KeyError uris` failure is also being seen in other PRs. likely a behavior change in marathon

```
app_name = 'elastic'

    def get_config(app_name):
        def fn():
            return sdk_cmd.request('get', api_url('apps/{}'.format(app_name)), retry=False)
    
        config = sdk_spin.time_wait_return(lambda: fn()).json()['app']
        # The configuration JSON that marathon returns doesn't match the configuration JSON it accepts,
        # so we have to remove some offending fields to make it re-submittable, since it's not possible to
        # submit a partial config with only the desired fields changed.
>       del config['uris']
E       KeyError: 'uris'

testing/sdk_marathon.py:17: KeyError
```

benclarkwood:
retest this please

benclarkwood:
retest this please

benclarkwood:
There is a known issue with Cosmos. That has been corrected upstream in Cosmos. When PR https://github.com/mesosphere/dcos-enterprise/pull/1036 catches a DC/OS train we'll be good to go.

benclarkwood:
Cosmos change was merged into DC/OS. Retrying.

benclarkwood:
Turns out the cosmos changes didn't actually make it in (Cosmos build process broke down). I expect this to keep failing intermittently.

nickbp:
retest this please (ccm flake again)

benclarkwood:
retest this please

benclarkwood:
Cosmos changes made it into dcos-ee.

PR #1072, opened 2017-06-02
Make Cassandra "auto_snapshot" option configurable

susanxhuynh:
Before, the Cassandra "auto_snapshot" setting was set to "true". This PR makes the setting configurable, with a default setting of "false".

### DIFF
--- a/frameworks/cassandra/README.md
+++ b/frameworks/cassandra/README.md
@@ -238,6 +238,7 @@ You can configure most of the settings exposed in Apache Cassandra's `cassandra.
 *   `dynamic_snitch_update_interval_in_ms`
 *   `dynamic_snitch_reset_interval_in_ms`
 *   `dynamic_snitch_badness_threshold`
+*   `auto_snapshot`
 *   `roles_update_interval_in_ms`
 *   `permissions_update_interval_in_ms`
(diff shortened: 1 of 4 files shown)


mohitsoni:
I think `auto_snapshot` should be set to `false` by default. Unless we figure out a way to actually do something with them. Also, AFAIK these snapshots are not deleted automatically, they need to be deleted manually.

susanxhuynh:
@mohitsoni Thanks, would you mind discussing the default value in the JIRA? I just cc'ed you.

susanxhuynh:
retest this please

012345:
This is exposed and configurable in beta_dse. This should remain set as true by default. This is ONLY referenced when tables are dropped or truncated and prevents unintentional data loss. This configuration can be changed post deployment and a rolling restart of the service will put it into effect.

auto_snapshot 
(Default: true) Enable or disable whether a snapshot is taken of the data before keyspace truncation or dropping of tables. To prevent data loss, using the default setting is strongly advised. If you set to false, you will lose data on truncation or drop.

susanxhuynh:
This change does not affect beta_dse. It only affects beta-cassandra, and the default is `false` for now because beta-cassandra does not provide a way to clear out automatic snapshots.

PR #1075, opened 2017-06-02
Delete directory in new-framework.sh that causes build.sh to fail

anita-tenjarla:
Delete vendor/gopkg/alecthomas directory in new-framework.sh script that will cause build.sh on newly created framework to fail


anita-tenjarla:
https://github.com/mesosphere/dcos-commons/pull/1081

PR #1082, opened 2017-06-05
Added API to retrieve IP on a CNI network.

asridharan:
This change is [Reviewable](https://reviewable.io/reviews/mesosphere/dcos-commons/1082)

### DIFF
--- a/sdk/bootstrap/main.go
+++ b/sdk/bootstrap/main.go
@@ -5,8 +5,8 @@ import (
 	"flag"
 	"fmt"
 	// TODO switch to upstream once https://github.com/hoisie/mustache/pull/57 is merged:
-	"github.com/nickbp/mustache"
 	"github.com/aryann/difflib"
+	"github.com/nickbp/mustache"
 	"io/ioutil"
 	"log"
 	"net"
@@ -14,6 +14,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"pkg/mesos"
 	"sort"
 	"strings"
 	"time"
@@ -23,9 +24,9 @@ import (
 
 const (
 	configTemplatePrefix = "CONFIG_TEMPLATE_"
-	resolveRetryDelay = time.Duration(1) * time.Second
-    dns_tld = "autoip.dcos.thisdcos.directory"
-    mesos_dns = "mesos"
+	resolveRetryDelay    = time.Duration(1) * time.Second
+	dns_tld              = "autoip.dcos.thisdcos.directory"
+	mesos_dns            = "mesos"
 )
 
 var verbose = false
@@ -49,8 +50,8 @@ type args struct {
 	// Install certs from .ssl into JRE/lib/security/cacerts
 	installCerts bool
 
-    // Get Task IP
-    getTaskIp bool
+	// Get Task IP
+	getTaskIp bool
 }
 
 func parseArgs() args {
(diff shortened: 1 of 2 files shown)


asridharan:
@ArtRand can you take a look at this?

sdk/bootstrap/main.go
```diff
@@ -5,15 +5,16 @@ import (
 	"flag"
 	"fmt"
 	// TODO switch to upstream once https://github.com/hoisie/mustache/pull/57 is merged:
```
nickbp: should move the comment as well

sdk/bootstrap/main.go
```diff
@@ -330,48 +331,43 @@ func isFile(path string) (bool, error) {
...
+	return ip.String()
 }
+
 // main
 
 func main() {
 	args := parseArgs()
 
-    libprocess_ip, found := os.LookupEnv("LIBPROCESS_IP")
+	libprocess_ip, found := os.LookupEnv("LIBPROCESS_IP")
```
nickbp: there are now two fetches of LIBPROCESS_IP, one here and one in `ip.go`. should this one go away?

sdk/bootstrap/src/pkg/mesos/ip.go
```diff
@@ -0,0 +1,41 @@
...
+// This helper function gives the current container's IP address, when run in a
+// given network, mnt and UTS namespace.  The function first checks the presence of
+// `LIBPROCESS_IP` if present it will return this as the container's IP address
+// else it will look at the `hostname` setup for the container to pick out the
+// IP address associated with the hostname.
+func ContainerIP() (containerIP net.IP, err error) {
+	ip := os.Getenv("LIBPROCESS_IP")
+	if ip != "" {
+		containerIP = net.ParseIP(ip)
+		if containerIP == nil {
```
nickbp: what happens when LIBPROCESS_IP is `0.0.0.0`?

asridharan: good catch !! Should have added that case.

sdk/bootstrap/main.go
asridharan: Not sure @ArtRand ??

asridharan: Actually didn't realize the logic over here. Was thinking @ArtRand might need to fix it in his patch. Will fix it in this patch itself.

asridharan:
@ArtRand @nickbp just fixed @nickbp comments. Most of the changes are due to the formatting that my VIM plugin introduced, based on the standard go formatting template. So please don't be put of by the number of lines of change.

sdk/bootstrap/main.go
```diff
@@ -4,16 +4,16 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
-	// TODO switch to upstream once https://github.com/hoisie/mustache/pull/57 is merged:
-	"github.com/nickbp/mustache"
 	"github.com/aryann/difflib"
+	"github.com/nickbp/mustache"
```
nickbp: the TODO should probably stick around...

sdk/bootstrap/main.go
```diff
@@ -329,49 +329,37 @@ func isFile(path string) (bool, error) {
...
+	if err != nil {
+		log.Fatalf("Cannot find the container's IP address: ", err)
+	}
 
-        err := os.Setenv("LIBPROCESS_IP", libprocess_ip)
-        if err != nil {
-            log.Fatalf("Failed to SET new LIBPROCESS_IP")
-        }
-    }
+	err = os.Setenv("LIBPROCESS_IP", libprocess_ip)
```
nickbp: Is this only going to apply to child processes of `bootstrap`? In practice bootstrap is run like this:

```
./bootstrap --foo && echo "other stuff here..."
```

sdk/bootstrap/main.go
```diff
@@ -329,49 +329,37 @@ func isFile(path string) (bool, error) {
...
 
 func main() {
 	args := parseArgs()
 
-    libprocess_ip, found := os.LookupEnv("LIBPROCESS_IP")
-
-    if !found {
-        log.Fatalf("Cannot find LIBPROCESS_IP")
-    }
+	libprocess_ip, err := GetLocalIP()
```
nickbp: Should this block only run when getTaskIp is enabled?

sdk/bootstrap/main.go
```diff
@@ -389,9 +377,9 @@ func main() {
 		log.Printf("Template handling disabled via -template=false: Skipping any config templates")
 	}
 
-	if (args.installCerts) {
+	if args.installCerts {
 		installDCOSCertIntoJRE()
 	}
-    log.Printf("Local IP --> %s", GetLocalIP())
+	log.Printf("Local IP --> %s", libprocess_ip)
```
nickbp: I think this could be omitted in favor of the handling within getTaskIp

nickbp:
I'm assuming the goal would be to allow upstream to do something like:

```
export MY_IP=$(bootstrap --get-ip)
```

nickbp:
FYI build is currently failing with:

```
Creating symlink from GOPATH=/root/go/src/github.com/mesosphere/dcos-commons to REPOPATH=/jenkins/workspace/dcos-commons/inf-pr
package pkg/mesos: unrecognized import path "pkg/mesos" (import path does not begin with hostname)
```

PR #1087, opened 2017-06-06
Remove modify_master.

benclarkwood:


jsrodman:
there's a bunch more simplification this opens for in configure_test_cluster, but that seems like another change.

PR #1105, opened 2017-06-09
[HDFS-462] Add readiness check to journal nodes

NimaVaziri:
The deployment plan now waits for the journal nodes to succeed their readiness check before proceeding to deploying the name nodes.

### DIFF
--- a/frameworks/hdfs/src/main/dist/svc.yml
+++ b/frameworks/hdfs/src/main/dist/svc.yml
@@ -49,6 +49,13 @@ pods:
           {{/TLS_ENABLED}}
         env:
           JOURNALNODE: true
+        readiness-check:
+          cmd: >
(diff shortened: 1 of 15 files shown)


gabrielhartmann:
@NimaVaziri:  We should write or update an integration test which does an update of HDFS and validate that the recovery plan never leaves it's COMPLETE state.

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -49,6 +49,13 @@ pods:
           {{/TLS_ENABLED}}
         env:
           JOURNALNODE: true
+        readiness-check:
+          cmd: >
+              READY=$(grep "IPC Server listener on {{TASKCFG_ALL_JOURNAL_NODE_RPC_PORT}}: starting" stderr| wc -l) &&
+              [ $READY -ge 1 ]
+          interval: 10
+          delay: 60
```
gabrielhartmann: delay seems long, is delay required?

PR #1107, opened 2017-06-09
INFINITY-1736 Fix tests

nickbp:
Rips out some things that have since been moved upstream into shakedown, and cleans up some other things to hopefully be a bit more maintainable and/or reliable.

- Cuts test runtime from ~3h41m to ~3h12m
- Drops `sdk_spin.py` in favor of upstreamed shakedown support. Switches all tests to use the latter.
- Shrinks `sdk_install.py` to reflect logic that has been upstreamed to shakedown.
- Restructures and "upstreams" job handling support from cassandra tests into `sdk_jobs.py`.
- Restructures calls in `sdk_plans.py`, adds logging including detailed plan status:
```
[25.7s/15m] spinning...
Waiting for deploy plan to have COMPLETE status:
deploy STARTING:
- node-deploy STARTING: node-0:[server]: COMPLETE, node-1:[server]: COMPLETE, node-2:[server]: STARTING
```
- Make noisy things like marathon config updates less noisy: Just print the new `env` instead of the whole job.
- Moves some tests for common SDK features from per-service tests into hello-world tests.
- Consolidates tests spanning multiple files in cases where the service doesn't really change (avoid unnecessary reinstall during test runs).
- Updates some cases of DIY in per-service tests with `sdk_x.py` equivalents.
- Adds way to disable S3/Azure backup tests, printing out instructions when running locally without required credentials.
- Other misc fixes to _many_ test flakes encountered while doing the above...


frameworks/cassandra/tests/jobs/verify-node-replace.json.template
```diff
@@ -2,15 +2,15 @@
   "description": "Cassandra node replace test verification",
   "id": "test.cassandra.verify-node-replace",
   "run": {
-    "cmd": "! cqlsh --no-color -e 'select peer from system.peers;' node-0-server.cassandra.autoip.dcos.thisdcos.directory 9042 | grep '{{NODE_IP}}' && [ 2 = $(cqlsh --no-color -e 'select peer from system.peers;' node-0-server.cassandra.autoip.dcos.thisdcos.directory 9042 | grep -E '^\\s+[0-9]' | wc -l | awk '{print $1}') ]",
+    "cmd": "cqlsh --no-color -e 'select peer from system.peers;' node-0-server.cassandra.autoip.dcos.thisdcos.directory 9042 > peers.txt; cat peers.txt; echo CHECKING FOR LACK OF {{NODE_ADDRESS}}; [ 0 = $(grep '{{NODE_ADDRESS}}' peers.txt | wc -l) ] && [ 2 = $(grep -E '[0-9.]+$' peers.txt | wc -l) ]",
```
gabrielhartmann: Would be good to move to `cqlsh -e 'select json ...'` soonish.  A lot of the grepping could be dropped in favor of better CQL queries and JSON parsing.

PR #1108, opened 2017-06-12
[ELASTIC-93] Release Elastic 1.0.13-5.4.1-beta

loren:


nickbp:
Assuming this is to release an elastic hotfix, I think one of the following options would make more sense. The choice to go with depends on whether recent SDK changes in the elastic release are acceptable:

- Backport the elastic hotfix to the prior elastic release: check out the last elastic release's tag, make a commit against that, and release against that tag (effectively without touching master)
- Release a new cut of elastic against SDK master

benclarkwood:
We're going to go with:
1) Cut a new release of the SDK
2) Release Elastic on that

PR #1112, opened 2017-06-13
[INFINITY-1659] enable run-all-tests

jsrodman:
This amounted to a *lot* less work than expected.

Unexpectedly (to me), almost all of our tests were already intended to perform cleanup (framework uninstall) on every test action.

Thus, this consists of:
* Add cleanup for the one test file that didn't have it.
* When test.py is run with --continue-on-error, invoke pytest without the --exitfirst flag.


jsrodman:
I'm kind of shocked at the idea that CI passed the first time.

PR #1114, opened 2017-06-13
[INFINITY-1762] Resource Refinement

gabrielhartmann:
Includes resource refinement / consumption of statically reserved resources, but with the capability turned off, pending merge of Mesos bits.


sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/Capabilities.java
```diff
@@ -37,6 +56,10 @@ public boolean supportsCNINetworking() throws IOException {
         return hasOrExceedsVersion(1, 9);
     }
 
+    public boolean supportsPreReservedResources() {
+        return false;
```
nickbp: Should this be checking 1.10? If it's still not in DCOS by default, could put it behind SchedulerFlags/an envvar to allow testing with the same build.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/EvaluationOutcome.java
```diff
@@ -24,6 +22,7 @@
 
     private final Type type;
     private final String source;
+    private final MesosResource mesosResource;
```
nickbp: How about giving EvaluationOutcome a Builder?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -87,7 +87,6 @@ public void frameworkMessage(SchedulerDriver driver,
     @Override
     public void disconnected(SchedulerDriver driver) {
         LOGGER.error("Disconnected from Master.");
-        SchedulerUtils.hardExit(SchedulerErrorCode.DISCONNECTED);
```
nickbp: Intentional removal?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
```diff
@@ -65,25 +62,37 @@ public DefaultPodSpec(
             @JsonProperty("task-specs") List<TaskSpec> tasks,
             @JsonProperty("placement-rule") PlacementRule placementRule,
             @JsonProperty("volumes") Collection<VolumeSpec> volumes,
+            @JsonProperty("pre-reserved-role") String preReservedRole,
```
nickbp: `prereserved-role`? `reserved-role`? having a dash between pre and reserved makes it look like pre is a word

(here and several cases below)

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: should be in a different PR

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/Capabilities.java
gabrielhartmann: I plan to have another PR, with the real released Mesos bits and at that point we'll determine what the best way is to determine whether support is enabled.  Not sure what Mesos has planned yet.  Is it a capability on Mesos endpoint or is it just based on DC/OS version?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
gabrielhartmann: `prereserved` isn't a word.  `reserved-role` is what `role` is.

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/Capabilities.java
gabrielhartmann: Mesos doesn't have a way expose capabilities.  I'll be based on version number.  Will update this when they PRs in DC/OS and Mesos are merged.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/EvaluationOutcome.java
gabrielhartmann: Happy to in another PR.  This is adding a single field.  I like the two `pass` and the one `fail` method.  Reads better in the code I think. I get that the `null` arguments in some places isn't great.  What would a builder alternative look like though?

```
return new Evaluation.Builder(this, true, "Reason of thing because %s, %s", reasonA, reasonB)
    .setMesosResource(mesosResource)
    .build();
```
vs
```
return pass(this,  null, "Reason of thing because %s, %s", reasonA, reasonB);
```

Here.  I'm adopting the rule, which we should follow everywhere that a builder shouldn't be constructable unless an immediate `build()` call generates a valid object.  We violate this rule all over the place, but I'd like to stop.

gabrielhartmann:
retest this please

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/EvaluationOutcome.java
nickbp: No rush just a thought, feel free to skip here. I'd just keep the current short names but have them return a builder:

common case: `[EvaluationOutcome.]pass(this, "description").build()`
additional data: `[EvaluationOutcome.]pass(this, "description").setMesosResource(mesosResource).build()`

gabrielhartmann: I like this.  I'll put it in a separate PR.  Really want to get this in.  It's blocking the PortsSpec bug fix, and I want to unblock @paegun, ASAP as well.

gabrielhartmann: https://jira.mesosphere.com/browse/INFINITY-1769

PR #1115, opened 2017-06-13
[CASSANDRA-623] Fixed cassandra.yaml "listen_address" to use LIBPROCESS_IP

susanxhuynh:
The "listen_address" in cassandra.yaml was previously set to "{{TASK_NAME}}.{{SERVICE_NAME}}.autoip.dcos.thisdcos.directory". However this dns name can sometimes be invalid during a "replace" operation, since it may still be pointing to the original task host IP.
This PR changes the "listen_address" to be based on LIBPROCESS_IP instead.


susanxhuynh:
retest this please

PR #1116, opened 2017-06-14
[HDFS-450] Add HDFS upgrade-downgrade integration test

susanxhuynh:
A test that does upgrade / downgrade between the currently released version of beta-hdfs and the version in master branch.


susanxhuynh:
Will re-open if/when upgrades are fixed.

NimaVaziri:
@susanxhuynh mind taking a look at this updated branch? Can't tag you as a reviewer since you authored the PR

NimaVaziri:
The majority of the `foo` => `sdk_foo` change was merged in https://github.com/mesosphere/dcos-commons/pull/1267 though the remaining portion is cleaned up here as an extension of the `sdk_test_upgrade` => `sdk_versions` modification.

frameworks/hdfs/tests/test_upgrade_downgrade.py
```diff
@@ -0,0 +1,36 @@
...
+
+import sdk_versions
+from tests.config import *
+
+@pytest.mark.upgrade
+@pytest.mark.sanity
+def test_upgrade():
+    options = {
+        "service": {
+            "name": "beta-{}".format(PACKAGE_NAME),
```
NimaVaziri: I'll rerun it to confirm but I remember when I actually ran the test without `beta-{}` there ended up being a discrepancy between service names so the polling for deployment and recovery would be polling using the wrong name and as a result would timeout.

NimaVaziri: Nevermind, confirmed the beta format isn't need. Reverted to just package name.

susanxhuynh: Cool, thanks, @NimaVaziri

NimaVaziri:
Closing this as the upgrade itself is handled in @nickbp's [upgrade PR](https://github.com/mesosphere/dcos-commons/pull/1339) and the journal node readiness check is handled in https://github.com/mesosphere/dcos-commons/pull/1351

PR #1119, opened 2017-06-14
KAF-353 - Add Kafka Readiness-Check, Using server.log

paegun:
* added kafka readiness-check, ensuring that the starting and started log lines are in the kafka server.log log file.
  * the majority of ways to check for readiness are written from the kafka client perspective. the use of the server.log file is IMO much preferable since the check is executing on the server, so should just bubble up to DC/OS the state of kafka readiness that was written to the server.log file.

### DIFF
--- a/frameworks/kafka/src/main/dist/svc.yml
+++ b/frameworks/kafka/src/main/dist/svc.yml
@@ -52,7 +52,7 @@ pods:
             # [2017-06-14 22:20:55,464] INFO [Kafka Server 1], started (kafka.server.KafkaServer)
 
             kafka_dir=$(ls -d kafka_* |head -n 1)
-            kafka_server_log_file=${kafka_dir}/log/server.log
+            kafka_server_log_file=${kafka_dir}/logs/server.log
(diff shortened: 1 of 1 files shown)


gabrielhartmann:
@paegun:  Could you please provide some example log snippets illustrating what this check should be detecting?

frameworks/kafka/src/main/dist/svc.yml
```diff
@@ -39,6 +39,27 @@ pods:
           server-properties:
             template: "{{CONFIG_TEMPLATE_PATH}}/server.properties.mustache"
             dest: "{{KAFKA_VERSION_PATH}}/config/server.properties"
+        readiness-check:
+          cmd: |
+            # since the server is just started, the starting and started log lines
```
gabrielhartmann: I would put this in a script file and include it as part of the distribution instead of putting the whole thing in the YAML.  It doesn't require any templating.

benclarkwood: This is ~20 lines of bash (~ 1/2 of it comments).

Why is the overhead of adding it as an asset worth it?

paegun:
retest this please

PR #1121, opened 2017-06-15
INFINITY-1677 Implement folder support for cassandra/elastic/hdfs

nickbp:
Updates cassandra/elastic/hdfs to support foldered service names (i.e. names with slashes). Each service needed to be made folder-safe in e.g. hostname references. Adds integration tests for each service to validate that they continue to be folder-safe.

- Adds a new default `FRAMEWORK_HOST` envvar to tasks which can be used for intra-service hostnames, while also ensuring that those hostnames are slash-safe.
- Updates existing hostname references across all services to use `FRAMEWORK_HOST`.
- Updates zk references to be folder-safe as well.
- Adds mode to `TemplateUtils` to throw when template values are missing, rather than silently making them empty. Mainly for HDFS' endpoint template rendering (avoid recurrences of HDFS-461), but may also be useful elsewhere.
- Updates docs to reflect the new `FRAMEWORK_HOST` envvar and other docs cleanup.


frameworks/cassandra/tests/config.py
```diff
@@ -2,32 +2,132 @@
...
+import sdk_utils as utils
 
 
 PACKAGE_NAME = 'cassandra'
+FOLDERED_SERVICE_NAME = '/test/integration/' + PACKAGE_NAME
+FOLDERED_SERVICE_AUTOIP_HOST = FOLDERED_SERVICE_NAME.replace('/', '') + '.autoip.dcos.thisdcos.directory'
+FOLDERED_SERVICE_VIP_HOST = FOLDERED_SERVICE_NAME.replace('/', '') + '.l4lb.thisdcos.directory'
 DEFAULT_TASK_COUNT = 3
 
+DEFAULT_NODE_ADDRESS = os.getenv('CASSANDRA_NODE_ADDRESS', 'node-0-server.cassandra.autoip.dcos.thisdcos.directory')
```
ArtRand: just an idea, but why not write to a random node?

frameworks/elastic/tests/config.py
```diff
@@ -10,12 +10,16 @@
 import sdk_utils
 
 PACKAGE_NAME = 'elastic'
+FOLDERED_SERVICE_NAME = '/test/integration/' + PACKAGE_NAME
```
ArtRand: this is in two (at least?) tests now, factor it out into `sdk_folders.py` or the like?

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
```diff
@@ -43,25 +49,23 @@ public static void main(String[] args) throws Exception {
...
-    private static String getCoreSiteXml() {
-        return renderTemplate(System.getProperty("user.dir") + "/hdfs-scheduler/core-site.xml");
+                .setEndpointProducer(HDFS_SITE_XML,
+                        EndpointProducer.constant(renderTemplate(HDFS_SITE_XML, serviceSpec.getName())))
+                .setEndpointProducer(CORE_SITE_XML,
+                        EndpointProducer.constant(renderTemplate(CORE_SITE_XML, serviceSpec.getName())));
     }
 
-    private static String renderTemplate(String pathStr) {
+    private static String renderTemplate(String filename, String serviceName) {
```
ArtRand: 👍

frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -24,36 +26,68 @@
...
+    assert expect == found
 
 
 @pytest.mark.skip(reason="HDFS-451")
 @pytest.mark.data_integrity
 @pytest.mark.sanity
 def test_integrity_on_data_node_failure():
     shakedown.wait_for(
-        lambda: write_data_to_hdfs("data-0-node.hdfs.autoip.dcos.thisdcos.directory", TEST_FILE_1_NAME),
+        lambda: write_data_to_hdfs('data-0-node.{}'.format(FOLDERED_SERVICE_AUTOIP_HOST), TEST_FILE_1_NAME),
```
ArtRand: Looking into this, because I'm doing a lot of the same things in the overlay tests. this is an example of when we want to use the `.mesos` DNS because we want the host IP to run the command on. Otherwise, when you call `shakedown.run_command_on_agent(<container_ip>...)` it won't work.

frameworks/cassandra/tests/test_sanity.py
```diff
@@ -1,35 +1,44 @@
...
+    plan.wait_for_completed_deployment(FOLDERED_SERVICE_NAME)
 
     tmp_dir = tempfile.mkdtemp(prefix='cassandra-test')
     for job in TEST_JOBS:
         jobs.install_job(job, tmp_dir=tmp_dir)
 
 
 def teardown_module(module):
-    install.uninstall(PACKAGE_NAME)
+    #install.uninstall(FOLDERED_SERVICE_NAME, package_name=PACKAGE_NAME)
```
nickbp: It isn't, I left it that way by accident. Uncommented.

frameworks/cassandra/tests/test_sanity.py
```diff
@@ -62,20 +79,20 @@ def test_repair_plan_completes():
...
+# use e.g. "TEST_TYPES=speedy and not aws and not azure":
 
 
 @pytest.mark.aws
 @pytest.mark.sanity
 def test_backup_and_restore_to_s3():
     key_id = os.getenv('AWS_ACCESS_KEY_ID')
     if not key_id:
-        assert False, 'AWS credentials are required for this test. Disable test with e.g. TEST_TYPES="sanity and not aws"'
+        assert False, 'AWS credentials are required for this test. Disable test with e.g. TEST_TYPES="speedy and not aws"'
```
nickbp: That was just me accidentally leaving some local changes in the PR. Removed..

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointUtils.java
```diff
@@ -58,15 +58,26 @@ public static String toEndpoint(String hostname, int port) {
...
      */
-    public static String toAutoIpEndpoint(String serviceName, String taskName, int port) {
+    public static String toAutoIpDomain(String serviceName) {
         // Unlike with VIPs and mesos-dns hostnames, dots are converted to dashes with autoip hostnames. See DCOS-16086.
-        String hostname = String.format("%s.%s.%s",
-                removeSlashes(replaceDotsWithDashes(taskName)),
-                removeSlashes(replaceDotsWithDashes(serviceName)),
-                AUTOIP_HOST_TLD);
-        return toEndpoint(hostname, port);
+        return String.format("%s.%s", removeSlashes(replaceDotsWithDashes(serviceName)), AUTOIP_HOST_TLD);
```
nickbp: There isn't, removed the `EndpointUtils` versions in favor of the `Constants` versions

nickbp:
Added a `sdk_hosts` utility for autoip/vip/mesosdns hostnames. Switched tests to use the utility, and switched host system lookups to use the mesosdns host instead of the autoip host.

nickbp:
OK found the cause: cassandra's svc.yml was still using scheduler envvars for things which were only available as task envvars. Testing locally worked, lets see if CI catches anything else (particularly any typos from the earlier `sdk_hosts` refactoring), but should otherwise be good to go.

benclarkwood:
retest this please (Haven't seen this HDFS flake before, and Loren upped some Elastic timeouts)

nickbp:
The failures aren't flakes, in cb0d07b I had changed how hosts were being handled, and that broke something for both of them. Fixed elastic locally (so far, running through it now) and diagnosing what I broke with HDFS.

PR #1122, opened 2017-06-15
[INFINITY-1734] Fix Deprecation of PortsSpec

gabrielhartmann:
As part of the very big refactor associated with resource reservation work, PortsSpec stopped being used.  However, we can't just remove it since it breaks upgrade from older versions of the framework that still used PortsSpec

This work detects situations where PortsSpec was used with more than one port, UNRESERVEs all those ports, and makes it clear that they must be re-reserved all within the confines of a single OfferEvaluation cycle.

### DIFF
--- a/frameworks/kafka/tests/test_upgrade.py
+++ b/frameworks/kafka/tests/test_upgrade.py
@@ -30,7 +30,6 @@ def teardown_module(module):
 
 @pytest.mark.sanity
 @pytest.mark.smoke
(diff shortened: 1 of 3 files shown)


benclarkwood:
@gabrielhartmann Can you please add a description of the work?

gabrielhartmann:
@benclarkwood: done

nickbp:
retest this please (ccm)

PR #1127, opened 2017-06-15
[INFINITY-576] C*, kafka, hdfs, elastic, overlay network support and tests

ArtRand:
This PR contains 2 changes:
1. Changes to bootstrap. 
2. Tests for the 4 major data services on the overlay network.

Modified bootstrap to check for LIBPROCESS_IP == 0.0.0.0. When this is the case (when the pod is on the overlay network). It runs hostname (within bootstrap) to get the container IP and sets LIBPROCESS_IP to the container IP. With DefaultExecutor there will be a new environment variable that contains the correct IP, in that case we can disable this behavior or set LIBPROCESS_IP to that new envvar and keep the current behavior. 

Overlay tests for a service include the following:
1. Deploy the service and make sure that all tasks have the `dcos` network correctly in their `NetworkInfo`.
2. Make sure the endpoints for the service are correct. 
3. Write, read, and delete data into the service (not HDFS, see HDFS-451)

- [x] Cassandra overlay tests and documentation
- [x] Kafka overlay tests and documentation
- [x] HDFS overlay tests and documentation
- [x] Elastic overlay tests and documentation


ArtRand:
retest this please

ArtRand:
retest this please

ArtRand:
retest this please (ccm)

ArtRand:
retest this please

ArtRand:
retest this please (unable to repro)

ArtRand:
retest this please (marathon)

frameworks/cassandra/universe/config.json
```diff
@@ -89,7 +89,7 @@
         "placement_constraint": {
           "type": "string",
           "description": "The Marathon-style placement constraint string controlling node placement",
-          "default": ""
+          "default": "hostname:MAX_PER:1"
```
benclarkwood: @ArtRand What effect does this change have? Limit us to one C* node per agent, so they can't by default stack?

frameworks/hdfs/docs/limitations.md
```diff
@@ -0,0 +1,9 @@
+---
+post_title: Limitations
+menu_order: 50
+feature_maturity: preview
+enterprise: 'no'
+---
+# Overlay networks
+
+When elastic is deployed on the `dcos` overlay network, the configuration cannot be updated to use the host network.
```
susanxhuynh: Typo: s/elastic/hdfs/

frameworks/cassandra/universe/config.json
ArtRand: Exactly, before CNI we effectively enforced this constraint with the port resources. Now we don't consume them, so we need/want to enforce it with placement.

frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -10,17 +10,7 @@
...
-    DEFAULT_TASK_COUNT
-)
-
-TEST_CONTENT_SMALL = "This is some test data"
-# TODO: TEST_CONTENT_LARGE = Give a large file as input to the write/read commands...
-TEST_FILE_1_NAME = "test_1"
-TEST_FILE_2_NAME = "test_2"
-HDFS_CMD_TIMEOUT_SEC = 5 * 60
-HDFS_POD_TYPES = {"journal", "name", "data"}
+from tests.config import *
```
jsrodman: Nope.

frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -42,18 +32,18 @@ def teardown_module(module):
 @pytest.mark.sanity
 def test_integrity_on_data_node_failure():
     shakedown.wait_for(
-        lambda: write_data_to_hdfs("data-0-node.hdfs.autoip.dcos.thisdcos.directory", TEST_FILE_1_NAME),
+        lambda: write_data_to_hdfs("data-0-node.hdfs.mesos", TEST_FILE_1_NAME),
         timeout_seconds=HDFS_CMD_TIMEOUT_SEC)
 
     # gives chance for write to succeed and replication to occur
-    time.sleep(5)
+    time.sleep(9)
```
jsrodman: Looks sketchy.  is there no way to wait for the work to be accomplished?

frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -68,7 +58,7 @@ def test_integrity_on_name_node_failure():
     This test checks that it is possible to write and read data after the first name node fails.
     """
     tasks.kill_task_with_pattern("NameNode", 'name-0-node.hdfs.autoip.dcos.thisdcos.directory')
-    time.sleep(1)  # give NameNode a chance to die
+    time.sleep(9)  # give NameNode a chance to die
```
jsrodman: Same.

frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -318,62 +308,3 @@ def replace_name_node(index):
     tasks.check_tasks_updated(PACKAGE_NAME, name_node_name, name_id)
     tasks.check_tasks_not_updated(PACKAGE_NAME, 'journal', journal_ids)
     tasks.check_tasks_not_updated(PACKAGE_NAME, 'data', data_ids)
-
```
jsrodman: Where did all this code go?  Dead?

frameworks/helloworld/tests/test_executor_volumes.py
```diff
@@ -26,6 +26,7 @@ def teardown_module(module):
 
 
 @pytest.mark.sanity
+@pytest.mark.executor_volumes
```
jsrodman: What is this mark? Was this for testing, or do we want to keep it?  pytest can invoke a module, and this seems to be a mark for the module, so it seems redundant.

frameworks/kafka/tests/test_cni.py
```diff
@@ -1,59 +0,0 @@
-import pytest
```
jsrodman: This PR is adding support for CNI, but we're deleting test_cni.py?  What?

frameworks/kafka/tests/test_overlay.py
```diff
@@ -0,0 +1,88 @@
...
+        service_name=SERVICE_NAME,
+        additional_options = {'service':{'virtual_network':True}})
+    plan.wait_for_completed_deployment(PACKAGE_NAME)
+
+
+def teardown_module(module):
+    install.uninstall(SERVICE_NAME, PACKAGE_NAME)
+
+
+@pytest.mark.overlay
```
jsrodman: These probably shouldn't exist.

frameworks/kafka/tests/test_overlay.py
```diff
@@ -0,0 +1,88 @@
...
+@pytest.mark.smoke
+@pytest.mark.sanity
+def test_service_overlay_health():
```
jsrodman: I can't tell if these test are new, but if they are, please add a docstring saying what they attest to.

testing/sdk_networks.py
```diff
@@ -1,10 +1,18 @@
...
+
 import shakedown
 
 
 def check_task_network(task_name, expected_network_name="dcos"):
     """Tests whether a task (and it's parent pod) is on a given network
     """
     _task = shakedown.get_task(task_id=task_name, completed=False)
+
+    if type(_task) == list:
```
jsrodman: type testing is usually wrong.  does get_task actually return an object OR a list?

From the code:

```pythondef get_tasks(task_id='', completed=True):
    """ Get a list of tasks, optionally filtered by task id.

        :param task_id: task ID
        :type task_id: str
        :param completed: include completed tasks?
        :type completed: bool

        :return: a tuple of tasks
        :rtype: tuple
    """
```
If this is right, your code will never get a list.  

Probably your code should be 

tasks = shakedown.get_tasks(...)
assert len(tasks) == 1, ...

jsrodman: Was it always wrong to get multiple tasks for a task_name?

testing/sdk_networks.py
```diff
@@ -1,10 +1,18 @@
...
+        assert len(_task) == 1, "Found too many tasks matching {}, got {}"\
+            .format(task_name, _task)
+        _task = _task[0]
```
jsrodman: No longer relevant, but definitely don't have the same variable be both a tuple and a tuple entry.

testing/sdk_networks.py
```diff
@@ -19,3 +27,23 @@ def check_task_network(task_name, expected_network_name="dcos"):
                     assert "name" not in network_info, \
                         "Task {task} has network name when it shouldn't has status:{status}" \
                         .format(task=task_name, status=status)
+
+
+def get_endpoints(endpoint_to_get, package_name, correct_count):
+    endpoints, _, rc = shakedown.run_dcos_command("{} endpoints {}".format(package_name, endpoint_to_get))
+    assert rc == 0, "Failed to get endpoints on overlay network"
+    endpoints = json.loads(endpoints)
+    assert len(endpoints) == correct_count, "Wrong number of endpoints, got {} should be {}" \
```
jsrodman: If this is "get_()" why is it validating?

testing/sdk_networks.py
```diff
@@ -19,3 +27,23 @@ def check_task_network(task_name, expected_network_name="dcos"):
...
+        .format(len(endpoints), correct_count)
+    return endpoints
+
+
+def check_endpoints_on_overlay(endpoints):
+    ip_addresses = [e.split(":")[0] for e in endpoints["address"]]
```
jsrodman: My opinion is this is doing too much for one line.  It's nice when your python is so simpleminded, the line and exception tell you what's actually wrong.  Maybe it's less essential for test code, but I'm doubtful.

testing/sdk_networks.py
```diff
@@ -19,3 +27,23 @@ def check_task_network(task_name, expected_network_name="dcos"):
...
+    assert len(set(ip_addresses).intersection(set(shakedown.get_agents()))) == 0, \
+        "IP addresses for this service should not contain agent IPs, IPs were {}".format(ip_addresses)
```
jsrodman: Overall this function is more than a little inscrutable.  An endpoint has an address?  We're actually not checking anything about the endpoints besides the addresses?  And why are agent addresses invalid?

I suspect this is really assert_endpoints using_overlay() or something, though I still don't know what an endpoint is in this context, or why it has an address. 

In my experience, programmers use "endpoint" to refer very sloppily to some variation on a URL.  If we mean that here, do we mean http://*address*/whatever/whaterver?  Or does this api resolve the name in the URL to its address?  Does DC/OS prohibit multihome hosts?

testing/sdk_networks.py
```diff
@@ -19,3 +27,23 @@ def check_task_network(task_name, expected_network_name="dcos"):
...
+
+    assert "dns" in endpoints, "Endpoints missing DNS key"
+    for dns in endpoints["dns"]:
+        assert "autoip.dcos.thisdcos.directory" in dns, "DNS {} is incorrect should have autoip.dcos.thisdcos." \
```
jsrodman: Does this want to be dns.endswith() ?

frameworks/kafka/tests/test_cni.py
ArtRand: The old `test_cni.py` is redundant at this point.

frameworks/kafka/tests/test_overlay.py
ArtRand: no? It makes it easier to run subsets of the tests, especially when CNI/overlay networks are still new and the SDK is changing.

testing/sdk_networks.py
ArtRand: Not quite, the DNS will be `<task_name>.<service_name>.autoip.dcos.thisdcos.directory:<port>`, `endswith()` requires that I know the port.

testing/sdk_networks.py
ArtRand: So this is a little curious to me as well. In the past this function returned a Task instance, now it returns a list of length 1 with that task. The list should always be length 1 in this case, because we're querying by task ID (which are unique). So I have the check for backwards compatibility. I like your assert(len(..)) == 1 idea though.

testing/sdk_networks.py
ArtRand: get the first element after a split is too complex?

frameworks/hdfs/tests/test_shakedown.py
ArtRand: moved to the config.py in that test dir, because it's used by the overlay and non-overlay tests

frameworks/hdfs/tests/test_shakedown.py
ArtRand: I can change it back, 1 second seemed pretty aggressive though given all the other times in the tests.

testing/sdk_networks.py
ArtRand: disagree, I think this is simpler, and easy to reason about.

ArtRand:
retest this please (ccm)

ArtRand:
retest this please (xpack)

benclarkwood:
@loren has a fix out for the xpack flakiness: https://github.com/mesosphere/dcos-commons/pull/1131

frameworks/hdfs/tests/test_shakedown.py
jsrodman: Doesn't look like config to me.

frameworks/kafka/tests/test_overlay.py
jsrodman: pytest path/test_overlay.py

pytest -k test_overlay

frameworks/kafka/tests/test_overlay.py
```diff
@@ -0,0 +1,91 @@
...
+    )
+    for task in broker_tasks:
+        networks.check_task_network(task)
+
+
+@pytest.mark.smoke
+@pytest.mark.sanity
+@pytest.mark.overlay
+def test_overlay_network_deployment_and_endpoints():
+    # double check
```
jsrodman: And this one

frameworks/kafka/tests/test_overlay.py
```diff
@@ -0,0 +1,91 @@
...
+    networks.check_endpoints_on_overlay(broker_endpoints)
+
+    zookeeper = service_cli('endpoints zookeeper', get_json=False)
+    assert zookeeper.rstrip() == 'master.mesos:2181/dcos-service-{}'.format(PACKAGE_NAME)
+
+
+@pytest.mark.sanity
+@pytest.mark.overlay
+def test_pods_restart_on_overlay():
+    restart_broker_pods()
```
jsrodman: And this one. Etc.

testing/sdk_networks.py
jsrodman: According to the docstring it is *never* a list, but a tuple. If you did want to do a type test, you should use isinstance()

testing/sdk_networks.py
jsrodman: In a language where you get no static support on type checking, having one local variable be two totally unrelated types is bad.

testing/sdk_networks.py
jsrodman: Failure scenarios in this one line. 
- endpoints might not be a mapping type
- endpoints might lack the key address
- the value at address might not be a sequence
- the elements of that value might not have the attribute split
- the attribute might not be callable
- the string might not have a colon

Not all the exceptions will be obvious

benclarkwood:
@ArtRand Xpack fix was merged.

ArtRand:
retest this please (xpack round 2)

frameworks/kafka/tests/test_overlay.py
ArtRand: I've been maintaining the pytest marks mostly because I still use TEST_TYPES and test.sh to test things locally.

PR #1129, opened 2017-06-16
change in testing/sdk_package 

OhRobin:
- Changing from "dcos dcos package describe" -> "dcos package describe"

- This is because sdk_cmd.run_cli = shakedown.run_dcos_command() NOT shakedown.run_command()


benclarkwood:
retest this please

PR #1133, opened 2017-06-20
Improve robustness of DefaultSchedulerTest

benclarkwood:
Verify that plan has completed before checking the expected results. Test suite has a 30 second per test timeout, so test will timeout despite infinite loop.

### DIFF
--- a/sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/DefaultSchedulerTest.java
+++ b/sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/DefaultSchedulerTest.java
@@ -872,6 +872,16 @@ public class DefaultSchedulerTest {
         taskIds.add(installStep(1, 0, getSufficientOfferForTaskB()));
         taskIds.add(installStep(1, 1, getSufficientOfferForTaskB()));
 
+        while (true) {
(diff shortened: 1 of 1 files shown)


sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/DefaultSchedulerTest.java
```diff
@@ -872,6 +872,16 @@ private void statusUpdate(Protos.TaskID launchedTaskId, Protos.TaskState state)
         taskIds.add(installStep(1, 0, getSufficientOfferForTaskB()));
         taskIds.add(installStep(1, 1, getSufficientOfferForTaskB()));
 
+        while (true) {
```
nickbp: `while (!defaultScheduler.deploymentPlanManager.getPlan().isComplete()) { ...sleep... }`?

Could also use a for loop to explicitly limit the number of iterations within the test:

`for (int i = 0; i < 10 && !defaultScheduler.deploymentPlanManager.getPlan().isComplete(); ++i) { ... sleep ...}`

benclarkwood: Yeah, the loop was more complicated at some point before I checked it in. Good catch.

PR #1134, opened 2017-06-20
[INFINITY-1787]Handle arg-provided requirements.txt to run-tests.py correctly.

jsrodman:
Confusion ensued from "requirements_txt" not being text, but a filename.  Handled partially but not fully.

### DIFF
--- a/tools/piputil.py
+++ b/tools/piputil.py
@@ -59,16 +59,16 @@ def pip_install_dir(path, requirements_filepath):
 
     subprocess.check_call(pip_cmd)
 
-def create_requirementsfile(filename, req_txt=None):
-    if not req_txt:
-        req_txt = ('''
+def create_requirementsfile(filename, req_text=None):
+    if not req_text:
+        req_text = ('''
 requests==2.10.0
 dcoscli==0.4.16
 dcos==0.4.16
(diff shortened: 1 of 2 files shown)


mgummelt:
Please prefix your PRs with [JIRA-###]

tools/piputil.py
```diff
@@ -59,16 +59,16 @@ def pip_install_dir(path, requirements_filepath):
 
     subprocess.check_call(pip_cmd)
 
-def create_requirementsfile(filename, req_txt=None):
-    if not req_txt:
-        req_txt = ('''
+def create_requirementsfile(filename, req_text=None):
+    if not req_text:
```
mgummelt: `if req_text is None`

jsrodman:
I always prefix my PRs with the INFINITY tag.

jsrodman:
The part I know how to test has already passed, merging.

tools/piputil.py
jsrodman: An empty requirements.txt file is not valid.

mgummelt: We should probably fail fast on invalid input, rather that silently converting it, but not a big deal.

PR #1135, opened 2017-06-20
Use default executor and task groups for launching pods

mrbrowning:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LaunchOfferRecommendation.java
```diff
@@ -56,13 +58,19 @@ public boolean shouldLaunch() {
...
-     * Returns the original, unpacked {@link TaskInfo} to be launched. This varies from the {@link TaskInfo} stored
-     * within the {@link Operation}, which is packed.
+     * Returns the {@link TaskInfo} to be launched.
      */
     public TaskInfo getTaskInfo() {
         return taskInfo;
     }
 
+    /**
+     * Returns the {@link TaskInfo} to be launched.
```
ArtRand: s`ExecutorInfo`/`TaskInfo`?

benclarkwood:
@gabrielhartmann @nickbp @ArtRand please review

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -69,7 +70,7 @@ pods:
             template: {{CONFIG_TEMPLATE_PATH}}/s3config.json
             dest: s3config.json
         readiness-check:
-          cmd: export TASK_IP=$(./bootstrap --get-task-ip) && ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool status | grep -q "UN  $TASK_IP"
+          cmd: export TASK_IP=$(./bootstrap --get-task-ip) && ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool status -p 7199 | grep -q "UN  $TASK_IP"
```
nickbp: should this use `{{TASKCFG_ALL_JMX_PORT}}`?

frameworks/cassandra/universe/marathon.json.mustache
```diff
@@ -45,13 +45,15 @@
...
     "TASKCFG_ALL_CASSANDRA_HEAP_NEW_MB": "{{nodes.heap.new}}",
+    "CASSANDRA_JAVA_URI": "{{resource.assets.uris.cassandra-jre-tar-gz}}",
     "CASSANDRA_HEAP_GC": "{{nodes.heap.gc}}",
     "PLACEMENT_CONSTRAINT": "{{nodes.placement_constraint}}",
     {{#service.virtual_networks}}
     "ENABLE_VIRTUAL_NETWORK": "yes",
     "VIRTUAL_NETWORK_NAME": "{{nodes.virtual_network}}",
     "CNI_PLUGIN_LABELS": "{{nodes.cni_plugin_labels}}",
     {{/service.virtual_networks}}
+    "TASKCFG_ALL_JAVA_HOME": "jre1.8.0_131/",
```
nickbp: To avoid dealing with updating this when the JRE version changes, should instead set `JAVA_HOME` like this within the `cmd`:

```
export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/)
... rest of cmd here ...
```

frameworks/elastic/universe/marathon.json.mustache
```diff
@@ -55,6 +56,7 @@
     {{/service.virtual_networks}}
 
     "DEPLOY_STRATEGY": "{{service.deploy_strategy}}",
+    "TASKCFG_ALL_JAVA_HOME": "jre1.8.0_131/",
```
nickbp: Same as above. Should do this dynamically within the `cmd` so that it's less work to bump the JRE (happens every month or two), and/or to even allow in-the-field upgrades of the JRE.

frameworks/hdfs/universe/marathon.json.mustache
```diff
@@ -26,6 +26,7 @@
     "HDFS_VERSION": "hadoop-2.6.0-cdh5.11.0",
     "JRE_VERSION": "jre1.8.0_131",
     "TASKCFG_ALL_JRE_VERSION": "jre1.8.0_131",
+    "TASKCFG_ALL_JAVA_HOME": "jre1.8.0_131/",
```
nickbp:
```
export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/)
```

frameworks/hdfs/universe/resource.json
```diff
@@ -4,6 +4,7 @@
       "jre-tar-gz": "{{jre-jce-unlimited-url}}",
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
       "hdfs-tar-gz": "https://downloads.mesosphere.com/hdfs/assets/hadoop-2.6.0-cdh5.11.0.tar.gz",
+      "hdfs-jre-tar-gz": "{{jre-jce-unlimited-url}}",
```
nickbp: This appears to be the same value as `jre-tar-gz` above. Are these expected to vary in practice?

frameworks/kafka/universe/marathon.json.mustache
```diff
@@ -68,6 +69,7 @@
     "TASKCFG_ALL_KAFKA_ADVERTISE_HOST" : "set",
     {{/kafka.kafka_advertise_host_ip}}
 
+    "TASKCFG_ALL_JAVA_HOME": "jre1.8.0_131/",
```
nickbp:
```
export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/)
```

frameworks/kafka/universe/resource.json
```diff
@@ -5,6 +5,7 @@
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
       "bootstrap-zip": "{{artifact-dir}}/bootstrap.zip",
       "kafka-tgz": "https://downloads.mesosphere.com/kafka/assets/kafka_2.11-0.10.2.1.tgz",
+      "kafka-jre-tar-gz": "{{jre-url}}",
```
nickbp: Same as `jre-tar-gz` above. Expected to vary?

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/DefaultSchedulerTest.java
```diff
@@ -61,6 +63,7 @@
  * This class tests the DefaultScheduler class.
  */
 @SuppressWarnings({"PMD.TooManyStaticImports", "PMD.AvoidUsingHardCodedIP"})
+@Ignore
```
nickbp: Does this disable `DefaultSchedulerTest`?

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/DefaultSchedulerTest.java
```diff
@@ -789,9 +796,9 @@ private void register() {
                 .setHostname(TestConstants.HOSTNAME)
                 .addAllResources(
                         Arrays.asList(
-                                ResourceTestUtils.getUnreservedCpu(TASK_A_CPU),
-                                ResourceTestUtils.getUnreservedMem(TASK_A_MEM),
-                                ResourceTestUtils.getUnreservedDisk(TASK_A_DISK)))
+                                ResourceTestUtils.getUnreservedCpu(TASK_A_CPU + 0.1),
```
nickbp: Ideally these 0.1/32/256 values should be consts somewhere. Looks like they're used here and in OfferTestUtils, maybe they could be placed in OfferTestUtils? Or are they in the main src somewhere as well?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -42,23 +42,28 @@
...
     public PodInfoBuilder(
             PodInstanceRequirement podInstanceRequirement,
             String serviceName,
             UUID targetConfigId,
             SchedulerFlags schedulerFlags,
-            Collection<Protos.TaskInfo> currentPodTasks)
-                    throws InvalidRequirementException {
+            Collection<Protos.TaskInfo> currentPodTasks,
+            Protos.FrameworkID frameworkID,
+            boolean useDefaultExecutor) throws InvalidRequirementException {
```
nickbp: IMO it'll be a little easier to read if this parameter is defined as an enum rather than a bool:

```
public enum ExecutorType {
  DEFAULT,
  CUSTOM
}
```

Then callers would look like this:
```
PodInfoBuilder b = new PodInfoBuilder(..., ExecutorType.DEFAULT);
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -136,19 +142,46 @@ public void addAssignedOverlayPort(long port) {
         return assignedOverlayPorts;
     }
 
-    private static Protos.TaskInfo getTaskInfo(
+    public void setExecutorVolume(VolumeSpec volumeSpec) {
+        // Volumes on the executor must be declared in each TaskInfo.ContainerInfo to be shared among them.
+        for (Protos.TaskInfo.Builder t : getTaskBuilders()) {
+            Protos.ContainerInfo.Builder builder = t.getContainerBuilder();
+
+            if (!builder.hasType()) {
```
nickbp: Should ContainerInfo.Type always be MESOS? If so it may make sense to just always set it here:

```
t.getContainerBuilder()
    .setType(Protos.ContainerInfo.Type.MESOS)
    .addVolumes(volume);
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -136,19 +142,46 @@ public void addAssignedOverlayPort(long port) {
...
+                builder.setType(Protos.ContainerInfo.Type.MESOS);
+            }
+            builder.addVolumes(getVolume(volumeSpec));
```
nickbp: Could call `getVolume(volumeSpec)` once and pass the resulting `Protos.Volume` to each `TaskInfo`. Not worried about efficiency; it'd just be a bit clearer that it's identical across the TaskInfos.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -163,91 +196,131 @@ public void addAssignedOverlayPort(long port) {
...
+                containerBuilder.addAllNetworkInfos(
+                        podSpec.getNetworks().stream()
+                                .map(PodInfoBuilder::getNetworkInfo).collect(Collectors.toList()));
+            }
+        } else {
+            // command and user:
+            Protos.CommandInfo.Builder executorCommandBuilder = executorInfoBuilder.getCommandBuilder().setValue(
+                    "export LD_LIBRARY_PATH=$MESOS_SANDBOX/libmesos-bundle/lib:$LD_LIBRARY_PATH && " +
+                    "export MESOS_NATIVE_JAVA_LIBRARY=$(ls $MESOS_SANDBOX/libmesos-bundle/lib/libmesos-*.so) && " +
+                    "export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/) && " +
```
nickbp: (not a thing to fix) here's an example of dynamically setting `JAVA_HOME` within the `cmd`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -163,91 +196,131 @@ public void addAssignedOverlayPort(long port) {
...
-                        "export JAVA_OPTS=\"-Xms128M -Xmx128M\" && " +
-                        "$MESOS_SANDBOX/executor/bin/executor");
+                .setExecutorId(Protos.ExecutorID.newBuilder().setValue("").build());
+        Protos.ContainerInfo.Builder containerBuilder = getContainerInfo(podSpec, true, false).toBuilder();
 
-        if (podSpec.getUser().isPresent()) {
-            executorCommandBuilder.setUser(podSpec.getUser().get());
-        }
+        executorInfoBuilder.getLabelsBuilder().addLabelsBuilder()
+                .setKey("DCOS_SPACE")
```
nickbp: This label name should probably be broken out into `LabelConstants`. Could make an `ExecutorLabelWriter` to handle setting it. See e.g. `SchedulerLabelWriter` which is confusingly used for writing TaskInfos (but from the Scheduler as opposed to from the custom Executor)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -42,23 +42,28 @@
...
             String serviceName,
             UUID targetConfigId,
             SchedulerFlags schedulerFlags,
-            Collection<Protos.TaskInfo> currentPodTasks)
-                    throws InvalidRequirementException {
+            Collection<Protos.TaskInfo> currentPodTasks,
+            Protos.FrameworkID frameworkID,
+            boolean useDefaultExecutor) throws InvalidRequirementException {
         PodInstance podInstance = podInstanceRequirement.getPodInstance();
+        this.useDefaultExecutor = useDefaultExecutor;
```
nickbp: Subjective: IMO should either have this be passed down the call stack (getTaskInfo() onwards). Could then continue to have static builder functions. At the moment though it's got a mix of both: e.g. `serviceName` being passed down the stack but `useDefaultExecutor` being set as a member.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -315,7 +388,7 @@ private static void extendEnv(Protos.CommandInfo.Builder builder, Map<String, St
         return builder.build();
     }
 
-    private static void setHealthCheck(
+    private void setHealthCheck(
```
nickbp: (cotd from subjective comment in constructor) For example this (and similar calls elsewhere) could continue being static if useDefaultExecutor was being passed in.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -355,52 +432,69 @@ private static void setReadinessCheck(
...
+        }
     }
 
     private static String getConfigTemplateDownloadPath(ConfigFileSpec config) {
         // Name is unique.
         return String.format("%s%s", CONFIG_TEMPLATE_DOWNLOAD_PATH, config.getName());
     }
 
-    private static Protos.ContainerInfo getContainerInfo(PodSpec podSpec, boolean addExtraParameters) {
+    private Protos.ContainerInfo getContainerInfo(
```
nickbp: Couldn't hurt to add javadoc explaining what the two booleans mean

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -28,18 +29,25 @@
...
     private Optional<String> resourceId;
 
-    public PortEvaluationStage(PortSpec portSpec, String taskName, Optional<String> resourceId, String portName) {
-        this.portName = portName;
+    public PortEvaluationStage(
+            PortSpec portSpec,
+            String taskName,
+            Optional<String> resourceId,
+            String portName,
+            boolean useDefaultExecutor) {
```
nickbp: Similarly to PodInfoBuilder this could be using an `ExecutorType` enum instead

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -187,28 +195,48 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
...
                     withPortEnvironmentVariable(executorBuilder.getCommandBuilder().getEnvironment(), port));
             if (useHostPorts) {
                 executorBuilder.addResources(resource);
             }
+        }
+    }
+
+    private void addReadinessCheckPort(Protos.TaskInfo.Builder taskBuilder, String name, String value) {
+        if (useDefaultExecutor) {
+            Protos.Environment.Builder envBuilder = taskBuilder.getCheckBuilder()
```
nickbp: Use `EnvUtils.withEnvVar(name, value)`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
```diff
@@ -51,14 +52,15 @@ public String toString() {
...
-    public TaskResourceMapper(TaskSpec taskSpec, Protos.TaskInfo taskInfo) {
+    public TaskResourceMapper(TaskSpec taskSpec, Protos.TaskInfo taskInfo, boolean useDefaultExecutor) {
         this.taskSpecName = taskSpec.getName();
         this.resourceSpecs = new ArrayList<>();
         this.resourceSpecs.addAll(taskSpec.getResourceSet().getResources());
         this.resourceSpecs.addAll(taskSpec.getResourceSet().getVolumes());
         this.resources = taskInfo.getResourcesList();
         this.taskEnv = EnvUtils.toMap(taskInfo.getCommand().getEnvironment());
         this.evaluationStages = getEvaluationStagesInternal();
+        this.useDefaultExecutor = useDefaultExecutor;
```
nickbp: Subjective: may as well just pass it into the static functions below. For example taskSpecName is already being passed in despite also being a member.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/taskdata/SchedulerLabelWriter.java
```diff
@@ -3,14 +3,14 @@
 import java.util.Optional;
 import java.util.UUID;
 
+import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
```
nickbp: The wrong VisibleForTesting (I get this too)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/taskdata/SchedulerLabelWriter.java
```diff
@@ -106,15 +106,15 @@ public SchedulerLabelWriter setTargetConfiguration(UUID targetConfigurationId) {
...
+        put(LabelConstants.READINESS_CHECK_LABEL, LabelUtils.encodeHealthCheck(readinessCheck.toBuilder().setConsecutiveFailures(0).build()));
         return this;
     }
 
     /**
      * Updates the stored readiness check, if any, to have the provided environment variable.
      * Does nothing if no readiness check is present.
      *
-     * @throws TaskException if parsing a previously set {@link HealthCheck} failed
+     * @throws TaskException if parsing a previously set {@link HealthCheck}
```
nickbp: the previous javadoc looks accurate?

frameworks/hdfs/universe/resource.json
mrbrowning: The former is the scheduler's JRE, and the latter is the task's JRE (which will be free from any framework-side constraints once we transition fully to the default executor). It seemed like it would cause confusion down the road if these two were coupled, since they have effectively nothing to do with each other, even though right now there is not a pressing reason to bring in two different JRE instances.

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/DefaultSchedulerTest.java
mrbrowning: Yes, there's some flakiness with mocks that @ArtRand is going to take a look at when he gets the chance, then it'll be reenabled.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
mrbrowning: I don't really understand the value of making these functions static at all, since they do in fact depend on an aspect of object state, so that's why I elected to make them not-static. I'll make that uniform.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/taskdata/SchedulerLabelWriter.java
mrbrowning: Not sure how this snuck in, probably just an accidental vim movement.

frameworks/hdfs/universe/resource.json
nickbp: :+1:

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
nickbp: I think it originated with the functions taking fewer params but now that there's 2-3 different members involved it'd probably make sense to move to non-static like you say.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
mrbrowning: I made a constant in `PodInfoBuilder` to handle this for now, deferring anything moving to `ExecutorLabelWriter` until after we're no longer bound to support both custom and default executors, since this entails some package visibility changes (don't want to have the scheduler package depend on the executor package, where those classes currently live, but moving them to common is outside the scope of this PR).

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
mrbrowning: I favor them being non-static -- the need to have the same data passed in to several of these static methods indicates to me that they are in fact functions of the object and not the class.

PR #1136, opened 2017-06-20
[INFINITY-1737] Enable statsd for Kafka

NimaVaziri:


frameworks/kafka/src/main/dist/svc.yml
```diff
@@ -34,7 +36,10 @@ pods:
           KAFKA_DISK_PATH: "{{BROKER_DISK_PATH}}"
           KAFKA_HEAP_OPTS: "-Xms{{BROKER_JAVA_HEAP}}M -Xmx{{BROKER_JAVA_HEAP}}M"
         goal: RUNNING
-        cmd: "./bootstrap -resolve=false && exec $MESOS_SANDBOX/{{KAFKA_VERSION_PATH}}/bin/kafka-server-start.sh $MESOS_SANDBOX/{{KAFKA_VERSION_PATH}}/config/server.properties"
+        cmd: >
+          find $MESOS_SANDBOX -maxdepth 1 -name "*.jar" | grep "statsd" | xargs -I "{}" mv "{}" $MESOS_SANDBOX/{{KAFKA_VERSION_PATH}}/libs/ &&
```
nickbp: should do `mv -v` here

nickbp:
Just in case, I reverified via the stub universe build that the agents were getting metrics. :+1:

porridge:
@nickbp and @triclambert since you reviewed this, do you happen to know where @NimaVaziri got https://downloads.mesosphere.com/kafka/assets/kafka-statsd-metrics2-0.5.3.jar from? Upstream is only 0.5.2: https://github.com/airbnb/kafka-statsd-metrics2/blob/master/gradle.properties#L1

Now I'd like to replace it with yet another version and I'm worried I might drop some important changes that Nima applied...

porridge:
Note for posterity:
I built upstream 0.5.2, and the contents of the resulting jar is byte-for-byte identical to the aforementioned "0.5.3", except for one class file, `MetricNameFormatter.class`.
Simple decompile shows no difference. After decompiling that class on both sides with `procyon -r -v -ln -cp -lv -ta` (raw bytecode output with some more details), there only difference is a single missing line number annotation in Nima's version:
```
@@ -233,7 +233,6 @@
           45: invokevirtual   java/lang/StringBuilder.append:(Ljava/lang/String;)Ljava/lang/StringBuilder;
               linenumber      56
           48: invokevirtual   java/lang/StringBuilder.toString:()Ljava/lang/String;
-              linenumber      49
           51: areturn        
     LocalVariableTable:
       Start  Length  Slot  Name        Signature
```
So I'm going to assume there were no functional changes and the file is safe to replace with 0.5.2.

porridge:
FTR, @elezar found the source for 0.5.3: https://bintray.com/airbnb/jars/kafka-statsd-metrics2/0.5.3#files/com%2Fairbnb%2Fkafka-statsd-metrics2%2F0.5.3 although it's still not clear what the exact difference is.

PR #1137, opened 2017-06-20
Fix CI error tools/piputil.py

mbdcos:


jsrodman:
Not sure why this error isn't seen elsewhere, but it's probably better to use b'python 3' if it's reliably a byte string.  I thought i checked the output value though.

mbdcos:
retest this please

mbdcos:
Byte or string, should be same type. You already have pip3 command, it will return before, so you do not see this error.

mbdcos:
@nickbp already fixed this. Closing....

PR #1141, opened 2017-06-21
INFINITY-1806 Reenable documented support for custom ZK location for kafka

nickbp:
Updates Kafka tests to exercise the support as well

### DIFF
--- a/frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/scheduler/KafkaService.java
+++ b/frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/scheduler/KafkaService.java
@@ -17,10 +17,14 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Collection;
 
+import org.apache.commons.lang3.StringUtils;
+
 /**
  * Kafka Service.
  */
 public class KafkaService extends DefaultService {
+    private static final String KAFKA_ZK_URI_ENV = "KAFKA_ZOOKEEPER_URI";
+
     public KafkaService(File pathToYamlSpecification) throws Exception {
         super(createSchedulerBuilder(pathToYamlSpecification));
     }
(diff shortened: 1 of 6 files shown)


frameworks/kafka/universe/marathon.json.mustache
```diff
@@ -59,6 +59,8 @@
     "CONFIG_TEMPLATE_PATH": "kafka-scheduler",
     "KAFKA_VERSION_PATH": "kafka_2.11-0.10.2.1",
 
+    "KAFKA_ZOOKEEPER_URI": "{{kafka.kafka_zookeeper_uri}}",
```
vishnu2kmohan: Shouldn't this be `"TASKCFG_ALL_KAFKA_ZOOKEEPER_URI": "{{kafka.kafka_zookeeper_uri}}",` since it needs to be rendered into every Kafka broker's `server.properties`?

benclarkwood: It gets set manually above in the Java

nickbp: It's being passed to the scheduler, so that the scheduler can fill in a reasonable value if it's empty/unset. The scheduler then passes the result to the tasks. See code in KafkaService.java above.

vishnu2kmohan: Thanks, still learning about the SDK, appreciate your patience.

nickbp: PS: Handling the default ZK path in java allows escaping out slashes in the service name, which is the case if it's installed in a folder

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/scheduler/KafkaService.java
```diff
@@ -31,7 +33,7 @@ public KafkaService(File pathToYamlSpecification) throws Exception {
 
         // Allow users to manually specify a ZK location for kafka itself. Otherwise default to our service ZK location:
         String kafkaZookeeperUri = System.getenv("KAFKA_ZOOKEEPER_URI");
```
gabrielhartmann: Should probably make this string a constant, but definitely not a blocker.

nickbp: Done

nickbp:
retest this please (test took too long)

nickbp:
`Velocity` is green, merging: looks like the build last time tested everything but then this time realized that only Kafka code changed, but the non Kafka statuses were regardless carried over in a pending state.

benclarkwood:
Yeah, there's no way to remove statuses :(

PR #1143, opened 2017-06-21
INFINITY-1761 Use 'nodetool' instead of direct task lookup in cassandra node replace test

nickbp:
The table can flake out and never update (or did once, at least...), whereas nodetool is hopefully more consistent(?). Waits for the old node to disappear and a new node to appear with UN status.

### DIFF
--- a/frameworks/cassandra/tests/test_recovery.py
+++ b/frameworks/cassandra/tests/test_recovery.py
@@ -51,9 +51,21 @@ def test_node_replace_replaces_node():
     # replaced from Cassandra's perspective)
     # Note: Task will sometimes flake out because the node list can take a minute or two to update.
     #       Therefore this job has restart.policy=ON_FAILURE
-    verify_replace_job = get_verify_node_replace_job(pod_host)
-    with jobs.InstallJobContext([verify_replace_job]):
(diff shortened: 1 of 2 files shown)


frameworks/cassandra/tests/test_recovery.py
```diff
@@ -51,9 +51,21 @@ def test_node_replace_replaces_node():
     # replaced from Cassandra's perspective)
     # Note: Task will sometimes flake out because the node list can take a minute or two to update.
     #       Therefore this job has restart.policy=ON_FAILURE
-    verify_replace_job = get_verify_node_replace_job(pod_host)
-    with jobs.InstallJobContext([verify_replace_job]):
-        jobs.run_job(verify_replace_job)
+    check_host = get_pod_host('node-0')
+    def fun():
+        status, stdout = shakedown.run_command_on_agent(check_host, "docker run -t --net=host pitrho/cassandra-nodetool nodetool -p 7199 status")
```
mrbrowning: This will likely not work in the future should Cassandra end up running on an overlay, should this be done with task exec to future proof it?

nickbp: Good point, done!

PR #1144, opened 2017-06-21
INFINITY-1793 Fix strict mode tests for foldered services (probably)

nickbp:
- Foldered services are exercised as `/test/integration/svcname` which maps to a role of `test__integration__svcname-role`
- Removes `svcname2` role setup, as `grep -R 2 frameworks/*/tests/` shows that tests aren't installing a separate `2` instance anymore. Comments imply this was once the case in `dcos-hdfs-service` which is now long gone.


nickbp:
retest this please (unit test timeout flake)

PR #1147, opened 2017-06-21
[INFINITY-1814] Improve service uninstall in tests.

benclarkwood:
Verify that service has ability to uninstall as well as whether the platform has the ability to uninstall.

### DIFF
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -12,6 +12,7 @@ import sdk_api
 import sdk_plan
 import sdk_tasks
 import sdk_utils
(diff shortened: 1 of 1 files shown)


benclarkwood:
Hm.

It occurs to me that I cannot check a Marathon app that hasn't yet been installed.

@nickbp Got any good ideas on how to check before the app is installed (since many of our "setups" uninstall the app first). I think it can safely run the "new" uninstall, since it's just a no-op.

benclarkwood:
Closing until I have time to circle back to this.

PR #1149, opened 2017-06-21
[INFINITY-1636] set port names in DiscoveryInfo

ArtRand:
This PR adds a `DiscoveryInfo` for pods that have specified `PortSpec`s, which is required for SDK-based services to interact with EdgeLB. Specifically, the `DiscoveryInfo` has a populated `name` record (https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L2834) that originates from the YAML ServiceSpec. 

Previously, the `DiscoveryInfo` was only populated when a `PortSpec` had a VIP (making it actually a `NamedVIPSpec`). Always having the `DiscoveryInfo` actually makes some of the logic around VIPs simpler. 

The implementation is unit tested, and this PR contains improved coverage of `NamedVIPEvaluationStage`. 

Behavior is integration tested: the correctness of the `TaskInfo`s is tested in the hello-world framework along with the correct generation of the Mesos-DNS SRV records. 

Thanks for the reviews.


ArtRand:
retest this please (ccm)

ArtRand:
retest this please (ccm x2)

ArtRand:
retest this please (unable to repro)

frameworks/helloworld/tests/test_overlay.py
```diff
@@ -111,3 +121,36 @@ def test_overlay_network():
...
+            check_task_ports(task, 1, ["test"])
+
+@pytest.mark.sanity
+@pytest.mark.overlay
+def test_srv_records():
+    fmk_srvs = networks.get_framework_srv_records(PACKAGE_NAME)
+    for task in TASKS_WITH_PORTS:
+        task_records = networks.get_task_record(task, fmk_srvs)
+        if task == "hello-overlay-0-server":
+            assert len([r for r in task_records if "dummy" in r["name"]]) == 1, "Missing SRV record for dummy and "\
```
mrbrowning: No need for `\` to split up lines while you're in parens, expressions can be split across lines and strings get auto-concatenated.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/NamedVIPEvaluationStage.java
```diff
@@ -44,21 +48,29 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
...
+                            .addAllLabels(EndpointUtils.createVipLabels(vipName, vipPort, onNamedNetwork));
             } else {
-                taskBuilder.setDiscovery(getVIPDiscoveryInfo(
-                        taskBuilder.getName(),
-                        vipName,
-                        vipPort,
-                        protocol,
-                        visibility,
-                        resource));
+                throw new IllegalStateException(String.format("Trying TaskBuilder missing DiscoveryInfo for port" +
```
mrbrowning: Is the "trying" supposed to be here?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -44,6 +46,18 @@ protected long getPort() {
...
 
+    protected long getResourcePort(Protos.Resource resource) {
+        return resource.getRanges().getRange(0).getBegin();
+    }
+
+    protected String getPortName() {
+        return portName;
+    }
+
+    protected boolean isUseHostPorts() {
```
mrbrowning: Maybe `isUsingHostPorts`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -137,6 +163,42 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
             taskBuilder.getCommandBuilder().setEnvironment(
                     withPortEnvironmentVariable(taskBuilder.getCommandBuilder().getEnvironment(), port));
 
+            //Protos.DiscoveryInfo.Builder discoveryInfoBuilder = taskBuilder.hasDiscovery() ?
+            //        taskBuilder.getDiscoveryBuilder() : Protos.DiscoveryInfo.newBuilder();
+
+            //Protos.Ports.Builder portsBuilder = discoveryInfoBuilder.getPortsBuilder();
```
mrbrowning: Drop these comments

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -137,6 +163,42 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
...
+                        .findFirst();
+                if (builderOptional.isPresent()) {
+                    //builderOptional.get()
+                    //        .setNumber((int) port)
+                    //        .setProtocol(DcosConstants.DEFAULT_IP_PROTOCOL);
+                    setPortBuilder(builderOptional.get(), port);
+                } else {
+                    portsBuilder.addPorts(makePortBuilder(port));
+                    //portsBuilder.addPortsBuilder()
+                    //        .setNumber((int) port)
```
mrbrowning: Same

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
```diff
@@ -230,9 +230,11 @@ private static OfferEvaluationStage toEvaluationStage(
             Optional<String> resourceId,
             Optional<String> persistenceId) {
         if (resourceSpec instanceof NamedVIPSpec) {
-            return new NamedVIPEvaluationStage((NamedVIPSpec) resourceSpec, taskSpecName, resourceId);
+            NamedVIPSpec namedVIPSpec = (NamedVIPSpec) resourceSpec;
```
mrbrowning: @gabrielhartmann part of the reason that specs were formerly able to create their own evaluation stages polymorphically was to prevent brittleness like this, where each new case is going to require another `if` clause...

testing/sdk_networks.py
```diff
@@ -60,3 +60,45 @@ def check_ip_addresses_on_overlay():
...
         assert "autoip.dcos.thisdcos.directory" in dns, "DNS {} is incorrect should have autoip.dcos.thisdcos." \
                                                         "directory".format(dns)
+
+
+def get_framework_srv_records(package_name):
+    cmd = "curl localhost:8123/v1/enumerate"
+    ok, out = shakedown.run_command_on_master(cmd)
+    assert ok, "Failed to get srv records. command was {}".format(cmd)
+    srvs = json.loads(out)
+    fmk_srvs = [f for f in srvs["frameworks"] if f["name"] == package_name]
```
mrbrowning: Maybe `framework_srvs`, otherwise it's super easy to read this as `framework_services`

frameworks/helloworld/src/main/dist/examples/overlay.yml
```diff
@@ -38,12 +38,14 @@ pods:
...
-            port: 4044
+          dummy:
+            port: 1025
+          dynport:
+            port: 0
     tasks:
       server:
         goal: RUNNING
-        cmd: "python3 http_py3responder.py 4044 hello from the overlay network!"
+        cmd: "python3 http_py3responder.py $PORT_DYNPORT hello from the overlay network!"
```
ArtRand: This tests that dynamic ports on the overlay network will not collide with user-specified ports. (Dynamic ports start at 1025, so we expect `$PORT_DYNPORT` to be 1026, which is tested by the `get-overlay` task below.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
ArtRand: well that's embarrassing..

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
benclarkwood: @mrbrowning Do you know why it got removed?

It seems reasonable to me for the `Spec`s to know how to generate their evaluation stages, but they are also in a very different package.

frameworks/helloworld/tests/test_overlay.py
ArtRand: The message string is not in parens...?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/NamedVIPEvaluationStage.java
ArtRand: Good question, removed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
ArtRand: This isn't actually used,  so I removed it.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
ArtRand: This was done during the "major" refactor that removed `DefaultOfferRequirementProvider` and the like. I can reinstate these methods into the `ResouceSpec` classes if y'all want.

mrbrowning: Probably because, like you imply, the logical distance between the two in the codebase is counterintuitively high, but as is in evidence here, evaluation stages directly depend on their attendant specs. We've just elected now to opt out of having the type system help us to manage those dependencies.

frameworks/helloworld/tests/test_overlay.py
mrbrowning: Whoops, misread the formatting. It could be, though, for pseudo-PEP8 compliance :)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
mrbrowning: I wouldn't worry about it here, it's a little tedious and outside of the scope of this change.

frameworks/helloworld/src/main/dist/examples/overlay.yml
ArtRand: Yes it is. When assigning dynamic ports on a virtual network that does not depend on Mesos resources, it should start at 1025:
https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/DcosConstants.java#L29

benclarkwood: Cool. The use of "should" is a little concerning but that looks like a "will"

ArtRand:
retest this please

PR #1150, opened 2017-06-22
[INFINITY-1818]To avoid cert trouble, download using curl.

jsrodman:


jsrodman:
@mbdcos I think this will fix what you saw.

jsrodman:
CLI download succeeded; tests passed so it worked.

PR #1154, opened 2017-06-22
Workaround security module inability to authorize multi-role frameworks.

gabrielhartmann:


benclarkwood:
It's a sufficient unblock for now as the only consumer of multi-role is edge-lb which is 1.10 only.

benclarkwood:
s/0.18.1/0.19.1

nickbp:
er yeah that

PR #1156, opened 2017-06-22
Move hello-world upgrade test into separate file

nickbp:
- If the test is run in test_sanity, it results in two hello-world instances being installed simultaneously: one with folders (from test_sanity) and one without (from the upgrade test run)
- Skip 2nd upgrade to test version in upgrade test run, as there aren't any followup tests which need the test version to be reinstalled

### DIFF
--- a/frameworks/helloworld/tests/test_sanity.py
+++ b/frameworks/helloworld/tests/test_sanity.py
@@ -9,7 +9,6 @@ import sdk_cmd as cmd
 import sdk_install as install
 import sdk_marathon as marathon
 import sdk_tasks as tasks
-import sdk_test_upgrade
 import sdk_utils
 from tests.config import (
     PACKAGE_NAME,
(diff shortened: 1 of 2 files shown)


frameworks/helloworld/tests/test_sanity.py
```diff
@@ -262,8 +261,6 @@ def fn():
     zk_config_new = shakedown.get_zk_node_data(zk_path)
     assert zk_config_old == zk_config_new
 
-
-@pytest.mark.upgrade
-@pytest.mark.sanity
-def test_upgrade_downgrade():
-    sdk_test_upgrade.upgrade_downgrade(PACKAGE_NAME, PACKAGE_NAME, DEFAULT_TASK_COUNT)
+    # Don't bother scaling back to 1 instance as we're about to be uninstalled anyway.
```
benclarkwood: Any reason to not just take this code out?

nickbp: Good point, done

nickbp:
retest this please (some weird flake in gradle build)

PR #1164, opened 2017-06-23
Fix possible null pool

gabrielhartmann:


benclarkwood:
@NimaVaziri Did you mean to push those HDFS changes here?

NimaVaziri:
@benclarkwood Eh this branch was created in the context of upgrading HDFS (GA to beta) but the main changes are not HDFS specific so I'll just move those commits (except the last one) to their relevant pre-existing branch.

PR #1169, opened 2017-06-24
Patch for PersistentLaunchRecorderTest

mellenburg:
This test would only fail when ordering was changed and would also
segfault the JVM. The segfault was due to an incompatibility with
gradle's native jansi module. This was fixed with explicitly importing
a newer module. The initial failure was due to
PeristentLaunchRecorderTest shadowing the beforeClass method of parent
DefaultCapabilities test suite. Parent classes were made final to avoid
future NPEs in other classes.

### DIFF
--- a/sdk/scheduler/src/test/java/com/mesosphere/sdk/state/PersistentLaunchRecorderTest.java
+++ b/sdk/scheduler/src/test/java/com/mesosphere/sdk/state/PersistentLaunchRecorderTest.java
@@ -29,7 +29,8 @@ public class PersistentLaunchRecorderTest extends OfferEvaluatorTestBase {
             .build();
 
     @BeforeClass
-    public static void beforeAll() throws Exception {
+    public static void beforeAll2() throws Exception {
+        // Use a new beforeAll name to avoid shadowing parent final void
         ClassLoader classLoader = PersistentLaunchRecorderTest.class.getClassLoader();
         File file = new File(classLoader.getResource("shared-resource-set.yml").getFile());
         RawServiceSpec rawServiceSpec = RawServiceSpec.newBuilder(file).build();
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/test/java/com/mesosphere/sdk/state/PersistentLaunchRecorderTest.java
```diff
@@ -29,7 +29,8 @@
             .build();
 
     @BeforeClass
-    public static void beforeAll() throws Exception {
+    public static void beforeAll2() throws Exception {
+        // Use a new beforeAll name to avoid shadowing parent final void
```
nickbp: - May make sense to just rename the one in OfferEvaluatorTestBase instead. Could see this coming up with other extending classes
- Comment should probably be in javadoc form (`/** foo */`)

sdk/scheduler/src/test/java/com/mesosphere/sdk/state/PersistentLaunchRecorderTest.java
```diff
@@ -28,6 +28,7 @@
             .setSlaveId(TestConstants.AGENT_ID)
             .build();
 
+    /** Use a new beforeAll name to avoid shadowing parent final void */
```
nickbp: should this now be in `DefaultCapabilitiesTestSuite.java`?

mellenburg: D'oh!

mellenburg: The parent function is now a final so it doesnt need to be commented anyway

PR #1170, opened 2017-06-26
[CASSANDRA-624] Fix airgapped cluster support in Cassandra (C* docker image & tar.gz)

susanxhuynh:
Fix airgapped cluster support in Cassandra. The cassandra docker image and cassandra tar-gz should be in resources.json.


benclarkwood:
What nick said.

benclarkwood:
retest this please

PR #1172, opened 2017-06-26
[INFINITY-1854] Flaky unit test

benclarkwood:
Raise timeout on individual step installs.

Previously, the timeout was increased on the overall deployment completion


nickbp:
feel free to just merge once unit tests pass

PR #1173, opened 2017-06-26
[INFINITY-1366] Change HDFS service user to 'nobody'

NimaVaziri:


frameworks/hdfs/universe/config.json
```diff
@@ -13,13 +13,18 @@
...
           "description": "The linux user used to run the scheduler and all executors.",
-          "default": "root"
+          "default": "nobody"
         },
         "principal" : {
           "description":"The principal for the HDFS service instance.",
           "type":"string",
           "default":""
         },
+        "mesos_api_version": {
```
nickbp: Is anything using this value? Not seeing it mentioned elsewhere

nickbp: Looks like there's supposed to be a `"MESOS_API_VERSION": "{{service.mesos_api_version}}"` in marathon.json. See examples in kafka, helloworld, cassandra.

nickbp: OK per conversation elsewhere this option will likely be removed from everything except hello-world soon. Can just add it here for now as a no-op for CI compatibility, and then remove it again in a later pass.

PR #1174, opened 2017-06-26
[HDFS-462] Add readiness check to journal nodes

NimaVaziri:
The deployment plan now waits for the journal nodes to succeed their readiness check before proceeding to deploying the name nodes.

### DIFF
--- a/frameworks/hdfs/src/main/dist/svc.yml
+++ b/frameworks/hdfs/src/main/dist/svc.yml
@@ -51,6 +51,13 @@ pods:
           {{/TLS_ENABLED}}
         env:
           JOURNALNODE: true
+        readiness-check:
+          cmd: >
+              READY=$(grep "IPC Server listener on {{TASKCFG_ALL_JOURNAL_NODE_RPC_PORT}}: starting" stderr| wc -l) &&
+              [ $READY -ge 1 ]
+          interval: 10
+          delay: 60
+          timeout: 60
   name:
     count: 2
     user: {{TASKCFG_ALL_TASK_USER}}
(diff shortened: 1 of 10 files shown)


NimaVaziri:
retest this please

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -51,6 +51,22 @@ pods:
           {{/TLS_ENABLED}}
         env:
           JOURNALNODE: true
+        # The existence of the journal-data/hdfs/current/committed-txid file indicates HDFS has already been deployed once
```
gabrielhartmann: I would extend this comment a bit more future readers.  It only describes the simple `if` case, might want to describe what the `else` clause is doing too.

frameworks/hdfs/universe/config.json
```diff
@@ -47,6 +47,17 @@
           "default": "serial",
           "enum" : ["parallel", "serial"]
         },
+        "journal_readiness_check": {
+          "type": "object",
+          "description": "Readiness check configuration properties for Journal Nodes.",
```
gabrielhartmann: This description is strange.  Maybe, "Determines whether to enable a readiness check for Journal nodes."

frameworks/hdfs/src/main/dist/svc.yml
NimaVaziri: At the surface it's obvious what it's doing, but yes it'd be helpful if it described _why_ it was doing that specifically.

NimaVaziri:
@gabrielhartmann Addressed comments

NimaVaziri:
@gabrielhartmann rebased on master

PR #1175, opened 2017-06-26
[INFINITY-1858 ] Lint frameworks for URIs / docker images outside of resource.json

benclarkwood:
Create new airgap linting tool.

Hook it into the framework build process.

This will fail until #1170 merges.

### DIFF
--- a/tools/airgap_linter.py
+++ b/tools/airgap_linter.py
@@ -17,28 +17,18 @@ def extract_uris(file_name):
     with open(file_name, "r") as file:
         lines = file.readlines()
 
-    matcher = re.compile(".*(https:\/\/|http:\/\/)([^\/]*)", re.IGNORECASE)
+    matcher = re.compile(".*https?:\/\/([^\/\?]*)", re.IGNORECASE)
     matches = []
     for line in lines:
+        # Do not grab comments
+        if line.startswith("*") or line.startswith("#") or line.startswith("//"):
+            continue
+
         match = matcher.match(line)
         if match:
-            matches.append(match.group(2))
+            matches.append(match.group(1))
 
     return matches
-
-
-def validate_uri(uri):
-    # URIs are valid if they:
-    #   - end with .mesos
-    #   - contain .dcos
-    if ".dcos" in uri:
-        return True
-
-    if uri.endswith(".mesos"):
-        return True
-
-    return False
-
 
 
 def validate_uris_in(file_name):
@@ -46,14 +36,15 @@ def validate_uris_in(file_name):
 
     bad_uri = False
     for uri in uris:
-        if not validate_uri(uri):
+        # A FQDN is a valid internal FQDN if it contains .dcos or ends with .mesos.
+        if not (".dcos" in uri or uri.endswith(".mesos")):
             print("Found a bad URI:", uri, "in:", file_name)
             bad_uri = True
 
     return not bad_uri
 
 
-def get_all_files_to_check(framework_directory):
+def get_files_to_check_for_uris(framework_directory):
     # There's a set of files that will always be present.
     files = [os.path.join(framework_directory, "universe", "config.json"),
              os.path.join(framework_directory, "universe", "marathon.json.mustache")]
@@ -61,15 +52,16 @@ def get_all_files_to_check(framework_directory):
     # Always check every file in the `dist` directory of the scheduler.
     dist_dir = os.path.join(framework_directory, "src", "main", "dist")
 
-    for file in os.listdir(dist_dir):
-        files.append(os.path.join(dist_dir, file))
+    for dp, dn, filenames in os.walk(dist_dir):
+        for file in filenames:
+            files.append(os.path.join(dp, file))
 
     return files
 
 
 def validate_all_uris(framework_directory):
     bad_file = False
-    files = get_all_files_to_check(framework_directory)
+    files = get_files_to_check_for_uris(framework_directory)
     for file in files:
         if not validate_uris_in(file):
             bad_file = True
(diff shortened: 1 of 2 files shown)


tools/airgap_linter.py
```diff
@@ -0,0 +1,136 @@
...
+import re
+import sys
+import os
+
+
+def extract_uris(file_name):
+    with open(file_name, "r") as file:
+        lines = file.readlines()
+
+    matcher = re.compile(".*(https:\/\/|http:\/\/)([^\/]*)", re.IGNORECASE)
```
jsrodman: I'm not sure I understand this regex.
It's looking for https:// or http:// followed by any number of non-slashes, and it is trying to match an entire line.

so, wouldn't something like
```
     my_url = "https://something.com/something/something"
```
be a non-match?  The entire line can't be matched, because although it contains https://, it has another slash after the base protocol identifier.

Oh, I guess .match only anchors at the start of the subject, not the end (what a weird interface?)

Still, you will get very odd results for things like

```
      my_url = "https://fruitloops.com" // This is a url that I love
      my_url2 = "https://somethingelse.com?param=arg&param2=arg2&param3&arg3"
```

I'd counter-suggest
```
    with open(file_name, "r") as source_f:
        url_matcher = re.compile(r"https?://([^/]+)/", re.IGNORECASE)
        matches = []
        for line in source_f: # iterates across readline
            for match in re.finditer(url_matcher, line):
                 matches.append(match.group(1))
        return matches
```

(aside, cpython implicitly compiles inline strings, but probably best to explicitly compile to avoid confusion)

tools/airgap_linter.py
```diff
@@ -0,0 +1,136 @@
...
+    matches = []
+    for line in lines:
+        match = matcher.match(line)
+        if match:
+            matches.append(match.group(2))
+
+    return matches
+
+
+def validate_uri(uri):
```
jsrodman: Seems this is actually validating the hostname / fqdn?

tools/airgap_linter.py
```diff
@@ -0,0 +1,136 @@
...
+    return not bad_uri
+
+
+def get_all_files_to_check(framework_directory):
+    # There's a set of files that will always be present.
+    files = [os.path.join(framework_directory, "universe", "config.json"),
+             os.path.join(framework_directory, "universe", "marathon.json.mustache")]
+
+    # Always check every file in the `dist` directory of the scheduler.
+    dist_dir = os.path.join(framework_directory, "src", "main", "dist")
```
jsrodman: Do we care about subdirectories?

tools/airgap_linter.py
```diff
@@ -0,0 +1,136 @@
...
+    #   - end with .mesos
+    #   - contain .dcos
+    if ".dcos" in uri:
+        return True
+
+    if uri.endswith(".mesos"):
+        return True
+
+    return False
+
```
nickbp: mega-nit: triple newline

tools/airgap_linter.py
```diff
@@ -0,0 +1,136 @@
...
+            matches.append(match.group(2))
+
+    return matches
+
+
+def validate_uri(uri):
+    # URIs are valid if they:
+    #   - end with .mesos
+    #   - contain .dcos
+    if ".dcos" in uri:
```
nickbp: `return '.dcos' in uri or uri.endswith('.mesos')` ?

tools/airgap_linter.py
```diff
@@ -0,0 +1,136 @@
...
+def validate_images(framework_directory):
+    svc_yml = os.path.join(framework_directory, "src", "main", "dist", "svc.yml")
+
+    with open(svc_yml, "r") as file:
+        lines = file.readlines()
+
+    bad_image = False
+    for line in lines:
+        if "image:" in line:
+            image_matcher = re.compile("\s*image:\s?(.*)", re.IGNORECASE)
```
jsrodman: Probably:
```
   image_matcher = re.compile("\s*image:\s?(.*)\s*$", re.IGNORECASE)
```
to avoid trailing whitespace, though i'd probably just, instead

```
    line = line.strip()
    image_matcher = re.compile("image:\s?(.*)$", re.IGNORECASE)
```

tools/airgap_linter.py
nickbp: at this point it could conceivably just be folded into `validate_uris_in`

tools/airgap_linter.py
```diff
@@ -0,0 +1,136 @@
...
+        return True
+
+    if uri.endswith(".mesos"):
+        return True
+
+    return False
+
+
+
+def validate_uris_in(file_name):
```
nickbp: feels like either this should be `validate_uris`, or `validate_all_uris` should be `validate_all_uris_in`

tools/airgap_linter.py
```diff
@@ -0,0 +1,136 @@
...
+    bad_uri = False
+    for uri in uris:
+        if not validate_uri(uri):
+            print("Found a bad URI:", uri, "in:", file_name)
+            bad_uri = True
+
+    return not bad_uri
+
+
+def get_all_files_to_check(framework_directory):
```
nickbp: `get_files_to_check_for_uris`? or something like that?

benclarkwood:
retest this please  (#1170 merged)

tools/airgap_linter.py
benclarkwood: I'm hesitant to change a regex that works (or at least mostly works. I'm making a minor edit to it. Let me know what you think (patch coming soon)

benclarkwood:
Holy beans! This caught something!

benclarkwood:
Of course, it was a false positive. Whoops.

tools/airgap_linter.py
benclarkwood: Yes, switched to a walk.

benclarkwood:
@jsrodman @nickbp Given that this only failed elastic due to cluster death (whee!), and it only actually touched the build phase: we good to merge it?

jsrodman:
I'm fine with it and will kick up cluster lifetime for pr runs.

PR #1184, opened 2017-06-27
[INFINITY-1853][INFINITY-1852] networknames labels cni cleanup

ArtRand:
This PR finishes the implementation of labels for CNI plugins. 

`labels` within the `NetworkInfo` on the `TaskInfo`'s `ContainerInfo` and `ExecutorInfo`'s `ContatinerInfo` are passed to the CNI plugin. 

Also change the default "port mapping" policy to `false` as "bridge-style" virtual networks seem to be the exception rather than the norm. 

Lastly, this PR contains the appropriate packaging updates for the above. 

Tested with unittests.


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -476,7 +477,15 @@ private static DefaultNetworkSpec from(
             builder.portMappings(Collections.emptyMap());
         }
 
-        //TODO(arand) implement network labels
+        if (!Strings.isNullOrEmpty(rawNetwork.getLabelsCsv())) {
+            Map<String, String> labels = new HashMap<>();
+            for (String kv : Arrays.asList(rawNetwork.getLabelsCsv().split(","))) {
+                labels.put(kv.split(":")[0], kv.split(":")[1]);
```
triclambert: Potential `ArrayIndexOutOfBoundsException`?

ArtRand: Yes! Added validator, confirmed that it will complain at compile time (instead of deployment time).

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -479,8 +479,8 @@ private static DefaultNetworkSpec from(
 
         if (!Strings.isNullOrEmpty(rawNetwork.getLabelsCsv())) {
             Map<String, String> labels = new HashMap<>();
-            for (String kv : Arrays.asList(rawNetwork.getLabelsCsv().split(","))) {
-                labels.put(kv.split(":")[0], kv.split(":")[1]);
+            for (String[] kv : rawNetwork.getValidadedLabels()) {
```
triclambert: Great, please add a test that fails as expected.

ArtRand:
retest this please (uninstall or timeout?)

ArtRand:
retest this please (jenkins snafu)

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -7,7 +7,8 @@ pods:
     placement: "{{PLACEMENT_CONSTRAINT}}"
     {{#ENABLE_VIRTUAL_NETWORK}}
     networks:
-      dcos:
+      {{VIRTUAL_NETWORK_NAME}}:
```
nickbp: should this be wrapped in a check for non-empty `{{VIRTUAL_NETWORK_NAME}}`? something like:
```
{{#VIRTUAL_NETWORK_NAME}}
  {{VIRTUAL_NETWORK_NAME}}:
    labels: {{CNI_PLUGIN_LABELS}}
{{/VIRTUAL_NETWORK_NAME}}
```
(here and below)

nickbp: oh i see in marathon.json there's a separate enabled/disabled check. nevermind.

ArtRand: No? Because we want virtual networks to be false by default, but we want the default virtual network to be `dcos`

frameworks/elastic/src/main/dist/svc.yml
```diff
@@ -6,7 +6,8 @@ pods:
     count: 3
     {{#ENABLE_VIRTUAL_NETWORK}}
     networks:
-      dcos:
+      {{VIRTUAL_NETWORK_NAME_MASTER}}:
```
benclarkwood: Why are these distinct network names? Will Elastic work if it has totally different networks for different node types? (I'd assume not?)

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -11,7 +11,8 @@ pods:
       {{#KEYTABS_URI}}- {{KEYTABS_URI}}{{/KEYTABS_URI}}
     {{#ENABLE_VIRTUAL_NETWORK}}
     networks:
-      dcos:
+      {{VIRTUAL_NETWORK_NAME_JOURNAL}}:
```
benclarkwood: Same question as Elastic.

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/DcosConstants.java
```diff
@@ -17,14 +17,14 @@
     public static final Boolean DEFAULT_GPU_POLICY = true;
     public static final String DEFAULT_IP_PROTOCOL = "tcp";
     public static final String DEFAULT_OVERLAY_NETWORK = "dcos";
+    public static final String DEFAULT_BRIDGE_NETWORK = "mesos-bridge";
     @edu.umd.cs.findbugs.annotations.SuppressFBWarnings
-    public static final Set<String> SUPPORTED_OVERLAY_NETWORKS = new HashSet<>(Arrays.asList(DEFAULT_OVERLAY_NETWORK));
+    public static final Set<String> SUPPORTED_OVERLAY_NETWORKS = new HashSet<>(
```
benclarkwood: Do we not support any other networks? What about Calico?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawNetwork.java
```diff
@@ -3,30 +3,31 @@
...
  */
 public class RawNetwork {
     private final List<Integer> hostPorts;
     private final List<Integer> containerPorts;
-    private final Map<String, String> labels;
+    private final String labelsCsv;
 
     @JsonCreator
     private RawNetwork(
             @JsonProperty("host-ports") List<Integer> hostPorts,
```
benclarkwood: We obviously aren't using this yet, so consider this non-applicable future commentary for the purposes of this PR.

Aren't these ports for doing mapping? Why are they separate lists?

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/DcosConstants.java
benclarkwood: Also, do we actually use this anywhere?

frameworks/elastic/src/main/dist/svc.yml
ArtRand: It should, actually.

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/DcosConstants.java
ArtRand: 1: No, because we don't have test clusters with anything other than `dcos` and `mesos-bridge`.
2: Yes there is a check and warning message.

frameworks/elastic/src/main/dist/svc.yml
benclarkwood: Ah right, because the other tasks are using navstar addresses.

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/DcosConstants.java
benclarkwood: Does it fail to launch? Or does it just warn the user it isn't supported?

frameworks/elastic/src/main/dist/svc.yml
benclarkwood: My only concern would be requiring the user to set a bunch of network names rather than just one.

But, this is the most configurable option so /shrug.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawNetwork.java
ArtRand: Yes, that's why it's a `Map` in `DefaultNetworkSpec`. The current (IMO, bad) yaml uses this:
```
networks:
   mesos-bridge:
        host-ports: [1025]
        container-ports: [80]
```
Thus the lists. I can change it in a another PR after input from product, but I think it's not worth re-resting this PR because it's not fully supported capability.

benclarkwood: Yup

PR #1185, opened 2017-06-27
[INFINTY-1841] Drop use of PYTHONPATH env var, and simply run pytest in-process.

jsrodman:


jsrodman:
wow, 30+ failures in helloworld.  I have no idea, but I'm paranoid so retrying.

jsrodman:
retest this please

nickbp:
that sounds like the module bringup function (installing the service) failed

PR #1186, opened 2017-06-27
[DCOS-16102] Add CockroachDB framework

springle:
Thanks for the feedback! I'm going to cancel this PR since we are moving the work out of dcos-commons.


frameworks/cockroachdb/src/main/dist/start.sh.mustache
```diff
@@ -0,0 +1,40 @@
...
+$MESOS_SANDBOX/$COCKROACH_VERSION/cockroach cert create-node \
+    localhost \
+    $(hostname) \
+    --certs-dir=certs \
+    --ca-key=my-safe-directory/ca.key
+
+# Start CockroachDB server
+$MESOS_SANDBOX/$COCKROACH_VERSION/cockroach start \
+    --logtostderr \
+    --insecure \
```
jgehrcke: :male_detective:

frameworks/cockroachdb/src/main/dist/svc.yml
```diff
@@ -0,0 +1,196 @@
...
+              port: {{COCKROACH_INTERNAL_PORT}}
+        volume:
+          path: "cockroach-data"
+          type: {{NODE_DISK_TYPE}}
+          size: {{NODE_DISK}}
+      sidecar-resources:
+        cpus: {{SIDE_CPUS}}
+        memory: {{SIDE_MEM}}
+    tasks:
+      node-init:
```
gpaul: are we sure this will never be run more than once? if this command is run in a directory where a cluster previously ran, the old cluster is effectively re-initialized (read: destroyed)

springle: This should never run more than once. If you look at the plans at the bottom of this file, the `node-init` task should only be run for Pod 0 during the deploy plan. All restores, upgrades, etc. use `node-join` exclusively to avoid re-initializing the cluster.

Fwiw, even if `node-init` were run again, this should only re-initialize the cluster in the case of a permanent failure where the persistent volume containing the cockroach-data directory is lost. Otherwise Cockroach will recover a list of the nodes to which it was connected and rejoin the cluster. But the permanent failure case is covered and would use `node-join` anyways.

LakeCarrot: About the re-initilized problem. when executing the `init` command, if it found there are some old-cluster's data in the directory it run, it will load the data instead of destroying it.

frameworks/cockroachdb/universe/config.json
```diff
@@ -0,0 +1,178 @@
...
+            "title": "Credential secret name (optional)",
+            "description": "Name of the Secret Store credentials to use for DC/OS service authentication. This should be left empty unless service authentication is needed.",
+            "type": "string",
+            "default": ""
+          },
+          "version":{
+            "title": "CockroachDB version",
+            "description": "Production release version of CockroachDB",
+            "type": "string",
+            "default": "1.0.1"
```
gpaul: 1.0.2 has been released, want to consider bumping this?

frameworks/cockroachdb/src/main/dist/svc.yml
gpaul: @LakeCarrot are you referring to https://github.com/cockroachdb/docs/issues/1288 ? That does not appear to exist yet.

frameworks/cockroachdb/src/main/dist/svc.yml
```diff
@@ -0,0 +1,196 @@
...
+        goal: FINISHED
+        configs:
+          aws:
+            template: "{{CONFIG_TEMPLATE_PATH}}/aws.sh.mustache"
+            dest: "aws.sh"
+        cmd: "./bootstrap && \
+              chmod +x $MESOS_SANDBOX/aws.sh && \
+              mkdir restore && \
+              $MESOS_SANDBOX/aws.sh s3 cp \
+              s3://$S3_BUCKET_NAME/$S3_DIR_PATH/$BACKUP_DIR/$DATABASE_NAME.sql \
```
gpaul: Do we want to sanitize the DATABASE_NAME, perhaps ?

frameworks/cockroachdb/src/main/dist/svc.yml
gpaul: @springle thanks for the response - that sounds sane.

frameworks/cockroachdb/universe/config.json
springle: done. (not pushing to this branch though, since we are moving the work out of dcos-commons)

frameworks/cockroachdb/src/main/dist/svc.yml
LakeCarrot: Thanks for pointer! @gpaul It seems that is the case. We told with cockroachDB people before and they said the `init` worked like what I said, I think they may include this function sooner. Then, the problem will be solved.

PR #1188, opened 2017-06-27
[INFINITY-1872] Plan name should always default to deploy.

ssk2:
The plan name we should always show is apparently always deploy.


benclarkwood:
retest this please

ssk2:
retest this please

PR #1189, opened 2017-06-27
INFINITY-1876 Specify minDcosVersion in package, not release script

nickbp:
The current minDcosVersion handling was only in the release script because our stub-universe.zips were stuck on universe-2.x, which didn't understand this field and which would refuse to install packages that contained it. So at the time we omitted it from stub-universe.zips and just added it when releasing them to universe.

Now that we're on universe-3.x packages in the form of stub-universe.jsons, we can specify it explicitly in the source-controlled package files. This PR also updates the release tooling to require that it already be populated, as it should currently be AT LEAST 1.8 for SDK-based services (leaving the field unset means the package can be installed back to 1.6 IIRC).

TODO separately: Remove the option/envvar from the jenkins release job as this makes the option a noop.

### DIFF
--- a/frameworks/cassandra/universe/package.json
+++ b/frameworks/cassandra/universe/package.json
@@ -1,16 +1,13 @@
 {
-  "packagingVersion": "2.0",
+  "packagingVersion": "3.0",
+  "minDcosReleaseVersion": "1.8",
   "name": "cassandra",
   "version": "{{package-version}}",
   "maintainer": "support@mesosphere.io",
-  "description": "An example implementation of a stateful service using the DC/OS SDK",
+  "description": "Apache Cassandra",
   "selected": false,
   "framework": true,
-  "tags": [
-    "example",
-    "reference",
-    "cassandra"
-  ],
+  "tags": ["cassandra"],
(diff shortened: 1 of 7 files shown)


frameworks/cassandra/universe/package.json
```diff
@@ -1,16 +1,13 @@
 {
-  "packagingVersion": "2.0",
+  "packagingVersion": "3.0",
+  "minDcosReleaseVersion": "1.8",
   "name": "cassandra",
   "version": "{{package-version}}",
   "maintainer": "support@mesosphere.io",
-  "description": "An example implementation of a stateful service using the DC/OS SDK",
+  "description": "Apache Cassandra",
```
jsrodman: Bonus!

frameworks/template/universe/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "packagingVersion": "2.0",
-  "minDcosReleaseVersion": "1.9",
+  "packagingVersion": "3.0",
+  "minDcosReleaseVersion": "1.8",
```
jsrodman: @ArtRand why was this 1.9 in the first place? 4c0f41248bc72bc028b94d879cfd6c01005fea11

nickbp:
Cluster timed out in the middle of otherwise-successful elastic tests. Given that everything else passed, and that the only change to elastic was the list of tags in the packaging, I'm gonna call this close enough.

PR #1130, opened 2017-06-17
[INFINITY-1778] Insulate sdk_install from failures

benclarkwood:
Retry install and uninstall 5 times, waiting 5 seconds between attempts. Insulates against Marathon flakes.

PR #1192, opened 2017-06-27
Revert "[INFINITY-1778] Insulate sdk_install from failures (#1130)"

benclarkwood:
This reverts commit f0a159709fcd9469e06b8c55d17e220871503b53.

### DIFF
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -22,18 +22,6 @@ def install(
         package_version=None,
         check_suppression=True,
         timeout_seconds=15 * 60):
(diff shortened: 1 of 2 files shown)


jsrodman:
Sadly, the dance of "what exceptions are good vs bad" is just a PITA.

benclarkwood:
retest this please

jsrodman:
The problem with the last run was specific to helloworld which we know has unsolved problems with startup that come and go.  essentially IMO the last run was "clean", and this is good to merge.

PR #1194, opened 2017-06-28
Add Secrets and DCOS_SPACE in Operation Guide

mbdcos:


nickbp:
- Maybe start with some examples of where a file secret or an env secret would be useful (e.g. ssl certs, signing keys, ...)
- Is this EE-only? If so, may make sense to mention it here. What happens if this is used on an Open cluster?

mbdcos:
Thank you @nick

PR #1196, opened 2017-06-28
universe add/remove via pytest

falseperfection:
This patch moves the logic that adds stub universe urls (currently in run_tests.py:setup_cli) into a pytest fixture. See: https://jira.mesosphere.com/browse/INFINITY-1884

Because fixtures and xunit setup/teardown calls don't play nicely together, it also refactors the setup_/teardown_module functions into fixtures. See: https://github.com/pytest-dev/pytest/issues/517


frameworks/cassandra/tests/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+@pytest.fixture(scope='session', autouse=True)
+def configure_universe(request):
+    stub_urls = {}
+
+    # prepare needed universe repositories
+    stub_universe_urls = os.environ.get('STUB_UNIVERSE_URL')
+    if not stub_universe_urls:
+        return
+    for url in stub_universe_urls.split(' '):
+        package_name = url.split('/')[4] # TODO: does this need to be a random string?
```
mellenburg: This is tricky, the string needs to match what is used inside the framework here, right?
https://github.com/mesosphere/dcos-commons/blob/master/frameworks/cassandra/tests/config.py#L11

In such a situation where there needs to be a shared constant and you are using pytest, I generally recommend a simple fixture like:
```
@pytest.fixture
def package_name():
    return 'cassandra'
```
but there may be more ways of doing that here

frameworks/cassandra/tests/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+        stub_urls[package_name] = url
+
+    # clear out the added universe repositores at testing end
+    def unconfigure_universe():
+        for name, url in stub_urls.items():
+            remove_package_cmd = 'dcos package repo remove {}'.format(name)
+            shakedown.run_dcos_command(remove_package_cmd)
+    request.addfinalizer(unconfigure_universe)
```
mellenburg: I find this method of teardown to be clunky and don't know why its not deprecated. I prefer using the `yield` syntax of a fixture:
```
@pytest.fixture
def configure_universe():
    # my setup code
    yield # test session/module/case happens
    # my teardown code
```

frameworks/cassandra/tests/conftest.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # clean up any duplicate repositories
+    # TODO: is this needed, or do duplicate repositories signify an actual error?
```
mellenburg: I think its necessary? We probably don't want to have a competing universe for our framework. I guess this is kinda covered by inserting the repo at position 0...

frameworks/cassandra/tests/conftest.py
```diff
@@ -0,0 +1,37 @@
+import os
+import json
+import pytest
+import shakedown
+
+
+@pytest.fixture(scope='session', autouse=True)
```
mellenburg: :+1: appropriate use of autouse I think

frameworks/cassandra/tests/conftest.py
falseperfection: But the current implementation installs with a random string, I thought.
https://github.com/mesosphere/dcos-commons/blob/master/tools/run_tests.py#L93
https://github.com/mesosphere/dcos-commons/blob/master/tools/run_tests.py#L281
https://github.com/mesosphere/dcos-commons/blob/master/tools/run_tests.py#L217

I'll have to look at your link more closely to see how these two different package names are related.

frameworks/cassandra/tests/conftest.py
falseperfection: I do it because, although prettier, using `yield` doesn't guarantee the teardown code runs. If the setup code throws an exception the teardown is skipped.

frameworks/cassandra/tests/conftest.py
falseperfection: I don't see the harm in it; I'll leave it.

frameworks/cassandra/tests/conftest.py
falseperfection: My suspicion is that the `package_name` used in the `dcos repo add` command is more correctly called the _repo_name_, and the installed repo provides a well-known package name.

falseperfection: Yes. It currently uses a random string. From the build logs:
https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=694157&_focus=988
```
[06:20:38][Step 1/1] Adding repository: testpkg-5iyxntqs https://infinity-artifacts.s3.amazonaws.com/autodelete7d/cassandra/20170627-062026-8JIXD6Ekopw6LSMu/stub-universe-cassandra.json
```

frameworks/cassandra/tests/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+    stub_universe_urls = os.environ.get('STUB_UNIVERSE_URL')
+    if not stub_universe_urls:
+        return
+    for url in stub_universe_urls.split(' '):
+        package_name = 'testpkg-'
+        package_name += ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(8))
+        stub_urls[package_name] = url
+
+    try:
+        # clean up any duplicate repositories
```
mellenburg: Cleaning up duplicates isnt bad, but why would we care if the URI is the same between them?

frameworks/cassandra/tests/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+def configure_universe(request):
+    stub_urls = {}
+
+    # prepare needed universe repositories
+    stub_universe_urls = os.environ.get('STUB_UNIVERSE_URL')
+    if not stub_universe_urls:
+        return
+    for url in stub_universe_urls.split(' '):
+        package_name = 'testpkg-'
+        package_name += ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(8))
```
mellenburg: I like having the non overlapping name, but this just needs to be coordinated with the tests

falseperfection: I'm going to change the variable name to `repo_name` because I think that's causing the confusion. There shouldn't need to be any coordination with the tests. This logic is identical to what's currently running via `run_tests.py`.

frameworks/cassandra/tests/conftest.py
falseperfection: Good question. Again: I copied the logic from `run_tests.py`. I can only speculate as to why, will need to ask about it.

frameworks/cassandra/tests/test_overlay.py
```diff
@@ -22,25 +22,27 @@
...
-
-
-def teardown_module(module):
-    install.uninstall(PACKAGE_NAME)
-
-    for job in TEST_JOBS:
-        jobs.remove_job(job)
-
+@pytest.fixture(scope='module', autouse=True)
+def configure_package():
```
mellenburg: Since we are making this a fixture, can we make this one not autouse, and just have it explicitly use `configure_universe`? E.G.
```
def configure_package(configure_universe):
    ...
```
This is technically the exact same thing as the autouse, but its a better precedent as autouse can easily get out of control

frameworks/cassandra/tests/conftest.py
mellenburg: Now that you are touching the setup functions in the test cases, this autouse isn't really appropriate as the downstream fixture can declare this one. See below.

frameworks/cassandra/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import sdk_repository as repo
+
+
+@pytest.fixture(scope='session')
+def configure_universe(request):
```
mellenburg: This doesn't need `request`, right?

testing/sdk_utils.py
```diff
@@ -1,4 +1,5 @@
 import sys
+
```
mellenburg: Nit (you REALLY don't need to respond to this): I would generally avoid making edits (even if stylistically correct) in unrelated files just to absolutely limit the sticker shock of any given diff. We will definitely need to do a huge style PR later

frameworks/elastic/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import sdk_repository as repo
+
+
+@pytest.fixture(scope='session')
+def configure_universe(request):
```
mellenburg: Not necessary?

frameworks/hdfs/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import sdk_repository as repo
+
+
+@pytest.fixture(scope='session')
+def configure_universe(request):
```
mellenburg: request not necessary?

frameworks/helloworld/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import sdk_repository as repo
+
+
+@pytest.fixture(scope='session')
+def configure_universe(request):
```
mellenburg: Not necessary?

frameworks/kafka/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import sdk_repository as repo
+
+
+@pytest.fixture(scope='session')
+def configure_universe(request):
```
mellenburg: request not necessary?

frameworks/template/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import sdk_repository as repo
+
+
+@pytest.fixture(scope='session')
+def configure_universe(request):
```
mellenburg: request not necessary?

frameworks/hdfs/tests/test_overlay.py
```diff
@@ -8,23 +8,24 @@
...
 
-def setup_module(module):
-    install.uninstall(PACKAGE_NAME)
-    utils.gc_frameworks()
-    install.install(PACKAGE_NAME, DEFAULT_TASK_COUNT,
-                    additional_options=networks.ENABLE_VIRTUAL_NETWORKS_OPTIONS)
 
-
-def setup_function(function):
+@pytest.fixture(autouse=True)
```
mellenburg: FYI, its entirely a matter of personal preference but instead of having multiple autouse fixtures, you can just have a global at the top of the test file:
```
pytestmark = pytest.mark.usefixtures('configure_package'. 'pre_test_setup')
```
See: https://docs.pytest.org/en/latest/fixture.html#usefixtures

mellenburg:
@falseperfection Overall it looks great! just had mostly inconsequential comments

testing/sdk_utils.py
falseperfection: Ha! Yeah I noticed that, I'll pull it. I had put a ton of code in there, pulled it out to `sdk_repository.py` and that change was left after the imports got cleaned out.

frameworks/cassandra/tests/conftest.py
falseperfection: Ugh. You're right. I forgot!

frameworks/hdfs/tests/test_overlay.py
falseperfection: I didn't know that. Cool. I do like having the decorator attached to the function though. I think it's a bit clearer to the next person reading the code.

frameworks/elastic/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import sdk_repository as repo
+
+
+@pytest.fixture(scope='session')
+def configure_universe():
```
falseperfection: Loops across the stub_urls defined in `STUB_UNIVERSE_URL`:

https://github.com/mesosphere/dcos-commons/blob/599380647c3d304aff1a6e2c40f0b0f97cd5813e/testing/sdk_repository.py#L44

PR #1197, opened 2017-06-28
Specify required Go version

benclarkwood:


benclarkwood:
@nickbp Do we specify the Go version elsewhere?

nickbp:
https://github.com/mesosphere/dcos-commons/blob/master/tools/build_go_exe.sh#L32

tools/build_go_exe.sh
```diff
@@ -30,19 +30,13 @@ fi
 GO_VERSION=$(go version | awk '{print $3}')
 UPX_BINARY="" # only enabled for go1.7+
```
nickbp: Now that the go-version-specific UPX checks are gone, the `UPX_BINARY` setting at L34 can be pulled up here. It'd probably make sense to leave a comment along the lines of `UPX only works on binaries produced by go1.7+, but we require go1.8+`.

Previously: (Is UPX installed PATH) AND (Is go 1.7+)
Now: (Is UPX installed in PATH)

nickbp:
Another case where a pytest failure was reported as success to github @jsrodman :

```
 generated xml file: /jenkins/workspace/dcos-commons/inf-pr/elastic-shakedown-report.xml 
==================================== ERRORS ====================================
____________________ ERROR at setup of test_service_health _____________________
...
_______________________ ERROR at setup of test_endpoints _______________________
...
others...
...
============================== 1 tests deselected ==============================
============= 5 passed, 1 deselected, 10 error in 1155.92 seconds ==============
Marathon apps: ['/test/integration/elastic']
[GH-STATUS]  success: Shakedown tests succeeded
```

nickbp:
However that said I don't think the above failure is caused by this PR if you should want to merge it.

benclarkwood:
I concur.

PR #1200, opened 2017-06-28
On pytest failure, bubble up failure.

jsrodman:


jsrodman:
The helloworld failure actually shows this fix works.

PR #1204, opened 2017-06-29
CI Refactor

mellenburg:
This PR checks in a Dockerfile that can be used to build all of the projects within this repo as well as a script that is run to test PRs. This PR removes a number of scripts that are not necessary or are overly specialized as a means to improve maintainability of this repo.


README.md
```diff
@@ -114,6 +114,33 @@ volume:
 ```
 
 ---
+### Integration Testing The Frameworks
+To run the framework integration tests hosted in this repository, use a Docker container from the [CI Dockerfile](Dockerfile) or similarly provisioned environment and do the following:
+1. Build and upload (to local or AWS) the desired framework
+  ```
+  export STUB_UNIVERSE_URL_PATH=foo/bar
+  cd frameworks/cassandra/ && ./build.sh aww
```
benclarkwood: aws

README.md
```diff
@@ -114,6 +114,33 @@ volume:
 ```
 
 ---
+### Integration Testing The Frameworks
+To run the framework integration tests hosted in this repository, use a Docker container from the [CI Dockerfile](Dockerfile) or similarly provisioned environment and do the following:
```
benclarkwood: How to use said docker container should be explicitly laid out.

jsrodman:
The sheer quantity of change here leaves me uable to evaluate whether existing desired behavior is being maintained. 

How are we retaining :
* dependcy evaluation for test execution? 
* auto validation of environment requirements? 
* testing of one or more frameworks with or without rebuilding?

jsrodman:
Does this jettison the possibility of running all the tests locally?

jsrodman:
Oh, this amount of aggressive change will need to come with a paired pull request against spark-build, which for better or worse uses some of the scripts being deleted here.

test.sh
```diff
@@ -1,80 +1,48 @@
...
-        run_framework_tests $framework
-    done
-fi
-
-# Tests succeeded. Out of courtesy, trigger a teardown of the cluster if we created it ourselves.
-# Don't wait for the cluster to complete teardown.
-if [ -n "${CLUSTER_CREATED}" ]; then
-    ${REPO_ROOT_DIR}/tools/launch_ccm_cluster.py trigger-stop ${CLUSTER_ID}
-fi
+# Create out test cluster
```
nickbp: How to run tests against an existing cluster? Common situation when testing changes locally before pushing.

test.sh
```diff
@@ -1,80 +1,48 @@
...
+dcos config set core.dcos_url $CLUSTER_URL
+dcos config set core.ssl_verify false
+tools/./dcos_login.py
+for url in `cat $UNIVERSE_URL_PATH`; do
+    dcos package repo add --index=0 `echo $url | cut -d / -f 5` $url
+done
+
+set +e
+py.test --teamcity -m "sanity" frameworks/$FRAMEWORK/tests
+dcos-launch delete
```
nickbp: How to delete the cluster only if tests succeeded and if the cluster was created by this script?

test.sh
```diff
@@ -1,80 +1,48 @@
...
+CLUSTER_URL=http://`dcos-launch describe | jq -r .masters[0].public_ip`
+dcos config set core.dcos_url $CLUSTER_URL
+dcos config set core.ssl_verify false
+tools/./dcos_login.py
+for url in `cat $UNIVERSE_URL_PATH`; do
+    dcos package repo add --index=0 `echo $url | cut -d / -f 5` $url
+done
+
+set +e
+py.test --teamcity -m "sanity" frameworks/$FRAMEWORK/tests
```
nickbp: How to specify a custom test label expression?

test.sh
```diff
@@ -1,80 +1,48 @@
...
+mkdir -p ~/.ssh/
+cat cluster_info.json | jq -r .ssh_private_key > ~/.ssh/id_rsa
+chmod 600 ~/.ssh/id_rsa
+
+# configure the dcos-cli/shakedown-backend
+CLUSTER_URL=http://`dcos-launch describe | jq -r .masters[0].public_ip`
+dcos config set core.dcos_url $CLUSTER_URL
+dcos config set core.ssl_verify false
+tools/./dcos_login.py
+for url in `cat $UNIVERSE_URL_PATH`; do
```
nickbp: An existing cluster may already have matching stub universes, this should detect and remove them (because Cosmos won't).

test.sh
```diff
@@ -1,80 +1,48 @@
...
+key_helper: true
+template_parameters:
+    AdminLocation: 0.0.0.0/0
+    PublicSlaveInstanceCount: 1
+    SlaveInstanceCount: 6
+ssh_user: core
+EOF
+
+dcos-launch create
+dcos-launch wait
```
nickbp: How often does this flake out? We currently automatically retry once when CCM flakes.

tools/cli_install.py
```diff
@@ -1,172 +0,0 @@
-#!/usr/bin/env python3
```
nickbp: Where will the CLI be installed during the test run? What version of the CLI is used?

test.sh
mellenburg: I have yet to see this ever flake out in dcos/dcos.git testing and its been running like that for a month

test.sh
mellenburg: This is being added to the test setup/teardown instead of being externalized

test.sh
mellenburg: This is not intended to be a one-size fits all entrypoint. The entrypoint should be pytest. Definitely needs better documentation on how to do setup, but the gist is that more and more of that should live with the tests so that there isnt a opaque/convoluted externalized setup

test.sh
mellenburg: This is intended only as a CI script (and discussion point for the moment). A local user should user dcos-launch directly instead of calling this

test.sh
mellenburg: That is roughly described in lines 31-44. The idea is that you will not be calling this script to call tests. Rather, you will call pytest on your tests and those tests will do the required setup rather than assuming the local dcos-cli is correctly configured.

Thus, the entire flow will be:
* use dcos-launch to create a cluster on your own
* build and upload a framework
* Run the test by providing a cluster URL, an SSH key, and the stub universe URL file path

test.sh
nickbp: Should include documentation in this PR describing how to do that. Would rather avoid people getting stuck once this is merged.

nickbp: Oh looks like it's already in the main `README.md`?

test.sh
nickbp: Some tests involve multiple stub universe paths, should probably allow for that here. In particular elastic/kibana are two separate packages.

test.sh
```diff
@@ -1,80 +1,48 @@
...
-    echo "Launched cluster: ${CLUSTER_INFO}"
-    # jq emits json strings by default: "value".  Use --raw-output to get value without quotes
-    export CLUSTER_URL=$(echo "${CLUSTER_INFO}" | jq --raw-output .url)
-    export CLUSTER_ID=$(echo "${CLUSTER_INFO}" | jq .id)
-    export CLUSTER_AUTH_TOKEN=$(echo "${CLUSTER_INFO}" | jq --raw-output .auth_token)
-    CLUSTER_CREATED="true"
-else
-    echo "Using provided CLUSTER_URL as cluster: $CLUSTER_URL"
-    CLUSTER_CREATED=""
+#!/bin/bash
```
nickbp: Sounds like this script is intended to only be used in CI. Maybe give it a CI-specific filename?

tools/run_tests.py
```diff
@@ -1,312 +0,0 @@
-#!/usr/bin/env python3
```
jsrodman: currently soak cluster automation relies on run_tests.py.  CC: @susanxhuynh 
If we drop this, we'll have to rejigger soak automation at the same time.

gabrielhartmann:
@mellenburg: Is this still needed?

benclarkwood:
I'm closing this for now, as I believe it needs a lot of renovation to move forward (if it even needs to).

PR #1214, opened 2017-07-05
Change menu order and move reference docs to new directory and menu.

ssk2:
I tidied up the top menu a little.

## Before
Documentation was a bag of different types of docs - tutorials/guides and API references:
image

## After
Documentation menu reordered such that the two guides are at the top:
image

Reference docs moved into a new subheading "Reference +" and sorted alphabetically:
image

I've also moved the reference pages under a "reference" subdirectory and updated the links within this repository. There is a re-direct for this but not for the Swagger / Javadoc.


ssk2:
Hm, this also exposes this UI quirk: 
image

Looks like it's creating empty pages for the redirected pages and including them here. I'll see if I can tweak the conditional.

PR #1215, opened 2017-07-05
[INFINITY-1911] disable overlay in strict

jsrodman:
Turn off overlay tests in strict until we can make them work.


jsrodman:
All the current pull request failures are timeout expired on uninstall.  None of these are relvant to this pull request.  Not really sure whether to wait or merge.  For now leaving this alone.

frameworks/cassandra/tests/test_overlay.py
```diff
@@ -43,6 +45,7 @@ def teardown_module(module):
 @pytest.mark.sanity
 @pytest.mark.smoke
 @pytest.mark.overlay
+@pytest.mark.skipif('os.environ.get("SECURITY") == "strict"')
```
mellenburg: Why are you using the odd, string-based syntax instead of simply entering this as python? I.E. just remove the outside single quotes

frameworks/cassandra/tests/test_overlay.py
```diff
@@ -3,6 +3,8 @@
 
 import shakedown
 
+# Do not use import *; it makes it harder to determine the origin of config
+# items
```
mellenburg: :+1: Yes, this forever and everywhere. If we can't purge these mistakes easily, then marking them until they are fixed is a good compromise

jsrodman: Your choice of red x make this a requested change.

frameworks/cassandra/tests/test_overlay.py
jsrodman: I find the pytest docs somewhat useless, so was following the odd example from shakedown's use of skipif for version numbers.  I'm not sure if they have a need for it (does it possibly delay execution?)  I don't.

frameworks/cassandra/tests/test_overlay.py
mellenburg: Not really. I did a review and the sum result of that review was a requested change. This was a comment in that review.

frameworks/cassandra/tests/test_overlay.py
mellenburg: I don't find the pytest documentation useless at all. Google "pytest skipif" will take you here, which has plenty of great examples:
https://docs.pytest.org/en/latest/skipping.html#id1

jsrodman:
As detail on pytest having bad documentation, it does not actually state that reason is a mandatory argument, while it does state that it's optional in another context several times.  This is pretty counterintuitive because keyword args are rarely required, and there is no complete reference, just a lot of discussion and partial coverage by example.  This is precisely what I feel is poor about pytest.

PR #1216, opened 2017-07-05
[SDK] Add tests for Soak cluster metronome job

mbdcos:


mbdcos:
ping @susanxhuynh

frameworks/helloworld/tests/test_soak.py
```diff
@@ -1,11 +1,74 @@
...
+    assert secret_content_alternative == task_exec(hello_tasks[0], "cat HELLO_SECRET2_FILE")
+
+    # revert back
+    cmd.run_cli("security secrets update --value=SECRET1 secrets/secret1")
+    cmd.run_cli("security secrets update --value=SECRET2 secrets/secret2")
+    cmd.run_cli("security secrets update --value=SECRET3 secrets/secret3")
+
+    # restart pods to retrieve new secret's content
+    cmd.run_cli('hello-world pods restart hello-0')
+    cmd.run_cli('hello-world pods restart world-0')
```
susanxhuynh: It would be nice to wait until restart is complete here. The reason is if something goes wrong during the restart, I would like this test to fail, not the next one.

frameworks/helloworld/tests/test_soak.py
```diff
@@ -1,11 +1,91 @@
 import pytest
+from shakedown import *
```
mellenburg: Can we please not use this pattern at all anymore? When we do 'from X import *` it makes it completely impossible to use a good linter or know where a given imported function actually came from.

I realize this pattern was set before, but its time to change the precedent and we gotta start somewhere. :D

benclarkwood: What is the pattern you would prefer to see?

Also, I will die before we do strict pep8 linting on our test code.

mbdcos: @mellenburg https://github.com/dcos/shakedown/blob/master/API.md#usage

mbdcos: will address this comment in the next PR

PR #1217, opened 2017-07-06
Add script to publish stub universe packages to Azure Blob storage

### LINKED ISSUE #1218: Add tool for uploading stub universe packages to Azure Blob Storage
For DC/OS clusters on Azure, it would be better to have an option to publish stub universe packages to an existing Blob Storage account, rather than publishing to an S3 bucket.

rozele:
When running DC/OS on Azure and using custom built DC/OS stub universe packages, it would be nice to leverage an existing Azure Blob Storage account rather than setting up an S3 bucket for hosting.

Fixes #1218

### DIFF
--- a/tools/publish_az.py
+++ b/tools/publish_az.py
@@ -1,0 +1,220 @@
+#!/usr/bin/env python3
+#
+# Uploads artifacts to Azure Blob Storage.
+# Produces a universe, and uploads it to Blob.
+# If running in jenkins ($WORKSPACE is defined), writes $WORKSPACE/stub-universe.properties
+#
+# Env:
+#   BLOB_CONTAINER (default: infinity-artifacts)
+#   BLOB_DIR_PATH (default: autdelete7d)
+#   ARTIFACT_DIR (default: ...blob.core.windows.net...)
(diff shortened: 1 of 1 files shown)


benclarkwood:
Thanks, @rozele. What do I need to setup Azure wise to test this?

rozele:
@benclarkwood You'll need to [create an Azure storage account](https://docs.microsoft.com/en-us/azure/storage/storage-create-storage-account#create-a-storage-account).  Locally, you'll need to install [Azure CLI 2.0](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli).  Once you do that, you'll need to set some environment variables like you do for the AWS CLI:
* AZURE_STORAGE_ACCOUNT - the storage account name
* AZURE_STORAGE_KEY - the storage account key
* BLOB_CONTAINER - the blob container name

You'll also need to make the blob container public, which you can do by running
```
az storage container set-permission -n $BLOB_CONTAINER --public-access blob
```

gabrielhartmann:
@benclarkwood: What's needed to check this in?

benclarkwood:
I'd like to see it successfully upload a package to Azure.

PR #1221, opened 2017-07-07
[INFINITY-1927] File path for Secrets can start with / if in a container

mbdcos:
If using an image, user may need to set for example a cert file in /etc/... 
We need to allow file path starting with /


mbdcos:
PR#1094

mbdcos:
My opinion is to make it simpler and remove relative path requirement.  We do not enforce whether an image exist or not. We do not enforce whether a user exist or not. We do not check whether task cmd is valid or not. We do not check whether user have permission to read the secret (not the file), we fail at launch. It is a scheduler. If configuration is wrong, launch or task will fail and we return the error message. Similarly, if user gives a path that he/she does not have permission, launch fails and returns an error.

PR #1225, opened 2017-07-08
[SDK] Avoid 'from X import *' in python integration tests

mbdcos:
We need to make sure integration tests with shadown.dcos_1_10 are not totally skipped...

### DIFF
--- a/frameworks/helloworld/tests/test_secrets.py
+++ b/frameworks/helloworld/tests/test_secrets.py
@@ -8,6 +8,7 @@ import sdk_install as install
 import sdk_plan as plan
 import sdk_tasks as tasks
 import sdk_marathon as marathon
+import sdk_utils as utils
 
 from tests.config import (
     PACKAGE_NAME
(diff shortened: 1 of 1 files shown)


mbdcos:
test_secrets tests are run in 1.10

frameworks/helloworld/tests/test_secrets.py
```diff
@@ -72,7 +72,7 @@ def teardown_module(module):
 @pytest.mark.sanity
 @pytest.mark.smoke
 @pytest.mark.secrets
-@dcos_1_10
+@pytest.mark.skipif('shakedown.dcos_version_less_than("1.10")')
```
benclarkwood: Cool. Can you use the annotation I created in sdk_utils for this? You just need to import shakedown (already done) and sdk_utils and it will work.

https://github.com/mesosphere/dcos-commons/blob/master/testing/sdk_utils.py#L46

benclarkwood:
Actually, I'm going through cleaning up tests right now anyways. Thanks for identifying these ones. I'll switch it to the helper function.

PR #1231, opened 2017-07-09
[HDFS-451] Add back removed constants

benclarkwood:


NimaVaziri:
The actual error is addressed in this commit as I forgot to update the referenced variable name: https://github.com/mesosphere/dcos-commons/pull/1174/commits/d1161fd7f40a1231f88d97182a03150248347b99

benclarkwood:
I don't understand. That commit seems to do the opposite of fix the error?

NimaVaziri:
Ah seemingly so. The [tests](https://teamcity.mesosphere.io/viewLog.html?buildId=702423&tab=buildResultsDiv&buildTypeId=DcosIo_DcosCommons_Hdfs_CompleteIntegrationTest#testNameId-7889024649195187153) were failing due to `NameError: name 'HDFS_CMD_TIMEOUT_SEC' is not defined` when `wait_for_failover_to_complete` was called. 

That was because I forgot to update `HDFS_CMD_TIMEOUT_SEC` to `DEFAULT_HDFS_TIMEOUT` which is what's used everywhere else. The other constants aren't needed in `test_shakedown.py` as they're imported on [L15](https://github.com/mesosphere/dcos-commons/blob/master/frameworks/hdfs/tests/test_shakedown.py#L15) from `config.py` which appropriately defines them.

benclarkwood:
Ah gotcha. I just did a `history` and added back what used to be there.

Can you open a PR (feel free to use this branch) to clean that up?

NimaVaziri:
The commit I referenced above that cleans it up is already in an open PR right now.

benclarkwood:
Gotcha

PR #1239, opened 2017-07-11
[KAF-362] Bump beta-kafka to 0.11.0.0 including protocol and log versions. 

mbdcos:


mbdcos:
This is not supposed to pass Kafka test. How possible? Are we sure upgrade test is real? It was doing write read tests after update, last time I remember. Now it just calls an sdk generic test. I do not know whether sdk_test is good or not?

gabrielhartmann:
@mbdcos:  I would add some detailed instructions for the set of operations users should perform in the install/upgrade cases.

If you're on interbroker protocol version `foo` and you are upgrading to a version with a default of version `bar` you should execute `dcos kafka --name update start --package-version=<p-q-r> --options=options.json` etc. etc. etc.

frameworks/kafka/universe/config.json
```diff
@@ -490,14 +490,14 @@
           "inter_broker_protocol_version":{
             "type":"string",
             "title":"inter.broker.protocol.version",
-            "description":"Specify which version of the inter-broker protocol will be used, which must align with log.message.format.version. This is typically bumped after all brokers were upgraded to a new version. Example of some valid values are: 0.8.0, 0.8.1, 0.8.1.1, 0.8.2, 0.8.2.0, 0.8.2.1, 0.9.0.0, 0.9.0.1, 0.10.0.0. Check ApiVersion for the full list.",
-            "default":"0.10.0.0"
+            "description":"Specify which version of the inter-broker protocol will be used, which must align with log.message.format.version. This is typically bumped serially one broker at a time, *after*all brokers were upgraded to a new version. Example of some valid values are: 0.8.0, 0.8.1, 0.8.1.1, 0.8.2, 0.8.2.0, 0.8.2.1, 0.9.0.0, 0.9.0.1, 0.10.0.0, 0.10.1.x, 0.10.2.x, 0.11.0.0. Check ApiVersion for the full list.",
```
elezar: There is a space missing in `*after*all`. Also, most other json we have (e.g. resources.json) has a space following a ":" whereas this file does not. Is there a specific reason for the difference?

gabrielhartmann:
@mbdcos: This is the PR's 1 month mark.  Is this still needed?

mbdcos:
I still need to test it with new update feature (v4 packaging). It is OK to close though but please keep the branch.

mbdcos:
Decision is to document the update process (another PR) from upgrading from 0.10 version to 0.11 version

- (step 1) Update to 0.11 version but use an options.json file to set protocol/log version to 0.10
- (step 2) Change protocol version to 0.11.0 and update again
- (step 3) Change log version to 0.11.0 and update again

At the end final updated service will be equal to 0.11 in terms of configuration

Only challenge in here is that there is nothing to prevent a user from upgrading from 0.10 version to 0.11 directly, there is nothing to enforce these three steps.

PR #1243, opened 2017-07-11
[INFINITY-1654][CASSANDRA] Set default and configurable update plan strategy

gabrielhartmann:


frameworks/cassandra/universe/config.json
```diff
@@ -51,6 +51,12 @@
           "enum": ["serial", "parallel"],
           "default": "serial"
         },
+        "update_strategy": {
+          "description": "Strategy to use for rolling out updates {serial, serial-canary}",
+          "type": "string",
+          "enum": ["serial", "serial-canary"],
```
ArtRand: why would you want to use `serial-canary` given that `serial` is an option? `serial-canary` is safer right? maybe make that the default?

PR #1245, opened 2017-07-11
INFINITY-1938 Remove TaskResource

nickbp:
Been deprecated for a while now. Removes remaining references to it in the (old) data store tutorial as well.


nickbp:
Similar story to #1247. Same pattern of HDFS flakes across both PRs.

PR #1246, opened 2017-07-11
[INFINITY-1939] Fix reservation detection

NimaVaziri:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResource.java
```diff
@@ -56,27 +56,23 @@ public Value getValue() {
...
         }
     }
 
     private String getRefinedRole() {
-        if (resource.getReservationsCount() > 0) {
-            return resource.getReservations(resource.getReservationsCount() - 1).getRole();
-        }
-
-        return Constants.ANY_ROLE;
+        return resource.getReservations(resource.getReservationsCount() - 1).getRole();
```
gabrielhartmann: in the `0` case we should still return `ANY_ROLE` to avoid relying on legacy behavior.  The role returned by `getLegacyRole` will not be a field for much longer.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
```diff
@@ -213,6 +214,7 @@ public void free(MesosResource mesosResource) {
     private void freeMergedResource(MesosResource mesosResource) {
         if (mesosResource.getResourceId().isPresent()) {
             dynamicallyReservedPoolByResourceId.remove(mesosResource.getResourceId().get());
+            logger.info("Resource still reserved? {}", dynamicallyReservedPoolByResourceId.containsKey(mesosResource.getResourceId().get()));
```
gabrielhartmann: `Unreserved resource: {}, true|false`

PR #1247, opened 2017-07-11
INFINITY-1940 Clean up label access classes

nickbp:
No actual functional changes...

- Rename SchedulerLabel[Writer|Reader] to TaskLabel[Writer|Reader]
- Rename ExecutorLabel[Writer|Reader] to ExecutorTaskLabel[Writer|Reader]
- Use underlying Label[Reader|Writer]s as a member rather than a base class, and remove TaskInfo references from there as well.
- Remove some remnants of direct TaskInfo label access (specifically some code was still doing it for a permanently-failed flag)


sdk/executor/src/main/java/com/mesosphere/sdk/offer/taskdata/ExecutorLabelWriter.java
```diff
@@ -1,19 +0,0 @@
-package com.mesosphere.sdk.offer.taskdata;
-
-/**
```
nickbp: (moved to ExecutorTaskLabelWriter -- not sure why Github isn't catching this)

nickbp:
Across two runs of HDFS tests, the only consistent failure was `test_metrics` which is known to be flaky. Calling it close enough for a refactor.

PR #1249, opened 2017-07-12
[WIP]  Experimenting with sdk upgrade downgrade test

mbdcos:


mbdcos:
It passed the  sdk upgrade downgrade test

PR #1250, opened 2017-07-12
make test.sh use pytest directly

falseperfection:
Streamline and update test.sh
https://jira.mesosphere.com/browse/INFINITY-1936

Based on INFINITY-1884 (ie, requires the new pytest-fixture method of stub-universe configuration).

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,80 +1,62 @@
 #!/usr/bin/env bash
 
-# This file contains logic for integration tests which are executed by CI upon pull requests to
-# dcos-commons. The script builds each framework, packages and uploads it, then runs its
-# integration tests against a newly-launched cluster.
+# Build a framework, package, upload it, and then run its integration tests.
+# (Or all frameworks depending on arguments.) Depends on a cluster (identified by CLUSTER_URL).
 
-# Exit immediately on errors -- the helper scripts all emit github statuses internally
+# Exit immediately on errors
 set -e
 
-function run_framework_tests {
-    framework=$1
-    FRAMEWORK_DIR=${REPO_ROOT_DIR}/frameworks/${framework}
+REPO_ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+FRAMEWORK_LIST=$(ls $REPO_ROOT_DIR/frameworks | sort)
 
-    # Build/upload framework scheduler artifact if one is not directly provided:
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -1,80 +1,62 @@
...
-    CLUSTER_CREATED=""
-fi
+    echo "Starting build for $framework at "`date`
+    export UNIVERSE_URL_PATH=${FRAMEWORK_DIR}/$framework-universe-url
+    ${FRAMEWORK_DIR}/build.sh aws
+    if [ ! -f "$UNIVERSE_URL_PATH" ]; then
+        echo "Missing universe URL file: $UNIVERSE_URL_PATH"
+        exit 1
+    fi
+    export STUB_UNIVERSE_URL=$(cat $UNIVERSE_URL_PATH)
```
nickbp: Could skip the build if STUB_UNIVERSE_URL is already defined. Imagining a case where someone's using this while working on tests (but without needing a rebuild each time). That said, I guess they could also just invoke py.test manually too.. any preference?

falseperfection: I took that block of logic out because I wasn't sure how to handle it for the case where all the frameworks were being tested. Would there just be one STUB_UNIVERSE_URL for them all? Or would one need to be defined for each separately? As soon as it seemed like there would be more complex modes to the script I opted for simplicity.

nickbp: FWIW it's also possible for a framework to have multiple stub universes -- elastic+kibana specifically. There's also other complications coming soon*

Maybe the way to go would be to have the build+repoadd be optional. A tester may already have the cluster in the state they want (including added stub universes, if any)

* It's looking likely that we'll end up needing to have separate stub universe outputs for 1.9 vs 1.10 due to the universe-4.x schism. Despite universe 4.x really just being a 3.1, with a couple fields added, they decided to make it mutually incompatible where 1.9 refuses to use 4.x and 1.10 refuses to honor to the additional 4.x fields if it's 3.x.

falseperfection: stub-universes are about to get added via pytest fixtures (see #1196). That simplification, plus making cluster spin-up external to the test script were the impetus for these changes.

benclarkwood: `test.sh` should definitely still support skipping the build step.

Can we just make setting STUB_UNIVERSE_URL mutually exclusive with running `all`?

falseperfection: Will do.

PR #1252, opened 2017-07-12
[INFINITY-1950] Validate no user change across service deployments

NimaVaziri:
From a Kafka deployment trying to prevent a user change:
```
WARN  2017-07-12 23:38:30,731 [main] com.mesosphere.sdk.config.DefaultConfigurationUpdater:updateConfiguration(110): New configuration failed validation against current target configuration 09586d73-4bac-4dcb-8bae-f42dd2e212f0, with 1 errors across 6 validators:
1: Field: 'user'; Transition: 'nobody' => 'root'; Message: 'Service user must remain the same across deployments. Expected: 'nobody', given: 'root''
```


sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/UserCannotChange.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public Collection<ConfigValidationError> validate(Optional<ServiceSpec> oldConfig, ServiceSpec newConfig) {
+        if (!oldConfig.isPresent()) {
+            return Collections.emptyList();
+        }
+
+        // user is uniform across pods...
+        String oldUser = oldConfig.get().getPods().get(0).getUser().get();
+        String newUser = newConfig.getPods().get(0).getUser().get();
```
gabrielhartmann: Can't just check the first pod.  Need to match up pods and make sure each individual pod doesn't change the user.

sdk/scheduler/src/test/java/com/mesosphere/sdk/config/validate/UserCannotChangeTest.java
```diff
@@ -0,0 +1,80 @@
...
+        MockitoAnnotations.initMocks(this);
+        when(mockOldPodSpec.getUser()).thenReturn(Optional.of(USER_A));
+        when(mockOldPodSpec.getType()).thenReturn(POD_TYPE_A);
+
+        when(mockNewPodSpec.getType()).thenReturn(POD_TYPE_B);
+    }
+
+    @Test
+    public void testSameUser() {
+        when(mockNewPodSpec.getUser()).thenReturn(Optional.of(USER_A));
```
gabrielhartmann: `when(mockNewPodSpec.getUser()).thenReturn(mockOldPodSpec.getUser());` is a bit neater and shows the intent better I think

sdk/scheduler/src/test/java/com/mesosphere/sdk/config/validate/UserCannotChangeTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+import static org.mockito.Mockito.when;
+
+public class UserCannotChangeTest {
```
gabrielhartmann: Moar tests:
1. compare pods not setting any user
1. compare multiple pods
1. compare pod with no user with one with a suer

sdk/scheduler/src/test/java/com/mesosphere/sdk/config/validate/UserCannotChangeTest.java
NimaVaziri: Although that does show the intent better, the testing library warns `thenReturn() may be missing` which makes sense, but I guess it's not smart enough to realize that it _is_ actually set via the before decorator.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/UserCannotChange.java
```diff
@@ -0,0 +1,98 @@
...
+        PodSpec oldPod = null, newPod;
+        String oldUser = "", newUser;
+
+        while (oldPods.hasNext() && newPods.hasNext()) {
+            oldPod = oldPods.next();
+            newPod = newPods.next();
+
+            // in case old user is set but not new user or reverse
+            if (!userStatesMatch(oldPod, newPod, errors)) {
+               return errors;
```
gabrielhartmann: Don't short circuit.  Return all errors on all pods.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/UserCannotChange.java
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+        // If the new config has more pods than the previous config than the new pods can set the user
+        // however the developer intends to. We're only concerned about not updating the user for existing pods
+        // in this validation.
+
+        return errors;
+    }
+
+    private boolean userStatesMatch(PodSpec oldPod, PodSpec newPod, List<ConfigValidationError> errors) {
```
gabrielhartmann: This should just return a list of errors.  Empty list, means no errors.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/UserCannotChange.java
```diff
@@ -0,0 +1,98 @@
...
+                                    oldPod.getUser().get(),
+                                    null)
+                    ));
+            return false;
+        }
+
+        return true;
+    }
+
+    private void checkForUserEquality(PodSpec oldPod, PodSpec newPod, List<ConfigValidationError> errors) {
```
gabrielhartmann: Probably don't need both methods.  Create a single method that handles Optional.  I think there's no need for the special logic here.

https://stackoverflow.com/questions/37004579/convenient-way-of-checking-equality-for-optionals

sdk/scheduler/src/test/java/com/mesosphere/sdk/config/validate/UserCannotChangeTest.java
gabrielhartmann: k

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/UserCannotChange.java
```diff
@@ -0,0 +1,98 @@
...
+
+        List<ConfigValidationError> errors = new ArrayList<>();
+
+        // The corresponding old and new pods must have the same user.
+        Iterator<PodSpec> oldPods = oldConfig.get().getPods().iterator();
+        Iterator<PodSpec> newPods = newConfig.getPods().iterator();
+        PodSpec oldPod = null, newPod;
+        String oldUser = "", newUser;
+
+        while (oldPods.hasNext() && newPods.hasNext()) {
```
gabrielhartmann: I don't think you should rely on oldPods and newPods being in the same order.  Consider in particular the addition of a new Pod at the beginning of the list.  This would throw things off.  Probably need to use Maps.

NimaVaziri:
@gabrielhartmann Addressed comments

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/UserCannotChange.java
```diff
@@ -0,0 +1,54 @@
...
+        Map<String, PodSpec> oldPods = new HashMap<>();
+        for (PodSpec oldPod : oldConfig.get().getPods()) {
+            oldPods.put(oldPod.getType(), oldPod);
+        }
+
+        for (PodSpec newPod : newConfig.getPods()) {
+            // If a new pod type is introduced, we don't validate how its user is set as we're only concerned
+            // about not updating the user for existing pods in this validation.
+            if (oldPods.containsKey(newPod.getType())) {
+                checkForUserEquality(oldPods.get(newPod.getType()), newPod, errors);
```
gabrielhartmann: Don't pass in the `errors` list to be modified.  Have the method return a list of errors.

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -261,8 +261,3 @@ def fn():
     # Verify ZK is unchanged
     zk_config_new = shakedown.get_zk_node_data(zk_path)
     assert zk_config_old == zk_config_new
-
-@pytest.mark.upgrade
-@pytest.mark.sanity
-def test_upgrade_downgrade():
-    sdk_test_upgrade.upgrade_downgrade(PACKAGE_NAME, PACKAGE_NAME, DEFAULT_TASK_COUNT)
```
nickbp: Is this expected to fail as of this PR? If so the test should probably just be disabled instead of deleted

NimaVaziri: Oh no it's just duped, the exact same test exists [here](https://github.com/mesosphere/dcos-commons/blob/master/frameworks/helloworld/tests/test_upgrade.py#L13)

PR #1254, opened 2017-07-13
Airgap linter should ignore id lines in files.

ssk2:
This should catch lines that contain `"id":`, which should just get JSON lines.

### DIFF
--- a/tools/airgap_linter.py
+++ b/tools/airgap_linter.py
@@ -23,6 +23,9 @@ def extract_uris(file_name):
         # Do not grab comments
         if line.startswith("*") or line.startswith("#") or line.startswith("//"):
             continue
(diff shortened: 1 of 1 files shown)


tools/airgap_linter.py
```diff
@@ -23,6 +23,9 @@ def extract_uris(file_name):
         # Do not grab comments
         if line.startswith("*") or line.startswith("#") or line.startswith("//"):
             continue
+        # Do not grab "id" lines
+        if "\"id\": " in line:
```
benclarkwood: LGTM. I'd probably not make it match on space.

ssk2: Good point. Amended.

PR #1255, opened 2017-07-13
INFINITY-1110 Clean up port offer evaluation: Remove task update, use DiscoveryInfo names

nickbp:
Ports without VIPs are now correctly grouped by port name rather than task type under `endpoints` lookup. Now a VIP is no longer needed to appear in `endpoints`. The port name field itself recently started being set as of PR #1149.

Other cleanup relating to port/VIP handling:
- For dynamic port lookup, prefer searching the DiscoveryInfo for a matching port name, instead of the task envvars. Leave envvar lookup as a temporary fallback for upgrade scenarios where the port name isn't set in the TaskInfo.
- Remove update logic from both port and VIP evaluation. TaskInfos are now created from scratch every time so any update logic is effectively dead code.
- Remove two YAML spec VIP fields which had long-standing TODOs to be removed. Add a temporary throw to explicitly complain if the fields are still present.


nickbp:
Going to reduce this PR to not change the current default behavior (additional non-VIP entries in `endpoints`) pending PM input. Changes to behavior, if any, will be done in a separate PR. Please hold..

nickbp:
OK back to prior default behavior.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -105,11 +117,35 @@ public void setExecutorBuilder(Protos.ExecutorInfo.Builder executorBuilder) {
         this.executorBuilder = executorBuilder;
     }
 
-    public Optional<String> getLastTaskEnv(String taskSpecName, String envName) {
+    /**
+     * This is the only carry-over from old tasks: If a port was dynamically allocated, we want to avoid reallocating
+     * it on task relaunch.
+     */
+    public Optional<Long> getTaskLastPort(String taskSpecName, String portName, String portEnvName) {
```
ArtRand: Maybe s/`getTaskLastPort`/`lookupTaskPortValue` ?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -158,37 +122,32 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
...
-                    portsBuilder.addPorts(makePortBuilder(port));
 
-                }
-            } else {
-                Protos.DiscoveryInfo.Builder discoveryInfoBuilder = Protos.DiscoveryInfo.newBuilder()
-                        .setVisibility(Protos.DiscoveryInfo.Visibility.FRAMEWORK)
+            if (!taskBuilder.hasDiscovery()) {
+                // Initialize with defaults:
+                taskBuilder.getDiscoveryBuilder()
+                        .setVisibility(Constants.DEFAULT_TASK_DISCOVERY_VISIBILITY)
```
ArtRand: The default for Marathon is `FRAMEWORK` visibility for ports to generate SRV records. We only increase visibility when there is a VIP on a port. IIUC to replicate Marathon's behavior the logic is:
port has VIP -> EXTERNAL
port does not have vip -> FRAMEWORK

there is a CI test for generating SRV records, so if that passes then this is all moot I suppose, but with noting in case we hit an error later. 

Looks like you have the visibility in the YAML?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -539,45 +526,48 @@ private static PortsSpec from(
...
                 final String vipName = StringUtils.isEmpty(rawVip.getPrefix()) ? name : rawVip.getPrefix();
+                RawPort matchingRawPort = rawPorts.get(vipName);
+                if (matchingRawPort != null && matchingRawPort != rawPort) {
+                    throw new IllegalArgumentException(String.format(
+                            "Provided VIP prefix '%s' in port '%s' conflicts with other port also named '%s'. " +
+                            "Expected VIP prefix to not collide with other ports' names.",
+                            vipName, name, vipName));
+                }
+                //TODO(nickbp): Ideally we would also check here that multiple VIPs don't share names with each other.
+                //              However elastic currently does this. Not sure why.
```
ArtRand: Does that work?!

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -41,7 +42,9 @@
     private final Map<String, Protos.TaskInfo.Builder> taskBuilders = new HashMap<>();
     private Protos.ExecutorInfo.Builder executorBuilder;
     private final PodInstance podInstance;
+    // TODO(nickbp): Remove this env storage after October 2017 when it's no longer used as a fallback for dynamic ports
```
nickbp: The env storage is _only_ currently used to autodetect prior values for dynamic ports. Once we've had a couple dcos-commons releases where stored TaskInfos have the port name set, it can be removed in favor of that mechanism instead.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
nickbp: It does work, but at the moment it results in somewhat confusing endpoints output. I will actually remove this TODO as elastic's behavior (having all 'data' nodes show up under a 'data' VIP hostname) does make some sense.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
nickbp: The YAML has an advertise bool but it's under the VIP. It's removed in this PR as it's not actually being used anywhere (it defaults to 'true' anyway, which is what you'd want if you were going to the trouble of creating a VIP). It's going to be promoted to the Port in a following behavior-changing PR.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
nickbp: SGTM. Went with `lookupPriorTaskPortValue`

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointsResource.java
```diff
@@ -188,28 +183,31 @@ public Response getEndpoint(
...
      *
+     * @param serviceName the name of the parent service
      * @param endpointsByName the map to write to
-     * @param taskInfo the task which has the port
-     * @param dnsHostPort the host:port value to advertise for directly connecting to the task
+     * @param taskInfo the task which has the port in question
+     * @param autoipHostPort the host:port value to advertise for connecting to the task over DNS
+     * @param ipHostPort the host:port value to advertise for connecting to the task's IP
      * @param portLabels list of any {@link Label}s which were present in the {@link Port}
      * @throws TaskException if no VIPs were found and the task type couldn't be extracted
```
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointsResource.java
```diff
@@ -188,28 +183,31 @@ public Response getEndpoint(
...
-            String addressHostPort,
-            List<Label> portLabels) throws TaskException {
-        // Search for any VIPs to add the above host:port against:
+            Port taskInfoPort,
+            String autoipHostPort,
+            String ipHostPort) throws TaskException {
+        // Search for any VIPs to list the port against:
         boolean foundAnyVips = false;
-        for (Label label : portLabels) {
+        for (Label label : taskInfoPort.getLabels().getLabelsList()) {
```
nickbp: Broke out into separate functions

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
ArtRand: Ok. I guess what I'm asking is that, in this refactor, if I specify a port and I don't want it advertised what is the VISIBILITY? In the past, it was FRAMEWORK, looks like it will be EXTERNAL now?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
ArtRand: 👍

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
nickbp: If visibility == EXTERNAL then advertise, else don't

nickbp:
```
[23:02:23][Step 4/4] frameworks/helloworld/tests/test_web_url.py::test_deploy 
[23:02:23][Step 4/4] frameworks.helloworld.tests.test_web_url.test_deploy (15s)
[23:02:38][Step 4/4] The build PR Jobs::Hello World PR #269 {buildId=706942} has been running for more than 130 minutes. Terminating...
[23:02:38][Step 4/4] Execution timeout
```

??

benclarkwood:
That should be fixed. Re-running.

PR #1259, opened 2017-07-14
[CASSANDRA-628] Allow configuration of scheduler log level

gabrielhartmann:
I tested it with `DEBUG` in `hello-world` and it did what I expect.

### DIFF
--- a/frameworks/cassandra/universe/config.json
+++ b/frameworks/cassandra/universe/config.json
@@ -55,6 +55,11 @@
           "description": "Enable DC/OS Virtual Networking",
           "type": "boolean",
           "default": false
+        },
+        "log_level": {
+          "description": "The log level for the DC/OS service.",
+          "type": "string",
+          "default": "INFO"
         }
       },
       "required": [
--- a/frameworks/cassandra/universe/config.json
+++ b/frameworks/cassandra/universe/config.json
@@ -59,6 +59,7 @@
         "log_level": {
           "description": "The log level for the DC/OS service.",
           "type": "string",
+          "enum": ["OFF", "FATAL", "ERROR", "WARN", "INFO", "DEBUG", "TRACE", "ALL"],
           "default": "INFO"
         }
       },
--- a/frameworks/elastic/universe-kibana/config.json
+++ b/frameworks/elastic/universe-kibana/config.json
@@ -18,6 +18,7 @@
         "log_level": {
           "description": "The log level for the DC/OS service.",
           "type": "string",
+          "enum": ["OFF", "FATAL", "ERROR", "WARN", "INFO", "DEBUG", "TRACE", "ALL"],
           "default": "INFO"
         }
       },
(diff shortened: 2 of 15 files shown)


frameworks/cassandra/universe/config.json
```diff
@@ -55,6 +55,11 @@
           "description": "Enable DC/OS Virtual Networking",
           "type": "boolean",
           "default": false
+        },
+        "log_level": {
```
benclarkwood: This should have an enforcement that restricts you to the strings logback actually understands (ERROR, DEBUG, WARN, INFO, TRACE?)

Also, should we only allow DEBUG and INFO? Customers might elect for a more restrictive log level than we want them to have.

sdk/common/src/main/resources/log4j2.xml
```diff
@@ -4,15 +4,14 @@
...
-
     <Appenders>
         <Console name="STDOUT" target="SYSTEM_OUT">
             <PatternLayout pattern="${PATTERN}"/>
         </Console>
     </Appenders>
 
     <Loggers>
-        <Root level="INFO">
+        <Root level="${env:FRAMEWORK_LOG_LEVEL}">
```
benclarkwood: This should have a default specified, in case a framework developer effs up: https://stackoverflow.com/questions/22028399/log4j2-property-substitution-default

frameworks/cassandra/universe/config.json
gabrielhartmann: I thought of this and opted not to make a decision by placing no restrictions.  Might be handy to be able do whatever we want in an emergency.  Changing this setting is basically an emergency operation.  So if you want to restrict to an enum I'd go with allowing all options.

benclarkwood: Yeah, let's do an enum with all options. I'd rather not suffer no logs due to fat fingering.


OFF
FATAL
ERROR
WARN
INFO
DEBUG
TRACE
ALL

from: https://logging.apache.org/log4j/2.x/manual/customloglevels.html

gabrielhartmann:
@benclarkwood:  Addressed comments.

frameworks/elastic/universe-kibana/config.json
```diff
@@ -14,6 +14,12 @@
           "description": "The user that runs the Kibana services and owns the Mesos sandbox.",
           "type": "string",
           "default": "nobody"
+        },
```
loren: Does this have any effect for Kibana as it's just a Marathon app and does not use the SDK?

benclarkwood: That is a good question. Depends how the Marathon app is configured I think? I think Marathon supports users. Not sure tho.

loren: Oops, that wasn't my question...somehow the comment got attached to the wrong line. I was asking about the relevance of log4j2 configuration w/r/t a Marathon app.

benclarkwood: Oh yes, that's irrelevant.

benclarkwood: (and should be removed)

PR #1262, opened 2017-07-14
[INFINITY-1968] Switch default to v0 for mesos api in HW

benclarkwood:
Create a v1 test (so that we have something that fails in strict)

### DIFF
--- a/frameworks/helloworld/tests/test_sanity.py
+++ b/frameworks/helloworld/tests/test_sanity.py
@@ -2,6 +2,7 @@ import json
 import re
 
 import dcos.marathon
+import os
 import pytest
 import shakedown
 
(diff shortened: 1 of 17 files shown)


frameworks/helloworld/tests/test_sanity.py
```diff
@@ -48,6 +49,24 @@ def test_install():
     check_running(FOLDERED_SERVICE_NAME)
 
 
+# Note: presently the mesos v1 api does _not_ work in strict mode.
+# As such, we expect this test to fail until it does in fact work in strict mode.
+@pytest.mark.sanity
+@pytest.mark.skipif(os.environ.get("SECURITY") != "strict")
```
nickbp: Should this be "== strict" ?

benclarkwood: I'd originally meant "We should run this in strict and see that it fails"

But, we should actually run this in all modes.

PR #1263, opened 2017-07-15
Example has single volume, but multiple volumes field name.

benclarkwood:


keithchambers:
It makes sense to deprecate `volume` in favor of `volumes` in the future.

Not sure what the motivation is for this change, but if the goal is to make all the examples consistent I'd recommend we go with `volumes`.

benclarkwood:
@keithchambers The example was incorrect, and I went with the simplest change to make it correct (volumes -> volume).

nickbp:
IIRC the reason for having `volume` was for simplifying the common case where tasks only needed a single volume. This is as opposed to say `ports` where 1. multiple of them is more common 2. the names are significant and are used for identifying the port

PR #1269, opened 2017-07-17
[CASSANDRA-631] Add an Offer Queue for more reliable Offer Processing

gabrielhartmann:
**Problem**:
We implicitly had a queue for processing Offers.  As each bundle of Offers came in we processed that bundle on a single thread of an executor service.  If another offer bundle came in it was queued.  In a large cluster it's possible to receive offers faster than we can process them.

In a scale test on a large cluster we saw that we attempted to consume Offers which had been in the queue so long that they were rescinded by Mesos.  We tried to process them anyway resulting in problems, but that's a different topic (https://jira.mesosphere.com/browse/INFINITY-1964).

**Solution**:
Build an explicit bounded queue (buffer) for offers.  We bound the queue to 100 offers by default.  This isn't currently configurable at runtime.  Until the need arises I'd like to keep it this way.  Offers are pushed onto the queue synchronously if there is capacity.  Placement of Offers into the in memory queue should happen pretty much instantaneously, so a separate thread isn't needed.  Another thread pulls offers off the queue as fast as it can.  On each iteration, this thread pulls _ALL_ offers out of the queue, so it processes 100 offers at a time at maximum.

**Test**:
Deployment of a 100 node Cassandra cluster failed on node-52 after ~3-4 hours of deploying without this change.  A 9 hour deployment of all 100 nodes succeeded with this change.

Note:
Please ignore the log level related changes.  They're based on an as yet unrelated but unmerged PR.  It'll go in before this PR.  It's somewhat related because somethings have been moved to `debug` log level.  It was observed that logs were pretty hard to follow in a very large service (C* with 100 nodes).

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
@@ -23,7 +23,7 @@ public abstract class AbstractScheduler implements Scheduler {
     // Mesos may call registered() multiple times in the lifespan of a Scheduler process, specifically when there's
     // master re-election. Avoid performing initialization multiple times, which would cause resourcesQueue to be stuck.
     private final AtomicBoolean isAlreadyRegistered = new AtomicBoolean(false);
-    protected final OfferQueue offerQueue;
+    protected final OfferQueue offerQueue = new OfferQueue();
     protected SchedulerDriver driver;
     protected DefaultReconciler reconciler;
 
@@ -32,7 +32,6 @@ public abstract class AbstractScheduler implements Scheduler {
      */
     protected AbstractScheduler(StateStore stateStore) {
         this.stateStore = stateStore;
-        this.offerQueue = new OfferQueue();
     }
 
     @Override
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
@@ -1,6 +1,8 @@
 package com.mesosphere.sdk.scheduler;
 
+import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.TextFormat;
+import com.mesosphere.sdk.offer.OfferUtils;
 import com.mesosphere.sdk.reconciliation.DefaultReconciler;
 import com.mesosphere.sdk.state.StateStore;
 import com.mesosphere.sdk.state.StateStoreUtils;
@@ -10,7 +12,11 @@ import org.apache.mesos.SchedulerDriver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.*;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 
 /**
  * Abstract {@link Scheduler} that provides some default behaviors around Mesos lifecycle events such as
@@ -26,12 +32,25 @@ public abstract class AbstractScheduler implements Scheduler {
     protected final OfferQueue offerQueue = new OfferQueue();
     protected SchedulerDriver driver;
     protected DefaultReconciler reconciler;
+    private Set<Protos.OfferID> offersInProgress = new HashSet<>();
+
+    /**
+     * Executor for handling TaskStatus updates in {@link #statusUpdate(SchedulerDriver, Protos.TaskStatus)}.
+     */
+    protected final ExecutorService statusExecutor = Executors.newSingleThreadExecutor();
+
+    /**
+     * Executor for processing offers off the queue in {@code processOffers()}.
+     */
+    protected final ExecutorService offerExecutor = Executors.newSingleThreadExecutor();
+
 
     /**
      * Creates a new AbstractScheduler given a {@link StateStore}.
      */
     protected AbstractScheduler(StateStore stateStore) {
         this.stateStore = stateStore;
+        processOffers();
     }
 
     @Override
@@ -64,7 +83,72 @@ public abstract class AbstractScheduler implements Scheduler {
         postRegister();
     }
 
+    private void processOffers() {
+        offerExecutor.execute(() -> {
+            while (true) {
+                // This is a blocking call which pulls as many elements from the offer queue as possible.
+                List<Protos.Offer> offers = offerQueue.takeAll();
+                processOfferSet(offers);
+                offersInProgress.removeAll(
+                        offers.stream()
+                                .map(offer -> offer.getId())
+                                .collect(Collectors.toList()));
+            }
+        });
+    }
+
+    protected abstract void processOfferSet(List<Protos.Offer> offers);
+
+    @Override
+    public void resourceOffers(SchedulerDriver driver, List<Protos.Offer> offers) {
+        if (!apiServerReady()) {
+            LOGGER.info("Waiting for API Server to start ...");
+            OfferUtils.declineOffers(driver, offers);
+            return;
+        }
+
+        // Task Reconciliation:
+        // Task Reconciliation must complete before any Tasks may be launched.  It ensures that a Scheduler and
+        // Mesos have agreed upon the state of all Tasks of interest to the scheduler.
+        // http://mesos.apache.org/documentation/latest/reconciliation/
+        reconciler.reconcile(driver);
+        if (!reconciler.isReconciled()) {
+            LOGGER.info("Waiting for task reconciliation to complete...");
+            OfferUtils.declineOffers(driver, offers);
+            return;
+        }
+
+        offersInProgress.addAll(
+                offers.stream()
+                        .map(offer -> offer.getId())
+                        .collect(Collectors.toList()));
+
+        for (Protos.Offer offer : offers) {
+            boolean queued = offerQueue.offer(offer);
+            if (!queued) {
+                LOGGER.warn("Failed to enqueue offer: '{}'", offer.getId().getValue());
+                OfferUtils.declineOffers(driver, Arrays.asList(offer));
+            }
+        }
+    }
+
     protected abstract void initialize(SchedulerDriver driver) throws InterruptedException;
+
+    protected abstract boolean apiServerReady();
+
+    /**
+     * All offers must have been presented to resourceOffers() before calling this.
+     * @throws InterruptedException
+     */
+    @VisibleForTesting
+    public void awaitOffersProcessed() throws InterruptedException {
+        while (!offersInProgress.isEmpty()) {
+            LOGGER.warn("Offers to be processed {} is non empty, sleeping for 1s ...", offersInProgress);
+            Thread.sleep(1000);
+        }
+
+        LOGGER.info("All offers processed.");
+    }
 
     @Override
     public void reregistered(SchedulerDriver driver, Protos.MasterInfo masterInfo) {
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
@@ -23,6 +23,7 @@ public abstract class AbstractScheduler implements Scheduler {
     // Mesos may call registered() multiple times in the lifespan of a Scheduler process, specifically when there's
     // master re-election. Avoid performing initialization multiple times, which would cause resourcesQueue to be stuck.
     private final AtomicBoolean isAlreadyRegistered = new AtomicBoolean(false);
+    protected final OfferQueue offerQueue;
     protected SchedulerDriver driver;
     protected DefaultReconciler reconciler;
 
@@ -31,6 +32,7 @@ public abstract class AbstractScheduler implements Scheduler {
      */
     protected AbstractScheduler(StateStore stateStore) {
         this.stateStore = stateStore;
+        this.offerQueue = new OfferQueue();
     }
 
     @Override
@@ -73,7 +75,8 @@ public abstract class AbstractScheduler implements Scheduler {
 
     @Override
     public void offerRescinded(SchedulerDriver driver, Protos.OfferID offerId) {
-        LOGGER.warn("Ignoring rescinded Offer: {}.", offerId.getValue());
+        LOGGER.info("Rescinding offer: {}.", offerId.getValue());
+        offerQueue.remove(offerId);
     }
 
     @Override
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
@@ -1,6 +1,6 @@
 package com.mesosphere.sdk.scheduler;
 
-import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.TextFormat;
 import com.mesosphere.sdk.offer.OfferUtils;
 import com.mesosphere.sdk.reconciliation.DefaultReconciler;
@@ -12,7 +12,10 @@ import org.apache.mesos.SchedulerDriver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -32,6 +35,8 @@ public abstract class AbstractScheduler implements Scheduler {
     protected final OfferQueue offerQueue = new OfferQueue();
     protected SchedulerDriver driver;
     protected DefaultReconciler reconciler;
+
+    private Object inProgressLock = new Object();
     private Set<Protos.OfferID> offersInProgress = new HashSet<>();
 
     /**
@@ -42,8 +47,7 @@ public abstract class AbstractScheduler implements Scheduler {
     /**
      * Executor for processing offers off the queue in {@code processOffers()}.
      */
-    protected final ExecutorService offerExecutor = Executors.newSingleThreadExecutor();
-
+    private final ExecutorService offerExecutor = Executors.newSingleThreadExecutor();
 
     /**
      * Creates a new AbstractScheduler given a {@link StateStore}.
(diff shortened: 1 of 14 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -23,6 +23,7 @@
     // Mesos may call registered() multiple times in the lifespan of a Scheduler process, specifically when there's
     // master re-election. Avoid performing initialization multiple times, which would cause resourcesQueue to be stuck.
     private final AtomicBoolean isAlreadyRegistered = new AtomicBoolean(false);
+    protected final OfferQueue offerQueue;
```
nickbp: Should offers be getting queued in `UninstallScheduler`? From its perspective, offerRescinded will attempt to remove offers that aren't being added. Would it make sense for `UninstallScheduler` to take advantage of this as well?

frameworks/cassandra/universe/config.json
```diff
@@ -55,6 +55,12 @@
           "description": "Enable DC/OS Virtual Networking",
           "type": "boolean",
           "default": false
+        },
+        "log_level": {
+          "description": "The log level for the DC/OS service.",
```
nickbp: This name and description imply that it'll carry through to the underlying service. Maybe go with `scheduler_log_level`/`The log level for the service scheduler` or similar

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -61,7 +60,11 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultScheduler.class);
 
-    protected final ExecutorService executor = Executors.newFixedThreadPool(1);
+    /**
+     * 1 Thread for processing offers off the queue in {@link #processOffers()}
```
nickbp: With this structure, both threads could be used by a single operation (e.g. two concurrent processOffers() calls taking up the whole pool). Would it make sense to have two isolated queues?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -587,12 +600,22 @@ protected DefaultScheduler(
...
+        processOffers();
     }
 
     @VisibleForTesting
-    void awaitTermination() throws InterruptedException {
-        executor.shutdown();
-        executor.awaitTermination(AWAIT_TERMINATION_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+    void awaitOffersProcessed() throws InterruptedException {
+        while (!offerQueue.isEmpty()) {
+            LOGGER.warn("Queue is NOT empty");
```
nickbp: Could include `... sleeping for 1s` in these messages

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * This class acts as a buffer of Offers from Mesos.  By default it holds a maximum of 100 Offers.
+ */
+public class OfferQueue {
+    public static final int DEFAULT_CAPACITY = 100;
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private final BlockingQueue<Protos.Offer> queue;
+
+    public OfferQueue () {
```
nickbp: extra space. `OfferQueue()`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.stream.Collectors;
+
+/**
+ * This class acts as a buffer of Offers from Mesos.  By default it holds a maximum of 100 Offers.
+ */
+public class OfferQueue {
+    public static final int DEFAULT_CAPACITY = 100;
```
nickbp: Could be package-private and `@VisibleForTesting`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Calling this method is a blocking Operation. It returns all Offers currently in the queue. It always returns at
+     * least one Offer.
+     */
+    public List<Protos.Offer> takeAll() {
+        List<Protos.Offer> offers = new LinkedList<>();
+        try {
+            offers.add(queue.take());
```
nickbp: Should add a comment here describing the blocking behavior/why both calls are needed

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+            queue.drainTo(offers);
+        } catch (InterruptedException e) {
+            logger.warn("Interrupted while waiting for offer in queue.");
+        }
+
+        return offers;
+    }
+
+    /**
+     * This method enqueues an Offer from Mesos if their is capacity.
```
nickbp: `there`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public void remove(Protos.OfferID offerID) {
+        Collection<Protos.Offer> offers = queue.parallelStream()
+                .filter(offer -> offer.getId().equals(offerID))
+                .collect(Collectors.toList());
+
+        boolean removed = queue.removeAll(offers);
+        if (!removed) {
+            logger.warn(
+                    String.format(
```
nickbp: Should be able to use `{}` style formatting here (no exception being passed)

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        return offers;
+    }
+
+    /**
+     * This method enqueues an Offer from Mesos if their is capacity.
+     * @return true if the Offer was successfully put in the queue, false otherwise
+     */
+    public boolean offer(Protos.Offer offer) {
```
nickbp: `put()`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * This method returns the number of elements in the queue.
+     */
+    public int getSize() {
+        return queue.size();
+    }
+
+    /**
+     * This method returns the remaining capacity in the queue.
```
nickbp: What happens if the capacity is exhausted? Blocking or dropped data? (Can't hurt to mention here)

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
nickbp: Can't hurt to also describe what happens if there isn't capacity. Interface implies the record isn't added?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/ParentElement.java
```diff
@@ -17,7 +18,7 @@
  * @param <C> the type of the child elements
  */
 public interface ParentElement<C extends Element> extends Element, Interruptible {
-
+    public static final Logger LOGGER = LoggerFactory.getLogger(ParentElement.class);
```
nickbp: `private`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public int getRemainingCapacity() {
```
nickbp: Given that this is only used for tests, it could be package-private + `@VisibleForTesting`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -0,0 +1,94 @@
...
+     * This method specifies whether any offers are in the queue.
+     */
+    public boolean isEmpty() {
+        return queue.isEmpty();
+    }
+
+    /**
+     * This method returns the number of elements in the queue.
+     */
+    public int getSize() {
```
nickbp: Given that this is only used for tests, it could be package-private + `@VisibleForTesting`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -587,12 +600,22 @@ protected DefaultScheduler(
...
+            Thread.sleep(1000);
+        }
+
+        while (!offersInProgress.isEmpty()) {
```
nickbp: There is a window/race between grabbing the offers from offerQueue (L782) and adding them to offersInProgress (L788) in which both lists would be empty.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: Yes, I remembered and then forgot again about the UninstallScheduler, thanks for reminding me.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
gabrielhartmann: The `processOffers` call never releases its thread and the ordering of Scheduler startup precludes this possibility right now, but we could have separate executor services to avoid future developers hitting this accidentally.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
gabrielhartmann: Borrowing words from here, where `offer` means non-blocking and `put` means blocking.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
gabrielhartmann: I would mention this over at `offer`.  Seems irrelevant to `getReminaingCapacity`'s behavior.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/ParentElement.java
gabrielhartmann: Best I can do in an interface is package-private apparently.

nickbp: Oh didn't realize it was an interface

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
nickbp: :+1:

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
nickbp: :+1:

gabrielhartmann:
@nickbp:  Comments addressed.  Probably worth another look at `AbstractScheduler` `UninstallScheduler` and `DefaultScheduler` changes.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -63,8 +83,73 @@ public void registered(SchedulerDriver driver, Protos.FrameworkID frameworkId, P
...
 
+    protected abstract boolean apiServerReady();
+
+    /**
+     * All offers must have been presented to resourceOffers() before calling this.
+     * @throws InterruptedException
+     */
+    @VisibleForTesting
+    public void awaitOffersProcessed() throws InterruptedException {
+        while (!offersInProgress.isEmpty()) {
```
nickbp: Once offersInProgress is lock protected, a Condition could be used to synchronously wait for offersInProgress to be empty: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Condition.html

in `processOffers()`:
```
lock.lock();
...
offersInProgress.removeAll(...);
if (offersInProgress.isEmpty()) {
  offersAreEmpty.signalAll();
}
...
lock.unlock();
```

here:

```
lock.lock();
while (!offersInProgress.isEmpty()) {
  offersAreEmpty.await();
}
lock.unlock();
```

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/OfferQueueTest.java
```diff
@@ -0,0 +1,132 @@
...
+public class OfferQueueTest {
+    @Test
+    public void testEmptyQueue() {
+        OfferQueue offerQueue = new OfferQueue();
+        Assert.assertTrue(offerQueue.isEmpty());
+    }
+
+    @Test
+    public void testEnqueueOffer() {
+        OfferQueue offerQueue = new OfferQueue();
```
nickbp: Instead of requiring access to OfferQueue.DEFAULT_CAPACITY, these tests could define their own TEST_CAPACITY and pass that to the OfferQueue constructor. Then DEFAULT_CAPACITY could remain `private`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -23,14 +29,28 @@
...
     private final AtomicBoolean isAlreadyRegistered = new AtomicBoolean(false);
+    protected final OfferQueue offerQueue = new OfferQueue();
     protected SchedulerDriver driver;
     protected DefaultReconciler reconciler;
+    private Set<Protos.OfferID> offersInProgress = new HashSet<>();
+
+    /**
+     * Executor for handling TaskStatus updates in {@link #statusUpdate(SchedulerDriver, Protos.TaskStatus)}.
+     */
+    protected final ExecutorService statusExecutor = Executors.newSingleThreadExecutor();
```
nickbp: I just realized that with this change we're moving to status handling and offer handling running on different threads. If we're going that route it'll be necessary to check if there's any unsafe behavior downstream. In theory the state store stuff should be threadsafe already but is there anything else?...

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: I don't want to put this more complicated logic which is really only needed for tests into the heart of the the offer processing loop.  Added locks, but keeping polling on `awaitOffersProcessed`.

nickbp: Yeah that's fair, :+1:

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: Spoke offline, should be good.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -63,8 +87,87 @@ public void registered(SchedulerDriver driver, Protos.FrameworkID frameworkId, P
...
+
+    /**
+     * All offers must have been presented to resourceOffers() before calling this.  This call will block until all
+     * offers have been processed.
+     * @throws InterruptedException
+     */
+    @VisibleForTesting
+    public void awaitOffersProcessed() throws InterruptedException {
+        while (true) {
+            synchronized (offersInProgress) {
```
nickbp: not inProgressLock like at L129?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -63,8 +87,87 @@ public void registered(SchedulerDriver driver, Protos.FrameworkID frameworkId, P
...
+                if (offersInProgress.isEmpty()) {
+                    LOGGER.info("All offers processed.");
+                    return;
+                }
+            }
+
+            LOGGER.warn("Offers to be processed {} is non empty, sleeping for 500ms ...", offersInProgress);
```
nickbp: May as well keep this within the sync block

gabrielhartmann: Want to release lock periodically to allow progress.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: my mistake

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: It looks to me that ultimately `TaskStatus` messages end up in the `update` call of `DeploymentStep` in all the `DefaultPlanCoordinator` case.  This updates only the status member variable which whose access is synchronized.  Will keep looking though.  Tests are not passing pretty broadly.  Trying to get a local repro to see what's happening.

gabrielhartmann:
So far no repro on my own cluster:
1. Kafka: no repro
2. Cassandra: ...

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: Mystery solved.  Adding unrelated lock in status handling in the DeploymentStep was the problem.  Fixed separately by @nickbp.  After that fix all tests pass, want to take one last look @nickbp?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
nickbp: I mean, the logger message where it's printing offersInProgress should probably be within the lock protecting that. Feel free to defer to separate PR

PR #1271, opened 2017-07-17
[INFINITY-1989] Add framework flag to just build the cli.

jsrodman:
Enable sane cli development cycles with a faster cli build iteration.

### DIFF
--- a/frameworks/cassandra/build.sh
+++ b/frameworks/cassandra/build.sh
@@ -1,6 +1,36 @@
 #!/bin/bash
 set -e
 
+usage() {
+    echo "Syntax: $0 [--cli-only] [aws|local]"
+}
+
+cli_only=
+while :; do
+    case $1 in
+        --help|-h|-\?)
+            usage
+            exit
+            ;;
+        --cli-only)
+            cli_only="true"
+            ;;
+        --)
+            shift
+            break
+            ;;
+        -*)
+            echo "unknown option $1" >&2
+            exit 1
+            ;;
+        *)
+            break
+            ;;
+    esac
+
+    shift
+done
+
 # capture anonymous metrics for reporting
 curl https://mesosphere.com/wp-content/themes/mesosphere/library/images/assets/sdk/build-sh-start.png >/dev/null 2>&1
 
@@ -9,7 +39,12 @@ ROOT_DIR="$(dirname "$(dirname $FRAMEWORK_DIR)")"
 export TOOLS_DIR=${ROOT_DIR}/tools
 BUILD_DIR=$FRAMEWORK_DIR/build/distributions
 PUBLISH_STEP=${1-none}
-${ROOT_DIR}/gradlew distZip -p ${ROOT_DIR}/sdk/executor
+
+if [ x"$cli_only" = xtrue ]; then
+    ${ROOT_DIR}/tools/build_framework.sh --cli-only $PUBLISH_STEP cassandra $FRAMEWORK_DIR $BUILD_DIR/executor.zip $BUILD_DIR/cassandra-scheduler.zip
+    exit $?
+fi
+
 ${ROOT_DIR}/tools/build_framework.sh $PUBLISH_STEP cassandra $FRAMEWORK_DIR $BUILD_DIR/executor.zip $BUILD_DIR/cassandra-scheduler.zip
 
 # capture anonymous metrics for reporting
--- a/tools/build_framework.sh
+++ b/tools/build_framework.sh
@@ -92,7 +92,7 @@ then
 fi
 
 # Ensure executor build up to date
-${ROOT_DIR}/gradlew distZip -p ${ROOT_DIR}/sdk/executor
+${REPO_ROOT_DIR}/gradlew distZip -p ${REPO_ROOT_DIR}/sdk/executor
 
 # Service (Java):
 ${REPO_ROOT_DIR}/gradlew -p ${FRAMEWORK_DIR} check distZip
--- a/tools/build_framework.sh
+++ b/tools/build_framework.sh
@@ -3,8 +3,39 @@
 # Prevent jenkins from immediately killing the script when a step fails, allowing us to notify github:
 set +e
 
+usage() {
+    echo "Syntax: $0 [--cli-only] <framework-name> </path/to/framework> [local|aws]"
+}
+
+cli_only=
+
+while :; do
+    case $1 in
+        --help|-h|-\?)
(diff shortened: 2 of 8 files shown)


jsrodman:
When you're working on the CLI and want to try stuff without waiting 10 minutes for the next cycle.

nickbp:
Like, `cd $GOPATH/src/github.com/mesosphere/dcos-commons/frameworks/$FRAMEWORK/cli/dcos-$FRAMEWORK && go build && ./dcos-$FRAMEWORK`

jsrodman:
If you want to roll that into a shellscript to ship in the repo or something instead of this, fine with me.  I've had to work out how to skip the unnecessary steps around 4 times now, so felt like it was time to not have to do that again.

nickbp:
Maybe have per-framework CLIs could have a `build.sh` that's invoked by the framework root `build.sh`es? Then if you just want the CLI you'd cd into the cli subdir and just build that subset from there

jsrodman:
I feel like the number of shell scripts that run shell scripts in the build is already beyond maintainable.  Adding another layer of shell indirection feels like a step backwards to me.  I'm okay with being overruled.

jsrodman:
Reading the script fragment you suggest, I have no idea at all how this command would know what workspace I'm using.

frameworks/cassandra/build.sh
```diff
@@ -9,7 +39,12 @@ ROOT_DIR="$(dirname "$(dirname $FRAMEWORK_DIR)")"
 export TOOLS_DIR=${ROOT_DIR}/tools
 BUILD_DIR=$FRAMEWORK_DIR/build/distributions
 PUBLISH_STEP=${1-none}
-${ROOT_DIR}/gradlew distZip -p ${ROOT_DIR}/sdk/executor
+
+if [ x"$cli_only" = xtrue ]; then
```
mellenburg: This can just be:
```
if $cli_only; then
```
E.G. https://github.com/mesosphere/tweeter/blob/master/cli_script.sh#L270

frameworks/cassandra/build.sh
```diff
@@ -1,6 +1,36 @@
 #!/bin/bash
 set -e
 
+usage() {
+    echo "Syntax: $0 [--cli-only] [aws|local]"
+}
+
+cli_only=
```
mellenburg: set this to false instead of nothing?

tools/build_go_exe.sh
```diff
@@ -28,6 +28,10 @@ fi
 GO_VERSION=$(go version | awk '{print $3}')
 # Note, UPX only works on binaries produced by Go 1.7+. However, we require Go 1.8+
 UPX_BINARY="$(which upx || which upx-ucl || echo '')"
+# For dev iteration; upx takes a long time; can set env var
+if [ -n "$CLI_BUILD_SKIP_UPX" ]; then
+    UPX_BINARY=
```
mellenburg: should this just be unset?

tools/build_framework.sh
```diff
@@ -34,6 +66,23 @@ _notify_github() {
...
 }
 
+# Used below in-order, but here for cli-only
+build_cli() {
+    # CLI (Go):
+    # /home/user/dcos-commons/frameworks/helloworld/cli => frameworks/helloworld/cli
+    REPO_CLI_RELATIVE_PATH="$(echo $CLI_DIR | cut -c $((2 + ${#REPO_ROOT_DIR}))-)"
+    ${TOOLS_DIR}/build_cli.sh ${CLI_EXE_NAME} ${CLI_DIR} ${REPO_CLI_RELATIVE_PATH}
+    if [ $? -ne 0 ]; then
+        _notify_github failure "CLI build failed"
```
jsrodman: Which script, build_framework, or _notify_github? If the latter, I must have not updated master before createing this branch.  Checking now.

frameworks/cassandra/build.sh
jsrodman: This is strongly un-idiomatic.  It happens to work, but I've never seen this pattern in 25 years of shell scripting.  it's fragile, "True" is false.  I'm not a fan.

frameworks/cassandra/build.sh
jsrodman: Shell typically uses non-empty as true, and empty as false.  I'd prefer to never use shell for build scripts but that ship seems to have sailed.

tools/build_go_exe.sh
jsrodman: The line above sets the value to '' when the binary isn't found, so this is consistent.

tools/build_framework.sh
jsrodman: Checked, it still seems used.  I don't understand this comment?

tools/build_go_exe.sh
lingmann: @jsrodman `which` is generally a bad way to test for the presence of binaries. Here is a pretty good deep dive on the negatives of doing this and some preferred & simpler approaches: https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

jsrodman: I'm not changing the logic at all involving the which command.

frameworks/cassandra/build.sh
lingmann: @mellenburg this approach is a pretty common pattern with shell programming. I tend to prefer using the modern bash [double-brack tests](http://tldp.org/LDP/abs/html/comparison-ops.html) which allows for more powerful and readable expressions, but the `x"$cli_only"...` test is also fine IMO.

tools/build_go_exe.sh
mellenburg: :+1: new JIRA is entirely appropriate

frameworks/cassandra/build.sh
mellenburg: Should have pointed out this was a style suggestion with no meat to it, but I see now there can be a lot more behind it, so do ignore and carry on to get your actual work done

tools/build_framework.sh
mellenburg: _notify_github is still living with this code but we now have that functionality handled by plugins. Ideally, that should always be handled by the plugins or on a higher level just so we don't conflate build scripts and our in house CI+review process.

Anyway, my suggestion is that we should probably work towards purging this and now that it doesnt actually functionally hook up to anything, whenever we touch something like this we can just let it disappear. Definitely not a big deal if it stays though

PR #1273, opened 2017-07-18
[INFINITY-1782] get ip addresses from statestore

ArtRand:
Use the `TaskStatus` stored in the stateStore (which should be the last `TaskStatus` with a IP address) as a fall-back for when the current `TaskStatus` doesn't gave a IP address when fetching endpoints. 

Tested with a unit test.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointsResource.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointsResource.java
@@ -10,7 +10,6 @@ import com.mesosphere.sdk.api.types.EndpointProducer;
 import com.mesosphere.sdk.offer.Constants;
 import com.mesosphere.sdk.offer.TaskException;
 import com.mesosphere.sdk.offer.taskdata.TaskLabelReader;
-import com.mesosphere.sdk.specification.TaskSpec;
 import com.mesosphere.sdk.state.StateStore;
 
 import com.mesosphere.sdk.state.StateStoreUtils;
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/api/EndpointsResource.java
```diff
@@ -180,6 +175,26 @@ public Response getEndpoint(@PathParam("name") String name) {
         return endpointsByName;
     }
 
+    private static List<String> getIpAddresses(Protos.TaskStatus taskStatus) {
```
benclarkwood: If you wanna just lean right into Java8, this works:
```
private static List<String> getIpAddresses(Protos.TaskStatus taskStatus) {
        if (taskStatus != null && taskStatus.hasContainerStatus() &&
                taskStatus.getContainerStatus().getNetworkInfosCount() > 0) {
            List<String> ipAddresses = taskStatus.getContainerStatus()
                    .getNetworkInfosList().stream()
                    .flatMap(networkInfo -> networkInfo.getIpAddressesList().stream())
                    .map(ipAddress -> ipAddress.getIpAddress())

            return ipAddresses;
        }

    }
```

benclarkwood: (Not a required change, I'm just always curious how java8y I can be)

ArtRand: nice

PR #1280, opened 2017-07-18
Lock AbstractStep.setStatus()

nickbp:
See surrounding code where locking is used

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/AbstractStep.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/AbstractStep.java
@@ -61,11 +61,14 @@ public abstract class AbstractStep extends DefaultObservable implements Step {
      * @param newStatus the new status to be set
      */
     protected void setStatus(Status newStatus) {
-        Status oldStatus = status;
-        status = newStatus;
-        logger.info("{}: changed status from: {} to: {} (interrupted={})",
-                getName(), oldStatus, newStatus, interrupted);
-
+        Status oldStatus;
+        synchronized (statusLock) {
+            oldStatus = status;
+            status = newStatus;
+            logger.info("{}: changed status from: {} to: {} (interrupted={})",
+                    getName(), oldStatus, newStatus, interrupted);
+        }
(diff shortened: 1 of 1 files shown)


benclarkwood:
Restarted C*, it hit a paramiko exception

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/AbstractStep.java
```diff
@@ -61,11 +61,14 @@ public Status getStatus() {
...
-
+        Status oldStatus;
+        synchronized (statusLock) {
+            oldStatus = status;
+            status = newStatus;
+            logger.info("{}: changed status from: {} to: {} (interrupted={})",
+                    getName(), oldStatus, newStatus, interrupted);
+        }
+        // Just in case, avoid possibility of deadlocks by calling out from outside the lock:
         if (!Objects.equals(oldStatus, newStatus)) {
```
benclarkwood: We're considering this `notifyObservers` safe outside of synchronization just because it will ultimately tell the observer the "true" latest status when it deals with the notification?

benclarkwood:
One quibble: is there a reason to not use a separate RWLock for each of the interrupted bit and the status field?

Not that any of these steps are having their synchronization hammered really.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/AbstractStep.java
nickbp: Yes. The locking is just to ensure that the Step itself behaves consistently wrt its state.

I'm avoiding locking during the notifyObservers call in case the side effect of that call is the callee coming back and changing the step's status or something (deadlock)

nickbp:
Both fields are used to determine the step's state so they're being treated as a single unit

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/AbstractStep.java
nickbp: This is actually more likely than it sounds. It could happen if the callee wants to get the overall status of the parent plan or parent phase.

PR #1281, opened 2017-07-18
Break down TaskTestUtils into non-common parts

nickbp:
Executor used one thing, Scheduler used the rest.


benclarkwood:
Restarted Elastic, failure was due to cluster not coming up

PR #1282, opened 2017-07-18
Yet more label cleanup, this time for non-TaskInfos

nickbp:
Should be the last of it?

- Move label handling for `DiscoveryInfo`s, `Resource`s, `Executor`s, ... into a new `AuxLabelAccess` (adjacent to current `TaskLabel[Reader|Writer]`)
- Move some new direct scheduler env access that snuck in, into `SchedulerFlags`


nickbp:
Turns out this effectively depends on #1281. Removing reviewers until that's in so that this is buildable.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -194,25 +193,26 @@ public void setExecutorVolume(VolumeSpec volumeSpec) {
...
+
+        Protos.Resource.DiskInfo.Builder diskInfoBuilder = resourceBuilder.getDiskBuilder();
+        diskInfoBuilder.getPersistenceBuilder()
+                .setId(persistenceId)
+                .setPrincipal(volumeSpec.getPrincipal());
+        diskInfoBuilder.getVolumeBuilder()
+                .setContainerPath(volumeSpec.getContainerPath())
+                .setMode(Protos.Volume.Mode.RW);
+
+        Protos.Resource.ReservationInfo.Builder reservationBuilder = resourceBuilder.addReservationsBuilder();
```
benclarkwood: Am I missing something? Shouldn't you need a call to setDisk?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -301,11 +295,14 @@ public void setExecutorVolume(VolumeSpec volumeSpec) {
             taskInfoBuilder.setDiscovery(getDiscoveryInfo(taskSpec.getDiscovery().get(), podInstance.getIndex()));
         }
 
+        //TODO(nickbp): This ContainerInfo handling has turned a bit spaghetti-like and needs cleaning up.
```
benclarkwood: I believe @mrbrowning plans to refactor a lot of this to be a lot more OOy.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -559,17 +548,18 @@ private static String getConfigTemplateDownloadPath(ConfigFileSpec config) {
                 && podSpec.getNetworks().isEmpty()
                 && podSpec.getRLimits().isEmpty()
                 && secretVolumes.isEmpty()) {
+            // Nothing left to do.
             return containerInfo.build();
         }
 
-        boolean shouldAddImage = podSpec.getImage().isPresent() && addExtraParameters &&
+        boolean shouldAddImage =
```
benclarkwood: nittiest-pick: Is this field used anywhere else? Can it just be put directly in the if?

benclarkwood:
@nickbp, can remove wip?

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/DcosConstants.java
```diff
@@ -4,51 +4,55 @@
...
-        switch (networkName) {
-            case DEFAULT_OVERLAY_NETWORK:
-                supportsPortMapping = false;
-                break;
-            case DEFAULT_BRIDGE_NETWORK:
-                supportsPortMapping = true;
-                break;
-            default:
-                supportsPortMapping = DEFAULT_PORT_MAPPING_POLICY;
+        if (networkName.equals(DEFAULT_BRIDGE_NETWORK)) {
```
mrbrowning: Hm, something feels strange about this check being in a plain-old-bag-of-constants class, since it now has an implicit runtime dependency.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceBuilder.java
```diff
@@ -255,27 +256,19 @@ public Resource build() {
...
                 .setRole(role)
                 .setType(Resource.ReservationInfo.Type.DYNAMIC)
-                .setLabels(
-                        Protos.Labels.newBuilder()
-                                .addLabels(Protos.Label.newBuilder()
-                                        .setKey(MesosResource.RESOURCE_ID_KEY)
-                                        .setValue(resId)))
-                .build();
+                .setPrincipal(principal.get());
+        AuxLabelAccess.setResourceId(reservationBuilder, resId);
```
mrbrowning: I wonder if things like this are actually more appropriately placed next to the resources they modify; `Label` isn't really a meaningful cross-cutting concern in our world, instead we have type-local concerns like `resource_id`. Isolating the fact that it's even implemented as a label under the hood seems like the logical endpoint of abstracting label access.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
nickbp: L201-L207 are filling in `resource.disk` by grabbing the mutable Disk.Builder from resourceBuilder

mrbrowning: Nope, the `getField` methods return a direct reference to the field being gotten, so you can mutate it directly without setting it again.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
mrbrowning: Aye.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
nickbp: It isn't. I'd normally have dropped the variable but the explicit name gives some more context on what the boolean logic is trying to determine

mrbrowning: I viscerally hate line breaks inside of if/loop condition blocks, so I elected to waste a byte of stack space to make it a little more readable.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
benclarkwood: Ugh, that just feels really dirty but ok.

nickbp: Java Proto builders effectively allow you to edit the tree in-place.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceBuilder.java
nickbp: Labels are somewhat special as they're stored in ZK (generally via TaskInfos, sometimes elsewhere), and reused by schedulers across releases. Moving labels into the special `taskdata` package at least makes it clearer where the boundary is between stuff that can change and stuff that needs a migration strategy.

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/DcosConstants.java
elezar: One comment regarding equality checks. Tools like error-prone suggest that:
```java
DEFAULT_BRIDGE_NETWORK.equals(networkName)
```
is used instead. This removes the need for an is-null check on the first variable (`networkName`).

benclarkwood: @elezar, can you open a PR adding error-prone into build.gradle?

PR #1284, opened 2017-07-19
[INFINITY-1884] Make test.sh use pytest directly

benclarkwood:


test.sh
```diff
@@ -1,80 +1,129 @@
...
+    pytest_m="$2"
+    shift # past argument
+    ;;
+    -k)
+    pytest_k="$2"
+    shift # past argument
+    ;;
+    -s)
+    security="strict"
+    shift # past argument
```
elezar: Isn't this going to shift past the next argument?

test.sh
```diff
@@ -1,80 +1,129 @@
...
+    $azure_args \
+    -e SECURITY=$security \
+    -e "PYTEST_K=$pytest_k" \
+    -e "PYTEST_M=$pytest_m" \
+    -e FRAMEWORK=$framework \
+    -e STUB_UNIVERSE_URL=$STUB_UNIVERSE_URL \
+    -v $(pwd):/build \
+    -v $ssh_path:/ssh/key \
+    -w /build \
+    michaelellenburg/dcos-commons:v0 \
```
elezar: We could add the docker image as an argument too (with a sane default).

benclarkwood: That seems like a reasonable thing to do.

test.sh
benclarkwood: Well, that would explain some things. (FWIW this is still a WIP, I extracted it out into a separate PR to get the base changes in)

cc @falseperfection

elezar: (No problem with regards to the WIP. Was just looking through things, and noticed that. Thought it may cause some unexpected behaviour).

falseperfection: Ha! Good catch, @elezar, copy/paste is so often my enemy.

benclarkwood:
Dupe of #1296

PR #1285, opened 2017-07-19
Fix local tests

elezar:
This PR fixes issues that came up running unit tests on a fresh environment. 

- A number of the tests in `DefaultPlanCoordinatorTest` were failing due to `leader.mesos` not being defined in my environment (thanks to @gabrielhartmann) for the quick fix suggestion.
- I'm not 100% happy with the solution of waiting a little longer for `SchedulerApiServerTest`, where the timeout was being triggered on my machine, but I couldn't come up with a better solution without taking a little more time. In theory, we would want to inject the dependency on the `JettyApiServer` in some way so that we can better control the behaviour and make the tests deterministic. I could also revert the changes to `SchedulerApiServerTest`.

### DIFF
--- a/sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/SchedulerApiServerTest.java
+++ b/sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/SchedulerApiServerTest.java
@@ -28,7 +28,7 @@ public class SchedulerApiServerTest {
         SchedulerApiServer schedulerApiServer = new SchedulerApiServer(12345, Collections.emptyList(),
                 Duration.ofMillis(TIMEOUT_MILLIS));
         new Thread(schedulerApiServer).start();
-        Thread.sleep(TIMEOUT_MILLIS * 10);
+        Thread.sleep(TIMEOUT_MILLIS * 50);
         Assert.assertTrue(schedulerApiServer.ready());
     }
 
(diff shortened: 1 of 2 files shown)


NimaVaziri:
lgtm

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/SchedulerApiServerTest.java
```diff
@@ -28,7 +28,7 @@ public void testApiServerReady() throws Exception {
         SchedulerApiServer schedulerApiServer = new SchedulerApiServer(12345, Collections.emptyList(),
                 Duration.ofMillis(TIMEOUT_MILLIS));
         new Thread(schedulerApiServer).start();
-        Thread.sleep(TIMEOUT_MILLIS * 10);
```
benclarkwood: Does the api server not have a "is running" check?

benclarkwood: It does: https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerApiServer.java#L56

You can poll on that.

elezar: You mean the `ready()` method? That's the method that causes the unexpected termination due to a timeout when `assertTrue(schedulerApiServer.ready());` is called on the line following the sleep.

A better solution to this may be to rather increase the timeout passed to the constrcutor on line 28. What are your thoughts on this @benclarkwood?

benclarkwood: Github lost my comments on this. Thanks github.

Yes, make the timeout in the constructor larger (30 seconds, the auto-kill of our unit tests), then just poll on it.

elezar: Thanks @benclarkwood. I've switched to polling. Would you have another look?

elezar:
@benclarkwood since the tests have now passed, I'm going to go ahead and merge this. Shouldn't affect anything that has been merged into master since the PR was created.

PR #1287, opened 2017-07-19
[INFINITY-1260][INFINITY-2019] Change secret_name to service_account_secret

mbdcos:
secret_name in config.json makes people confused about file-based secrets. I saw at least two times in slack channel. User is thinking this is the file-based secret name. Just giving it a better name


mbdcos:
@benclarkwood There is a discussion already going on at INFINITY-2019  This is the first set of PRs. Yes, all services will change.


I definitely agree it is not a easy change because we need to update our strict mode scripts. But in terms of config.json, I think it is an easy change. Here are my comments below. 

This should be no different than any other config parameter change. I strongly support changing secret_name to something more meaningful as soon as possible. 
1) If you are not running in strict mode, you do not even care. I doubt there are many users running in strict mode
2) It does not even affect application or scheduler. We have been adding/removing/changing parameters in config.json 
3) I do not know any service that can upgrade from old service to new service. Only hello-world :) So it is the best time to make this change before betas become permanent release.
4) Lets say old hello-world is upgraded to new hello-world, this change will be documented similar to any other change as we document for any update
5) If KafkaUpgrade() was not dropped, this was just a one line change in Spec Upgrader in auto config update 
6) Naming is wrong. It is super confusing. Just look at Marathon docs and Marathon secrets, you will see why  people think this is file-based secret.

mbdcos:
This should be no different than any other config parameter change. I strongly support changing secret_name to something more meaningful as soon as possible. 
1) If you are not running in strict mode, you do not even care. I doubt there are many users running in strict mode
2) It does not even affect application or scheduler. We have been adding/removing/changing parameters in config.json 
3) I do not know any service that can upgrade from old releases  to new versions (betas). Only hello-world :) So it is the best time to make this change before betas become permanent release.
4) Lets say old hello-world is upgraded to new hello-world, this change will be documented similar to any other change as we document for any update. If user is running in strict mode, and service-account-secret-name is empty, then can not even start the scheduler.
5) If KafkaUpgrade() was not dropped, this was just a one line change in Spec Upgrader in Auto Config Upgrade. It does not affect scheduler or application, so it is easy. 
6) Naming is wrong. It is super confusing. Please take a look at Marathon docs and Marathon secrets, you will see why  people think this is file-based secret :)

mbdcos:
```{
  "type":"object",
    "properties":{
      "service":{
        "type":"object",
        "description": "DC/OS service configuration properties",
        "properties":{
          "name" : {
            "description":"The name of the service instance",
            "type":"string",
            "default":"hello-world"
          },
          "sleep" : {
            "description":"The sleep duration in seconds before tasks exit.",
            "type":"number",
            "default":1000
          },
          "spec_file" : {
            "description":"The name of the service spec yaml file.",
            "type":"string",
            "default":"svc.yml"
          },
          "mesos_api_version" : {
            "description":"Configures the Mesos API version to use. Possible values: V0 (non-HTTP), V1 (HTTP)",
            "type":"string",
            "default":"V1"
          },
          "service-account-secret": {
            "description":"Name of the Secret Store credentials to use for DC/OS service authentication. This should be left empty unless service authentication is needed.",
            "type":"string",
            "default":""
          },
          "user": {
            "description": "The user that the service will run as.",
            "type": "string",
            "default": "nobody"
          },
          "principal": {
            "description": "The principal for the service instance.",
            "type": "string",
            "default": ""
          },
          "log_level": {
            "description": "The log level for the DC/OS service.",
            "type": "string",
            "enum": ["OFF", "FATAL", "ERROR", "WARN", "INFO", "DEBUG", "TRACE", "ALL"],
            "default": "INFO"
          }
        }
````


Instead of *service-account-secret*, it is requested to change it to "_service-account-keypair_"

benclarkwood:
IIRC it's that other docs refer to it as a keypair. cc @somikbehera

adam-mesos:
The contents of the service account secret is a json blob that includes the private key, the login endpoint (bouncer URL), and the service account id. It is more accurate to call it the "service account credential secret" than to discuss private vs. public keys. In theory, we could switch to a non-PKI credential.
Also note that we might switch to using file-based secrets for service account credentials. We only use env-based secrets here because that's what we had in 1.8. We may not have time to move to FBS for service-account credential secrets in 1.10, but might in 1.11.
Also note that any universe package might load one or more additional secrets, e.g. for TLS certificates, database passwords, kerberos keytabs, etc. So "secret_name" is certainly not specific enough, but if it's namespaced in the service.secret_name, to complement the service.principal field, then perhaps it is reasonable to expect that the "service" prefix implies "service account"?

benclarkwood:
@somikbehera and @keithchambers made the call. I will defer to their decision. They are free to change their decision.

mbdcos:
Thank you @adam-mesos for confirming :)

I think there might be a mismatch in product jargon/language and DC/OS document and cli language.  

I am repeating again to clarify in details:


I want to change “secret_name” to something more specific.  (pls see description of this PR). Please do not see this is under "service" field  in config.json, and assume that it is a secret for the task, mostly getting confused after file-based secret is documented in Marathon.

We only expose service account secret name to user. Marathon config is internal so environment based and file based do not matter for us. This can change in the future.

instead of “service.secret_name”
Here are some alternatives I propose:
 service._service_account_secret_ (PR)
 service._sa_secret_  (similar to dcos CLI
 service._account_secret_ (I liked this most)
….
@adam-mesos what is your suggestion?



Naming should make some sense intuitively.  service.**secret_name** (the existing config.json) is much better than service.service-account-key-pair.  The name "key-pair" is really confusing. What is a key-pair? there is no key pair in any of our docs or samples?  How will people know that this is actually "**service account secret**"?

In dcos, we have a special secret exactly called “service account secret”. It is created with “create-sa-secret” command (see dcos security secrets -h). I had explained all those with more details in ticket as well.  

It is called service account secret everywhere else, so call it service account secret. It is best not invent a new name.

Product people: There might be a mismatch in product jargon and dc/os cli and doc jargon.  Coordinate with me if you questions.

elezar:
@mbdcos while looking into (COPS-553)[https://jira.mesosphere.com/browse/COPS-553], we discovered that there is some inconsistency with the documentation. Is there a follow-up PR or issue to ensure that any docs or examples are also updated?

benclarkwood:
I'm running https://teamcity.mesosphere.io/viewLog.html?buildId=741656&tab=queuedBuildOverviewTab to test this in strict mode.

elezar:
Since this PR doesn't cover spark (which isn't an SDK framework), is the change there covered elsewhere?

nickbp:
Spark currently lives in a separate repo: https://github.com/mesosphere/spark-build

mbdcos:
closing this PR in favor of #1432

PR #1289, opened 2017-07-20
Flatten data storage types: ConfigStore, StateStore, SchemaVersionStore

nickbp:
The rest of the scheduler depends heavily on current behavior in these, it's infeasible to have alternate implementations at this level. In practice other implementations would be at the key/value Persister level instead.


sdk/scheduler/src/main/java/com/mesosphere/sdk/state/DefaultSchemaVersionStore.java
```diff
@@ -1,98 +0,0 @@
-package com.mesosphere.sdk.state;
```
nickbp: Reviewer note: this was moved to `SchemaVersionStore.java` without further changes

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/DefaultStateStore.java
```diff
@@ -1,417 +0,0 @@
-package com.mesosphere.sdk.state;
```
nickbp: Reviewer note: this was moved to `StateStore.java` without further changes (outside of a new TODO around task name retrieval in `storeStatus()`)

PR #1296, opened 2017-07-20
[Infinity-1884] Have test.sh use pytest directly

benclarkwood:


benclarkwood:
Note, this is the new interface for test.sh:
```
Mesospheres-MacBook-Pro-2:dcos-commons benwood$ ./test.sh 
Usage: ./test.sh [-m MARKEXPR] [-k EXPRESSION] [-p PATH] [-s] all|<framework-name>
-m passed to pytest directly [default -m "sanity and not azure"]
-k passed to pytest directly [default NONE]
-p PATH to cluster SSH key [default ~/.ssh/ccm.pem]
-s run in strict mode (sets $SECURITY="strict")
Cluster must be created and $CLUSTER_URL set
AWS credentials must exist in the variables:
      $AWS_ACCESS_KEY_ID
      $AWS_SECRET_ACCESS_KEY
Azure tests will run if thses variables are set:
      $AWS_ACCESS_KEY_ID
      $AWS_SECRET_ACCESS_KEY
      $AZURE_CLIENT_ID
      $AZURE_CLIENT_SECRET
      $AZURE_TENANT_ID
      $AZURE_STORAGE_ACCOUNT
      $AZURE_STORAGE_KEY
  (changes the -m default to "sanity")
Set $STUB_UNIVERSE_URL to bypass build
  (invalid when building all frameworks)
Current frameworks:
       cassandra
       elastic
       hdfs
       helloworld
       kafka
       template
```

jsrodman:
I'm a little confused by the support for -p path and framework name.  how do these interact?

test-runner.sh
```diff
@@ -0,0 +1,65 @@
...
+# Build a framework, package, upload it, and then run its integration tests.
+# (Or all frameworks depending on arguments.) Expected to be called by test.sh
+
+# Exit immediately on errors
+set -e
+
+REPO_ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+if [ "$FRAMEWORK" = "all" ]; then
+    # randomize the FRAMEWORK_LIST
```
jsrodman: This was only desired for flushing out staleness in CI.  I think the randomization can die now, if you're willing to change that.

test.sh
```diff
@@ -1,80 +1,132 @@
...
-            exit 1
-        fi
-        export STUB_UNIVERSE_URL=$(cat $UNIVERSE_URL_PATH)
-        rm -f $UNIVERSE_URL_PATH
-        echo "Built/uploaded stub universe: $STUB_UNIVERSE_URL"
-    else
-        echo "Using provided STUB_UNIVERSE_URL: $STUB_UNIVERSE_URL"
-    fi
+if [ "$#" -eq "0" ]; then
+    usage
```
jsrodman: This is weird.  Usage and continue on no arguments?
I'd default to all, but if we don't want that it should exit.

-----

OOOH, you have  usage that exits.  Somewhat awkward to read at the call site. It's usually better to have the caller perform the exit, since help shouldn't return error.

test.sh
```diff
@@ -1,80 +1,132 @@
...
 
-    echo Security: $SECURITY
-    if [ "$SECURITY" = "strict" ]; then
-        ${REPO_ROOT_DIR}/tools/setup_permissions.sh root ${framework}-role
-        # include foldered roles (tests exercise with /test/integration/svcname):
-        ${REPO_ROOT_DIR}/tools/setup_permissions.sh root test__integration__${framework}-role
-    fi
+if [ -z "$CLUSTER_URL" ]; then
+    echo "Cluster not found. Create and configure one then set \$CLUSTER_URL."
+    exit 1
```
jsrodman: Really?  We don't want to be able to have the cluster brought up the same way as CI does for reproding CI stuff?

test.sh
```diff
@@ -1,80 +1,132 @@
...
+    -e AZURE_TENANT_ID=$AZURE_DEV_TENANT_ID \
+    -e AZURE_STORAGE_ACCOUNT=$AZURE_DEV_STORAGE_ACCOUNT \
+    -e AZURE_STORAGE_KEY=$AZURE_DEV_STORAGE_KEY
+EOFF
+)
+    pytest_m="sanity"
+fi
 
-cd $REPO_ROOT_DIR
+while [[ $# -gt 1 ]]
```
jsrodman: I don't see any long options here.  Why write custom getopts instead of just using the provided one?  I don't feel strongly enough to encourage extra work but I don't like the pattern.

jsrodman:
Not sure anything really requires change, but I'd like to hear your thoughts.

benclarkwood:
-p is how you specify the path to the ssh key that should be mounted in the docker container.

It does not interface with framework name at all.

test.sh
benclarkwood: I had told @falseperfection that we'd just do "bring your own cluster" for now.

AFAIK no one is actually launching clusters with test.sh locally.

benclarkwood:
As the creator of the PR, I can't green check mark it, but I am a 👍

PR #1298, opened 2017-07-20
INFINITY-2031 User-facing changes to 'endpoints' API

nickbp:
This covers changes that affect the EndpointsResource command. Changes for the yaml spec will be done in a separate PR.

- Files VIPs against port name rather than VIP name
- Removes 'vips' list removed due to the above
- Adds validation that port names (which are used as endpoint names across the service) cannot be duplicated. This validation will later be loosed to only apply when a port is configured as advertisable (the current default)
- Update elastic to have e.g. 'coordinator-http' and 'coordinator-transport' instead of just 'http' and 'transport' in its endpoints names, fixing its current conflicting endpoint names. Note that actual hostnames are unaffected. See endpoints example in updated elastic docs.


nickbp:
Do not merge yet. Despite the status, fixes are still needed for helloworld. See prior commit where helloworld had failed (and then not re-run)

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -108,7 +111,43 @@ private static void verifyDistinctDiscoveryPrefixes(Collection<RawPod> rawPods)
...
+            }
+        }
+        if (!portNameDuplicates.isEmpty()) {
+            //TODO(nickbp): Only check this when ports are flagged with 'advertise: true' (the current default)
+            throw new IllegalArgumentException(String.format(
+                    "Duplicate port/endpoint names across different tasks: %s", portNameDuplicates));
+        }
+    }
+
+    private static <T> void collectKeys(Map<String, T> map, Set<String> seenNames, Set<String> duplicates) {
```
elezar: A style comment from my side. This method is a utility method, but mentions ports, names, and keys. We could consider changing this to:
```
private static <T> void collectKeys(Map<String, T> map, Set<String> uniques, Set<String> duplicates) {
  if (map == null) {
    return;
  }
  for (String key : map.keySet()) {
    if (!uniques.add(key)) {
      dupplicates.add(key);
    }
  }
}
```

nickbp: Forgot to generify the variable names when generifying the function. Will do so in the following PR as this one is intended to make the upcoming cut

docs/pages/reference/yaml-reference.md
```diff
@@ -230,7 +230,7 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
     * `ports`
 
-      The ports which your service will be using to accept incoming connections. Each port is given a unique name as follows:
+      The ports which your service will be using to accept incoming connections. Each port is given a unique name across the service, and this name is what's used to advertise that port in the `endpoints` listing. Ports may be defined a task as follows:
```
benclarkwood: Follow on PR: "defined in a task as..."

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -117,7 +156,7 @@ private static ReadinessCheckSpec from(RawReadinessCheck rawReadinessCheck) {
                 .build();
     }
 
-    private static DiscoverySpec from(RawDiscovery rawDiscovery) {
+    private static DiscoverySpec convertDiscovery(RawDiscovery rawDiscovery) {
```
benclarkwood: I applaud this renaming 👍  🥇

benclarkwood:
Verified checks.

PR #1299, opened 2017-07-20
Detect and fix user not being set

NimaVaziri:
The commons used in universe packages doesn't set the user at the pod level, and Mesos interprets this as it should set user to `root`.

Upon an update from release to master, even overriding the default user from "nobody" to "root"  pod-level user-change validation fails as the comparison is actually `null` vs "root".

This fixes that by detecting if the pod-level user has been set.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
@@ -30,6 +30,7 @@ import java.util.*;
 public class DefaultConfigurationUpdater implements ConfigurationUpdater<ServiceSpec> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultConfigurationUpdater.class);
+    private static final String USER = "root";
 
     private final StateStore stateStore;
     private final ConfigStore<ServiceSpec> configStore;
@@ -91,6 +92,8 @@ public class DefaultConfigurationUpdater implements ConfigurationUpdater<Service
             printConfigDiff(targetConfig.get(), targetConfigId, candidateConfigJson);
         }
 
+        fixLastDeploymentUser(targetConfig);
+
         // Check for any validation errors (including against the prior config, if one is available)
         // NOTE: We ALWAYS run validation regardless of config equality. This allows the configured
         // validators to always have a say in whether a given configuration is valid, regardless of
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
```diff
@@ -128,6 +128,11 @@ public String getType() {
     }
 
     @Override
+    public void setUser(String user) {
```
nickbp: instead of adding a setter, have the calling code create a new builder from the current copy, edit the user in the builder, and then create a new copy with the user set.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
```diff
@@ -143,6 +146,25 @@ public UpdateResult updateConfiguration(ServiceSpec candidateConfig) throws Conf
...
+            return;
+        }
+
+        for (PodSpec pod : targetConfig.get().getPods()) {
+            if (!pod.getUser().isPresent()) {
+                pod.setUser(USER);
+            }
+        }
+    }
+
```
nickbp: extra newline

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
```diff
@@ -143,6 +146,25 @@ public UpdateResult updateConfiguration(ServiceSpec candidateConfig) throws Conf
...
+     * @param targetConfig The previous service spec config
+     */
+    private void fixLastDeploymentUser(Optional<ServiceSpec> targetConfig) {
+        if (!targetConfig.isPresent()) {
+            return;
+        }
+
+        for (PodSpec pod : targetConfig.get().getPods()) {
+            if (!pod.getUser().isPresent()) {
+                pod.setUser(USER);
```
nickbp: this can create a new ServiceSpec with the changes/fixes applied

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
```diff
@@ -25,7 +25,7 @@
     @NotNull
     @Size(min = 1)
     private final String type;
-    private final String user;
+    private String user;
```
nickbp: actually, is there an annotation where we can have a default value of `root` when `user` is unset?

https://stackoverflow.com/questions/29176325/give-a-default-value-for-an-attribute-if-the-value-is-null-in-json-by-jackson

NimaVaziri:
Holding off on this for now as it's dependent on DE changes that aren't in its branch yet

PR #1307, opened 2017-07-24
[INFINITY-2019] Change "principle" to "service-account" 

mbdcos:
Do we really want to do this? "principle" is used in many places including docs/tests....
cc: @gabrielhartmann


benclarkwood:
Yes, product requested this change as it brings our docs/api in line with the rest of the product.

mbdcos:
I am not convinced with converting "principle" keyword  to "service account" keyword.

PR #1308, opened 2017-07-24
[INFINITY-1953] Remove user field from pod specs

mrbrowning:
Now that we set the user on the framework itself, there's no need to have a separate user field on pod specs. This change depends on the default executor, so I'll rebase once that's merged.


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawPod.java
```diff
@@ -19,7 +19,6 @@
     private final String image;
     private final WriteOnceLinkedHashMap<String, RawNetwork> networks;
     private final WriteOnceLinkedHashMap<String, RawRLimit> rlimits;
-    private final String user;
```
mrbrowning: We have to leave it on for now to retain upgradeability from versions that don't have a service user. I can add a deprecation notice.

PR #1312, opened 2017-07-24
[INFINITY-2062][INFINITY-2069] Detect no user from previous service spec

NimaVaziri:
The commons used in universe packages doesn't set the user, and Mesos interprets this as it should set user to root.

Upon an update from release to master, even overriding the default user from "nobody" to "root" doesn't work as the user-change validation fails since the comparison is actually null vs "root".

This fixes that by detecting if the user has been set.

This also adds resiliency to a hello-world test to prevent flakiness, as it also happened to be a blocker for this PR.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
@@ -9,6 +9,7 @@ import com.mesosphere.sdk.config.validate.ConfigValidator;
 import com.mesosphere.sdk.offer.TaskException;
 import com.mesosphere.sdk.offer.taskdata.TaskLabelReader;
 import com.mesosphere.sdk.offer.taskdata.TaskLabelWriter;
+import com.mesosphere.sdk.scheduler.SchedulerUtils;
 import com.mesosphere.sdk.specification.DefaultPodSpec;
 import com.mesosphere.sdk.specification.DefaultServiceSpec;
 import com.mesosphere.sdk.specification.PodSpec;
@@ -33,7 +34,6 @@ import java.util.*;
 public class DefaultConfigurationUpdater implements ConfigurationUpdater<ServiceSpec> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultConfigurationUpdater.class);
-    private static final String USER = "root";
 
     private final StateStore stateStore;
     private final ConfigStore<ServiceSpec> configStore;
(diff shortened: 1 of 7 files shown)


mrbrowning:
@NimaVaziri you'll probably need to fix this to be based off of master, this includes un-squashed commits from the default executor branch before it was merged.

NimaVaziri:
@mrbrowning Yeah, I'll just rebased off master and force-push after your PR gets merged

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
```diff
@@ -30,6 +28,7 @@
 public class DefaultConfigurationUpdater implements ConfigurationUpdater<ServiceSpec> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultConfigurationUpdater.class);
+    private static final String USER = "root";
```
mrbrowning: This is already set as a constant in `SchedulerUtils`, it'd be good to go with one or the other.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
```diff
@@ -143,6 +144,35 @@ public UpdateResult updateConfiguration(ServiceSpec candidateConfig) throws Conf
     }
 
     /**
+     * Detects whether the previous {@link ServiceSpec} set the user. If it didn't, we set it to "root"
+     * as Mesos treats a non-set user as "root".
+     *
+     * @param targetConfig The previous service spec from the config
+     */
+    private Optional<ServiceSpec> fixLastServiceSpec(Optional<ServiceSpec> targetConfig) {
```
mrbrowning: A more descriptive name might be helpful here, since this function only updates the user field, and that only when it's not already set.

NimaVaziri: D'oh...will do, forgot to update that.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/DefaultConfigurationUpdater.java
elezar: Thanks for pointing this out @mrbrowning. I would argue for sticking (and possibly renaming) the one in `DefaultConfigurationUpdater` instead of the one in `SchedulerUtils`. With the current implementation (using the one in `SchedulerUtils`), we end up importing the class **only** for this variable (and not the other utility methods). How do we handle other such constants?

NimaVaziri: It seems like changing the variable itself to `DEFAULT_USER` makes more sense as setting a user at the scheduler level will propagate the user down to tasks thereby setting the user for the whole service.

In that case, moving it to [DcosConstants](https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/DcosConstants.java) seems reasonable as the class generically "encapsulates constants common to DC/OS and its services."

PR #1316, opened 2017-07-25
Explain file path for Secrets

mbdcos:


docs/pages/operations-guide.md
```diff
@@ -320,7 +320,7 @@ name: secret-app/instance1
 pods:
```
sascala: This comment is actually in reference to the paragraph above ("The SDK allows secrets...). I'd like to offer users a way to decide between environment variable-based secrets, file-based secrets, or both. When I asked @adam-mesos about this when writing the file-based secrets section of the DC/OS docs, I learned that there isn't really a reason why you'd prefer an environment-variable based secret over a file-based secret. Is this also true in the context of the SDK?

Below is my (minimal) introductory text for file-based secrets. Would it make sense to incorporate some of that here?

```
File-based secrets

You can reference the secret as a file for increased security from other processes, or if your service needs to read secrets from files mounted in the container.

Referencing a file-based secret can be particularly useful for:

Kerberos keytabs or other credential files.
SSL certificates.
Configuration files with sensitive data.
File-based secrets are available in the sandbox of the task ($MESOS_SANDBOX/<configured-path>).
```

mbdcos: @sascala Yes, incorporating some of it here makes sense definitely. 

You can reference the secret as a file for increased security from other processes, or if your service needs to read secrets from files mounted in the container.

Referencing a file-based secret can be particularly useful for:

Kerberos keytabs or other credential files.
SSL certificates.
Configuration files with sensitive data.

mbdcos: If your application is designed to read private information from an environment variable (most are designed to read private information from a file), you may prefer environment-variable based secret. Alternatively, you can use file-based secret, and read from the secret file and put to an environment variable using an initialization script.  File-based and environment-based secrets are doing the same thing. 

Mesos protobuf definition lets you define both environment and file based secrets even for the same secret, so our SDK yaml will also let you do that.

"The SDK allows secrets to be exposed to pods as a file and/or as an environment variable."

PR #1319, opened 2017-07-25
[INFINITY-1794][Elastic] Re-enabled overlay tests for strict

NimaVaziri:
All overlay tests passed in CI


benclarkwood:
Can you reenable these in all the other frameworks? Or do you want to do Elastic as a trial balloon?

NimaVaziri:
Confirmed with Elastic, have others branches for respective frameworks

PR #1324, opened 2017-07-25
[INFINITY-2052] Upgrade C*

NimaVaziri:
Placeholder PR until an amend to link to prod bucket


benclarkwood:
@NimaVaziri I think this has become lost in the matrix. Can you push a typo fix commit or some such to force TC to pick it back up?

NimaVaziri:
If it's stale, just rebasing on master and force pushing does the trick.

PR #1326, opened 2017-07-25
create service account and set permissions

falseperfection:
Moves service account creation and permission setting to a `pytest` fixture.

Two follow-up tickets:
- INFINITY-2066: Move sdk_security.py to dcos_test_utils for API calls
- INFINITY-2065: Implement security cleanup in pytest fixtures

--------------------
Edit 07/27 - previous notes and questions from WIP review below...

Some notes:
1. Only configured for `cassandra` right now
2. Has a few TODOs marked in the source
3. Doesn't detect if running in `strict` mode (see question 6 below)

I have a bunch of questions:

1. What should I use instead of straight `requests` calls? Is there an easily available library? Preferably something that understands what ACLs and USERs are.
2. Is there a prefered Exception class to use throughout? If not I'll make something like `sdk_security.SecurityException()`
3. Is there a better way to get the `dcosurl` and `token` than calling the `dcoscli` directly?
4. Have I understood the vocabulary and named variables correctly (`acl`, `user`, `role`, etc)?
5. Ideally, I'd like to have tests describe the permissions they need instead of using a generic security model for all tests, but I'm thinking that needs to be a separate block of work. Agree?
6. What's the best way to detect if we're in `strict` mode? Should I just rely on the `SECURITY` envvar, or is there something more authoritative?

### DIFF
--- a/frameworks/cassandra/tests/conftest.py
+++ b/frameworks/cassandra/tests/conftest.py
@@ -1,12 +1,12 @@
 import pytest
-import sdk_repository as repo
+import sdk_repository
+import sdk_security
 
 
 @pytest.fixture(scope='session')
 def configure_universe():
-    stub_urls = {}
-    try:
-        stub_urls = repo.add_universe_repos()
-        yield # let the test session execute
-    finally:
-        repo.remove_universe_repos(stub_urls)
+    yield from sdk_repository.universe_session()
+
+@pytest.fixture(scope='session')
+def configure_security(configure_universe):
+    yield from sdk_security.security_session('cassandra')
--- a/testing/sdk_security.py
+++ b/testing/sdk_security.py
@@ -1,0 +1,239 @@
+import logging
+import os
+from typing import List, Tuple
+
+import requests
+import shakedown
+
+log = logging.getLogger(__name__)
+
+
+def grant(dcosurl: str, headers: dict, user: str, acl: str, description: str, action: str="create") -> None:
+    log.info('Granting permission to {user} for {acl}/{action} ({description})'.format(
+        user=user, acl=acl, action=action, description=description))
+
+    # TODO(kwood): INFINITY-2066 - Use dcos_test_utils instead of raw requests
+
+    # Create the ACL
+    create_endpoint = '{dcosurl}/acs/api/v1/acls/{acl}'.format(dcosurl=dcosurl, acl=acl)
+    r = requests.put(create_endpoint, headers=headers, json={'description': description}, verify=False)
+    # 201=created, 409=already exists
+    assert r.status_code == 201 or r.status_code == 409, '{} failed {}: {}'.format(
+        create_endpoint, r.status_code, r.text)
+
+    # Assign the user to the ACL
+    assign_endpoint = '{dcosurl}/acs/api/v1/acls/{acl}/users/{user}/{action}'.format(
+        dcosurl=dcosurl, acl=acl, user=user, action=action)
+    r = requests.put(assign_endpoint, headers=headers, verify=False)
+    # 204=success, 409=already exists
+    assert r.status_code == 204 or r.status_code == 409, '{} failed {}: {}'.format(
+        create_endpoint, r.status_code, r.text)
+
+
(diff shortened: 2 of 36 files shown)


testing/sdk_security.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def get_dcos_credentials():
+    dcosurl, err, rc = shakedown.run_dcos_command('config show core.dcos_url')
+    token, err, rc = shakedown.run_dcos_command('config show core.dcos_acs_token')
+    headers = {
+        'Content-Type': 'application/json',
+        'Authorization': 'token={}'.format(token.strip()),
+    }
+    creds = { 'dcosurl': dcosurl.strip(), 'headers': headers }
```
mellenburg: Could we perhaps just do `return dcosurl.strip(), headers`?

When the variables are actually used later, they are split up anyway:
```
dcosurl = creds['dcosurl']
headers = creds['headers']
```
Removing that dict would just be one less data structure to mess with. 

Also, I think we should move towards adding type hints for our method arguments in which case the choice is between:
`(creds: typing.Dict[str, typing.Union[str, typing.Dict[str, str]])` vs `(dcosurl: str, headers: dict)` so the latter would be preferable

frameworks/cassandra/tests/conftest.py
```diff
@@ -1,12 +1,19 @@
...
+import sdk_security
 
 
 @pytest.fixture(scope='session')
-def configure_universe():
+def configure_session():
     stub_urls = {}
     try:
-        stub_urls = repo.add_universe_repos()
+        stub_urls = sdk_repository.add_universe_repos()
```
mellenburg: Rather than having to repeat this `try: finally:` form in all of the conftest.py files, we could add a few methods
1. A method in sdk_repository:
```
def universe_session():
    stub_urls = {}
    try:
        stub_urls = add_universe_repos()
        yield
    finally:
        remove_universe_repos(stub_urls)
```
2. a method in sdk_security:
```
def service_account_session(account_name: str):
    try:
        setup_security(account_name)
        yield
    finally:
        cleanup_security(account_name)
```
3. Now all the conftest.py files should look like:
```
import pytest
import sdk_repository
import sdk_security

def configure_universe():
    yield from sdk_repository.universe_session()

def configure_session(configure_universe):
    yield from sdk_security.service_account_session('cassandra')
```

mellenburg: Also dependencies can be managed by pytest.mark.usefixtures or autouse=true instead of explicit fixture dependencies in the tests, but I don't have a strong preference either way at this point in time

mellenburg:
1. dcos-test-utils has clear abstractions for ACS sessions and users
2. the preferred exception in pytest is `assert ...` because pytest can do some nifty introspection allowing you to get tons of info out without having to craft a specialized error message. So, I would just stick with something like `assert not rc, 'Failed to create service account!'` However, if you plan to explicitly catch the errors from the sdk_security module somewhere, then by all means feel free to add a module-specific exception
3. I think this should be an explicit argument for the test rather than relying on state of the CLI that is magically set before the test, but that is out of scope for the time being and what you did here is totally fine
4. Vocab looks pretty good!
5. Yes, this is an awesome idea. However, that might get complicated as tests have been taking the generic model for granted. Definitely worth adding to the backlog though
6. Using the SECURITY env var is the best idea for now I think. I will almost always prefer explicit, required declaration of a test setting over implicit retrieval.

elezar:
For changes to `test.sh`, it may be worth it to look at #1343 as well.

NimaVaziri:
@falseperfection just to clarify, this is still WIP, correct?

testing/sdk_security.py
```diff
@@ -180,6 +180,16 @@ def setup_security(framework_name: str) -> None:
     sdk_utils.out('Setting up strict-mode security')
     create_service_account(service_account_name='service-acct', secret_name='secret')
     grant_permissions(
+        linux_user='nobody',
```
NimaVaziri: 👍

PR #1331, opened 2017-07-25
[KAF-350] Extend Graceful Shutdown configuration option for Kafka service

paegun:
* For Graceful Shutdown SDK MVP feature, wire Graceful Shutdown
  `kill-grace-period` through to Mesos Default Executor.
  * Using the same plumbing for Default Executor, support Custom
    Executor by ensuring that it performs the semantic SIGTERM ..
    SIGKILL and logs helpful operator messages.
* For test coverage, extend Graceful Shutdown `kill_grace_period`
  configuration option for Hello World service.
* For service availability (faster startup), extend Graceful Shutdown
  `kill_grace_period` for Kafka service.


frameworks/helloworld/tests/test_recovery.py
```diff
@@ -28,6 +30,23 @@ def configure_package(configure_security):
         sdk_install.uninstall(PACKAGE_NAME)
 
 
+# the only need for having a different WORLD_KILL_GRACE_PERIOD is/was to ensure that
+# the test of grace properly fails to detect a SIGTERM-dependent log message,
+# so environment variable is a fine-enough toggle.
```
gabrielhartmann: I don't understand this comment, can you please clarify?

frameworks/kafka/tests/test_availability.py
```diff
@@ -0,0 +1,117 @@
...
+
+def setup_module(module):
+    options = {
+        "brokers": {
+            "kill_grace_period": BROKER_KILL_GRACE_PERIOD
+        }
+    }
+
+    sdk_install.uninstall(SERVICE_NAME, PACKAGE_NAME)
+    sdk_utils.gc_frameworks()
```
gabrielhartmann: Please don't do this.  We lost logs in CI because of this.  Let the cluster's fill.  Mesos can GC.

sdk/executor/src/main/java/com/mesosphere/sdk/executor/ProcessTask.java
```diff
@@ -62,6 +62,19 @@ protected ProcessTask(
         this.taskInfo = taskInfo;
         this.processBuilder = processBuilder;
         this.exitOnTermination = exitOnTermination;
+        this.taskKillGracePeriod = getTaskKillGracePeriod(taskInfo);
+    }
+
+    private Duration getTaskKillGracePeriod(Protos.TaskInfo taskInfo) {
```
gabrielhartmann: Please remove `null` checks and use protobuf `hasKillPolicy()` and `hasGracePeriod()` style instead.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -492,6 +493,28 @@ private void setReadinessCheck(
         }
     }
 
+    private static void setTaskKillGracePeriod(
+            Protos.TaskInfo.Builder taskInfoBuilder,
+            TaskSpec taskSpec) throws InvalidRequirementException {
+        Integer taskKillGracePeriodSecondsOptional = taskSpec.getTaskKillGracePeriodSeconds();
```
gabrielhartmann: Why is this called `Optional`?  Usually we name things that are of the type `Optional<>` with that suffix.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultTaskSpec.java
```diff
@@ -44,6 +51,14 @@
     @Valid
     private final Collection<ConfigFileSpec> configFiles;
 
+    @DecimalMin("0")
+    @DecimalMax("1209600") //<< two weeks
```
gabrielhartmann: Why is this the max?  Are we avoiding overflow since it's unit is nanoseconds?  It's fine if it's just a reasonable max as well.

paegun: This max of two weeks is from the product specification, so yes a reasonable max.

frameworks/kafka/tests/test_availability.py
paegun: Agreed. I was simply following how other tests behave. Otherwise said, I was preferring consistency over what I felt correct.

sdk/executor/src/main/java/com/mesosphere/sdk/executor/ProcessTask.java
paegun: Thank you much. I was here and below again following how I understood what is the consistent way of handling optional protobuf fields. Your clarification here and below to base such decisions upon whether the model field is `Optional<>` helps a bunch.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
paegun: Again, thank you for the clarification to base such decisions upon whether the model field is Optional<> helps a bunch.

PR #1336, opened 2017-07-26
[INFINITY-2061] Fix target config selection for pod recovery

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -436,4 +438,30 @@ private static void logOutcome(StringBuilder stringBuilder, EvaluationOutcome ou
 
         return null;
     }
+
+    @VisibleForTesting
+    UUID getTargetConfig(PodInstanceRequirement podInstanceRequirement, Collection<Protos.TaskInfo> taskInfos) {
+        if (podInstanceRequirement.getRecoveryType().equals(RecoveryType.NONE) || taskInfos.isEmpty()) {
```
elezar: Just for my understanding:
- This also handles the initial case where a pod is first launched?
- The `taskInfos` passed are the tasks defined in the spec, and the `recoveryType` is explicitly set where required (defaulting to `NONE`)?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -436,4 +438,30 @@ private static void logOutcome(StringBuilder stringBuilder, EvaluationOutcome ou
...
+            // representative of the whole of the pod. If tasks in the same pod with a goal
+            // state of RUNNING had different target configurations this should be rectified
+            // in any case, so it is doubly proper to choose a single target configuration as
+            // representative of the whole pod's target configuration.
+
+            Protos.TaskInfo taskInfo = taskInfos.stream().findFirst().get();
+            try {
+                return new TaskLabelReader(taskInfo).getTargetConfiguration();
+            } catch (TaskException e) {
+                logger.error(
```
elezar: Should we add information on the exception here too?

nickbp: May make sense to log the `targetConfigId` as well

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
gabrielhartmann: Correct.  PodInstanceRequirements [default to NONE as recoveryType](https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/PodInstanceRequirement.java#L98).  When type is `NONE`, that's the initial launch condition.

PR #1338, opened 2017-07-26
[INFINITY-2008] Disable Kafka upgrade/downgrade in 1.8

benclarkwood:
Currently released beta-kafka is set to 1.9. Can't do this test in 1.8 anymore.


nickbp:
Talked offline about #1339, should be able to have it do a normal install on 1.8, or an install+upgrade on non-1.8 (i.e. have a version-specific check at the start of sanity deciding which to perform)

benclarkwood:
Will be addressed in #1339

PR #1339, opened 2017-07-26
[INFINITY-2064] Test upgrade (but not downgrade) in all sanity tests*

nickbp:
\* Except template

Upgrade is now a key part of sanity tests: Before running the tests, we install the universe version, and then upgrade that to the test version.

Upgrade fixes included:
- Broke up upgrade test library into separate upgrade/downgrade operations. Soak tests continue having full upgrade+downgrade while integration tests now just have upgrade until we're closer to exiting beta.
- Elastic had the test disabled. Effectively reenabled.
- HDFS didn't have the test at all. Added.

Other cleanup:
- Cleanup of unused imports as found by `pyflakes3`
- Reenabling `is_suppressed` check across the board, except for specific cases where the scheduler isn't expected to complete deployment (renamed flag from `check_suppression` to `wait_scheduler_idle`)
- Removal of no-op `wait_for_completed_deployment` following install operations: Effectively checked via `is_suppressed`.
- Removal of unused `sdk_package` library


frameworks/helloworld/tests/test_canary_strategy.py
```diff
@@ -28,7 +27,7 @@ def configure_package(configure_universe):
                 'hello': {'count': 4},
                 'world': {'count': 4}
             },
-            check_suppression=False)
+            wait_scheduler_idle=False)
```
benclarkwood: Nice rename 👍

frameworks/cassandra/tests/test_sanity.py
```diff
@@ -28,14 +28,12 @@ def configure_package(configure_universe):
...
-        sdk_install.install(
+        sdk_upgrade.test_upgrade(
+            "beta-{}".format(PACKAGE_NAME),
             PACKAGE_NAME,
             DEFAULT_TASK_COUNT,
             service_name=FOLDERED_SERVICE_NAME,
-            additional_options={"service": { "name": FOLDERED_SERVICE_NAME } },
-            check_suppression=False)
-        sdk_plan.wait_for_completed_deployment(FOLDERED_SERVICE_NAME)
+            additional_options={"service": {"name": FOLDERED_SERVICE_NAME} })
```
NimaVaziri: Well, except for Elastic apparently.

nickbp: Looks like this is a problem with Kafka. I have a Kafka test fix that I'll be pushing in a bit.

Cassandra meanwhile is already defaulting to user=nobody in the latest universe release. Cassandra has a different issue with volumes moving from tasks to executors.

benclarkwood:
Blocked by #1312

nickbp:
Some work is still needed in this PR for HDFS. It can't be tested as part of sanity like the others because the released version is too old to support folders. There may yet be other issues to be found in local testing, as hdfs didn't have any upgrade tests before

Cassandra is just blocked on the volume migration fix

gabrielhartmann:
#1346 should fix C* upgrade

NimaVaziri:
FYI: both of the blockers are merged

PR #1340, opened 2017-07-26
Add retry for determining cluster id in CCM extension script

benclarkwood:


tools/ccm_extend.py
```diff
@@ -17,10 +18,28 @@
...
+# until it receives it. Try for 60 minutes, because we really don't want to waste a cluster we just created.
+give_up_time = time.time() + 60 * 60
+
+while give_up_time > time.time():
+    print("Looking up cluster id.")
+    # Get all the clusters
+    all_clusters = json.loads(requests.get(base_url+'/api/cluster/active/all', headers=headers).text)
+    for cluster in all_clusters:
+        if cluster['name'] == cluster_name:
+            cluster_id = cluster['id']
```
elezar: We should break on finding the cluster with the desired name.

elezar: For my info. Is there a specific reason that this file is a linear script?

tools/ccm_extend.py
```diff
@@ -17,10 +18,28 @@
...
-for cluster in all_clusters:
-    if cluster['name'] == cluster_name:
-        cluster_id = cluster['id']
+# It may take some time for CCM to have all of the clusters information. Poll
+# until it receives it. Try for 60 minutes, because we really don't want to waste a cluster we just created.
+give_up_time = time.time() + 60 * 60
+
+while give_up_time > time.time():
+    print("Looking up cluster id.")
+    # Get all the clusters
```
elezar: These are the "active" cluster, right?

tools/ccm_extend.py
elezar: `cluster_id` is not initialised.

tools/ccm_extend.py
```diff
@@ -17,10 +18,28 @@
...
+
+    print("Cluster id not found. Sleeping for 60 seconds...")
+    time.sleep(60)
+
+if cluster_id is None:
+    print("Never retrieved cluster id from CCM!")
+    os.exit(1)
 
 # This API call increases the cluster length by an additional 4 hours.
 # See: https://ccm.mesosphere.com/api-docs/#!/cluster/Cluster_Detail_PUT
```
elezar: On the line below, the spaces on either side of the `=` are not pythonic for specifying keyword arguments and the `data` dictionary is missing a space after the `:`.

tools/ccm_extend.py
```diff
@@ -17,10 +18,28 @@
...
-    if cluster['name'] == cluster_name:
-        cluster_id = cluster['id']
+# It may take some time for CCM to have all of the clusters information. Poll
+# until it receives it. Try for 60 minutes, because we really don't want to waste a cluster we just created.
+give_up_time = time.time() + 60 * 60
+
+while give_up_time > time.time():
+    print("Looking up cluster id.")
+    # Get all the clusters
+    all_clusters = json.loads(requests.get(base_url+'/api/cluster/active/all', headers=headers).text)
```
elezar: doesn't `requests` return a response object that already has a `json()` method? http://docs.python-requests.org/en/master/

tools/ccm_extend.py
```diff
@@ -17,10 +18,28 @@
...
+
+    if cluster_id is not None:
+        print("Cluster id found!")
+        break
+
+    print("Cluster id not found. Sleeping for 60 seconds...")
+    time.sleep(60)
+
+if cluster_id is None:
+    print("Never retrieved cluster id from CCM!")
```
elezar: Should we add the cluster name to the output?

tools/ccm_extend.py
benclarkwood: It's a linear script because it's just used as a hack in CI.

tools/ccm_extend.py
benclarkwood: Again, just a hack in CI, so cluster is obvious in context.

tools/ccm_extend.py
benclarkwood: Noted (but also gleefully copied from elsewhere)

tools/ccm_extend.py
benclarkwood: Copied this from elsewhere

tools/ccm_extend.py
```diff
@@ -17,12 +18,28 @@
...
+    print("Looking up cluster id.")
+    # Get all active clusters
+    all_clusters = json.loads(requests.get(base_url+'/api/cluster/active/all', headers=headers).text)
+    for cluster in all_clusters:
+        if cluster['name'] == cluster_name:
+            cluster_id = cluster['id']
+            print("Found cluster id!")
+            break
+
+    print("Cluster id not found. Sleeping for 60 seconds...")
```
elezar: We do need a second break here tough, otherwise we're always going to wait 60 minutes. (the break above only breaks out of the `for` loop).

benclarkwood:
Verified in the Hello World PR tests that the script is running

PR #1342, opened 2017-07-26
[INFINITY-2030] Add state store utils unit tests

elezar:
This PR adds unit tests for `StateStoreUtils` (see https://jira.mesosphere.com/browse/INFINITY-2030).

The main focus was `fetchTasksNeedingRecovery()`, but some smaller methods have also been covered. Since this is my first PR, the following would be appreciated:

- Notes/comments on style
- Notes/comments on the setup of the config, store, task info and task status objects.
- Notes/comments on the granularity of the commits (should a squash the edits instead).


sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreUtils.java
```diff
@@ -33,7 +39,7 @@
     private static final String UNINSTALLING_PROPERTY_KEY = "uninstalling";
     private static final String LAST_COMPLETED_UPDATE_TYPE_KEY = "last-completed-update-type";
     private static final String PROPERTY_TASK_INFO_SUFFIX = ":task-status";
-    private static final int MAX_VALUE_LENGTH_BYTES = 1024 * 1024; // 1MB
+    protected static final int MAX_VALUE_LENGTH_BYTES = 1024 * 1024; // 1MB
```
nickbp: `@VisibleForTesting`

elezar: Yes, that was the first thing I tried, but then I get the following:
```
Error:(79, 53) java: MAX_VALUE_LENGTH_BYTES has private access in com.mesosphere.sdk.state.StateStoreUtils
```
in my unit tests. Is there some limitation where one can only annotate a non-primitive data type in Java?

nickbp: the annotation doesn't make it accessible, it's mainly just a note to say that it's non-private in order to be accessed by tests. it'd need to be marked with the annotation while also being set protected

elezar: Ah. Ok, I misunderstood. I have made it private again and added a getter which is better practice in general, right? Or would you be happier with `@VisibleForTesting protected`?

nickbp:
Either's fine, I don't think it's a big deal either way. The getter does imply a bit more formality than these internal tests really warrant

elezar:
Merging since the CI is **GREEN**.

PR #1343, opened 2017-07-26
Read AWS credentials from ~/.aws/credentials if it exists

elezar:
This PR adds support for reading AWS credentials if they are specified in the `~/.aws/credentials` file. This removes the need to set the relevant environment variables.


test.sh
```diff
@@ -52,8 +52,18 @@ if [ -z "$CLUSTER_URL" ]; then
 fi
 
 if [ -z "$AWS_ACCESS_KEY_ID" -o -z "$AWS_SECRET_ACCESS_KEY" ]; then
-    echo "AWS credentials not found (\$AWS_ACCESS_KEY_ID and \$AWS_SECRET_ACCESS_KEY)."
-    exit 1
+    CREDENTIALS_FILE="$HOME/.aws/credentials"
+    if  [ -e "$CREDENTIALS_FILE" ]
```
nickbp: may as well use `-f` to validate that it's a regular file

test.sh
```diff
@@ -52,8 +52,18 @@ if [ -z "$CLUSTER_URL" ]; then
...
+        AWS_ACCESS_KEY_ID=$( grep -oE "^aws_access_key_id\s*=\s*\S+" $CREDENTIALS_FILE | sed $SED_ARGS )
+        AWS_SECRET_ACCESS_KEY=$( grep -oE "^aws_secret_access_key\s*=\s*\S+" $CREDENTIALS_FILE | sed $SED_ARGS )
+    fi
+    if [ -z "$AWS_ACCESS_KEY_ID" -o -z "$AWS_SECRET_ACCESS_KEY" ]; then
+        echo "AWS credentials not found (\$AWS_ACCESS_KEY_ID and \$AWS_SECRET_ACCESS_KEY)."
+        exit 1
+    fi
 fi
 
 security="permissive"
```
nickbp: these constants could be moved above `usage()`, and then included in `usage()`'s output. for example:
- log $ssh_path as the default directly
- log $pytest_m as the default pytest args

also fix `thses` => `these` while in there

test.sh
```diff
@@ -42,26 +47,30 @@ function usage()
...
 if [ -z "$CLUSTER_URL" ]; then
     echo "Cluster not found. Create and configure one then set \$CLUSTER_URL."
     exit 1
 fi
 
 if [ -z "$AWS_ACCESS_KEY_ID" -o -z "$AWS_SECRET_ACCESS_KEY" ]; then
-    echo "AWS credentials not found (\$AWS_ACCESS_KEY_ID and \$AWS_SECRET_ACCESS_KEY)."
-    exit 1
+    CREDENTIALS_FILE="$HOME/.aws/credentials"
+    if  [ -f "$CREDENTIALS_FILE" ]; then
```
benclarkwood: Does this fail if the user has multiple profiles?

elezar: Yes. Most-likely. Suggestions? (apart from implementing a config file parser in bash).

benclarkwood: Eh. Probably not worth solving, but it would be nice to just tell them an error if they have multiple.

elezar: OK. Just implemented that and pushed.

elezar:
This should be rechecked when #1369 is merged.

elezar:
Would anyone be agains merging this with the CI in the current state? The changes are limited to `test.sh` (which, as far as I am aware has no influence on the CI system).

As it is, I keep having to copy my modified version to a new branch when I want to run it.

PR #1347, opened 2017-07-27
[INFINITY-2038] Reenable DC configuration options

mbdcos:
- Brings back the DC/rack config options
- Adds `ignore_dc` option to have cassandra allow renames if needed

### DIFF
--- a/frameworks/cassandra/docs/limitations.md
+++ b/frameworks/cassandra/docs/limitations.md
@@ -5,6 +5,30 @@ feature_maturity: preview
 enterprise: 'no'
 ---
 
+* Data Center Name
+
+Data Center Name:
+
+The name of the data center cannot be changed after Cassandra installation. `service.data_center` and `service.rack` options are not allowed to be modified once Cassandra is installed. 
+
+```
+"service": {
+……
+      data_center": {
+          "description": "The name of the data center this cluster is running in",
+          "type": "string",
+          "default": "datacenter1"
+        },
+        "rack": {
+          "description": "The name of the rack this cluster is running on",
+          "type": "string",
+          "default": "rack1"
+        },
+……
+}
+```
+
+
 * Seed Node Replace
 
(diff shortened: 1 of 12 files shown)


nickbp:
Added validation, tested by hand:

```
WARN  2017-07-28 01:10:32,132 [main] com.mesosphere.sdk.config.DefaultConfigurationUpdater:updateConfiguration(112): New configuration failed validation against current target configuration c60cf165-3401-4ad5-89bd-62b7c51b6460, with 1 errors across 8 validators:
1: Field: 'node.server.env.CASSANDRA_LOCATION_DATA_CENTER'; Transition: 'datacenter1' => 'datacenter2'; Message: 'Env value CASSANDRA_LOCATION_DATA_CENTER cannot change'
[...]
WARN  2017-07-28 01:10:32,730 [main] com.mesosphere.sdk.scheduler.DefaultScheduler:build(347): Failed to update configuration due to errors with configuration c60cf165-3401-4ad5-89bd-62b7c51b6460: [Field: 'node.server.env.CASSANDRA_LOCATION_DATA_CENTER'; Transition: 'datacenter1' => 'datacenter2'; Message: 'Env value CASSANDRA_LOCATION_DATA_CENTER cannot change']
```

elezar:
OK, I also tested the following:
1. Build cassandra and add the stub universe to a cluster.
2. Download the config JSON (`dcos describe`)
3. Rename the `data_center` and check that `update` fails.
4. Rename the `rack` and check that `update` fails.

I would also suggest that we add an integration test for this and possibly unit tests at a config level. Should these be added now, or should we create another JIRA ticket to track this?

nickbp:
Filed INFINITY-2087 for writing cassandra-specific tests. These could be in the form of integration tests, but I think it'd be great if we did it as unit tests instead, possibly extending to handling config.json in java test code.

elezar:
OK. @gabrielhartmann and I checked the initial settings too. Looks good.

PR #1348, opened 2017-07-27
[INFINITY-1921] Fix type error (str -> int) in hello-world secrets soak tests

susanxhuynh:
Encountered this error in soak tests:
`TypeError: '>=' not supported between instances of 'int' and 'str'`

Needed to convert string to int.


mbdcos:
hello-world secrets soak tests does not run in CI

PR #1349, opened 2017-07-27
[INFINITY-2082] Remove second scheduler used for testing locking before end of test

elezar:
As discussed with @benclarkwood, this addressed a flaky test. Due to the presence of a second scheduler, a timeout could occur in `test_sanity.test_lock_teardown`.

This removes the second scheduler before terminating the test.

### DIFF
--- a/frameworks/helloworld/tests/test_sanity.py
+++ b/frameworks/helloworld/tests/test_sanity.py
@@ -271,6 +271,7 @@ def test_lock():
 
     # Scale to 2 instances
     labels = app["labels"]
+    original_labels = labels.copy()
     labels.pop("MARATHON_SINGLE_INSTANCE_APP")
     marathon_client.update_app(FOLDERED_SERVICE_NAME, {"labels": labels})
     shakedown.deployment_wait()
@@ -288,6 +289,5 @@ def test_lock():
     assert zk_config_old == zk_config_new
 
     # In order to prevent the second scheduler instance from obtaining a lock, we undo the "scale-up" operation
-    labels = app["labels"]
-    marathon_client.update_app(FOLDERED_SERVICE_NAME, {"labels": labels, "instances": 1})
+    marathon_client.update_app(FOLDERED_SERVICE_NAME, {"labels": original_labels, "instances": 1})
     shakedown.deployment_wait()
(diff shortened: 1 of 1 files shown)


frameworks/helloworld/tests/test_sanity.py
```diff
@@ -284,6 +284,11 @@ def fn():
 
     shakedown.wait_for(lambda: fn())
 
+    # In order to prevent the second scheduler instance from obtaining a lock, we undo the "scale-up" operation
```
benclarkwood: I would place this after the asserts.

Additionally, you should add back the MARATHON_SINGLE_INSTANCE_APP label.

elezar: I was under the understanding that `app["labels"]` did not change, but maybe because python is pretty reference-y, the pop would have affected the original object. I will make it more explicit.

elezar: **UPDATE:** confirmed that `app["labesl"]` is modified. Using a copy instead.

elezar:
OK, so the "scale-down" fails because the previous deployment is still in progress:
```
kwargs = {'json': {'instances': 1, 'labels': {'DCOS_COMMONS_API_VERSION': 'v1', 'DCOS_COMMONS_UNINSTALL': 'true', 'DCOS_PACKAGE_FRAMEWORK_NAME': 'hello-world', 'DCOS_PACKAGE_IS_FRAMEWORK': 'true', ...}}, 'params': None, 'timeout': 5}
url = 'https://52.39.136.249/service/marathon/v2/apps/hello-world'
text = '{"message":"App is locked by one or more deployments. Override with the option \'?force=true\'. View details at \'/v2/deployments/<DEPLOYMENT_ID>\'.","deployments":[{"id":"5b9125d5-65e2-4bb2-a20f-2888d687dc85"}]}'
json_body = {'deployments': [{'id': '5b9125d5-65e2-4bb2-a20f-2888d687dc85'}], 'message': "App is locked by one or more deployments. Override with the option '?force=true'. View details at '/v2/deployments/<DEPLOYMENT_ID>'."}
message = 'Changes blocked: deployment already in progress for app.'
```
Is it safe to wait on the scale up? I assume not, because the whole point is that the second instance goes into a failure loop right?

PR #1353, opened 2017-07-28
Fix crash on no user

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -443,4 +444,35 @@ private static void logOutcome(StringBuilder stringBuilder, EvaluationOutcome ou
...
+            // representative of the whole of the pod. If tasks in the same pod with a goal
+            // state of RUNNING had different target configurations this should be rectified
+            // in any case, so it is doubly proper to choose a single target configuration as
+            // representative of the whole pod's target configuration.
+
+            Protos.TaskInfo taskInfo = taskInfos.stream().findFirst().get();
+            try {
+                return new TaskLabelReader(taskInfo).getTargetConfiguration();
+            } catch (TaskException e) {
+                logger.error(
```
elezar: Doesn't logger automatically format the strings? I've definitely seen something like:
```
logger.error("print {} and {}", "foo", "bar");
```
So that would mean:
```java
logger.error("Falling back to current target configuration '{}'. Failed to determine target configuration for task: {}, with exception: {}", targetConfigId,  TextFormat.shortDebugString(taskInfo), e);
```
instead of the nested format.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -75,7 +76,7 @@ public DefaultServiceSpec(
                 ? DcosConstants.MESOS_MASTER_ZK_CONNECTION_STRING : zookeeperConnection;
         this.pods = pods;
         this.replacementFailurePolicy = replacementFailurePolicy;
-        this.user = user;
+        this.user = StringUtils.isEmpty(user) ? DcosConstants.DEFAULT_SERVICE_USER : user;
```
elezar: Do we allow a username that contains only spaces? I checked with a quick unit test, and this failed. We could consider using `isBlank(user)` here too.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
nickbp: You'd think it would, but the calls where you pass an exception don't support formatting. The calls support formatting, or passing the exception, but not both.

`Logger.java`:
```
    /**
     * Log an exception (throwable) at the ERROR level with an
     * accompanying message.
     *
     * @param msg the message accompanying the exception
     * @param t   the exception (throwable) to log
     */
    public void error(String msg, Throwable t);
```

PR #1356, opened 2017-07-28
Fix selection of the service level user

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -75,10 +75,22 @@ public DefaultServiceSpec(
...
         this.pods = pods;
         this.replacementFailurePolicy = replacementFailurePolicy;
-        this.user = user;
+        this.user = getUser(user, pods);
         ValidationUtils.validate(this);
     }
 
+    static String getUser(String user, List<PodSpec> podSpecs) {
+        if (StringUtils.isBlank(user)) {
+            if (podSpecs.get(0).getUser().isPresent()) {
```
nickbp: search for the first/any podSpec with a user present?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -75,10 +75,22 @@ public DefaultServiceSpec(
                 ? DcosConstants.MESOS_MASTER_ZK_CONNECTION_STRING : zookeeperConnection;
         this.pods = pods;
         this.replacementFailurePolicy = replacementFailurePolicy;
-        this.user = user;
+        this.user = getUser(user, pods);
         ValidationUtils.validate(this);
     }
 
+    static String getUser(String user, List<PodSpec> podSpecs) {
```
nickbp: could add @VisibleForTesting here

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -75,10 +75,23 @@ public DefaultServiceSpec(
...
         ValidationUtils.validate(this);
     }
 
+    @VisibleForTesting
+    static String getUser(String user, List<PodSpec> podSpecs) {
+        if (!StringUtils.isBlank(user)) {
+            return user;
+        }
+
+        if (podSpecs.get(0).getUser().isPresent()) {
```
elezar: If `podSpecs` is an empty list, then this is going to throw on the call to `get(0)`.

We could also use:
```java
return podSpecs.get(0).getUser().orElse(DcosConstants.DEFAULT_SERVICE_USER);
```
instead of the `isPresent()` check. (this is more a question for me to get to know the preferred style).

nickbp: Wouldn't get(0) still throw in that example?

NimaVaziri: `podSpecs` will never be empty by definition of a service spec.

gabrielhartmann:
@nickbp: Can't you unapprove?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
gabrielhartmann: Moot point now, but an empty list of PodSpecs can never happen.

PR #1357, opened 2017-07-30
[INFINITY-2088] De-flake test_secrets.test_secrets_config_update

NimaVaziri:
This fix passed CI in the upgrade PR.

The logs showed that the error which caused the failure _did_ occur but then the retry kicked in and led to a pass.

This PR also does a bit of clean up wrt naming in the secret tests

### DIFF
--- a/frameworks/helloworld/tests/test_secrets.py
+++ b/frameworks/helloworld/tests/test_secrets.py
@@ -2,6 +2,7 @@ import pytest
 import shakedown
 import time
 import json
+from retrying import retry
 
 import sdk_cmd
 import sdk_install
--- a/frameworks/helloworld/tests/test_secrets.py
+++ b/frameworks/helloworld/tests/test_secrets.py
@@ -1,5 +1,7 @@
 import pytest
-import shakedown # required by @sdk_utils.dcos_X_Y_or_higher
+import shakedown
+import time
+import json
 
 import sdk_cmd
 import sdk_install
@@ -16,7 +18,7 @@ NUM_HELLO = 2
 NUM_WORLD = 3
 
 secret_content_default = "hello-world-secret-data"
-secret_content_alternative = "hello-world-secret-data-alternative"
+secret_content_alternative = secret_content_default + "-alternative"
 
 secret_options = {
         "service": {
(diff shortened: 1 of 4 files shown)


testing/sdk_cmd.py
```diff
@@ -22,6 +23,7 @@ def fn():
         return fn()
 
 
+@retry(stop_max_attempt_number=5, stop_max_delay=60)
```
elezar: Is it required to add retry logic here **and** in the secrets tests?

test.sh
```diff
@@ -18,6 +18,7 @@ function usage()
     echo "Usage: $0 [-m MARKEXPR] [-k EXPRESSION] [-p PATH] [-s] all|<framework-name>"
     echo "-m passed to pytest directly [default -m \"sanity and not azure\"]"
     echo "-k passed to pytest directly [default NONE]"
+    echo "-f module to test passed directly to pytest [default test_*.py]"
```
elezar: The default specified here does not agree with that in the bash script where this option is handled.

elezar: Does the -k option not already provide this functionality?

frameworks/helloworld/tests/test_secrets.py
```diff
@@ -373,13 +375,11 @@ def delete_secrets_all(path_prefix=""):
...
+@retry(stop_max_attempt_number=5, stop_max_delay=60)
+def read_secret(task_name, command):
     lines = sdk_cmd.run_cli("task exec {} {}".format(task_name, command)).split('\n')
     print(lines)
     for i in lines:
-        # ignore text starting with:
-        #    Overwriting Environment Variable ....
-        #    Overwriting PATH ......
-        if not i.isspace() and not i.startswith("Overwriting"):
+        if i.startswith(secret_content_default):
```
elezar: We could consider processing `i.strip()` here instead.

benclarkwood: 👍  to that

test.sh
benclarkwood: I can't recall if -k matches test names or module_name.test_name

benclarkwood: Assuming the latter, we probably don't need -f then.

testing/sdk_cmd.py
NimaVaziri: The failures I saw had separate call chains as `run_cli` is also called outside of the `read_secret` function in the secrets tests.

test.sh
NimaVaziri: "pytest -k stringexpr # only run tests with names that match the
                      # "string expression", e.g. "MyClass and not method"
                      # will select TestMyClass.test_something
                      # but not TestMyClass.test_method_simple"

Seems like since the tests in secrets are prefixed with `test_secrets`, we could use -k in this scope. But we don't seem to be following that convention (yet) so that functionality doesn't seem to target all tests at the module-level if not all the tests are prefixed wrt module.

NimaVaziri: As for the default, leaving that option out tests all the modules in the test directory, effectively being the same as `test_*.py`

NimaVaziri:
@elezar yeah, valid concern - we should add some logic around the asynchrony. I just didn't introduce it in this PR because of scope (in spite of modifying the test options, that was more in line with lowering local-dev iteration overhead which could be useful to others)

test.sh
elezar: The module name is the same as the python file name, and so `-k` will be able to do what we want it to do. I don't have a strong feeling on this though.

testing/sdk_cmd.py
elezar: One concern is that this is also called from **almost all** other tests, where we may have unintended side-effects from calling the same CLI command multiple times. 

With the retry logic here, we assume that either the return codes are consistent and represent the actual cluster state or the CLI commands idempotent and we don't mind calling them multiple times.

test.sh
NimaVaziri: Yup, just tried that locally and it's fine, not sure why it didn't work earlier when I tested it (probably didn't do it right). Anyhow, thanks for pointing it out - removed.

testing/sdk_cmd.py
NimaVaziri: Even though the other invocations of this don't seem to be causing any harm since the tests are passing, it's not safe enough to just assume it'll be fine without proper confirmation - removing for now as well. If the flakes pop up again, we'll revisit.

frameworks/helloworld/tests/test_secrets.py
NimaVaziri: Added

NimaVaziri:
Merging this as the failures are not relevant to this PR's scope.

PR #1363, opened 2017-07-31
[INFINITY-2085] Update Docs: Default Executor does not support Absolute Path for Secrets

mbdcos:
After switching to Default Executor, Absolute Path support for Secrets (supported only if there is a docker image) is dropped. All Secret files are relative to the SANDBOX.

### DIFF
--- a/docs/pages/operations-guide.md
+++ b/docs/pages/operations-guide.md
@@ -320,7 +320,7 @@ You can reference the secret as a file if your service needs to read secrets fro
 * SSL certificates.
 * Configuration files with sensitive data.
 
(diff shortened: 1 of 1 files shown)


docs/pages/operations-guide.md
```diff
@@ -374,11 +374,9 @@ On the other hand, the secret with path `secret-svc/instance1/Secret_Path2` cann
 
   
 
-### Absolute and Relative File Paths for Secrets
+**Note:** Absolute paths, with leading slash character, are not allowed.   The file paths for secret is relative path to sandbox, the secret file is placed under the sandbox.
```
sascala: Suggest:

Absolute paths (paths with a leading slash) to secrets are not supported. The secret file is located within the sandbox, therefore the file path for a secret must be relative to the sandbox.
- Supported: `secret-dvc/Secret_Path1`
- Not supported: `/secret-dvc/Secret_Path1`

mbdcos: Technically speaking, it is not because secret file is located within the sandbox. We enforce to place it within the sandbox.

mbdcos:
Ping @sascala

PR #1365, opened 2017-07-31
The plans endpoint should return errors for plans with errors

elezar:
In a quick discussion with @gabrielhartmann regarding the plans endpoint, he pointed out that `/plans/{planName}` returns `ACCEPTED (202)` even in all non-complete states and suggested an error code be returned when this is not the case. This PR is a quick attempt at this.

The behaviour is the same, but a `417` is now returned when the plan is in an error state.

This may, however, be not really RESTful, and it may be better to return `200` with some part of the payload indicating an error state.


sdk/scheduler/src/test/java/com/mesosphere/sdk/api/PlansResourceTest.java
```diff
@@ -165,6 +192,7 @@ public void testInterrupt() {
     }
 
     @Test
+
```
NimaVaziri: Unintended new line?

sdk/scheduler/src/test/java/com/mesosphere/sdk/api/PlansResourceTest.java
```diff
@@ -15,18 +19,24 @@
...
-import static com.mesosphere.sdk.api.ResponseUtils.*;
 
 public class PlansResourceTest {
-    @Mock private Plan mockPlan;
-    @Mock private Strategy<Phase> mockPlanStrategy;
-    @Mock private Phase mockPhase;
-    @Mock private Strategy<Step> mockPhaseStrategy;
-    @Mock private Step mockStep;
-    @Mock private PlanScheduler planScheduler;
+    @Mock
```
NimaVaziri: Is there a specific style convention behind this change?

elezar: I was just using @benclarkwood's IntelliJ settings. Not sure what he was using for this. I have reverted so as to decrease the noise in this PR.

PR #1367, opened 2017-07-31
[INFINITY-1977] Clarify purpose in description of SDK CLI config and state subcommands

paegun:
* Move the config and state subcommands under a new debug subcommand

### DIFF
--- a/cli/commands/config.go
+++ b/cli/commands/config.go
@@ -7,11 +7,11 @@ import (
 	"gopkg.in/alecthomas/kingpin.v3-unstable"
 )
 
-type configHandler struct {
+type debugConfigHandler struct {
 	ShowID string
 }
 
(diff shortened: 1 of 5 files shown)


cli/commands/config.go
```diff
@@ -51,18 +51,19 @@ func (cmd *configHandler) handleTargetID(a *kingpin.Application, e *kingpin.Pars
 	return nil
 }
 
+// TODO: deprecated -> commands/debug.go, handlers ^^ should remain here to organize the debug sub-command tree, seealso commands/state.go
```
nickbp: typo: seealso

nickbp:
Any opinion on pulling the bandaid and just moving the commands now? (with option of showing a 'this command has moved' error when the old commands are run for a couple releases)

paegun:
@nickbp, my opinion is to be trustworthy and reasonable (as in, can be reasoned about w/ little effort) wrt versioning. Deprecation should be employed here, not move w/o prior notification.

PR #1369, opened 2017-07-31
[INFINITY-2096][INFINITY-2109] Wait for recovery plan to start before checking success

elezar:
This PR attempts to address the flakiness of the Cassandra recovery tests (see https://jira.mesosphere.com/browse/INFINITY-2096). It was noticed that the default state of the recovery plan is `COMPLETE` which lead to the wrong things being checked at the various stages of the integration tests.

In addition, checking for `IN_PROGRESS` is not a reliable way to check if a deployment plan was started, and as such support for multiple conditions was added to the `sdk_plan.wait_for_plan_status` function.


testing/sdk_install.py
```diff
@@ -12,15 +12,30 @@
 import sdk_plan
 import sdk_utils
 
+from retrying import retry
 
-def install(
+
+@retry(stop_max_attempt_number=3, retry_on_exception=lambda e: isinstance(e, dcos.errors.DCOSException))
```
NimaVaziri: 👍

elezar:
I'm pretty sure I added those imports! Thanks for picking them up...

NimaVaziri:
@elezar seems like C* failures are fixed with this PR, should be able to merge once conflicts are resolved

elezar:
@NimaVaziri if the tests pass, then I'm going to merge. The kafka error seems to be addressed in #1372.

elezar:
@NimaVaziri, I reverted the changes to the hdfs tests (they failed, and I would rather address them properly in a separate PR).

NimaVaziri:
Fair enough, I have another (hopefully final) fix for that error...

PR #1370, opened 2017-07-31
[CASSANDRA-643] Extend seed node replacement timeout

gabrielhartmann:


elezar:
Looking at the history of this test, this should not take more than 10 minutes to complete. I can integrate this timeout increase into #1369 and then close this. @gabrielhartmann what are your thoughts?

gabrielhartmann:
@elezar:  It shouldn't take more than 15 minutes, but it does.  I've run it manually many times and all seems well.  The logs indicate that the recovery plan is proceeding.  It always takes longer on the last node than the others.

Feel free to integrate into another PR if you like, I'd be interested to see if it fails with this 20 minute timeout.

gabrielhartmann:
Looks like trying to replace a seed node immediately after replacing a regular node causes C* to be unhappy.  I changed the test ordering.  The test environment is unrealistic with the extreme proximity of operations in time.

gabrielhartmann:
@NimaVaziri: You should re-review

gabrielhartmann:
Moved to #1372.

PR #1372, opened 2017-08-01
[INFINITY-2104] Update upgrade tests to handle latest released versions of frameworks

gabrielhartmann:


frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -21,19 +21,20 @@ def configure_package(configure_universe):
...
-        #    DEFAULT_TASK_COUNT,
-        #    service_name=FOLDERED_SERVICE_NAME,
-        #    additional_options={"service": {"name": FOLDERED_SERVICE_NAME}})
-        sdk_install.install(
-            PACKAGE_NAME,
-            DEFAULT_TASK_COUNT,
-            service_name=FOLDERED_SERVICE_NAME,
-            additional_options={"service": { "name": FOLDERED_SERVICE_NAME } })
+        if shakedown.dcos_version_less_than("1.9"):
+            # Last beta-kafka release (1.1.25-0.10.1.0-beta) excludes 1.8. Skip upgrade tests with 1.8 and just install
```
nickbp: is this also the case for hdfs? (comment references kafka -- should probably update to reference the hdfs version involved)

frameworks/cassandra/tests/test_recovery.py
```diff
@@ -27,6 +27,17 @@ def configure_package(configure_universe):
 
 @pytest.mark.sanity
 @sdk_utils.dcos_1_9_or_higher # dcos task exec not supported < 1.9
+def test_node_replace_replaces_seed_node():
```
NimaVaziri: I'm not quite sure why the re-ordering would necessarily fix the issue but let's see what C* says.

NimaVaziri: I'm not quite sure why the re-ordering would necessarily fix the issue but let's see what C* says.

gabrielhartmann: Logs from C* indicate that it hasn't made up its mind about ring membership.  The destructive modifications of the ring seem to fast.

NimaVaziri: That makes sense.

NimaVaziri:
Either this PR or https://github.com/mesosphere/dcos-commons/pull/1369 will make it into master first, in which case the other PR needs to be rebased as they both touch a C* test...

frameworks/cassandra/tests/test_recovery.py
elezar: Is the ring membership part of the recovery success check? I still think that the flakiness is better handled by the improved recovery check in #1369, but I'm not against changing the order of the tests. Another issue could be the fact that we set one of the nodes to unusable in the other test, and that makes this deployment more tricky (also solved by switching the order).

NimaVaziri: Both of the proposed changes seem to introduce more resiliency for good reasons.

frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -21,19 +21,20 @@ def configure_package(configure_universe):
...
-            additional_options={"service": { "name": FOLDERED_SERVICE_NAME } })
+        if shakedown.dcos_version_less_than("1.9"):
+            # HDFS ugprade before 1.8 is not supported.
+            sdk_install.install(
+                PACKAGE_NAME,
+                DEFAULT_TASK_COUNT,
+                service_name=FOLDERED_SERVICE_NAME,
+                additional_options={"service": { "name": FOLDERED_SERVICE_NAME } })
+        else:
+            sdk_upgrade.test_upgrade(
```
elezar: What is the difference between upgrade used here and install used above? I see that this fails In the CI.

frameworks/cassandra/tests/test_recovery.py
gabrielhartmann: I agree with @NimaVaziri.  In the repro I saw, a series of valid operations against C* resulted in a cluster state it didn't like.  It wasn't a test artifact or readiness check artifact.  C* itself reported that all nodes were `UN` up and normal.  This designation of health appears to have some nuance because despite this report C* was unable to handle the replacement of a node due to an inconsistency in its internal state.

frameworks/hdfs/tests/test_shakedown.py
gabrielhartmann: Install uses the released version from the Universe, upgrade uses the stub-universe generated by this PR.

frameworks/cassandra/tests/test_recovery.py
gabrielhartmann: Now that I think about it, I believe there's another way to determine whether a node has joined a ring successfully or not.  `nodetool gossipinfo` I think.  Created [CASSANDRA-644]( https://jira.mesosphere.com/browse/CASSANDRA-644) to look into it.

NimaVaziri:
Merging this as the fix for the HDFS failure will be applied in a separate PR.

PR #1373, opened 2017-08-01
[INFINITY-2112] Fix Hello-World 1.8 permissive tests

NimaVaziri:
1.8 permissive CI status: https://teamcity.mesosphere.io/viewLog.html?buildId=725211&tab=buildResultsDiv&buildTypeId=DcosIo_DcosCommons_Sdk_DcOs18Permissive


frameworks/helloworld/tests/test_sanity.py
```diff
@@ -29,12 +29,26 @@ def configure_package(configure_universe):
...
-            PACKAGE_NAME,
-            PACKAGE_NAME,
-            DEFAULT_TASK_COUNT,
-            service_name=FOLDERED_SERVICE_NAME,
-            additional_options={"service": {"name": FOLDERED_SERVICE_NAME}})
+        if shakedown.dcos_version_less_than("1.9"):
+            # Note: The hello-world package in DC/OS 1.8 Universe is 1.1.3 and doesn't
+            # have a /suppressed endpoint. It also defaults to user=root, whereas the
+            # following release will default to user=nobody (and this value can't change in an upgrade)
+            sdk_upgrade.test_upgrade(
```
elezar: In #1372 @gabrielhartmann used `install` in this part of the if statement. That seems like an inconsistency, so I wouldn't mind some clarification as to which is the desired implementation.

testing/sdk_install.py
```diff
@@ -41,14 +41,15 @@ def install(
         timeout_sec=timeout_seconds,
         expected_running_tasks=running_task_count)
 
+    # Regardless of wait_scheduler_idle, it's safe to wait for the deployment from an install to complete
+    # before proceeding. This can take a while, default is 15 minutes. for example with HDFS, we can hit
+    # the expected total task count via FINISHED tasks, without actually completing deployment.
+    sdk_utils.out("Waiting for {}/{} to finish deployment plan...".format(package_name, service_name))
+    sdk_plan.wait_for_completed_deployment(service_name, timeout_seconds)
```
elezar: Do we not test deployments that fail? In that case, would the completed test pass?

testing/sdk_install.py
```diff
@@ -41,14 +41,15 @@ def install(
...
+
     # 2. Wait for the scheduler to be idle (as implied by deploy plan completion and suppressed bit)
     # This should be skipped ONLY when it's known that the scheduler will be stuck in an incomplete state.
     if wait_scheduler_idle:
```
elezar: This flag now only controls whether we check for the `suppressed` endpoint. Should the parameter not be renamed? Also, is there a way to discover if the endpoint is present? We could check for a 404 and not wait on that?

frameworks/helloworld/tests/test_canary_strategy.py
```diff
@@ -34,6 +34,7 @@ def configure_package(configure_universe):
         sdk_install.uninstall(PACKAGE_NAME)
 
 
+@sdk_utils.dcos_1_9_or_higher
```
elezar: Are all tests in this module now 1.9 or higher? Is there a way to specify this requirement at a module level instead of per test?

frameworks/helloworld/tests/test_sanity.py
NimaVaziri: Oh yes, that's right. Will correct to `install` after #1372 is merged and rebase.

testing/sdk_install.py
NimaVaziri: Looks like the failure case is covered by your change in #1369 so will just rebase to include that.

NimaVaziri:
Closing this in favor of https://github.com/mesosphere/dcos-commons/pull/1398

PR #1375, opened 2017-08-02
[INFINITY-2121] Fix ResourceSpec comparison

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -225,25 +224,7 @@ public static boolean areDifferent(TaskSpec oldTaskSpec, TaskSpec newTaskSpec) {
     }
 
     private static boolean areDifferent(ResourceSpec oldResourceSpec, ResourceSpec newResourceSpec) {
-        Protos.Value oldValue = oldResourceSpec.getValue();
```
nickbp: dedicated function is probably overkill now

elezar:
@gabrielhartmann should we merge this?

PR #1379, opened 2017-08-02
[INFINITY-1988] Fix locking around suppress/revive

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -213,16 +216,46 @@ public void error(SchedulerDriver driver, String message) {
         SchedulerUtils.hardExit(SchedulerErrorCode.ERROR);
     }
 
+    protected void suppressOrRevive(PlanCoordinator planCoordinator) {
+        synchronized (suppressReviveLock) {
```
elezar: Question: I've seen `synchronized (this)` this used in some places online (e.g. http://winterbe.com/posts/2015/04/30/java8-concurrency-tutorial-synchronized-locks-examples/). Is there an advantage over using an explicit object other than readability (and the fact that one can only handle one lock with `synchronized (this)`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -213,16 +216,46 @@ public void error(SchedulerDriver driver, String message) {
...
+                } else {
+                    suppress();
+                }
+            }
+        }
+    }
+
     void suppress() {
         LOGGER.info("Suppressing offers.");
+        setOfferMode(true);
```
elezar: Should this not replace the following two lines?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: Both of the reasons you stated are the reasons.
1. Readability
1. Avoids accidentally having methods with different purposes using `this` to lock and unintentionally conflicting.

gabrielhartmann:
@elezar:  Thanks for catching the failure to delete those lines.  Fixed.

gabrielhartmann:
@elezar:  I doubt it, but it's possible.  Will have to investigate.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -213,16 +216,44 @@ public void error(SchedulerDriver driver, String message) {
...
-        driver.reviveOffers();
-        StateStoreUtils.setSuppressed(stateStore, false);
+        setOfferMode(false);
+    }
+
+    private void setOfferMode(boolean suppressed) {
+        synchronized (suppressReviveLock) {
+            if (suppressed) {
+                LOGGER.info("Suppressing offers.");
+                driver.suppressOffers();
```
ArtRand: unhandled `StateStoreException` here yes?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -213,16 +216,44 @@ public void error(SchedulerDriver driver, String message) {
...
+                StateStoreUtils.setSuppressed(stateStore, true);
+            } else {
+                LOGGER.info("Reviving offers.");
+                driver.reviveOffers();
```
ArtRand: and here

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: Yes, by design.  If you can't write down your suppress/revive state, the Scheduler has no business staying up.

ArtRand: Yeah, I agree, maybe have a catch here to make it easier to debug?

gabrielhartmann: Would be inconsistent.  We could do that for every StateStore call, but we don't.  I don't know why I'd do it here and not everywhere else.

PR #1380, opened 2017-08-02
[INFINITY-2109] fix default recovery wait

elezar:
This PR addressed https://jira.mesosphere.com/browse/INFINITY-2109 and does not consider the default recovery plan state as a completed state.


elezar:
This needs more work, so I'm going to close this PR now.

PR #1383, opened 2017-08-03
Unify logging

mellenburg:
logging is much more useful when it is stratified and stamped with contextual information, so use the standard python logger

This PR also adds a conftest.py file to the root of the repository for setting up logging in all of the integration tests. Additionally, this file allows the user to set the log level at one of DEBUG, INFO, WARNING, ERROR, CRITICAL, or EXCEPTION with the TEST_LOG_LEVEL environment variable


frameworks/helloworld/tests/test_placement.py
```diff
@@ -239,7 +242,7 @@ def setup_constraint_switch():
     agents = shakedown.get_private_agents()
     some_agent = agents[0]
     other_agent = agents[1]
-    print("agents", some_agent, other_agent)
+    log.info('Agents: {} {}'.format(some_agent, other_agent))
```
elezar: The python logger supports format string out the box (sort of). This line can be replaced with:
```python
log.info('Agents: %s %s', some_agent, other_agent)
```

testing/sdk_cmd.py
```diff
@@ -27,6 +29,6 @@ def run_cli(cmd, print_output=True):
     if ret != 0:
         err = 'Got error code {} when running command "dcos {}":\nstdout: "{}"\nstderr: "{}"'.format(
             ret, cmd, stdout, stderr)
-        sdk_utils.out(err)
+        log.info(err)
```
elezar: Instead of using `info()` here, one could use `error`. This does not terminate, but does provide some extra context.

testing/sdk_install.py
```diff
@@ -13,7 +14,7 @@
 import sdk_plan
 import sdk_utils
 
-from retrying import retry
+log = logging.getLogger(__name__)
```
elezar: `retry` is still used below.

testing/sdk_cmd.py
mellenburg: This PR was simply intended just to cover the unification but I'm glad to add whatever levels you care to suggest

testing/sdk_install.py
mellenburg: Actually, this retry is already import *above*. 
https://github.com/mesosphere/dcos-commons/pull/1383/files#diff-a4edd5b45d2ef4fda8fd5a8e90348736R11

This  a good example of why we should be using a style checker in this repo... They have a good purpose. But that can be done later; there is no point in cleaning up stuff we might just toss out and we have some other things to take care of first.

elezar: Sorry, my bad. And yes, we should add flake8 or something similar for the python code.

testing/sdk_cmd.py
elezar: No, let's get this in first and the address the other changes in another PR.

frameworks/elastic/tests/test_shakedown.py
```diff
@@ -6,13 +8,15 @@
 import sdk_upgrade
 from tests.config import *
 
+log = logging.getLogger(__name__)
+
 FOLDERED_SERVICE_NAME = sdk_utils.get_foldered_name(PACKAGE_NAME)
```
elezar: `sdk_utils` has been removed from `tests.config`, so one should probably import it here (failing TC test pointed this out)

mellenburg: A great example of why not to ever do `from ____ import *` I have a hard time not fixing things if they are getting touched so I made those imports explicit but I won't complain if someone really wants that in another PR

frameworks/elastic/tests/test_shakedown.py
```diff
@@ -161,12 +172,12 @@ def test_plugin_install_and_uninstall(default_populated_index):
     config = sdk_marathon.get_config(FOLDERED_SERVICE_NAME)
     config['env']['TASKCFG_ALL_ELASTICSEARCH_PLUGINS'] = plugin_name
     sdk_marathon.update_app(FOLDERED_SERVICE_NAME, config)
-    check_plugin_installed(plugin_name, service_name=FOLDERED_SERVICE_NAME)
+    config.check_plugin_installed(plugin_name, service_name=FOLDERED_SERVICE_NAME)
```
loren: There's already a dict named `config` in this method, so the test fails with 
```
[09:12:32] :     [Step 4/4] >       config.check_plugin_installed(plugin_name, service_name=FOLDERED_SERVICE_NAME)
[09:12:32] :     [Step 4/4] E       AttributeError: 'dict' object has no attribute 'check_plugin_installed'
```

mellenburg: @loren Thanks for that catch! I'll just remove that renaming and live with the `from config import *` for the time being

elezar:
For reference, we're trying to address the hdfs test failures in #1387.

mellenburg:
I've settled on the current interface and I would love to ship this when the tests go green

elezar:
@NimaVaziri merged #1384 which gets the C* tests green, and #1387 is in fligt to get the HDFS tests to behave. Merging from master (and resolving conflicts) should already help the tests to be in a better state. I think @loren is working on the Elastic ones too.

frameworks/elastic/tests/test_shakedown.py
elezar: Yes, it's difficult to find a balance. I've started removing the `*` imports when I touch a file, but since this PR is already pretty big we can address that in a later one here.

conftest.py
```diff
@@ -0,0 +1,26 @@
...
+    'WARNING',
+    'ERROR',
+    'CRITICAL',
+    'EXCEPTION')
+
+assert log_level in log_levels, '{} is not a valid log level. ' \
+    'Use one of: {}'.format(log_level, ', '.join(log_levels))
+
+logging.basicConfig(
+    format='[%(asctime)s|%(name)s|%(levelname)s]: %(message)s',
```
elezar: I was just wondering if it would be useful to add filename and/or line number to the output. Is it important for us to be able to track this? Not a strong feeling on this though.

mellenburg: This already has `name` which is the module name (i.e. the filename). I think line number is a bit overkill as long as you add any kind of context to your log statements and can therefore search for them. This is a bit crowded with the full acstime, so I think I want to leave this as is

elezar: No problem. I missed `name` late last night when I was looking at this.

elezar: Just for the record, we're **really** excited by this!

mellenburg:
@elezar Is this something we should consider merging before stabilizing hdfs/cassandra/elastic? This can definitely help debug said tests  :)

Otherwise, just waiting to rerun tests when I have better expectations of success

NimaVaziri:
@mellenburg C* has been more or less stable with being green, the fix out for HDFS fixes the failing tests as well (they current failures are due to a known bug with overlay making it not working properly - the breaking change will be reverted)

Elastic is still flaky so it's likely that we'll merge this to be able to better investigate its failures, but we'll reassess on that shortly when the other fixes have gone in.

testing/sdk_tasks.py
```diff
@@ -20,7 +22,7 @@ def fn():
                 running_task_names.append(t['name'])
             else:
                 other_tasks.append('{}={}'.format(t['name'], t['state']))
-        sdk_utils.out('Waiting for {} running tasks, got {} running/{} total:\n- running: {}\n- other: {}'.format(
+        log.info('Waiting for {} running tasks, got {} running/{} total:\n- running: {}\n- other: {}'.format(
```
NimaVaziri: I this it's better to control higher level logging such as on this line and turn off lower level logging such as on a few lines below where we specify that `noisy=True` in `shakedown.wait_for`. Knowing that we're "Waiting for <something specific>" is far less noisier than generic spins so this allows greater flexibility for providing more accurate logging info based on the context.

elezar:
I think the last failing HDFS test was addressed as part of #1387, I have restarted the tests.

elezar:
@mellenburg I'll leave it to you to do the honors here!

elezar:
Thanks @mellenburg! Can we also delete the branch?

PR #1384, opened 2017-08-03
[INFINITY-2131][INFINITY-2003][Cassandra] Fix C* strict tests

NimaVaziri:
This fixes the C* tests by setting the user properly for a strict mode cluster. Previously the jobs were failing because the writer couldn't be successful writing as "root".

This also:
- fixes a few utility names and prevents an exception from being thrown if it tries to remove C* jobs but none exist if the installation wasn't successful.
- extends the timeout for a node replace test as that commit got dropped in 
https://github.com/mesosphere/dcos-commons/pull/1370

This passes all C* strict clusters:
- 1.8: https://teamcity.mesosphere.io/viewLog.html?buildId=727876&tab=buildResultsDiv&buildTypeId=DcosIo_DcosCommons_Cassandra_DcOs18Strict
- 1.9: https://teamcity.mesosphere.io/viewLog.html?buildId=727875&tab=buildResultsDiv&buildTypeId=DcosIo_DcosCommons_Cassandra_DcOs19Strict
- master: https://teamcity.mesosphere.io/viewLog.html?buildId=727956&tab=buildResultsDiv&buildTypeId=DcosIo_DcosCommons_Cassandra_DcOsMasterStrict

### DIFF
--- a/frameworks/cassandra/tests/test_recovery.py
+++ b/frameworks/cassandra/tests/test_recovery.py
@@ -11,6 +11,8 @@ import sdk_plan
 import sdk_tasks
 import sdk_utils
 
+TIMEOUT_SECONDS = 20 * 60
+
 from tests.config import (
     PACKAGE_NAME,
     DEFAULT_TASK_COUNT
@@ -42,6 +44,7 @@ def test_node_replace_replaces_seed_node():
 
 
 @pytest.mark.sanity
+@pytest.mark.local
 @sdk_utils.dcos_1_9_or_higher  # dcos task exec not supported < 1.9
 def test_node_replace_replaces_node():
     pod_to_replace = 'node-2'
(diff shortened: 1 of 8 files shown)


frameworks/cassandra/tests/test_recovery.py
```diff
@@ -11,6 +11,8 @@
 import sdk_tasks
 import sdk_utils
 
+TIMEOUT_SECONDS = 20 * 60
```
elezar: Is this timeout only used while waiting for recovery? If so, either rename it, or specify it closer to the point where it is used.

testing/sdk_install.py
```diff
@@ -173,7 +173,7 @@ def get_package_options(additional_options={}):
         return additional_options
 
 
-def _merge_dictionary(dict1, dict2):
+def _merge_dictionaries(dict1, dict2):
```
elezar: Something to consider for the future is the `dict.update()` method in python.

NimaVaziri: 👍 to that

elezar:
We could merge this and address the hdfs failures separately.

frameworks/cassandra/tests/test_recovery.py
NimaVaziri: Updated it to specify for recovery.

NimaVaziri:
The HDFS failures are addressed in https://github.com/mesosphere/dcos-commons/pull/1387

PR #1385, opened 2017-08-03
[CASSANDRA-352] Add recommendation for deadline IO scheduler.

triclambert:


keithchambers:
IIRC DC/OS installer and cloud templates set the scheduler to deadline by default. This was introduced for stateful workloads like Cassandra. Assuming this is the case, I think we can assume deadline is set for DC/OS install and skip the recommendation.

triclambert:
The default is not set to deadline.  Therefore, I believe we should document this.

PR #1387, opened 2017-08-03
[INFINITY-2005][HDFS] Add an expect_recovery step for HDFS integration tests

elezar:
This PR adds a helper to make it clear that an operation should trigger a FULL recovery plan.

While going through the test logs with @NimaVaziri, it was detected that for certain node replacement operations (including name node replacement), the recovery plan was still marked as `COMPLETE` from a previous operation, thus passing the health check. Depending on timing, this manifested as longer waits on the number of instances which could trigger timeouts.

The addition of a `expect_recovery` helper builds on #1369.


frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -12,7 +12,13 @@
 import sdk_tasks
 import sdk_upgrade
 import sdk_utils
-from tests.config import *
```
mellenburg: Yessss, this everywhere! Thanks!

frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -250,8 +251,15 @@ def test_bump_data_nodes():
...
-    check_healthy(count=DEFAULT_TASK_COUNT + 1)
-    sdk_tasks.check_tasks_not_updated(FOLDERED_SERVICE_NAME, 'data', data_ids)
+    try:
+        check_healthy(count=DEFAULT_TASK_COUNT + 1)
+        sdk_tasks.check_tasks_not_updated(FOLDERED_SERVICE_NAME, 'data', data_ids)
+
+    finally:
+        # TODO(elezar, nima) If one of the operations above (after the initial bump) fail, then
+        # the test will be terminated, and the number of tasks will remain at DEFAULT_TASK_COUNT + 1
+        sdk_marathon.bump_task_count_config(FOLDERED_SERVICE_NAME, 'DATA_COUNT', delta=-1)
```
NimaVaziri: The deployment complains about the updated config as the data nodes can't actually be scaled down.

elezar: Thanks @NimaVaziri. That makes sense now. Note that subsequent health checks should then probably check for the new number of expected tasks. Would it make sense to update `DEFAULT_TASK_COUNT` (or have an `EXPECTED_TASK_COUNT` variable) to ensure that this is the case for subsequent tests?

PR #1388, opened 2017-08-03
[INFINITY-2136][Hello-World] Fix for test_overlay_network flake

NimaVaziri:
This is a fix for: https://teamcity.mesosphere.io/viewLog.html?buildId=727870&buildTypeId=DcosIo_DcosCommons_Sdk_HelloWorldPrTest&guest=1


frameworks/helloworld/tests/test_overlay.py
```diff
@@ -73,16 +73,14 @@ def test_overlay_network():
...
         sdk_plan.wait_for_completed_recovery(PACKAGE_NAME, timeout_seconds=60)
+        sdk_tasks.check_running(PACKAGE_NAME, len(EXPECTED_TASKS))
     except TimeoutExpired:
         pass
 
     # test that the tasks are all up, which tests the overlay DNS
     framework_tasks = [task for task in shakedown.get_service_tasks(PACKAGE_NAME, completed=False)]
     framework_task_names = [t["name"] for t in framework_tasks]
 
-    for expected_task in EXPECTED_TASKS:
```
elezar: This changes what is being tested quite significantly. The check that has been moved into the `try`-`except` block now only checks for the number of tasks and not specific tasks. The question is what the expected behaviour is.

NimaVaziri: Given that all these tests run serially, it's expected to have all the tasks for this deployment be available and no other tasks. Since only these tasks are running and there's only one instance of each, the count is effectively the same as checking for the names.

elezar: Yes, but in the new implementation we **ignore** a failure now because of the try-catch block.

NimaVaziri: Right, good catch. I'll update that :P

NimaVaziri: @elezar Removed this try-catch as it doesn't make sense to proceed when tasks haven't deployed yet do another check immediately afterward to confirm for tasks' presence.

frameworks/helloworld/tests/test_overlay.py
```diff
@@ -69,20 +69,16 @@ def test_overlay_network():
...
     # Due to DNS resolution flakiness, some of the deployed tasks can fail. If so,
-    # we wait for them to redeploy, but if they don't fail we still want to proceed.
-    try:
-        sdk_plan.wait_for_in_progress_recovery(PACKAGE_NAME, timeout_seconds=60)
-        sdk_plan.wait_for_completed_recovery(PACKAGE_NAME, timeout_seconds=60)
-    except TimeoutExpired:
-        pass
+    # we wait for them to redeploy, but if they can't then we won't proceed as we
+    # expect the tasks to be running for this test.
+    sdk_plan.wait_for_in_progress_recovery(PACKAGE_NAME, timeout_seconds=60)
```
elezar: Are we really expecting a recovery here? I don't see what what operation before would cause this.

NimaVaziri: There's no explicit operation here that would trigger a recovery. It's more that the tasks implicitly fail (as stated by DNS resolution flakiness) thus starting a recovery. As for the timeout, since Hello-World tasks don't actually do much the come up pretty fast so recovery for them shouldn't actually take longer than a minute (if things go as expected, if not than it'll timeout as expected)

elezar: Yes, but now we are **explicitly** waiting on a recovery to start, and will fail if this does not occur (i.e. our DNS is not flaky) since we no longer have this timeout in a try-except block.

NimaVaziri: Hmm perhaps just waiting for all the tasks to come up is all that's needed.

If they need to recover from a re-deployment, they can do that. Otherwise, we can proceed. In either case, we simply wait for all of them to come up regardless of recovery or not.

NimaVaziri:
@elezar nuking this as it doesn't address the real root cause of the problem. @gabrielhartmann is working on a fix for that.

PR #1389, opened 2017-08-03
[INFINITY-2135] Fix resource refinement on pod level volumes

gabrielhartmann:
The role for all resources consuming pre-reserved resources should
conform to the `pre-reserved-role/service-role` pattern.  VolumeSpecs
generated from the pod level were not conformant.

For example the YAML below was generating a RESERVE operation with a role of `hello-world-role`, when it should have been `slave_public/hello-world-role`.

```yaml
  hello:
    pre-reserved-role: slave_public
    count: {{HELLO_COUNT}}
    placement: {{HELLO_PLACEMENT}}
    volume:
      path: pod-container-path
      type: ROOT
      size: {{HELLO_DISK}}
```
Additionally, the resource refinement test wasn't really exercising resource refinement, which is now fixed.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
@@ -73,12 +73,14 @@ public class YAMLToInternalMappers {
         List<PodSpec> pods = new ArrayList<>();
         final LinkedHashMap<String, RawPod> rawPods = rawServiceSpec.getPods();
         for (Map.Entry<String, RawPod> entry : rawPods.entrySet()) {
+            String podName = entry.getKey();
+            RawPod rawPod = entry.getValue();
             pods.add(convertPod(
-                    entry.getValue(),
+                    rawPod,
                     fileReader,
-                    entry.getKey(),
-                    taskEnvRouter.getConfig(entry.getKey()),
-                    role,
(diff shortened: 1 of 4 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -465,7 +466,7 @@ private static DefaultVolumeSpec convertVolume(
                 rawVolume.getSize(),
                 volumeTypeEnum,
                 rawVolume.getPath(),
-                role,
+                ResourceUtils.getRole(preReservedRole, role),
```
nickbp: Does this need to be done elsewhere? Like NamedVIPSpec or PortSpec below...

gabrielhartmann: Yes, I've been thinking how I can make sure I'm not missing a spot.  I think I have a way, testing it out now.

nickbp: To avoid mixing up one `String role` with another `String role`, there could be a `Role` class/object that has to be passed around, which would contain the combination of the original `String role` + `String preReservedRole`. This would turn it into a compile-time error if something is expecting a preprocessed role (as manifested by a `Role` object).

PR #1390, opened 2017-08-04
[INFINITY 1416 ] Fix up GPU opt-in policy

ArtRand:


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -72,17 +73,16 @@ public DefaultService(DefaultScheduler.Builder schedulerBuilder) throws Exceptio
...
-            }
-        }
-        return false;
+        boolean usesGpus = serviceSpec.getPods().stream()
+                .flatMap(podSpec -> podSpec.getTasks().stream())
+                .flatMap(taskSpec -> taskSpec.getResourceSet().getResources().stream())
+                .collect(Collectors.toList()).stream()
+                .filter(resourceSpec -> resourceSpec.getName().equals("gpus"))
+                .filter(resourceSpec -> resourceSpec.getValue().getScalar().getValue() >= 1)
+                .count() >= 1;
```
elezar: Instead of using `filter(condition).count()` here, one could use `anyMatch(condition)`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -72,17 +73,16 @@ public DefaultService(DefaultScheduler.Builder schedulerBuilder) throws Exceptio
...
+        // control automatic opt-in to scarce resources (GPUs) here. If the framework specifies GPU resources >= 1
+        // then we opt-in to scarce resource, otherwise follow the default policy (which as of 8/3/17 was to opt-out)
+        return usesGpus || DcosConstants.DEFAULT_GPU_POLICY;
```
elezar: Since `DEFAULT_GPU_POLICY` is `false` (and is `static final`), does this not mean that this conditional or statement is not required? Or is there some way that `DEFAULT_GPU_POLICY` could change?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
```diff
@@ -72,17 +73,16 @@ public DefaultService(DefaultScheduler.Builder schedulerBuilder) throws Exceptio
...
-                        return true;
-                    }
-                }
-            }
-        }
-        return false;
+        boolean usesGpus = serviceSpec.getPods().stream()
+                .flatMap(podSpec -> podSpec.getTasks().stream())
+                .flatMap(taskSpec -> taskSpec.getResourceSet().getResources().stream())
+                .collect(Collectors.toList()).stream()
```
elezar: Is the collection to a list strictly required? Could one not skip this step as the output of the previous step is already a stream of `resourceSpec`s?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultService.java
ArtRand: This method is really a trivial hack for when the `DEFAULT_GPU_POLICY` is `false`, obviously when it defaults to true, it's a no-op. Until we default to `true` with anti-affinity offer evaluation this is the solution.

ArtRand:
Hey @elezar, thanks for the reviews. I addressed your comments (good catches). Importantly I changed one of the unit tests in `DefaultServiceSpecTest.java` to explicitly test the policy.

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultServiceSpecTest.java
```diff
@@ -83,7 +83,8 @@ public void validSimple() throws Exception {
         File file = new File(classLoader.getResource("valid-simple.yml").getFile());
         DefaultServiceSpec serviceSpec = DefaultServiceSpec.newGenerator(RawServiceSpec.newBuilder(file).build(), flags).build();
         Assert.assertNotNull(serviceSpec);
-        Assert.assertFalse(DefaultService.serviceSpecRequestsGpuResources(serviceSpec));
```
elezar: Do we have a test of a service spec for which `serviceSpecRequestsGpuResources()` does not return `false`?

ArtRand: We do: https://github.com/mesosphere/dcos-commons/blob/8585d5a787bd8d8d9e19627f98926f17ccfd3eb6/sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultServiceSpecTest.java#L91

elezar: Sorry, I missed it! Thanks for the link.

elezar:
Looks good @ArtRand. Didn't want to take the joy of a merge away from you!

PR #1391, opened 2017-08-04
[INFINITY-2138] add config.json cni-plugin-labels sections

ArtRand:


triclambert:
Let's move this up from the pod level.  /cc @keithchambers

elezar:
With regards to:
> Let's move this up from the pod level. /cc @keithchambers

That's what INFINITY-1649 does, right?

triclambert:
That is correct.

elezar:
For the record, this is being addressed in #1412.

PR #1394, opened 2017-08-07
[INFINITY-2149] Ensure all plugins get installed successfully

loren:
- If a plugin fails to install, exit the script immediately with an error so the task fails and can be restarted


loren:
I fixed a problem, but not the one causing the errors. Going to fix that, too, and then reopen PR.

frameworks/elastic/tests/test_sanity.py
```diff
@@ -1,5 +1,3 @@
-import logging
```
NimaVaziri: Was this unintentional? 

`logging` is being used on L12

NimaVaziri: Oh importing everything from `tests.config` includes an import of `logging`

That's not immediately clear, hence we should do a cleanup in a separate PR.

loren: Yes, I saw another PR (I forget whose) that tried to remove `from tests.config import *` from all the shakedown files. Not sure what became of it.

PR #1395, opened 2017-08-07
[INFINITY-2157] Reset to PENDING on TASK_LOST in DeploymentStep

gabrielhartmann:
The deployment step should go back to PENDING on TASK_LOST.  I saw this in the wild and forced it back to PENDING with the `plan force-restart` command and it worked.  All other failure cases are handled this way, TASK_LOST should work the same way.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
@@ -162,7 +162,7 @@ public class DeploymentStep extends AbstractStep {
                 }
                 break;
             default:
-                logger.warn("Failed to process unexpected state: " + status.getState());
+                logger.error("Failed to process unexpected state: " + status.getState());
         }
 
         setStatus(getStatus(tasks));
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
```diff
@@ -136,6 +136,7 @@ public synchronized void update(Protos.TaskStatus status) {
             case TASK_FAILED:
             case TASK_KILLED:
             case TASK_KILLING:
+            case TASK_LOST:
```
elezar: It would be good to add a unit tests to prevent regressions here. Also, considering that this is an error, would should the log level of the `default` case not be increased?

gabrielhartmann:
@elezar: Just FYI. We decided the health-check bug that Art found wasn't really a bug.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
gabrielhartmann: Done and done.

PR #1396, opened 2017-08-08
Add implicit fixture to capture scheduler logs

mellenburg:
This fixture will output a file of the form "{test_name}_{scheduler_task_id}.log" for any test that fails. TeamCity will pickup and host all .log files

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -6,7 +6,9 @@ E.G. py.test frameworks/<your-frameworks>/tests
 """
 import logging
 import os
+import subprocess
 
+import pytest
 
 log_level = os.getenv('TEST_LOG_LEVEL', 'INFO').upper()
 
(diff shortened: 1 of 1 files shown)


mellenburg:
Luckily a test flaked out so we could see this patch in action:
https://teamcity.mesosphere.io/viewLog.html?buildId=731543&buildTypeId=DcosIo_DcosCommons_Hdfs_BuildAndTestPr&tab=artifacts

test_modify_app_config_test_integration_hdfs failed and so an artifact was uploaded which was called:
test_modify_app_config_test_integration_hdfs.6dad5029-7bf8-11e7-aee1-4e663b6df1e8.log

conftest.py
```diff
@@ -24,3 +26,54 @@
...
+    for task_str in tasks:
+        task = task_str.split()
+        if len(task) < 5:
+            continue
+        if task[2] == user:
+            yield task[4]
+
+
+def get_task_logs_for_id(task_id: str, lines: int=1000000):
+    return subprocess.check_output([
```
elezar: How do we want to handle the failures of the `check_output()` calls? I mean in the end the test has already failed, but we may want to add a try-except block to ensure that we don't add additional python stacktrace output to the TC log.

conftest.py
```diff
@@ -24,3 +26,54 @@
...
+        'dcos', 'task', 'log', task_id, '--lines', str(lines)]).decode()
```
elezar: Is there a way to paginate this output? I assume that the large number of lines specified should be enough.

conftest.py
```diff
@@ -24,3 +26,54 @@
 logging.basicConfig(
     format='[%(asctime)s|%(name)s|%(levelname)s]: %(message)s',
     level=log_level)
+
+
+def get_task_ids_for_user(user: str):
+    """ This function uses dcos task WITHOUT the JSON options because
+    that can return the wrong user for schedulers
+    """
+    tasks = subprocess.check_output(['dcos', 'task']).decode().split('\n')
```
elezar: Same question regarding what to do if `check_output` raises an exception.

conftest.py
mellenburg: Pytest actually does not do the introspected stack trace when errors occur in setup or teardown. Nonetheless, wrapping this could potentially decrease confusion, so we can wrap the entire thing up in one of these:
```
try:
    # do stuff
except subprocess.CalledProcessError as ex:
    raise Exception('Tasks logs could not be retrieved: ' + str(ex.output))
```

conftest.py
elezar: **For clarity:** Here I mean that we may not be getting the entire log, but I suppose in most cases the cause of the error should be in the last lines.

conftest.py
elezar: Sounds good. I don't see this as a strict requirement, and we could first check whether this is a problem or not.

conftest.py
benclarkwood: Interesting. Given that we're asking for 1000000 lines I'd expect more log output. Seeing ~ 1 MB per test failure.

elezar: I think it depends on how many lines there actually are. @mellenburg mentioned that he hasn't really seen more than ~5000 lines.

benclarkwood:
Given that this clearly works as demonstrated by it capturing logs on error, I'm going to merge it.

conftest.py
NimaVaziri: I've seen log files near 100k but that hasn't been common.

benclarkwood: I suspect that `task log` may not fetch across log rotations

NimaVaziri: I wouldn't be surprised if it did a concat before pulling though.

NimaVaziri: Ah, but looks like it does not. Hmm...

conftest.py
```diff
@@ -24,3 +26,54 @@
...
+def get_scheduler_logs_on_failure(request):
+    """ Scheduler should be the only task running as root
+    """
+    yield
+    for report in ('rep_setup', 'rep_call', 'rep_teardown'):
+        if not hasattr(request.node, report):
+            continue
+        if not getattr(request.node, report).failed:
+            continue
+        # Scheduler should be the only task running as root
```
NimaVaziri: This might not necessarily hold though, especially as of recently when we default the scheduler to "nobody" via the Marathon app definition.

benclarkwood: Oh.

I guess I don't know why this works then.

benclarkwood: Maybe Marathon always launches tasks as a single user? In the permissive case `root`?

mellenburg: In order to get a strong guarantee we will want to reimplement this at the module level where we can explicitly use the service name to grab the right task. However, some individual tests need have different scheduler names so really we need to be able to do the individual test cases as well.

Pytest fixtures allow overriding, so as we need those we can individually add them, but for now this is the simplest way to get this merged for people to use

PR #1398, opened 2017-08-08
[INFINITY-2112] [INFINITY-2154] Fix helloworld tests in 1.8

elezar:
@NimaVaziri this is a PR to your branch fix_HW_1.8 to get the 1.8 tests to pass.

You already have a PR open as #1373 but this does the following:

- Merges master into your branch
- Moves the wait for deployment back into the original `wait_scheduler_idle` block (you had some failing tests because of this)
- Update some formatting in `sdk_install.py`.

I don't know if PRs to other branches also trigger the CI system, but let's see.


elezar:
**UPDATE:** ok it seems as if not all tests are triggered.

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -32,12 +32,20 @@ def configure_package(configure_security):
...
         sdk_install.uninstall(FOLDERED_SERVICE_NAME, package_name=PACKAGE_NAME)
 
-        sdk_upgrade.test_upgrade(
-            PACKAGE_NAME,
-            PACKAGE_NAME,
-            DEFAULT_TASK_COUNT,
-            service_name=FOLDERED_SERVICE_NAME,
-            additional_options={"service": {"name": FOLDERED_SERVICE_NAME}})
+        if shakedown.dcos_version_less_than("1.9"):
+            # Hello-world before 1.9 isn't supported.
```
elezar: @NimaVaziri is this comment correct? How can we still install it if it's not supported?

elezar:
@NimaVaziri would it make sense to merge this now?

As a general question, is the only way to test on 1.8 to wait for the nightlies (other than running these locally?)

frameworks/helloworld/tests/test_sanity.py
NimaVaziri: I set this due to https://github.com/mesosphere/dcos-commons/commit/aa4ecd0cb7aec5f6592292be579bec9047c5f078

Let's double check & confirm to make sure.

NimaVaziri: That commit is in fact misleading as we should be supporting 1.8 as well.

elezar: Does this mean that the CI passing is a false positive?

NimaVaziri: Basically yes, as we should be testing it (regardless of it being able to pass without us testing it)

NimaVaziri:
1.8 permissive build: https://teamcity.mesosphere.io/viewLog.html?buildId=734180&

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -32,12 +32,14 @@ def configure_package(configure_security):
     try:
         sdk_install.uninstall(FOLDERED_SERVICE_NAME, package_name=PACKAGE_NAME)
 
+        service_config = {"service": {"name": FOLDERED_SERVICE_NAME, "user": "root"}}
```
elezar: Is this the way to do this here? If we don't specify "root" as the user, then the upgrade fails, but it feels as if this is not the ideal solution since we don't want to only test root installs forever.

Thoughts @benclarkwood @NimaVaziri?

benclarkwood: The path forward is:

1. Do this for now
2. Make sure we publish a Hello World with 1.8 as its min version that has nobody as default (make a ticket!)
3. Stop doing this once that version is released.

elezar: Is this a standard INFINITY ticket? What should a PR to universe contain?

benclarkwood: Create a ticket for changing the version 1.8 in the package.json. Assign to me.

We'll then naturally produce a HW package at the end of the sprint.

elezar: https://jira.mesosphere.com/browse/INFINITY-2196 has been created.

elezar: For the record, this PR already makes the change to `1.8` at https://github.com/mesosphere/dcos-commons/pull/1398/files#diff-163288c917fefa95b80dcc812d66d234R3

elezar: I have created mesosphere/universe#1350 for the version change.

elezar:
Any final comments here?

PR #1399, opened 2017-08-08
Cleanup reporting for conftest.py

mellenburg:
Continuation of https://github.com/mesosphere/dcos-commons/pull/1396
* wraps potential failures of dcos task commands with a readable expcetion
* limits lines collected to 10k. With 1M lines output, its very possible to have multiple GB of log collection (~50 tests * 1M logs lines @ 100-300MB) at the end of a test. 10k lines will limit this liability to a more manageable ~100MB

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,7 @@
-""" This file configures python logging for the pytest framework
-integration tests
+""" This file configures the pytest driven integration tests with:
+* logging functionality
+* global fixtures
+* custom pytest hooks
 
 Note: pytest must be invoked with this file in the working directory
 E.G. py.test frameworks/<your-frameworks>/tests
@@ -7,6 +9,7 @@ E.G. py.test frameworks/<your-frameworks>/tests
 import logging
 import os
 import subprocess
+import typing
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


mellenburg:
@benclarkwood its alright that #1396 merged, but as pointed out in the description, back of the envelope calculations show that the current settings could create an untenable amount of output in the worst case scenario. If certain tests need the additional logging, we can override at the module level

benclarkwood:
FWIW I didn't see any log captures that were > 1.5 MB. I suspect `task log` isn't capturing anything beyond the current stdout and stderr files.

mellenburg:
@benclarkwood elastic errors are unrelated to this PR. Shall I continue retrying or can we :ship: ?

benclarkwood:
We are in front load the pain mode for now...

conftest.py
```diff
@@ -41,14 +47,20 @@ def get_task_ids_for_user(user: str):
...
 
-def get_task_logs_for_id(task_id: str, lines: int=1000000):
+def get_task_logs_for_id(task_id: str) -> str:
+    """ Thin wrapper to return the logs for a given task.
+    Note: 10k lines is used as this will typically be 1-3MB. Higher volume
+    log pulls could result in reporting difficulties for other tests or CI
+    """
     return subprocess.check_output([
-        'dcos', 'task', 'log', task_id, '--lines', str(lines)]).decode()
+        'dcos', 'task', 'log', task_id, '--lines', '10000']).decode()
```
benclarkwood: I'd actually like to leave this really high given the behavior of `dcos task log` is just to grab the last stdout file. WDYT?

mellenburg: You make a good point. Will amend

mellenburg:
Not much point in merging this at the moment. Can be included in a bigger cleanup later....

PR #1401, opened 2017-08-08
[INFINITY-2153] fix flaky secret tests

elezar:
This PR tries to improve the robustness of secret creation and deletion.

The main change is that the state of the secret being deleted or created is checked after the operation to determine success.

That is to say:
- when setting secret `s1` to value `v1`, the creation is seen as successful if `secrets get s1` returns `v1` after the `create` operation. 
- when deleting secret `s1`, the deletion is seen as successful if `secret get s1` returns empty after the `delete` operation.

The second point is perhaps too flexible, since the only error we should really allow in this case is the `404` indicating that the secret does not exist.


benclarkwood:
@elezar is this still a WIP?

elezar:
@benclarkwood this test doesn't seem to be flaking as much anymore, so we could also just close it.  Thoughts?

elezar:
Closing this PR. Will readdress this if the helloworld secrets tests fail again.

PR #1402, opened 2017-08-08
login to cluster and configure cli

falseperfection:
Currently CLI configuration is left to the user or whatever script is calling tests. This is an essential part of running the test and is therefore now captures in test setup

### DIFF
--- a/frameworks/cassandra/tests/conftest.py
+++ b/frameworks/cassandra/tests/conftest.py
@@ -1,10 +1,15 @@
 import pytest
+import sdk_login
 import sdk_repository
 import sdk_security
 
 
 @pytest.fixture(scope='session')
-def configure_universe():
+def configure_login():
+    yield from sdk_login.login_session()
+
+@pytest.fixture(scope='session')
+def configure_universe(configure_login):
     yield from sdk_repository.universe_session()
 
 @pytest.fixture(scope='session')
--- a/test.sh
+++ b/test.sh
@@ -15,11 +15,12 @@ FRAMEWORK_LIST=$(ls $REPO_ROOT_DIR/frameworks | sort)
 
 function usage()
 {
-    echo "Usage: $0 [-m MARKEXPR] [-k EXPRESSION] [-p PATH] [-s] all|<framework-name>"
+    echo "Usage: $0 [-m MARKEXPR] [-k EXPRESSION] [-p PATH] [-s] [-o] all|<framework-name>"
     echo "-m passed to pytest directly [default -m \"sanity and not azure\"]"
     echo "-k passed to pytest directly [default NONE]"
     echo "-p PATH to cluster SSH key [default ~/.ssh/ccm.pem]"
     echo "-s run in strict mode (sets \$SECURITY=\"strict\")"
+    echo "-o Testing against an Open cluster not Enterprise"
     echo "Cluster must be created and \$CLUSTER_URL set"
     echo "AWS credentials must exist in the variables:"
     echo "      \$AWS_ACCESS_KEY_ID"
@@ -61,6 +62,7 @@ pytest_m="sanity and not azure"
 pytest_k=""
 azure_args=""
 ssh_path="${HOME}/.ssh/ccm.pem"
+DCOS_ENTERPRISE=True
 
 # If AZURE variables are given, change default -m and prepare args for docker
 if [ -n "$AZURE_DEV_CLIENT_ID" -a -n "$AZURE_DEV_CLIENT_SECRET" -a \
@@ -94,6 +96,10 @@ case $key in
     security="strict"
     [[ $CLUSTER_URL == https* ]] || echo "CLUSTER_URL must be https in strict mode" && exit 1
     ;;
+    -o)
+    DCOS_ENTERPRISE=False
+    shift # past argument
+    ;;
     -p)
     ssh_path="$2"
     shift # past argument
@@ -124,6 +130,7 @@ docker run --rm \
     -e CLUSTER_URL=$CLUSTER_URL \
     $azure_args \
     -e SECURITY=$security \
+    -e DCOS_ENTERPRISE=DCOS_ENTERPRISE \
     -e "PYTEST_K=$pytest_k" \
     -e "PYTEST_M=$pytest_m" \
     -e FRAMEWORK=$framework \
--- a/test.sh
+++ b/test.sh
@@ -130,7 +130,7 @@ docker run --rm \
     -e CLUSTER_URL=$CLUSTER_URL \
     $azure_args \
     -e SECURITY=$security \
-    -e DCOS_ENTERPRISE=DCOS_ENTERPRISE \
+    -e DCOS_ENTERPRISE=$DCOS_ENTERPRISE \
     -e "PYTEST_K=$pytest_k" \
     -e "PYTEST_M=$pytest_m" \
     -e FRAMEWORK=$framework \
--- a/testing/sdk_login.py
+++ b/testing/sdk_login.py
@@ -28,6 +28,7 @@ def login(dcosurl: str, username: str, password: str, is_enterprise: bool) -> st
 
 
 def configure_cli(dcosurl: str, token: str) -> None:
+    # TODO(kwood): INFINITY-2182 - Replace shakedown.run_dcos_command with subprocess invocation
     out, err, rc = shakedown.run_dcos_command('config set core.dcos_url {}'.format(dcosurl))
     assert not rc, 'Failed to set core.dcos_url: {}'.format(err)
     out, err, rc = shakedown.run_dcos_command('config set core.ssl_verify false')
@@ -36,18 +37,14 @@ def configure_cli(dcosurl: str, token: str) -> None:
     assert not rc, 'Failed to set core.dcos_acs_token: {}'.format(err)
 
 
-def logout(dcosurl: str):
-    pass
-
-
 def login_session() -> None:
     """Login to DC/OS.
 
     Behavior is determined by the following environment variables:
     CLUSTER_URL: full URL to the test cluster
+    DCOS_ENTERPRISE: if set to any value assume an EE instance
     DCOS_LOGIN_USERNAME: the EE user (defaults to bootstrapuser)
     DCOS_LOGIN_PASSWORD: the EE password (defaults to deleteme)
-    DCOS_ENTERPRISE: determine how to authenticate (defaults to false)
     DCOS_ACS_TOKEN: bypass auth and use the user supplied token
 
     This should generally be used as a fixture in a framework's conftest.py:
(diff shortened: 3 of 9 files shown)


testing/sdk_login.py
```diff
@@ -0,0 +1,81 @@
...
+def configure_cli(dcosurl: str, token: str) -> None:
+    out, err, rc = shakedown.run_dcos_command('dcos config set core.dcos_url {}'.format(dcosurl))
+    assert rc, 'Failed to set core.dcos_url: {}'.format(err)
+    out, err, rc = shakedown.run_dcos_command('dcos config set core.ssl_verify false')
+    assert rc, 'Failed to set core.ssl_verify: {}'.format(err)
+    out, err, rc = shakedown.run_dcos_command('config set core.dcos_acs_token {}'.format(token))
+    assert rc, 'Failed to set core.dcos_acs_token: {}'.format(err)
+
+
+def logout(dcosurl: str):
```
mellenburg: Do we even care? Seems OK to just leave off now and forever

testing/sdk_login.py
```diff
@@ -0,0 +1,81 @@
+import logging
+import os
+
+import requests
+import shakedown
```
mellenburg: I would prefer if we do not use shakedown in this case. See below

testing/sdk_login.py
```diff
@@ -0,0 +1,81 @@
...
+
+    headers = {'Content-Type': 'application/json',}
+    login_endpoint = '{dcosurl}/acs/api/v1/auth/login'.format(dcosurl=dcosurl)
+    r = requests.post(login_endpoint, headers=headers, json=payload, verify=False)
+    assert r.status_code == 200, '{} failed {}: {}'.format(login_endpoint, r.status_code, r.text)
+
+    return r.json()['token']
+
+
+def login(dcosurl: str, username: str, password: str, is_enterprise: bool, user_token: str) -> str:
```
mellenburg: Why bother passing user_token to this function at all if it just gets passed through untouched?

testing/sdk_login.py
```diff
@@ -0,0 +1,81 @@
...
+        dcos_login_username = os.environ.get('DCOS_LOGIN_USERNAME', __CLI_LOGIN_EE_USERNAME)
+        dcos_login_password = os.environ.get('DCOS_LOGIN_PASSWORD', __CLI_LOGIN_EE_PASSWORD)
+        dcos_enterprise = os.environ.get('DCOS_ENTERPRISE', False) and True
+        dcos_acs_token = os.environ.get('DCOS_ACS_TOKEN')
+        token = login(
+            dcosurl=cluster_url,
+            username=dcos_login_username,
+            password=dcos_login_password,
+            is_enterprise=dcos_enterprise,
+            user_token=dcos_acs_token
```
mellenburg: Rather than have the login function use the `user_token` argument, I would prefer that we move the if statement out of that function to here. E.G.:
```
if not dcos_acs_token:
    dcos_acs_token = login(...)
configure_cli(dcos_url, dcos_acs_token)
```
It makes a little more sense to me given that the dcos_acs_token can never really be used in the login process. It just negates the need to login at all. WDYT?

testing/sdk_login.py
```diff
@@ -0,0 +1,81 @@
...
+    return user_token
+
+
+def configure_cli(dcosurl: str, token: str) -> None:
+    out, err, rc = shakedown.run_dcos_command('dcos config set core.dcos_url {}'.format(dcosurl))
+    assert rc, 'Failed to set core.dcos_url: {}'.format(err)
+    out, err, rc = shakedown.run_dcos_command('dcos config set core.ssl_verify false')
+    assert rc, 'Failed to set core.ssl_verify: {}'.format(err)
+    out, err, rc = shakedown.run_dcos_command('config set core.dcos_acs_token {}'.format(token))
+    assert rc, 'Failed to set core.dcos_acs_token: {}'.format(err)
```
mellenburg: can we remove the shakedown call from this? in this case here, the function is literally equivalent to:
```
import subprocess

def run_dcos_command(cmd_str):
    subprocess.check_call(['dcos'] + cmd_str.split())

def configure_cli(dcosurl: str, token: str) -> None:
    run_dcos_command('config set core.dcos_url {}'.format(dcosurl))
    run_dcos_command('config set core.ssl_verify false')
    run_dcos_command('config set core.dcos_acs_token {}'.format(token))
```
That will automatically raise an assertion that contains the output of the failed command. If you want to embellish the errors then you can do so in run_dcos_command with:
```
try:
    # do stuff in run_dcos_command
except subprocess.CalledProcessError as ex:
    raise Exception('Failed to {}: {}'.format(cmd_str, ex.stdout.decode()))
```
I don't think its really necessary (i.e. the former is enough of an implementation), but I leave that to you to decide

mellenburg: Also, I kept the command input as a string for consistency with the current example, but I think it would be best to pass through a list to subprocess.check_call. Thats only a small preference though. The important thing is that the command is passed as a list to `check_call` as passing a str is platform dependent in execution

testing/sdk_login.py
```diff
@@ -0,0 +1,81 @@
...
+        dcos_acs_token = os.environ.get('DCOS_ACS_TOKEN')
+        token = login(
+            dcosurl=cluster_url,
+            username=dcos_login_username,
+            password=dcos_login_password,
+            is_enterprise=dcos_enterprise,
+            user_token=dcos_acs_token
+        )
+        configure_cli(dcosurl, token)
+        yield
```
mellenburg: Not required at all, but what do you think about yielding the acs token here? That way any test that needs it can just take it from this fixture instead of having to make a call to retrieve it with shakedown, which in turn calls out to the dcoscli

testing/sdk_login.py
```diff
@@ -0,0 +1,81 @@
...
+    finally:
+        logout(dcosurl=cluster_url)
```
mellenburg: Unless you see a clear benefit to logging out, I would be fine with dropping the whole `try:finally` clause here

frameworks/cassandra/tests/conftest.py
```diff
@@ -1,10 +1,15 @@
 import pytest
+import sdk_login
 import sdk_repository
 import sdk_security
 
 
 @pytest.fixture(scope='session')
-def configure_universe():
+def configure_login():
```
mellenburg: This invocation is identical in all of the framework test suites so I am fine if you make it an implicit (i.e. autouse) or explicit fixture at the top-level and thereby avoid having to make all these small changes.

One advantage of doing this at the top-level conftest.py with implicit injection is that you could cd down to this individual directory and login yourself and not deal with this new interface or simply avoid having the test reconfigure your local CLI

However, that's a pretty specific scenario and stuff at the top level might just need to be copied out anyway if we split these tests into their own repos. Furthermore the function you wrote already has a clear override mechanism so its not strictly necessary.

testing/sdk_login.py
falseperfection: Because even people as perfect as me do dumb things sometimes. :)

testing/sdk_login.py
falseperfection: I prefer consistency over correctness most of the time. There are `shakedown.run_docs_command` calls throughout. If we change here, I want to see it changed everywhere. That means we'd need a shared `run_dcos_ command`, bringing us very nearly back to where we are now. Just instead of `shakedown.run_dcos_command` it would be `sdk_utils.run_dcos_command`. I'm not sure I understand the value there. (Other than getting rid of the _word_ shakedown, but I have no issue with shakedown.)

testing/sdk_login.py
falseperfection: Yeah. I think I'm over thinking about cleanup. I believe it's a good practice, but I don't see us doing it anytime soon. (Nobody but me cares, and I don't have a strongly compelling use-case.)

testing/sdk_login.py
falseperfection: Well obviously. ;)

testing/sdk_login.py
mellenburg: That's a fair point and I do mind just pushing that out to another issue where we apply it consistently. I definitely have a tendency to push for "landing" paradigms over maintaining consistency...

Given that we can agree to table this, my qualm with shakedown in this case is the accessibility of the interface. Once I use shakedown for a CLI command, I am extremely limited in what I can  do with that command: https://github.com/dcos/shakedown/blob/master/shakedown/dcos/command.py#L103

By using subprocess, I have the ability to:
* declare a timeout
* arbitrarily redirect stdout, stdin, and stderr
* implicitly raise an expception on failure
* specifically control the logging statements around this command
* specify if the command should be shell interpreted
* control the specific environment for that command
* provide my command in an unambiguous argument list

By using shakedown:
* I MUST submit my command as a string that will be split up again anyway. This makes it harder to provide intricate arguments.
* I MUST push all output to either subprocess.PIPE or stdout. This prevents me from streaming the output
* I MUST declare with a keyword arg that failures in the command should raise a local exception
* If I want to access any of the powerful subprocess options ( https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module ) then I need to make a commit to shakedown and then wait for a dependency update in this repo. This means I cannot control: timeouts, working dir, command entrypoint (shell), or custom decoders for I/O (might be useful for mesos task log retrieval)

TL;DR I see the argument on maintaining consistency so we can handle this later, but I really care about making sure tests can directly access the powerful, native subprocess interface instead of using a customized, restrictive interface

testing/sdk_login.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @pytest.fixture(scope='session')
+    def configure_login():
+        yield from sdk_login.login_session()
+    """
+    try:
+        cluster_url = os.environ.get('CLUSTER_URL')
+        dcos_login_username = os.environ.get('DCOS_LOGIN_USERNAME', __CLI_LOGIN_EE_USERNAME)
+        dcos_login_password = os.environ.get('DCOS_LOGIN_PASSWORD', __CLI_LOGIN_EE_PASSWORD)
+        dcos_enterprise = os.environ.get('DCOS_ENTERPRISE', False) and True
```
mellenburg: wont this evaluate true if you set 'DCOS_ENTERPRISE=false'?

mellenburg: My preference for this pattern is generally:
```
dcos_enterprise = os.getenv('DCOS_ENTERPRISE', 'true') == 'true'
```
Also, I intentionally went with enterprise as the default to be consistent with how the testing suite is biased now, even though that feels like a faux pas

falseperfection: Yes, it will. Because it's a boolean, the idiom I was going for was "if the variable is defined," not the value it's set to. So things like case and "1" vs "True" vs "Yes" didn't have to be worried about. Either way needs to be made more clear.

frameworks/cassandra/tests/conftest.py
falseperfection: I don't want to do top-level because I like thinking of the frameworks as separate entities as much as possible. If anything, I'd rather go back to how I had the setup working originally:

```
def configuire_session():
  try:
    login_config = sdk_login.configure_login()
    repo_config = sdk_repository.configure_repository()
    security_config = sdk_security.configure_security('cassandra')
    yield
  finally:
    sdk_security.cleanup_security(security_config)
    sdk_repository.cleanup_repository(repo_config)
    sdk_login.cleanup_login(login_config)
```

I think that makes the setup clean, explicit, and minimizes the boilerplate, but I'm not tied to any implementation. I'll leave it as-is as a matter of inertia.

testing/sdk_login.py
falseperfection: Ticket created, and my next comment added there - https://jira.mesosphere.com/browse/INFINITY-2182

test.sh
```diff
@@ -124,6 +130,7 @@ docker run --rm \
     -e CLUSTER_URL=$CLUSTER_URL \
     $azure_args \
     -e SECURITY=$security \
+    -e DCOS_ENTERPRISE=DCOS_ENTERPRISE \
```
elezar: You will need to access the variable value here: `"$DCOS_ENTERPRISE"`

falseperfection: 👍 Good catch

benclarkwood:
I think this is redundant?

PR #1407, opened 2017-08-09
[INFINITY-1851] Make cassandra backup work in soak.

jsrodman:
* Ensure clean database state; delete the tests tables before running the
  test, so repeated runs don't hide real problems.
* When triggering the backup plan, use the actual service name, not the
  package name.
* Since adjusting the test code anyway, drop import *


jsrodman:
This change isn't actually sufficient to make cassandra backup reliable again, but i have manually demonstrated the problems identified here are actually fixed.  We just are still failing to run the verify_deletion job until INFINITY-2174 is merged.

jsrodman:
@mrbrowning : You might have opinions on this, but not officially making you a reviewer for the moment due to vacations.

frameworks/cassandra/tests/config.py
```diff
@@ -1,17 +1,21 @@
...
 import sdk_plan
 import sdk_utils
 
+log = logging.getLogger(__name__)
 
 PACKAGE_NAME = 'cassandra'
 DEFAULT_TASK_COUNT = 3
 FOLDERED_SERVICE_NAME = sdk_utils.get_foldered_name(PACKAGE_NAME)
 DEFAULT_CASSANDRA_TIMEOUT = 600
+SERVICE_NAME = os.environ.get('SERVICE_NAME') or PACKAGE_NAME
```
NimaVaziri: 1. Seems like this doesn't actually get used
2. @mellenburg is working on standardizing the consistency across service names and IIRC we're going with `FOLDERED_SERVICE_NAME` so we should probably use that everywhere

frameworks/cassandra/tests/config.py
```diff
@@ -89,6 +93,15 @@ def run_backup_and_restore(
     delete_data_job = get_delete_data_job(node_address=job_node_address)
     verify_deletion_job = get_verify_deletion_job(node_address=job_node_address)
 
+
+    # Ensure the keyspaces we will use aren't present.
+    try:
+        jobs.run_job(delete_data_job)
+    except:
```
NimaVaziri: :+1: to the try-catch block though is there a more specific exception we can catch here?

jsrodman: I have no idea of the total set of exceptions that can come out here.  The correct design would be to write a job that doesn't fail if the data isn't there which would remove the need for the try block, but i don't see a way to do that in cqlsh.

Presumably if we're tossing an exception for reasons unrelated to the delete action, all the other jobs are going to be failing too.

I could probably spend a while trying to investigate the possible exceptions, but I suspect I'll be incomplete and this will introduce unintended flakiness.

frameworks/cassandra/tests/config.py
jsrodman: The changes to pass in SERVICE_NAME are over in soak-cluster-configurations.  FOLDERED_SERVICE_NAME is NOT used by soak-cluster-configurations at present.

I could change this to SOAK_SERVICE_NAME or something to make it clear what this escape hatch is initially for, though it seemed like a generic escape hatch to me for the service name diverging from the package name.

jsrodman: The related pull is now open at https://github.com/mesosphere/soak-cluster-configurations/pull/149

NimaVaziri: Yeah until we the change in soak-cluster-configurations gets in, setting this to `SOAK_SERVICE_NAME` makes it more explicitly clear as to why we have two difference service names in the general `config.py` file

frameworks/cassandra/tests/config.py
NimaVaziri: Fair. At this time having the stack trace give info (even if it's about an unexpected failure) can possibly help diagnose problems out of this scope though I suspect the extent of that help won't go too far.

frameworks/cassandra/tests/test_soak.py
```diff
@@ -6,7 +6,7 @@
 import pytest
 import shakedown
 
-from tests.config import *
+import tests.config
```
mellenburg: Thanks!

NimaVaziri:
@jsrodman have we confirmed this works in soak?

jsrodman:
@NimaVaziri I no longer remember whether I proved this works in a CCM cluster using the soak automations or whether I ran it in a real soak cluster.  I think the former, but not certain at this time.

NimaVaziri:
@susanxhuynh do you have any insight on this change's status in soak?

elezar:
@susanxhuynh could we merge master into this?

susanxhuynh:
@NimaVaziri Soak doesn't run this until it gets merged into master, unless @jsrodman made a special change to test against this branch.

@jsrodman Please fix the merge conflict.

NimaVaziri:
Oh the JIRA for this was closed, I assumed that was because this PR got merged...

jsrodman:
Holy crow it passed.

PR #1409, opened 2017-08-09
Fix inverted section mustache rendering

gabrielhartmann:
The environment for mustache rendering should not consist entirely of
strings, but boolean objects in the case of "true" or "false" values

### DIFF
--- a/sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/yaml/TemplateUtilsTest.java
+++ b/sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/yaml/TemplateUtilsTest.java
@@ -100,4 +100,58 @@ public class TemplateUtilsTest {
                 Collections.emptyMap(),
                 TemplateUtils.MissingBehavior.EXCEPTION));
     }
+
+    @Test
+    public void testApplyTrueEnvVarToSection() throws IOException {
+        String filename = "test-render-inverted.yml";
+        String yaml = getYamlContent(filename);
+        Assert.assertTrue(yaml.contains("ENABLED"));
+
+        Map<String, String> envMap = new HashMap<>();
+        envMap.put("ENABLED", String.valueOf(true));
+        String renderedYaml = TemplateUtils.applyEnvToMustache(
+                filename, yaml, envMap, TemplateUtils.MissingBehavior.EMPTY_STRING);
+
+        Assert.assertTrue(renderedYaml.contains("cmd: ./enabled true"));
+        Assert.assertFalse(renderedYaml.contains("cmd: ./disabled"));
+        Assert.assertFalse(renderedYaml.contains("ENABLED"));
+    }
+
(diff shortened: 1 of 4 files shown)


sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/yaml/TemplateUtilsTest.java
```diff
@@ -100,4 +100,55 @@ public void testApplyMissingExceptionDisableBlockPasses() throws IOException {
...
+        String yaml = getYamlContent(filename);
+        Assert.assertTrue(yaml.contains("ENABLED"));
+
+        Map<String, String> envMap = new HashMap<>();
+        envMap.put("ENABLED", String.valueOf(true));
+        String renderedYaml = TemplateUtils.applyEnvToMustache(
+                filename, yaml, envMap, TemplateUtils.MissingBehavior.EMPTY_STRING);
+
+        Assert.assertTrue(renderedYaml.contains("cmd: ./enabled true"));
+        Assert.assertFalse(renderedYaml.contains("ENABLED"));
```
elezar: Should we check to ensure that `cmd: ./disabled true` is missing from this too? (also for `cmd: ./enabled false` in the test below)?

gabrielhartmann: Good idea.

gabrielhartmann:
@elezar:  Regarding:
```
 What is the expected behaviour if ENABLED="foobarbaz"
```
This evaluates to `true`.  That's just standard mustache behavior.  This PR is getting the library to execute correct mustache behavior environment context in the way it expects.

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/yaml/TemplateUtilsTest.java
```diff
@@ -100,4 +100,57 @@ public void testApplyMissingExceptionDisableBlockPasses() throws IOException {
...
+        String yaml = getYamlContent(filename);
+        Assert.assertTrue(yaml.contains("ENABLED"));
+
+        Map<String, String> envMap = new HashMap<>();
+        envMap.put("ENABLED", "");
+        String renderedYaml = TemplateUtils.applyEnvToMustache(
+                filename, yaml, envMap, TemplateUtils.MissingBehavior.EMPTY_STRING);
+
+        Assert.assertTrue(renderedYaml.contains("cmd: ./disabled"));
+        Assert.assertFalse(renderedYaml.contains("cmd: ./disabled false"));
```
elezar: Here `cmd: ./enabled` should also be missing, right?

gabrielhartmann: Added

PR #1410, opened 2017-08-10
[INFINITY-2005][HDFS] Name node test failure fix

NimaVaziri:
The journal nodes _might_ be out of quorum if they've gone through a restart or crash (which happens currently through the `test_kill_all_journalnodes` test). In a successive test (`test_permanently_replace_namenodes`), upon a replace of a name node pod, the name node bootstrap task can't bootstrap and goes into an infinite restart loop trying to contact the journal nodes.

This updates how we account for journal nodes being ready in order to retain quorum such that name nodes, as a result, can function correctly.


frameworks/hdfs/tests/config.py
```diff
@@ -74,7 +75,23 @@ def run_hdfs_command(svc_name, command):
     return rc, output
 
 
-def check_healthy(count=DEFAULT_TASK_COUNT):
-    sdk_plan.wait_for_completed_deployment(PACKAGE_NAME, timeout_seconds=25 * 60)
-    sdk_plan.wait_for_completed_recovery(PACKAGE_NAME, timeout_seconds=25 * 60)
-    sdk_tasks.check_running(PACKAGE_NAME, count)
+def check_healthy(service_name=PACKAGE_NAME, count=DEFAULT_TASK_COUNT, recovery_expected=False):
```
elezar: I think I would argue for making `service_name` a non-default parameter. This would remove the chance of forgetting to pass in the correct value (generally `FOLDERED_SERVICE_NAME`).

frameworks/hdfs/tests/config.py
```diff
@@ -74,7 +75,23 @@ def run_hdfs_command(svc_name, command):
...
+    # TODO(elezar, nima) check_healthy also check for complete deployment, and this should not
+    # affect the state of recovery.
+    check_healthy(service_name=service_name, count=DEFAULT_TASK_COUNT, recovery_expected=True)
+
+
+def get_pod_type_instances(pod_type_prefix, service_name=FOLDERED_SERVICE_NAME):
+    pod_types = sdk_cmd.convert_output_to_list(
+        sdk_cmd.run_cli("hdfs --name={} pod list".format(service_name))
+    )
+    return [pod_type for pod_type in pod_types if pod_type_prefix in pod_type]
```
elezar: The argument name `pod_type_prefix` implies that the condition for the generator expression should be:
```
[pod_type for pod_type in pod_types if pod_type.startswith(pod_type_prefix)]
```
It could be that I'm missing something though. 

Note, in this function we default to `FOLDERED_SERVICE_NAME` which is inconsistent with the other functions here.

testing/sdk_cmd.py
```diff
@@ -32,3 +33,12 @@ def run_cli(cmd, print_output=True):
         log.error(err)
         raise dcos.errors.DCOSException(err)
     return stdout
+
+
+def convert_output_to_list(output):
```
elezar: The only place this is currently used is where the list of pods of a particular type is returned. Would it not make more sense to wrap `sdk_cmd.run_cli()` with this instead of requiring that this be done at each point where the function is called?

NimaVaziri: Yeah I thought about that, then I ended up considering the case where the output of `run_cli` is not expected to be a list, in which case the wrap would break.

frameworks/hdfs/tests/test_shakedown.py
```diff
@@ -14,67 +14,61 @@
...
 import sdk_upgrade
 import sdk_utils
-from tests.config import (
-    DEFAULT_HDFS_TIMEOUT,
-    DEFAULT_TASK_COUNT,
-    FOLDERED_SERVICE_NAME,
-    PACKAGE_NAME,
-    ZK_SERVICE_PATH
-)
+import tests.config
```
NimaVaziri: Nope, was just reviewing https://github.com/mesosphere/dcos-commons/pull/1407/files right before getting to this section and wanted to ensure consistency, though the previous version seems to be less verbose so I'll revert.

frameworks/hdfs/tests/config.py
NimaVaziri: Yeah, that works better. I updated the parameter name afterwards to better match the behavior, should have updated the if clause as well.

frameworks/hdfs/tests/config.py
NimaVaziri: :+1:

NimaVaziri: Will do the same with `expected_recovery` to enforce this behavior upstream.

frameworks/hdfs/tests/config.py
```diff
@@ -74,7 +75,23 @@ def run_hdfs_command(svc_name, command):
...
+
+
+def expect_recovery(service_name=PACKAGE_NAME):
+    # TODO(elezar, nima) check_healthy also check for complete deployment, and this should not
+    # affect the state of recovery.
+    check_healthy(service_name=service_name, count=DEFAULT_TASK_COUNT, recovery_expected=True)
+
+
+def get_pod_type_instances(pod_type_prefix, service_name=FOLDERED_SERVICE_NAME):
+    pod_types = sdk_cmd.convert_output_to_list(
```
elezar: Looking at this again, I don't know if this is going to do what we want. The output from `run_cli` is a string (newline-separated), but we don't split the output here meaning that the output of this function will always be a list with a single string element.

frameworks/hdfs/tests/test_shakedown.py
elezar: For the record, I don't have a strong opinion. Was just asking.

testing/sdk_cmd.py
elezar: Yes, I just looked at it again. Missed the fact that the output is just a string and not the lines from the output.

frameworks/hdfs/tests/config.py
NimaVaziri: I've tested this locally and `ast.literal_eval()` takes care of that for us. It turns a list of elements represented as a newline-separated string into a list type. We then do the strip on each element in the list compression in `convert_output_to_list` in case there's extra whitespace (though that's just precautionary).

elezar: Ok, that isn't entirely clear. I have just tested here:
```
python3 -c "import ast; ast.literal_eval('foo\nbar')"
```
and get the output:
```
  File "<string>", line 1, in <module>
  File "/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ast.py", line 48, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 2
    bar
      ^
SyntaxError: unexpected EOF while parsing
```
Am I missing something?

NimaVaziri: Oh totally missed this. It looks like the input in that example wasn't correctly formatted.
```
Python 2.7.10 (default, Feb  7 2017, 00:08:15)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ast
>>> pods = """[
...   "data-0",
...   "data-1",
...   "data-2",
...   "journal-0",
...   "journal-1",
...   "journal-2",
...   "name-0",
...   "name-1"
... ]"""
>>> ast.literal_eval(pods)
['data-0', 'data-1', 'data-2', 'journal-0', 'journal-1', 'journal-2', 'name-0', 'name-1']
```

```
Python 3.6.1 (v3.6.1:69c0db5050, Mar 21 2017, 01:21:04)
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ast
>>> pods = """[
...   "data-0",
...   "data-1",
...   "data-2",
...   "journal-0",
...   "journal-1",
...   "journal-2",
...   "name-0",
...   "name-1"
... ]"""
>>> ast.literal_eval(pods)
['data-0', 'data-1', 'data-2', 'journal-0', 'journal-1', 'journal-2', 'name-0', 'name-1']
```

benclarkwood:
@elezar good with the changes?

frameworks/hdfs/tests/config.py
elezar: Ah, I see. It works because the output of the command is `json`. Silly me.

testing/sdk_cmd.py
```diff
@@ -34,3 +35,12 @@ def run_cli(cmd, print_output=True, return_stderr_in_stdout=False):
     if return_stderr_in_stdout:
         stdout = stdout + "\n" + stderr
     return stdout
+
+
+def convert_output_to_list(output):
```
elezar: Now that I have a clearer picture of what this does, I think the name is a bit misleading, as it REQUIRES that the input already be the string representation of a python list. Not a blocker though.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -148,7 +148,7 @@ def test_kill_all_journalnodes():
 
     for journal_pod in get_pod_type_instances("journal", FOLDERED_SERVICE_NAME):
         sdk_cmd.run_cli('hdfs --name={} pod restart {}'.format(FOLDERED_SERVICE_NAME, journal_pod))
-        # wait for the entire service to recover as name nodes are affected as well when journal nodes go down
+        # wait for the entire service to recover as name nodes will crash if quorum is lost among journal nodes
         expect_recovery(service_name=FOLDERED_SERVICE_NAME)
```
elezar: Is recovery expected here though? If the quorum is not lost, then the namenodes should stay up, right?

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -332,32 +342,19 @@ def test_metrics():
     sdk_metrics.wait_for_any_metrics(FOLDERED_SERVICE_NAME, "journal-0-node", DEFAULT_HDFS_TIMEOUT)
 
 
+
```
gabrielhartmann: Too many new lines

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -143,10 +146,11 @@ def test_kill_all_journalnodes():
     journal_ids = sdk_tasks.get_task_ids(FOLDERED_SERVICE_NAME, 'journal')
     data_ids = sdk_tasks.get_task_ids(FOLDERED_SERVICE_NAME, 'data')
 
-    for host in shakedown.get_service_ips(FOLDERED_SERVICE_NAME):
-        sdk_tasks.kill_task_with_pattern('journalnode', host)
+    for journal_pod in get_pod_type_instances("journal", FOLDERED_SERVICE_NAME):
+        sdk_cmd.run_cli('hdfs --name={} pod restart {}'.format(FOLDERED_SERVICE_NAME, journal_pod))
+        # wait for the entire service to recover as name nodes will crash if quorum is lost among journal nodes
+        expect_recovery(service_name=FOLDERED_SERVICE_NAME)
```
gabrielhartmann: Leave this outside the loop.  Let the name nodes crash.  It's fine.  The service should recover.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -159,9 +163,9 @@ def test_kill_all_namenodes():
     name_ids = sdk_tasks.get_task_ids(FOLDERED_SERVICE_NAME, 'name')
     data_ids = sdk_tasks.get_task_ids(FOLDERED_SERVICE_NAME, 'data')
 
-    for host in shakedown.get_service_ips(FOLDERED_SERVICE_NAME):
-        sdk_tasks.kill_task_with_pattern('namenode', host)
-    expect_recovery()
+    for name_pod in get_pod_type_instances("name", FOLDERED_SERVICE_NAME):
+        sdk_cmd.run_cli('hdfs --name={} pod restart {}'.format(FOLDERED_SERVICE_NAME, name_pod))
+        expect_recovery(service_name=FOLDERED_SERVICE_NAME)
```
gabrielhartmann: Same here, kill both at the same time.

PR #1411, opened 2017-08-10
[KAF-341] Add TLS support in Kafka

mhrabovcin:


frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
```diff
@@ -84,17 +90,38 @@ public JSONObject getTopic(String topicName) throws Exception {
...
+
+        List<String> ids = zkClient.getChildren().forPath(IDS_PATH);
+        for (String id : ids) {
+            byte[] bytes = zkClient.getData().forPath(IDS_PATH + "/" + id);
+            JSONObject broker = new JSONObject(new String(bytes, StandardCharsets.UTF_8));
+            String mappedProtocolName = broker.getJSONObject("listener_security_protocol_map").getString(protocolName);
+            StreamSupport.stream(broker.getJSONArray("endpoints").spliterator(), false)
+                    .map(endpoint -> endpoint.toString())
+                    .filter(endpoint -> endpoint.startsWith(mappedProtocolName))
+                    .map(endpoint -> endpoint.substring((mappedProtocolName + "://").length()))
```
elezar: In python, I would suggest `endpoint.lstrip(mappedProtocolName + "://")` to make it clearer what is being done here. Is there something similar in Java?

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
```diff
@@ -84,17 +90,38 @@ public JSONObject getTopic(String topicName) throws Exception {
...
+                    .forEach(endpoint -> endpoints.add(endpoint));
```
elezar: According to: https://stackoverflow.com/questions/14830313/retrieving-a-list-from-a-java-util-stream-stream-in-java-8 this `.forEach` may be racy for parallel stream.

We could consider something like:
```java
endpoints.addAll(StreamSupport.stream(...).map(...).collect(Collectors.toList());
```
Does someone else have any insights into this?

frameworks/kafka/src/main/dist/server.properties.mustache
```diff
@@ -27,9 +27,28 @@ broker.id={{POD_INSTANCE_INDEX}}
 #     listeners = security_protocol://host_name:port
 #   EXAMPLE:
 #     listeners = PLAINTEXT://your.host.name:9092
+{{^KAFKA_ENABLE_TLS}}
```
nickbp: FWIW naming aside, svc.yml templating is based on the scheduler's environment, while templating in this config file is based on the task's environment

frameworks/kafka/universe/marathon.json.mustache
```diff
@@ -61,6 +61,12 @@
     "CNI_PLUGIN_LABELS": "{{brokers.cni_plugin_labels}}",
     {{/service.virtual_networks}}
 
+    {{#service.tls}}
+    "ENABLE_TLS": "yes",
+    "BROKER_PORT_TLS": "{{brokers.port_tls}}",
+    "TASKCFG_ALL_KAFKA_ENABLE_TLS": "{{service.tls}}",
```
nickbp: If you wanted, ENABLE_TLS could be removed in favor of having svc.yml's templating use `TASKCFG_ALL_KAFKA_ENABLE_TLS` directly. Then it'd be a single value across the board.

frameworks/helloworld/tests/test_tls.py
```diff
@@ -263,8 +264,9 @@ def test_changing_discovery_replaces_certificate_sans(hello_world_service):
     config = sdk_marathon.get_config(PACKAGE_NAME)
     config['env']['DISCOVERY_TASK_PREFIX'] = DISCOVERY_TASK_PREFIX + '-new'
     sdk_marathon.update_app(PACKAGE_NAME, config)
+    sdk_tasks.check_tasks_updated(PACKAGE_NAME, 'discovery', original_tasks)
```
benclarkwood: I suggested this solution. Checking the tasks are updated should be sufficient.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
```diff
@@ -26,6 +28,10 @@
     private static final String IDS_PATH = "/brokers/ids";
     private static final String TOPICS_PATH = "/brokers/topics";
 
+    private final String PROTOCOL_NAME_PLAINTEXT = "PLAINTEXT";
+    private final String PROTOCOL_NAME_TLS = "SSL";
+
```
benclarkwood: extra newline

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
```diff
@@ -84,17 +90,38 @@ public JSONObject getTopic(String topicName) throws Exception {
...
+        try {
+            return getBrokerEndpoints(PROTOCOL_NAME_TLS);
+        } catch (Exception ex) {
+            log.error("Failed to retrieve broker TLS endpoints with exception: ", ex);
+        }
+        return Collections.emptyList();
+    }
+
+    private List<String> getBrokerEndpoints(String protocolName) throws Exception {
+        List<String> endpoints = new ArrayList<>();
```
benclarkwood: Method local variables should be final if possible.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
benclarkwood: Yes, `.collect(...)` is the cleaner / safer way of doing this.

What you wrote, @elezar, will work.

frameworks/kafka/tests/test_tls.py
```diff
@@ -0,0 +1,100 @@
...
+
+    sdk_cmd.run_cli("package install --yes dcos-enterprise-cli")
+
+
+@pytest.fixture(scope='module')
+def service_account(dcos_security_cli):
+    """
+    Creates service account with `hello-world` name and yields the name.
+    """
+    name = 'kafka'
```
benclarkwood: Essentially all test suites first uninstall and then reinstall the framework. Should be fine.

frameworks/kafka/tests/test_tls.py
```diff
@@ -0,0 +1,100 @@
...
+    endpoint_tls = service_cli(
+        'endpoints {name}'.format(name=BROKER_TLS_ENDPOINT)
+    )
+    assert len(endpoint_tls['dns']) == DEFAULT_BROKER_COUNT
+
+
+@pytest.mark.tls
+@pytest.mark.smoke
+@pytest.mark.sanity
+@sdk_utils.dcos_1_10_or_higher
```
benclarkwood: I thought this required a shakedown import?

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
elezar: Could we add that to our `findBugs` definition. Also, could we have a repo with our "shared" intellij settings?

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
elezar: Also, I think `addAll()` accepts a `Collection`, so one could even use `Collectors.toCollection()`. Does anyone know if there may be some performance benefit to this?

frameworks/kafka/tests/test_tls.py
elezar: It does, but it seems as if this is imported already.

frameworks/kafka/docs/configure.md
```diff
@@ -294,6 +303,45 @@ To configure it:
...
+{
+    "service": {
+        "tls": true
+    }
+}
+```
+
+Once the TLS is enabled, Kafka will do inter-node communication over secure TLS connections. Additional TLS port named `broker-tls` will be available for clients connecting to the service. Only the [`TLSv1.2`](https://www.ietf.org/rfc/rfc5246.txt) is supported. The non-TLS `broker` port will be still available.
+
+Enabling the TLS is possible only in `permissive` or `strict` cluster security modes. Both modes **require** a [service account](https://docs.mesosphere.com/service-docs/kafka/kafka-auth/). Additionally, the service account **must have** the `dcos:superuser` permission. If the permission is missing the Kafka scheduler will not abe able to provision TLS artifacts.
```
mhrabovcin: It is possible to launch Kafka with TLS in both `permissive` and `strict` security modes.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
```diff
@@ -84,17 +90,38 @@ public JSONObject getTopic(String topicName) throws Exception {
...
+
+        List<String> ids = zkClient.getChildren().forPath(IDS_PATH);
```
mhrabovcin: I'll keep it as it is if you don't mind as the `listBrokers` swallows ZK error.

frameworks/kafka/src/main/java/com/mesosphere/sdk/kafka/api/KafkaZKClient.java
mhrabovcin: Not in std library - https://stackoverflow.com/questions/3100295/java-equivalent-of-pythons-string-lstrip

frameworks/kafka/tests/test_tls.py
mhrabovcin: The `sdk_utils` package imports the `shakedown`.

mhrabovcin:
I've implemented requested changes @benclarkwood @elezar @nickbp

frameworks/kafka/src/main/dist/server.properties.mustache
```diff
@@ -27,9 +27,28 @@ broker.id={{POD_INSTANCE_INDEX}}
...
 listeners=PLAINTEXT://:{{KAFKA_BROKER_PORT}}
+{{/KAFKA_ENABLE_TLS}}
 
-# Hostname and port the broker will advertise to producers and consumers. If not set, 
+{{#KAFKA_ENABLE_TLS}}
+############################# TLS Settings #############################
+listeners=PLAINTEXT://:{{KAFKA_BROKER_PORT}},SSL://:{{KAFKA_BROKER_PORT_TLS}}
```
benclarkwood: @mhrabovcin Let's make TLS configurable like this:

- "is it enabled?"
- "leave plaintext enabled?" This can be accomplished using an inverted section:
```
{{#KAFKA_ENABLE_TLS_ALLOW_PERMISSIVE}}
listeners=PLAINTEXT://:{{KAFKA_BROKER_PORT}},SSL://:{{KAFKA_BROKER_PORT_TLS}}
{{/KAFKA_ENABLE_TLS_ALLOW_PERMISSIVE}}
{{^KAFKA_ENABLE_TLS_ALLOW_PERMISSIVE}}
listeners=SSL://:{{KAFKA_BROKER_PORT_TLS}}
{{/KAFKA_ENABLE_TLS_ALLOW_PERMISSIVE}}

File: frameworks/kafka/src/main/dist/svc.yml [hunk 13]
```diff
@@ -31,6 +31,15 @@ pods:
             vip:
```
benclarkwood: We should also make the broker plain text port only be advertised if TLS is enabled and plaintext comms are still allowed.

frameworks/kafka/tests/test_tls.py
benclarkwood: I'll figure out Python imports someday :/

mhrabovcin:
@benclarkwood I've added an option to run Kafka only over TLS.

nickbp:
Are all of these configurations expected to be used in practice? (More of a Kafka question):
- TLS only
- TLS and plaintext
- Plaintext only

benclarkwood:
@nickbp I believe so. I think the scenarios are:

1. Plaintext only: You're running dc/os open (no TLS) or you're totally firewalled and don't care about encryption

2. Plaintext and TLS: You allow plaintext comms within the cluster to save on CPU cycles and only route TLS externally

3. TLS Only: Zero trust security model / it's 2017 and the overhead of TLS is pretty low at this point.

benclarkwood:
Looks great, @mhrabovcin!

PR #1414, opened 2017-08-10
get all task logs, not just scheduler

falseperfection:


gabrielhartmann:
@falseperfection: Do you work at Mesosphere?

PR #1415, opened 2017-08-10
[INFINITY-2186] Add retry to uninstall for marathon flakes

benclarkwood:
Uninstall can fail if Cosmos can't talk to Marathon when the uninstall is initiated. This can cause test cleanups to leak. 

I've added a retry that reattempts uninstall if the error encountered contains "marathon".

Ideally, I'd just retry on any 5XX response, but Cosmos returns a 400 :/

### DIFF
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -83,7 +83,20 @@ def install(
         package_name, service_name, shakedown.pretty_duration(time.time() - start)))
 
 
-def uninstall(
+@retry(stop_max_attempt_number=5, wait_fixed=5000, retry_on_exception=lambda e: isinstance(e, dcos.errors.DCOSException))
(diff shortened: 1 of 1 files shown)


testing/sdk_install.py
```diff
@@ -102,6 +115,9 @@ def uninstall(
         except (dcos.errors.DCOSException, ValueError) as e:
             log.info('Got exception when uninstalling package, ' +
                           'continuing with janitor anyway: {}'.format(e))
+            if 'marathon' in str(e):
```
nickbp: just in case: this also applies to the old janitor uninstall flow in cosmos?

benclarkwood: Yep, it's actually a package describe call that is made before the package uninstall call.

PR #1417, opened 2017-08-10
[INFINITY-2187] Process TaskStatus messages synchronously

gabrielhartmann:
Implicit acknowledgement of TaskStatus messages requires that exit from
the Scheduler `statusUpdate` function implies complete processing of the
message.  Processing on an async thread breaks this assumption.

It's also making the DefaultSchedulerTest.testConfigurationUpdate non-deterministic. 

NOTE:  The diff looks weird.  I just removed the executor service and stopped wrapping calls to `statusUpdate` with use of that service.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
@@ -770,45 +770,43 @@ public class DefaultScheduler extends AbstractScheduler implements Observer {
 
     @Override
     public void statusUpdate(SchedulerDriver driver, Protos.TaskStatus status) {
-        statusExecutor.execute(() -> {
-            LOGGER.info("Received status update for taskId={} state={} message={} protobuf={}",
-                    status.getTaskId().getValue(),
-                    status.getState().toString(),
-                    status.getMessage(),
-                    TextFormat.shortDebugString(status));
+        LOGGER.info("Received status update for taskId={} state={} message={} protobuf={}",
+                status.getTaskId().getValue(),
+                status.getState().toString(),
+                status.getMessage(),
+                TextFormat.shortDebugString(status));
 
-            // Store status, then pass status to PlanManager => Plan => Steps
-            try {
-                stateStore.storeStatus(status);
-                planCoordinator.getPlanManagers().forEach(planManager -> planManager.update(status));
-                reconciler.update(status);
+        // Store status, then pass status to PlanManager => Plan => Steps
+        try {
+            stateStore.storeStatus(status);
+            planCoordinator.getPlanManagers().forEach(planManager -> planManager.update(status));
+            reconciler.update(status);
 
-                if (StateStoreUtils.isSuppressed(stateStore)
-                        && !StateStoreUtils.fetchTasksNeedingRecovery(stateStore, configStore).isEmpty()) {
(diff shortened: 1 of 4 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -770,45 +770,43 @@ public boolean apiServerReady() {
...
+                    status.getContainerStatus().getNetworkInfosCount() > 0 &&
+                    status.getContainerStatus().getNetworkInfosList().stream()
+                            .anyMatch(networkInfo -> networkInfo.getIpAddressesCount() > 0)) {
+                // Map the TaskStatus to a TaskInfo. The map will throw a StateStoreException if no such
+                // TaskInfo exists.
+                try {
+                    Protos.TaskInfo taskInfo = StateStoreUtils.getTaskInfo(stateStore, status);
+                    StateStoreUtils.storeTaskStatusAsProperty(stateStore, taskInfo.getName(), status);
+                } catch (StateStoreException e) {
+                    LOGGER.warn("Unable to store network info for status update: " + status, e);
```
elezar: This warning is a little confusing as the exception that is raised doesn't have anything to do with network information directly (and neither does the call to `storeTaskStatusAsProperty`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -770,45 +770,43 @@ public boolean apiServerReady() {
...
+            if (StateStoreUtils.isSuppressed(stateStore)
+                    && !StateStoreUtils.fetchTasksNeedingRecovery(stateStore, configStore).isEmpty()) {
+                revive();
+            }
+
+            // If the TaskStatus contains an IP Address, store it as a property in the StateStore.
+            // We expect the TaskStatus to contain an IP address in both Host or CNI networking.
+            // Currently, we are always _missing_ the IP Address on TASK_LOST. We always expect it
+            // on TASK_RUNNINGs
+            if (status.hasContainerStatus() &&
```
elezar: I don't see that we check for the requirement that you mention (the IP address on`TASK_RUNNING`). Is the failure of this conditional statement an error state at all?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
```diff
@@ -182,20 +182,18 @@ protected void processOfferSet(List<Protos.Offer> offers) {
...
-
-            try {
-                stateStore.storeStatus(status);
-                reconciler.update(status);
-            } catch (Exception e) {
-                LOGGER.warn("Failed to update TaskStatus received from Mesos. "
-                        + "This may be expected if Mesos sent stale status information: " + status, e);
-            }
-        });
+        LOGGER.info("Received status update for taskId={} state={} message='{}'",
```
elezar: This log message prints a different amount of information than that for the `DefaultScheduler`. Is that intentional?

If it is important to have this kind of logging uniform, one option would be to have `statusUpdate` implemented as a template method in the parent calling hooks (e.g. `logStatusUpdate`, `processStatusUpdate`) to do the real work and logging.

PR #1420, opened 2017-08-10
Delete doc for unsupported feature.

triclambert:


gabrielhartmann:
@sascala: Will this have the desired effect everywhere?

PR #1423, opened 2017-08-12
INFINITY-1964 Automatically replace a pod if its initial launch fails

nickbp:
When a task is launched for the first time at a given location (either no preceding tasks, or all tasks marked permanently failed) then the following TASK_STAGING TaskStatus is flagged with an additional `initial_launch` label.

When the task launches, either Mesos will send us a TASK_RUNNING TaskStatus, at which point the above `initial_launch` label is overwritten, or Mesos will send a failure state, at which point we see the `initial_launch` label and determine that the task can be safely replaced elsewhere.

This flow is intended to keep tasks from getting stuck on hosts which are failing to launch processes. If the task had never even successfully entered a RUNNING state, then the scheduler will move on with a clean slate, as there's no data to be saved by keeping the task at the failed location.


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/taskdata/AuxLabelAccess.java
```diff
@@ -59,6 +59,37 @@ public static void setNetworkLabels(Protos.NetworkInfo.Builder networkInfoBuilde
...
+     * <ul>
+     * <li>This is the first time it's been launched on this machine (initial deployment or pod replacement)</li>
+     * <li>The scheduler hasn't yet received a TaskStatus update for this task from Mesos, indicating a successful or
+     * failed launch.</li>
+     * </ul>
+     * When this is the case, the task could be relaunched elsewhere without worrying about losing persistent data.
+     */
+    public static boolean isInitialLaunch(Protos.TaskStatus taskStatus) {
+        String val = LabelUtils.toMap(taskStatus.getLabels()).get(LabelConstants.INITIAL_LAUNCH_LABEL);
+        return val != null && val.equals(LabelConstants.BOOLEAN_LABEL_TRUE_VALUE);
```
elezar: Yes, I've seen tools like `errorProne` suggest that `LabelConstants.BOOLEAN_LABEL_TRUE_VALUE.equals(val)` is a better form of this.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/OfferAccepter.java
```diff
@@ -23,30 +21,18 @@
...
-    @Inject
-    public OfferAccepter(Set<OperationRecorder> recorders) {
-        this.recorders = recorders;
-    }
-
     public List<OfferID> accept(SchedulerDriver driver, List<OfferRecommendation> recommendations) {
-        return accept(driver, recommendations, getFilters());
-    }
-
-    public List<OfferID> accept(SchedulerDriver driver, List<OfferRecommendation> recommendations, Filters filters) {
```
nickbp: Yes

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/taskdata/TaskLabelWriter.java
```diff
@@ -36,38 +36,25 @@ public TaskLabelWriter(TaskInfo.Builder taskInfoBuilder) {
...
      */
     public TaskLabelWriter setPermanentlyFailed() {
         writer.put(LabelConstants.PERMANENTLY_FAILED_LABEL, LabelConstants.BOOLEAN_LABEL_TRUE_VALUE);
         return this;
     }
 
     /**
-     * Ensures that the task is not identified as permanently failed.
-     */
-    public TaskLabelWriter clearPermanentlyFailed() {
```
nickbp: Correct. Rather than explicitly clearing the bit, it's getting overwritten when the new task is launched.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/PersistentLaunchRecorder.java
```diff
@@ -35,56 +37,81 @@ public void record(OfferRecommendation offerRecommendation) throws Exception {
         LaunchOfferRecommendation launchOfferRecommendation = (LaunchOfferRecommendation) offerRecommendation;
         Protos.TaskInfo taskInfo = launchOfferRecommendation.getStoreableTaskInfo();
 
-        Protos.TaskStatus taskStatus = null;
+        Optional<PodInstance> podInstance = getPodInstance(taskInfo);
+        final boolean isInitialLaunch;
+        if (!podInstance.isPresent()) {
```
nickbp: This is doing a reverse lookup to find the ServiceSpec/PodSpec entry that a given TaskInfo is associated with.

It could be missing due to:
- TaskInfo without the expected labels (bad data, or extremely old data)
- ServiceSpec has been changed since this task was launched to no longer have this PodSpec

TODO: Put the above in a comment in the code

nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/recovery/FailureUtils.java
```diff
@@ -26,64 +29,16 @@ public static boolean isLabeledAsFailed(Protos.TaskInfo taskInfo) {
...
-    public static Protos.TaskInfo markFailed(Protos.TaskInfo taskInfo) {
-        return taskInfo.toBuilder()
-                .setLabels(new TaskLabelWriter(taskInfo)
-                        .setPermanentlyFailed()
-                        .toProto())
-                .build();
-    }
-
-    public static void markFailed(PodInstance podInstance, StateStore stateStore) {
+    public static void markFailed(StateStore stateStore, PodInstance podInstance) {
```
nickbp: Done

benclarkwood:
LGTM.

NimaVaziri:
@nickbp looks like this should be good to merge upon rebase + conflict resolution since the HDFS fix is now in.

PR #1424, opened 2017-08-14
[DCOS-17885] Remove base64 decoding from the SDK

mhrabovcin:
We'll be using base64 decoding workaround that is in private mesos secrets module - https://jira.mesosphere.com/browse/DCOS-17621


benclarkwood:
@ArtRand can you leverage this module in spark?

PR #1428, opened 2017-08-15
removed import * and isort

falseperfection:
Making `import *` lines explicit before working on extracting import-time cluster dependencies.


falseperfection:
I've caught a few missed import fixes, so will need at least one more full round of tests.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -14,70 +11,62 @@
...
-    DEFAULT_TASK_COUNT,
-    FOLDERED_SERVICE_NAME,
-    PACKAGE_NAME,
-    ZK_SERVICE_PATH,
-    check_healthy,
-    expect_recovery,
-    get_pod_type_instances
-)
+import shakedown
+from tests import config
```
NimaVaziri: Is there an obvious advantage that I'm failing to see over importing this way as opposed to what we're currently doing?

falseperfection: The short answer is making calls to outside modules more explicit and easier to track in a consistent way. `dcos/dcos` is using the Google Python Style Guide (with a few local exceptions -- like 120 character line length), and it uses package- and module-level importing.

PR #1430, opened 2017-08-15
[ELASTIC-116] Deploy Elastic with 0 ingest nodes by default

loren:


loren:
@benclarkwood I agree. It's essentially an unsupported scale down that is hardcoded into the framework. At least when the scheduler fails, it'll say why in the logs (i.e. `PodSpecsCannotShrink`). LMK if there's a better way we should handle it.

benclarkwood:
@loren is this ready to go?

loren:
@benclarkwood Yes it is. I was just waiting to see if elastic tests passed.

PR #1431, opened 2017-08-15
[ELASTIC-117] Automatic management of gateway settings (node count)

loren:


frameworks/elastic/src/main/dist/elasticsearch.yml
```diff
@@ -21,3 +21,7 @@ node.data: {{DATA_ENABLED}}
 node.ingest: {{INGEST_ENABLED}}
 metrics.statsd.host: {{STATSD_UDP_HOST}}
 metrics.statsd.port: {{STATSD_UDP_PORT}}
+gateway.expected_master_nodes: 3
```
benclarkwood: Won't this cause rolling restarts of everything when you add a single data node?

loren: It will, yes. According to Elasticsearch docs, a full cluster restart is required when these settings change. https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-gateway.html
I will ask around though to see if that really has to happen on every single node type. Seems like the sort of thing that could be specified just on the masters' `elasticsearch.yml` files.

loren: https://discuss.elastic.co/t/gateway-expected-nodes-and-full-cluster-restart/97151

benclarkwood: Given that this is the documented way, I'm okay with this for now. Interested to see what they say back.

loren: https://discuss.elastic.co/t/gateway-expected-nodes-and-full-cluster-restart/97151/2?u=loren

I changed the template to only apply to masters.

PR #1434, opened 2017-08-16
[INFINITY-2216] Improve toxic sidecar test

elezar:
This PR improves the toxic sidecar test by reading the contents of a file created by the sidecar. This is to prevent timing issues with checking if the `plan` is in the `STARTING` state as the transitions from `STARTING` to `PENDING` are often too fast to capture.


elezar:
Testing a fix in https://teamcity.mesosphere.io/viewLog.html?buildId=744562&buildTypeId=DcosIo_DcosCommons_Sdk_DcOs19Permissive&tab=buildLog which failed this morning.

frameworks/helloworld/src/main/dist/examples/sidecar.yml
```diff
@@ -40,7 +40,7 @@ pods:
       sidecar-toxic:
         goal: FINISHED
         cmd: >
-          echo 'I'm addicted to you / Don't you know that you're toxic?'
+          echo "I'm addicted to you / Don't you know that you're toxic?" > hello-container-path/toxic-output-$POD_INSTANCE_INDEX
```
elezar: It could be that this is failing due to the POD index.

benclarkwood: Yes. POD_INSTANCE_INDEX is only in the scheduler IIRC.

elezar:
I have kicked-off the hello-world-19 tests again. They failed due to a recovery test, and the toxic sidecar test passed.

PR #1435, opened 2017-08-16
[INFINITY-2205] Uninstall after reconciliation.

benclarkwood:
Create task kill steps for each task, processing kills as part of the plan.

This guarantees reconciliation is complete before the uninstall scheduler does anything.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/ResourceCleanupStep.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Step which implements the uninstalling of a particular reserved resource. For instance, persistent
+ * volumes and cpu.
+ */
+public class ResourceCleanupStep extends UninstallStep {
+    /**
+     * Creates a new instance with the provided {@code resourceId} and initial {@code status}.
+     */
+    public ResourceCleanupStep(String resourceId, Status status) {
```
nickbp: Status could be determined internally. status is being derived from the `resourceId` value

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/TaskKillStep.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Step which issues kill commands to all .
+ */
+public class TaskKillStep extends UninstallStep {
+
+    private TaskKiller taskKiller;
+    private Protos.TaskID taskID;
+
+    public TaskKillStep(Status status, Protos.TaskID taskID) {
```
nickbp: Given that the passed status is always PENDING, could skip the param and just set that internally

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
```diff
@@ -94,19 +93,28 @@ private Plan getPlan() {
 
         List<Phase> phases = new ArrayList<>();
 
+        // First, we kill all the tasks, so that we may release their reserved resources.
+        List<Step> taskKillSteps = stateStore.fetchStatuses().stream()
```
nickbp: Any reason for getting the ids from the statuses rather than the infos? Should probably have a comment explaining why

benclarkwood: This was fancier before. It should come from the Infos.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/ResourceCleanupStep.java
benclarkwood: I think it was done this way to not be passing the `StateStore` down to the individual resource cleaning steps (this is just a rename of `UninstallStep` so I could make a baseclass).

nickbp: No I mean upstream, it's doing this:

benclarkwood:
@nickbp feedback addressed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/ResourceCleanupStep.java
benclarkwood: Oh, whoops.

Indeed it could be. I am hesitant to re-roll CI to change this...

nickbp: I have a followup PR relating to plan changes where I was about to do this anyway, so I can do it then instead :)

PR #1432, opened 2017-08-15
[INFINITY-1260][INFINITY-2019] Change secret_name to service_account_secret

mbdcos:
secret_name in config.json makes people confused about file-based secrets. I saw at least two times in slack channel. User is thinking this is the file-based secret name. Just giving it a better name.

@keithchambers agreed `secret_name` -> `service_account_secret` change

related PR #1287

PR #1438, opened 2017-08-16
[INFINITY-1260] Change "principal" to "service_account" in config.json options 

mbdcos:
This changes only the keyword in config.json (what a typical user sees), not in yaml so SDK code should not be affected.

@keithchambers principle -> service_account change

related PR #1432

### DIFF
--- a/frameworks/cassandra/universe/config.json
+++ b/frameworks/cassandra/universe/config.json
@@ -15,8 +15,8 @@
           "type": "string",
           "default": ""
         },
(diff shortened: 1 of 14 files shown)


mbdcos:
I do not see any doc change, please comment if you find any.

mbdcos:
https://teamcity.mesosphere.io/viewQueued.html?itemId=745699&tab=queuedBuildOverviewTab

PR #1440, opened 2017-08-17
protect from subprocess failure

falseperfection:
Tests were failing because of an unhandled exception.

See: https://jira.mesosphere.com/browse/INFINITY-2231


conftest.py
```diff
@@ -44,7 +38,8 @@ def get_task_ids(user: str=None):
 def get_task_logs_for_id(task_id: str, lines: int=1000000):
     try:
         task_logs = subprocess.check_output([
-            'dcos', 'task', 'log', task_id, '--lines', str(lines)]).decode()
+            'dcos', 'task', 'log', task_id, '--lines', str(lines), 'stderr'
```
benclarkwood: We should capture both

cc @falseperfection

NimaVaziri: FWIW we don't currently use stdout for anything. Only somewhat useful info is confirming all expected artifacts are fetched properly.

conftest.py
```diff
@@ -42,8 +36,13 @@ def get_task_ids(user: str=None):
...
 def get_task_logs_for_id(task_id: str, lines: int=1000000):
-    return subprocess.check_output([
-        'dcos', 'task', 'log', task_id, '--lines', str(lines)]).decode()
+    try:
+        task_logs = subprocess.check_output([
+            'dcos', 'task', 'log', task_id, '--lines', str(lines), 'stderr'
+        ]).decode()
+        return task_logs
+    except subprocess.CalledProcessError:
+        return None
```
elezar: Note that we now fail silently. Would some logging output be useful to see what caused the failure?

conftest.py
benclarkwood: The scheduler nearly exclusively logs to stdout.

falseperfection: No worries: I'll get them both.

conftest.py
falseperfection: 👍

conftest.py
NimaVaziri: Right, I was thinking non-scheduler tasks for some reason...

PR #1441, opened 2017-08-17
[ELASTIC-126] Remove health checks from Elastic

loren:


keithchambers:
If we are confident that this makes Elastic more reliable, then I am comfortable removing the health checks.

deric:
What was the reasoning behind removing health checks? I understand that sometimes deploy might take longer than expected and killing nodes before other nodes boot up is not very effective.

Wouldn't be to prolong grace period until deploy is finished and then start executing health checks? Currently the only health checks executed is `/v1/plans/deploy` which returns "healthy" even though cluster deployment wasn't finished:
```
> GET /v1/plans/deploy HTTP/1.1
< HTTP/1.1 202 Accepted
"status" : "IN_PROGRESS"
...
```

PR #1444, opened 2017-08-17
Give "principal" in additional options, to install a previous version.

mbdcos:
New version uses `service_account` instead of `principal` in config.json.


mbdcos:
strict mode (local with test.sh) works!

benclarkwood:
Thanks for making quick work of this, @mbdcos!

PR #1445, opened 2017-08-17
Strict temp hack

benclarkwood:
Get around the fact that old packages don't have the new parameters by hardcoding a hack.


benclarkwood:
Not needed. Simpler fix in #1444

PR #1446, opened 2017-08-18
INFINITY-2282 Exercise new upgrade flows across all services

nickbp:
Includes a rename workaround: Cosmos doesn't support upgrades across package names, which causes problems for our "beta-pkg" => "pkg" flow.
To get around this, the test will perform surgery on the Scheduler's labels to make it appear to be from the target package.


nickbp:
Rather than depending on this name hack, I'm going to update the PR to just use beta-foo for our in-development package names.

nickbp:
Breaking out tests for just hello-world in a separate PR. This one will come later with wider changes.

mbdcos:
@nickbp 
"Cosmos doesn't support upgrades across package names, which causes problems for our "beta-pkg" => "pkg" flow."

`$ docs kafka update start --options=options.json --package-version=2.0.0-0.11.0.0`

So this will not work anymore?

nickbp:
The name of the CLI itself doesn't matter as long as the `--name=x` parameter is correct. The command just triggers an RPC to cosmos containing the name of the service to be updated, not the name of the CLI itself. Technically you could do a beta-cassandra update using a beta-kafka CLI.

PR #1447, opened 2017-08-18
[INFINITY-2232] Do not log TLS errors for tasks without TLS config

mhrabovcin:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -278,7 +278,10 @@ private static void logOutcome(StringBuilder stringBuilder, EvaluationOutcome ou
             Collection<Protos.TaskInfo> allTasks) {
         Map<String, ResourceSet> resourceSets = getNewResourceSets(podInstanceRequirement);
 
-        Optional<TLSEvaluationStage.Builder> tlsBuilder = getTLSEvaluationStageBuilderFromEnvironment(schedulerFlags);
+        Optional<TLSEvaluationStage.Builder> tlsBuilder = Optional.empty();
+        if (!TaskUtils.getTasksWithTLS(podInstanceRequirement).isEmpty()) {
```
elezar: One could also modify `getTLSEvaluationStageBuilderFromEnvironment` to accept the podInstanceRequirements as well and perform this conditional check. Was there a specific reason for not doing this? I only ask, because this would have required fewer changes in the two places where the function is called?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -91,6 +91,19 @@ private TaskUtils() {
     }
 
     /**
+     * Returns all the {@link TaskSpec} that have TLS configuration.
+     *
+     * @param serviceSpec A ServiceSpec defining service.
+     * @return A list of the task specs.
+     */
+    public static List<TaskSpec> getTasksWithTLS(PodInstanceRequirement podInstanceRequirement) {
```
benclarkwood: As Evan pointed out, let's add a unit test for this.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
benclarkwood: I agree, let's internalize the check to the helper method.

PR #1429, opened 2017-08-15
[KAF-362] Add upgrade documentation for Kafka upgrade from 0.10.2.x to 0.11.0

mbdcos:
Upgrading from 0.10 version to 0.11 version (See PR #1239)
(step 1) Update to 0.11 version but use an options.json file to set protocol/log version to 0.10
(step 2) Change protocol version to 0.11.0 and update again
(step 3) Change log version to 0.11.0 and update again

PR #1449, opened 2017-08-18
[INFINITY-2236] Add release notes explaining upgrade in strict mode

mbdcos:


triclambert:
@sascala we should make this clear which versions are impacted by this upgrade.  

@keithchambers you way want to weigh in here.

keithchambers:
Generally speaking -- I would like to limit (or completely avoid) documentation that discusses how users migrate from pre-GA to GA packages. With this in mind -- I have a couple questions:

* Is the purpose of this documentation to describe how you would upgrade from pre-GA packages to GA packages?

* If we were to assume that upgrade from pre-GA packages to GA packages is not supported, would this documentation (or some subset of this documentation) still be useful to the reader?

frameworks/cassandra/RELEASE_NOTES.md
```diff
@@ -0,0 +1,50 @@
+### Package Upgrade in Strict Mode
+
+If you are upgrading a service running in `strict` mode to the new package version (this version), you need to specify service account details as options during the package update process.
+
+
+You need to give service account details in options in order to authenticate and run in DC/OS `strict` mode. Or you set them in the GUI. Here is how a service is installed using a previous package version:
+
```
sascala: "You can provide service account details in a JSON options file or via the DC/OS GUI."

Is it helpful to show a user the old config? If it's not crucial, I suggest removing it. Users like to cut-and-paste and could end up with exactly the wrong configuration.

mbdcos:
This document is about upgrade using the new update feature. Yes, that is only available from beta to GA.  

For example, if we want to upgrade Kafka from  1.1.26-0.10.1.0-beta to 2.0.0-011.0,  this document explains why/how you need to give service_account details again (only for strict mode).

If upgrade from beta to GA is not supported, this document is still helpful. User may still want to upgrade. We probably will support upgrade between future releases. Similar scenario may come in the future. We may in the future change some other config options. Moreover, upgrading  with a newly added config option is a very good example. It kind of shows how the new update feature operates.

frameworks/cassandra/RELEASE_NOTES.md
```diff
@@ -0,0 +1,50 @@
...
+ $ cat options.json
+  {
+        "service": {
+            "principal": "this_is_your_service_account_id",
+            "secret_name": "this_is_your_sa_secret_path"
+        }
+    }
+```
+
```
sascala: To provide your service account details, add the following parameter to a JSON options file.

**Note:** The syntax for specifying service account details has changed from the previous version. The `principal` and `secret_name` parameters have changed to `service_account` and `service_account_secret`. If you are upgrading your service from the previous version, you will need to modify your JSON options file.

frameworks/cassandra/RELEASE_NOTES.md
```diff
@@ -0,0 +1,50 @@
...
+ $ cat options.json
+  {
+        "service": {
+            "service_account": "this_is_your_service_account_id",
+            "service_account_secret": "this_is_your_sa_secret_path"
+        }
+    }
+```
+
+**Note**: "principal and secret_name" options are changed to "service_account and service_account_secret".
```
sascala: Suggest removing this note and replacing with the one above.

frameworks/cassandra/RELEASE_NOTES.md
```diff
@@ -0,0 +1,50 @@
...
+
+If you are upgrading your service to the new package version, you need to set these options again.
```
sascala: ^ Also suggest removing this; this is covered in the note I wrote above.

frameworks/cassandra/RELEASE_NOTES.md
```diff
@@ -0,0 +1,50 @@
+### Package Upgrade in Strict Mode
```
sascala: Would you mind putting the following comment tags around this text? In the future I'm going to try to create just one document rather than duplicating info.

sascala:
It looks to me like these instructions are appropriate to the security section of the DC/OS site.

mbdcos:
@sascala  to clarify:

If we did not change `principal/secret_name` to `service_account/service_account_secret`:
normal upgrade procedure (operations-guide.md) is to do the following:
```
  $ docs cassandra update start  -package-version=<new_package_version_number>
```
Whatever config options you gave in the previous install are carried over.

Since `principal/secret_name` changed to `service_account/service_account_secret`,  new package does not read `principal/secret_name` anymore (even if those are carried over). 

Purpose of this document it to specify "If upgrading in strict mode, you need to give  `service_account/service_account_secret` again (as if you are installing fresh) in upgrade (because they do not exist in previous install).

```
  $ docs cassandra update start  --options=options.json --package-version=<new_package_version_number>

  $ cat options.json
  {
        "service": {
            "service_account": "this_is_your_service_account_id",
            "service_account_secret": "this_is_your_sa_secret_path"
        }
    }
```

mbdcos:
Fresh install (new):
```
 $ docs package install  cassandra --options=options.json ` --package-version=<new_package_version_number>

 $ cat options.json
  {
        "service": {
            "service_account": "this_is_your_service_account_id",
            "service_account_secret": "this_is_your_sa_secret_path"
        }
    }
```


Fresh install (old):
```
 $ docs package install  cassandra --options=options.json ` --package-version=<previous_package_version_number>

 $ cat options.json
  {
        "service": {
            "principal": "this_is_your_service_account_id",
            "secret_name": "this_is_your_sa_secret_path"
        }
    }
```

mbdcos:
Modified Cassandra RELEASE_NOTES.md. @sascala can you take a look?

frameworks/cassandra/RELEASE_NOTES.md
sascala: The second **Note:** Repeats information from the one above.

frameworks/cassandra/RELEASE_NOTES.md
```diff
@@ -31,12 +20,16 @@ In this new version, you need to set service account details as follows.
...
+
+**Note:** The syntax for specifying service account details has changed from the previous version. The `principal` and `secret_name` parameters have changed to `service_account` and `service_account_secret`. If you are upgrading your service from the previous version, you will need to specify service account details in a JSON options file.
+
+
 **Note**: "principal and secret_name" options are changed to "service_account and service_account_secret".
 
-If you are upgrading your service to the new package version, you need to set these options again.
+If you are upgrading your service in strict mode to the new package version, set `service_account` and `service_account_secret` options.
   
 ```
```
sascala: This code and text is a duplicate of what is above.

triclambert:
These release notes documents need a heading that captures exactly which version we're talking about.  Saying "the latest" will be confusing—we should use specific version numbers.  Also, this only applies to a beta package upgrade...  do we want to handle those differently, @keithchambers?

triclambert:
Maybe every repo should have an UPGRADE.md doc that we can link out to from the RELEASE_NOTES.md doc, LIMITATIONS.md, etc?  That way we can minimize the details there, and we can specifically codify version upgrade paths in UPGRADE.md (e.g. x -> y, do _this_; x -> z, do _that_).

mbdcos:
#1429 Kafka upgrade.md

keithchambers:
This is a good point Chris. Update documentation should be part of the main service documentation. Release notes should capture new feature, enhancements to existing features, bug fixes, break changes.

mbdcos:
upgrade instructions will not be in release notes

PR #1286, opened 2017-07-19
Fix apiPort breakage

NimaVaziri:
Fixes the breakage caused by merging a commit which included `apiPort` after `apiPort` had been removed from the `ServiceSpec1`

https://github.com/mesosphere/dcos-commons/commit/19f5541d4b3db9a215913a9ff120e0d5b185f5a2

PR #1450, opened 2017-08-18
[INFINITY-2075] Fix DefaultSchedulerTest.testConfigurationUpdate

benclarkwood:
Wait for task to be starting before sending TASK_RUNNING status as that is what happens in the real world.


NimaVaziri:
@benclarkwood given this only touches the SDK unit tests (which passed), should we merge or rerun the failed tests?

benclarkwood:
@NimaVaziri I see no reason to break protocol for this. Kicked off re-runs.

NimaVaziri:
Fair. Just reminded me of https://github.com/mesosphere/dcos-commons/pull/1286, save for the breakage.

benclarkwood:
True. That was at peak testing frustration. I want to get a sense for where we are at now (I think we are better than we were then).

PR #1451, opened 2017-08-18
Fix permanent replacement of tasks with the wrong config

gabrielhartmann:


gabrielhartmann:
I'm closing this and will come back to it.  Not going to make it into GA and it's not super urgent.

PR #1457, opened 2017-08-21
[INFINITY-2054] Extend Replace flag for SDK CLI Update subcommand

paegun:


docs/pages/operations-guide.md
```diff
@@ -571,6 +571,12 @@ If you are missing mandatory configuration parameters, the `update` command will
 $ dcos dse update start --options=options.json --package-version="1.1.6-5.0.7"
 ```
 
+If you want to update the configuration options as a full replace, instead of the default hash merge of configuration
```
triclambert: @keithchambers ^

keithchambers: I recommend describing the default behavior, how --replace if different than the default behavior, when you would want to use --replace over the default.

~ "The default behavior on update is to merge ‘Default’, ‘Stored’ and ‘Provided’ configurations, in that order, and then validate against the schema. In some situations, such as when a schema option has been removed, the default behavior might result in an invalidate configuration. You can work around this with the ‘--replace’ argument which, when specified, will override the ‘Stored’ options with the ‘Provided’ options."

PR #1459, opened 2017-08-21
[INFINITY-2099] Improve Plan subcommand documentation

paegun:
* Provide, since it otherwise doesn't exist, Plan subcommand
  documentation. Include cross-command (see also) bits where relevant.

### DIFF
--- a/docs/pages/operations-guide.md
+++ b/docs/pages/operations-guide.md
@@ -916,7 +916,8 @@ Display the status of the plan with the provided plan name
 dcos kafka plan status deploy
 ```
 
-NOTE: the --json flag, though not default is helpful in extracting phase UUIDs.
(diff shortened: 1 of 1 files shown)


docs/pages/operations-guide.md
```diff
@@ -901,6 +901,77 @@ $ dcos node ssh --master-proxy --leader "docker run mesosphere/janitor /janitor.
...
+```
+
+### Status
+Display the status of the plan with the provided plan name
+
+```bash
+dcos kafka plan status deploy
+```
+
+NOTE: the --json flag, though not default is helpful in extracting phase UUIDs.
```
nickbp: May be worth pointing out this bit of background: The UUIDs can be useful for avoiding race conditions where you get plan status, then plan status changed (e.g. a task was relaunched), then you sent a command. In theory the UUIDs should have changed in that scenario, and you'd get an error, whereas if you referred to the step by name then the race could occur.

PR #1462, opened 2017-08-21
py.test install() skip marathon deployment wait.

jsrodman:
Depends upon a change to shakedown; skip the marathon deployment
wait since it can fail in a way that is nearly identical to what we want to
retry.

### DIFF
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -34,7 +34,8 @@ def retried_shakedown_install(
         options_json=merged_options,
         wait_for_completion=True,
         timeout_sec=timeout_seconds,
-        expected_running_tasks=expected_running_tasks)
+        expected_running_tasks=expected_running_tasks,
+        wait_for_marathon_deployment=False)
 
 
 def install(


jsrodman:
Depends upon: https://github.com/dcos/shakedown/pull/167

mellenburg:
I realize that the precedent of using this repository as an upstream library has already been set, but I would strongly prefer that we avoid doing any more changes here to affect soak cluster testing. If we absolutely need to upstream something, we can find a way, but even then I would prefer copy/pasting all the shared code into soak then rely on a unpinned, moving target

jsrodman:
@mellenburg I don't understand. This change is not soak cluster specific. This is specific to sdk apps, but only being switched on in a minimal way in initially since that is the environment I'm reviewing actively. 

We're working around misbehavior in adminrouter which is exposed by cosmos which is exposed by shakedown. The bad behavior could easily result in incorrect reinstall in ci as well. 

The workaround is inelegant, so I won't be surprised if there is pushback or someone points out a better path, but if this is the best we can do so long as adminrouter and cosmos can flake this way, i expect to switch this on for all sdk_install paths. 

As for pin vs unpin in not following. Currently we have essentially two test sets in dcos commons, ci and soak. We could certainly do the work to create a split where the soak tests live with the soak configuration, but I'm not willing to stay copy pasting or vendoring. But we deploy nightly packages to soak, so that is "unpinned" as well. We may well have, or want to have, tests that track those packages. 

All that said, I don't see what risks this change imparts. Are you just expressing that you want to see more clear separation of soak tests and ci tests in general? I can agree with that goal.

mellenburg:
@jsrodman Sorry, I went about this backwards, my comment had more to do with https://github.com/dcos/shakedown/pull/167

Where you say:

> I'm trying to avoid failure on success, which has occurred in soak clusters.

So I was just thinking about that as an end-goal.

As for what would be a better solution, I would recommend looking at: https://github.com/mesosphere/dcos-commons/blob/master/testing/sdk_install.py#L66

This method could be changed to do a simple non-blocking package install POST followed by whatever criteria to check and retry. Its worth noting that there is yet another conditional statement about waiting deployment, so we have a total of three high-level wait calls anytime we wait:
1. wait_for_service_tasks_running: https://github.com/dcos/shakedown/blob/master/shakedown/dcos/package.py#L125
2. https://github.com/dcos/shakedown/blob/master/shakedown/dcos/package.py#L128
3. sdk_plan.wait_for_plan: https://github.com/mesosphere/dcos-commons/blob/master/testing/sdk_install.py#L72

Instead, we could contain all of the wait-specific logic here and no one needs to go out of their way to see how their tests are installing packages

jsrodman:
@mellenburg 

I don't agree that it makes sense to move wait-for-install-to-complete logic OUT of shakedown.  It's reasonable for waiting primitives to be available there that can be used across teams, projects, etc.

PR #1463, opened 2017-08-22
[INFINITY-2264, INFINITY-2270] Suppress transition from REVIVED to WAITING_FOR_OFFER

gabrielhartmann:
While testing we ran into this situation:

```
INFO  2017-08-22 01:44:32,946 [Thread-108] com.mesosphere.sdk.state.StateStore:storeStatus(198): Storing status 'TASK_KILLED' ...
...
INFO  2017-08-22 01:44:33,002 [pool-8-thread-1] com.mesosphere.sdk.scheduler.recovery.DefaultRecoveryPlanManager:getRecoveryRequirements(212): New pods needing recovery: []
...
INFO  2017-08-22 01:44:33,004 [pool-12-thread-1] com.mesosphere.sdk.scheduler.SuppressReviveManager:setOfferMode(210): Suppressing offers.
...
INFO  2017-08-22 01:44:37,241 [Thread-110] com.mesosphere.mesos.HTTPAdapter.MesosToSchedulerDriverAdapter:received(181): Received event of type: HEARTBEAT
INFO  2017-08-22 01:44:52,243 [Thread-111] com.mesosphere.mesos.HTTPAdapter.MesosToSchedulerDriverAdapter:received(181): Received event of type: HEARTBEAT
```

Because suppress/revive is now polling there's a possible race between notification of failure and detection of need to revive.  In short it needs to detect failures in the same way the RecoveryManager does to ensure reviving.  We should change plan updates to be less dependent on the Offer cycle, but that's out of scope at this late date.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SuppressReviveManager.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SuppressReviveManager.java
@@ -2,9 +2,13 @@ package com.mesosphere.sdk.scheduler;
 
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
+import com.mesosphere.sdk.offer.TaskException;
 import com.mesosphere.sdk.offer.TaskUtils;
+import com.mesosphere.sdk.scheduler.plan.Plan;
 import com.mesosphere.sdk.scheduler.plan.PlanManager;
 import com.mesosphere.sdk.scheduler.plan.PlanUtils;
+import com.mesosphere.sdk.specification.ServiceSpec;
+import com.mesosphere.sdk.state.ConfigStore;
 import com.mesosphere.sdk.state.StateStore;
 import com.mesosphere.sdk.state.StateStoreUtils;
 import org.apache.mesos.Protos;
@@ -35,6 +39,7 @@ public class SuppressReviveManager {
     private final Object suppressReviveLock = new Object();
 
     private final Object stateLock = new Object();
+    private final ConfigStore<ServiceSpec> configStore;
     private AtomicReference<State> state = new AtomicReference<>(State.INITIAL);
 
     /**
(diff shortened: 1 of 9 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SuppressReviveManager.java
```diff
@@ -131,7 +131,14 @@ private void transitionState(State target) {
                     logger.error("Invalid state transition.  End state should never be INITIAL");
                     return;
                 case WAITING_FOR_OFFER:
-                    revive();
+                    switch (current) {
+                        case REVIVED:
+                            logger.debug("Already revived, no need to revive again and wait for offers.", target);
```
nickbp: missing param in template? (or does this work like this?)

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/SuppressReviveManagerTest.java
```diff
@@ -141,6 +149,21 @@ public Boolean call() throws Exception {
                 });
     }
 
+    private static void waitStateNeverOccurs(
+            SuppressReviveManager suppressReviveManager,
+            SuppressReviveManager.State state) {
+
+        Awaitility.await()
```
nickbp: if SuppressReviveManager entered WAITING_FOR_OFFER, i assume it wouldn't automatically go back to REVIVED, would it? if so then this 5s delay could be omitted...

quick way to check would be reverting the SuppressReviveManager change, and disabling this wait, and then seeing if the assert on L117 is enough to catch the issue

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SuppressReviveManager.java
benclarkwood: I think it just gets appended IIRC

gabrielhartmann: Unintentional regardless. I messed up.  I'll fix it.

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/SuppressReviveManagerTest.java
gabrielhartmann: Thanks.  I was going to write that the wait is needed because of async `TASK_FAILED` handling, but that's not true.  The EventBus in the test posts synchronously so we're good.

PR #1465, opened 2017-08-22
Graceful Shutdown Test Change

benclarkwood:
1. Make the test a sanity test
2. Fix up sdk_cmd import to new style.


elezar:
@benclarkwood I think these are failing the nightlies, and at least SOME of these changes may need to be backported to the @0.30.X branch.

PR #1473, opened 2017-08-23
[INFINITY-2286] Ensure tasks which should not be running are not recovered

gabrielhartmann:


gabrielhartmann:
Updating integration test but ready for review.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -585,11 +585,23 @@ protected void initialize(SchedulerDriver driver) throws InterruptedException {
...
                 .map(taskInfo -> taskInfo.getTaskId())
                 .collect(Collectors.toSet());
 
+        // Clear the TaskIDs from the TaskInfos so we drop all future TaskStatus Messages
+        Set<Protos.TaskInfo> cleanedTaskInfos = taskInfos.stream()
+                .map(taskInfo -> taskInfo.toBuilder())
+                .map(builder -> builder.setTaskId(Protos.TaskID.newBuilder().setValue("")).build())
+                .collect(Collectors.toSet());
+
+        cleanedTaskInfos.forEach(taskInfo -> stateStore.clearTask(taskInfo.getName()));
```
benclarkwood: This is just to drop the existing TaskStatuses?

gabrielhartmann: Yes.  Like it was never launched.

benclarkwood: One other Q: Can TaskKiller fail to send the task kill message and we leak a task?

benclarkwood: (I know this conforms to our standard, "Write it down, then do it" pattern)

frameworks/helloworld/tests/test_enable_disable.py
```diff
@@ -16,33 +16,35 @@ def configure_package(configure_security):
...
-            additional_options={ "service": { "spec_file": "examples/enable-disable.yml" } })
+            6,
+            additional_options={ "service": { "spec_file": "examples/enable-disable.yml" }, "hello": { "count": 3 } } )
 
         yield # let the test session execute
     finally:
         sdk_install.uninstall(PACKAGE_NAME)
 
 
+@pytest.mark.gabriel
```
nickbp: leftover here and below

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
nickbp: I think if we failed to kill it (e.g. scheduler restarted just before) then the running status would be restored in task reconciliation. So it wouldn't go completely missing but it may end up in the task still running (unless there's something that would garbage collect it at that point..)

gabrielhartmann: No task reconciliation would occur, because the TaskID is gone, so there's nothing to reconcile with.  The way out of this situation is to manually kill tasks.

1. The modifications to StateStore say this thing has never been launched.
1. The modifications which can be done in the plan to disable a task say don't deploy this task.

If you crash the scheduler in the middle we won't get around to killing tasks because lost the TaskID.  So you've have to manually kill it.  You could `pod restart` and that should work.

benclarkwood:
Yeah, that sounds right to me, @nickbp

PR #1474, opened 2017-08-23
testing: Allow to override service account name and secret name in STRICT security mode

mhrabovcin:
https://jira.mesosphere.com/browse/DCOS-18083

### DIFF
--- a/frameworks/helloworld/tests/test_tls.py
+++ b/frameworks/helloworld/tests/test_tls.py
@@ -37,16 +37,7 @@ DISCOVERY_TASK_PREFIX = 'discovery-prefix'
 
 
 @pytest.fixture(scope='module')
-def dcos_security_cli():
-    """
-    Installs the dcos enterprise cli.
-    """
-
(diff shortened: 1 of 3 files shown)


frameworks/helloworld/tests/test_tls.py
```diff
@@ -37,16 +37,7 @@
 
 
 @pytest.fixture(scope='module')
-def dcos_security_cli():
```
benclarkwood: @mhrabovcin why drop this?

benclarkwood:
Should this fix all the TLS tests in strict?

frameworks/helloworld/tests/test_tls.py
mhrabovcin: We're dropping this becuase `sdk_security.create_service_account` does the same thing so its a dupe.

mhrabovcin:
@benclarkwood Yes, I've tested both helloworld and kafka in strict security mode, both tests passed.

PR #1475, opened 2017-08-23
fetch rotated task logs on failure

falseperfection:
https://jira.mesosphere.com/browse/INFINITY-2249

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -64,20 +64,33 @@ def pytest_runtest_makereport(item, call):
     setattr(item, "rep_" + rep.when, rep)
 
 
-@pytest.fixture(autouse=True)
-def get_task_logs_on_failure(request):
-    """ Scheduler should be the only task running as root
-    """
-    yield
+def get_rotating_task_log_lines(task_id: str, task_file: str):
+    rotated_filenames = [task_file,]
+    rotated_filenames.extend(['{}.{}'.format(task_file, i) for i in range(1, 10)])
+    for filename in rotated_filenames:
+        lines = get_task_logs_for_id(task_id, filename)
+        if not lines:
+            return
+        yield lines
+
+
+def is_test_failure(request):
     for report in ('rep_setup', 'rep_call', 'rep_teardown'):
         if not hasattr(request.node, report):
             continue
         if not getattr(request.node, report).failed:
             continue
+        return True
+    return False
+
+
+@pytest.fixture(autouse=True)
+def get_task_logs_on_failure(request):
+    yield
+    if is_test_failure(request):
         for task_id in get_task_ids():
             for task_file in ('stderr', 'stdout'):
-                log_name = '{}_{}_{}.log'.format(request.node.name, task_id, task_file)
-                task_logs = get_task_logs_for_id(task_id, task_file=task_file)
-                if task_logs:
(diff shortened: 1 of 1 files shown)


falseperfection:
Agreed @mellenburg. I can put the `*.log => task_logs.tgz` line back in the TeamCity config to handle compression there. I don't see a way to compress each file individually, without knowing the filenames ahead of time.

benclarkwood:
The only log that will inflate much is the scheduler log (we don't do anything to the tasks such that they'd write a lot of logs)

I'd expect at most 20MB of logs for a failed test.

conftest.py
```diff
@@ -64,20 +64,33 @@ def pytest_runtest_makereport(item, call):
...
 
 
-@pytest.fixture(autouse=True)
-def get_task_logs_on_failure(request):
-    """ Scheduler should be the only task running as root
-    """
-    yield
+def get_rotating_task_log_lines(task_id: str, task_file: str):
+    rotated_filenames = [task_file,]
+    rotated_filenames.extend(['{}.{}'.format(task_file, i) for i in range(1, 10)])
```
elezar: Is there a reason that we limit this to `9`?  (We probably won't need more than this depending on the log rotation frequency though).

elezar: Just a thought, isn't this going to append the logs out of order? The file with no enumeration contains the most recent logs, but this is being returned first, followed by the oldest(?) logs?

conftest.py
```diff
@@ -64,20 +64,33 @@ def pytest_runtest_makereport(item, call):
...
+    yield
+    if is_test_failure(request):
         for task_id in get_task_ids():
             for task_file in ('stderr', 'stdout'):
-                log_name = '{}_{}_{}.log'.format(request.node.name, task_id, task_file)
-                task_logs = get_task_logs_for_id(task_id, task_file=task_file)
-                if task_logs:
+                for log_lines in get_rotating_task_log_lines(task_id, task_file):
+                    log_name = '{}_{}_{}.log'.format(request.node.name, task_id, task_file)
                     with open(log_name, 'w') as f:
```
elezar: Here we probably want to use `open(log_name, 'a')` instead. It is my understanding from: https://docs.python.org/3/library/functions.html#open that `open(log_name, 'w')` will clear the file before writing to it.

conftest.py
falseperfection: The lines aren't appended, each log file is downloaded as a separate file. `9` was just the number in the ticket. If we find we need more we can change it, or, perhaps, add a call to look up how many rotated log files already exist.

conftest.py
falseperfection: You're right that `'a'` would append. My intent was to open a separate file for each task log. So the names will look like: `<node.name>_<task_id>_stdout.log`, `<node.name>_<task_id>_stdout.1.log`, etc. Good catch!

benclarkwood:
@elezar Logrotate only leaves 9 rotate files.

conftest.py
elezar: @benclarkwood said that there are only 9 files, so this is fine.

elezar:
@benclarkwood @falseperfection can we delete this branch?

benclarkwood:
si

PR #1477, opened 2017-08-23
get additional mesos and cassandra info on failure

falseperfection:
Need to update TeamCity configuration to take `*_slaves.json` and `*_state.json` as artifacts.

### DIFF
--- a/testing/sdk_utils.py
+++ b/testing/sdk_utils.py
@@ -43,6 +43,20 @@ def dcos_version_less_than(version):
     return shakedown.dcos_version_less_than("1.10")
 
 
+def is_test_failure(pytest_request):
+    '''Determine if the test run failed using the request object from pytest.
+    The reports being evaluated are set in conftest.py:pytest_runtest_makereport()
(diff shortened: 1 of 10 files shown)


benclarkwood:
@falseperfection do you mean "cluster" in the ticket title?

falseperfection:
No, I meant cassandra. The ticket has these commands in it:
- `dcos cassandra plan list`
- `dcos cassandra plan show <plan-name>`

testing/sdk_utils.py
```diff
@@ -37,11 +37,20 @@ def get_zk_path(service_name):
     return "test__integration__" + service_name
 
 
+def is_test_failure(request):
```
mellenburg: This feels kinda awkward here, but I get why you need it here. Could you please add a doc-string explaining that this is for ingesting pytest plugin requests, not the beloved requests lib used throughout this repo

benclarkwood:
There were test failures for C*, but I don't see mesos state?

falseperfection:
Sure enough. I need to add those files as artifacts (they're .json not .log). Also, it looks like I'm not saving state per-test. I'll update both things.

PR #1481, opened 2017-08-24
Use Mesos version 1.4.0-rc1

gabrielhartmann:


gabrielhartmann:
Confirmed Kafka with this `dcos-commons` build and I've been running DSE with rc1 for two days.

PR #1484, opened 2017-08-24
[INFINITY-2293] Document how to use binary secrets in SDK-based services

mbdcos:


docs/pages/operations-guide.md
```diff
@@ -345,6 +340,23 @@ The path of a secret defines which service IDs can have access to it. You can th
 
 **Note:** Absolute paths (paths with a leading slash) to secrets are not supported. The file path for a secret must be relative to the sandbox.
 
+### Binary Secrets
+
+When you need to store binary files into DC/OS secrets store, for example a Kerberos keytab file, your file needs to be Base64 encoded as specified in RFC 4648. You can use standard `base64` utility.
```
karya0: s/Base64 encoded/Base64-encoded/

docs/pages/operations-guide.md
```diff
@@ -345,6 +340,23 @@ The path of a secret defines which service IDs can have access to it. You can th
...
+
+``` 
+$  base64 -i krb5.keytab -o kerb5.keytab.base64-encoded
```
karya0: `-o <filename>` is not available on Linux :(.

Also note that on Linux, `base64` inserts line-feeds in the encoded data and that is not RFC4648-compliant. The fix there is to use `-w 0` to disable line-wrapping. (I am not sure what happens on MAC.) We should explicitly mention the same here.

mbdcos: Default is `0` in BSD (no line-wrapping).   Parameter is `-b` not -w.  `-o` is available in BSD (Mac).  Linux is stdout only.  Interestingly  BSD, Linux, FreeBSD all different. That is why I specifically mentioned RFC 4648 :)

docs/pages/operations-guide.md
```diff
@@ -345,6 +340,32 @@ The path of a secret defines which service IDs can have access to it. You can th
...
+``` 
+$  base64 -i krb5.keytab -o kerb5.keytab.base64-encoded 
+```
+
+`base64` command line utility in Linux inserts line-feeds in the encoded data by default. Disable line-wrapping via  `-w 0` argument.  Here is a sample base64 command in Linux.
+
+``` 
+$  base64 -w 0 -i krb5.keytab > kerb5.keytab.base64-encoded 
+```
+
```
karya0: Too many new lines?

PR #1486, opened 2017-08-24
Fix release script for .zip universes

nickbp:
zipfile expects a seekable file-like object


nickbp:
Merging as this isn't exercised by CI

PR #1488, opened 2017-08-24
[INFINITY-2309] Fix handling of DeploymentStep status changes.

benclarkwood:
We were previously broadly handling DeploymentStep status changes. That broad handling left open time periods where a STARTING task would never get reset to PENDING on failure / completion.

This corrects that by explicitly handling each possible combination of task statuses in a DeploymentStep.


nickbp:
retriggering helloworld-19, appears to have been a flake with a pod replace

elezar:
This is a pretty reliable flake as that test has failed for every build on this branch. Our assumption is currently that there is a dependence between the said test, and the restart test that was previously disabled for 1.9.

benclarkwood:
I've inserted this hack: https://github.com/mesosphere/dcos-commons/pull/1488/files#diff-16ea3ae0dc6e52aaee4bfd980d5693fcR188

To prove out what was happening. With this change, tests pass locally for me.

elezar:
@benclarkwood seems as if this fixed the helloworld tests, but we have now broken almost everything else. This indicates that you are right in you code comment that this is probably not the best place to implement the state transition check.

benclarkwood:
Neat!

benclarkwood:
Actually, CI is just generally broken. Will re-run once #1538 merges.

elezar:
Seems as if #1538 was merged. I have merged master again to kick the CI.

benclarkwood:
Putting this back to WIP. There's a change I need to make to the logic of my fix.

benclarkwood:
This is ready to be reviewed now.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
```diff
@@ -181,17 +181,17 @@ private Status getStatus(Map<Protos.TaskID, TaskStatusPair> tasks) {
...
-                // Keep and log current status
-                return super.getStatus();
-            }
+
+            statuses.add(status);
         }
 
-        return Status.COMPLETE;
+
+        return Collections.min(statuses);
```
elezar: This implies that the statuses have a concept of ordering -- I assume that this is defined by the order of declaration of the enum class. Is there a particular ordering that we depend on for this to work correctly here? For example, if one task is `COMPLETE` and another is `IN_PROGRESS` this here will return `COMPLETE`? It would be good to get some unit tests in place that capture the desired behaviour for the `Enum` then.

benclarkwood: DeploymentStep will only ever return ERROR through COMPLETE. IN_PROGRESS is reserved for steps and phases.

Yes, there should be a unit test for this. It _is_ exercised by different unit tests, but I will make the `getStatus` method package-private so I can hit it in a unit test and validate it fully..

benclarkwood:
@elezar feedback addressed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
```diff
@@ -176,25 +177,26 @@ private void setTaskStatus(Protos.TaskID taskID, Status status) {
...
-            if (!status.equals(Status.COMPLETE)) {
-                // Keep and log current status
-                return super.getStatus();
-            }
+
+            statuses.add(status);
         }
 
-        return Status.COMPLETE;
+        return statuses.stream().min(Comparator.comparing(Status::getOrder)).get();
```
gabrielhartmann: I think this behavior which we spoke about offline is fine, but should have significant comment with examples etc.

gabrielhartmann: Also, there's a change in behavior where the status of super is never consulted.  The status of super could be lower in order.  Should that ever override the Step's behavior?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Status.java
```diff
@@ -63,7 +63,20 @@
      *
      * This value is only returned and never set to a variable.
      */
-    IN_PROGRESS;
+    IN_PROGRESS(7);
```
nickbp: no 6?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Status.java
```diff
@@ -63,7 +63,20 @@
...
+
+    /**
+     * Note, this field is currently ONLY consumed by
```
nickbp: The suspense is killing me

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Status.java
```diff
@@ -63,7 +63,20 @@
...
+     */
+    private final int order;
+
+    Status(final int order) {
```
nickbp: What sets this value?

benclarkwood:
@nickbp convinced me to change this to a _very_ explicit handling of Statuses. It is a good change as it relies less on magic Enumery. Thanks!

@gabrielhartmann @elezar please review.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
benclarkwood: super.getStatus() is ultimately just the _present_ status of the task.

The change I've made to not rely on the enum having a correct order does re-introduce the fallback.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
```diff
@@ -176,25 +177,48 @@ private void setTaskStatus(Protos.TaskID taskID, Status status) {
...
+
+            statuses.add(status);
+        }
+
+        // A DeploymentStep should have the "least" status of its consituent tasks.
+        // We can rely on the order of Status as returned by its getOrder method.
+        // For example:
+        // 1 PENDING, 2 STARTING => PENDING
+        // 2 STARTING, 1 COMPLETE => STARTING
+        // 3 COMPLETE => COMPLETE
```
gabrielhartmann: I could read this, but someone new to the code might not be able to quickly understand that those numbers mean counts of Tasks.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
```diff
@@ -176,25 +177,48 @@ private void setTaskStatus(Protos.TaskID taskID, Status status) {
...
+        } else if (statuses.contains(Status.STARTING)) {
+            return Status.STARTING;
+        } else if (statuses.contains(Status.COMPLETE) && statuses.size() == 1) {
+            return Status.COMPLETE;
         }
 
-        return Status.COMPLETE;
+        // If we don't explicitly handle the new status,
+        // we will simply return the previous status.
+        logger.warn("The minimum status of the set of task statuses, {}, is not explicitly handled. " +
```
nickbp: This be noisy in the case where `contains(COMPLETE) && statuses.size() != 1`. I assume this code shouldn't be reachable in practice?

Maybe add an explicit case for that above? Something like:
```
} else if (statuses.contains(Status.COMPLETE)) {
    if (statuses.size() == 1) {
        return Status.COMPLETE;
    } else {
        // ?
    }
}
```

benclarkwood: It should not be reachable in practice.

The else I would put is just what's outside the the if/else if and then PMD will yell at me that I can collapse ifs.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
elezar: Mentioning the `getOrder` method is no longer applicable, right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
```diff
@@ -176,25 +177,48 @@ private void setTaskStatus(Protos.TaskID taskID, Status status) {
...
+        if (statuses.contains(Status.ERROR)) {
+            return Status.ERROR;
+        } else if (statuses.contains(Status.PENDING)) {
+            return Status.PENDING;
+        } else if (statuses.contains(Status.PREPARED)) {
+            return Status.PREPARED;
+        } else if (statuses.contains(Status.STARTING)) {
+            return Status.STARTING;
+        } else if (statuses.contains(Status.COMPLETE) && statuses.size() == 1) {
+            return Status.COMPLETE;
```
elezar: OK. Was a bit confused, but then I realised this is equivalent to **ALL** tasks being `COMPLETE`. We could use a `Map<Status, Integer>` instead of a set and count the number of tasks in each state, and then only return `COMPLETE` if the number of tasks in this state is the total number of tasks.

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/plan/DeploymentStepTest.java
```diff
@@ -45,6 +43,85 @@ public void beforeEach() {
     }
 
     @Test
+    public void testGetStatusReturnsMinimumState() {
+        // No tasks
+        Map<Protos.TaskID, DeploymentStep.TaskStatusPair> tasks = new HashMap<>();
+        Assert.assertEquals(Status.PENDING, getPendingStep().getStatus(tasks));
+
+        // 1 PREPARED and 1 ERROR and 1 COMPLETE tasks
```
elezar: Should these be separate tests? (nit).

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
benclarkwood: I could add a comment if you like. I prefer set over map. Less complex.

benclarkwood:
@elezar I updated the comments slightly. I was too lazy to split the test apart.

benclarkwood:
[Finish changes for min version fix](https://trello.com/c/4k2UxM8p/3-finish-changes-for-min-version-fix)

PR #1426, opened 2017-08-15
[CASSANDRA-598] Add TLS support in Cassandra

mhrabovcin:

PR #1497, opened 2017-08-25
[INFINITY-2306] Correct closing brace in Cassandra Moustache setting 

elezar:
While reviewing #1426 @mhrabovcin found an error in `cassandra.yaml` file where `CASSANDRA_COMMITLOG_TOTAL_SPACE_IN_MB` was not being correctly closed and thus not applied to `commitlog_total_space_in_mb`.

For the record:
```sh
grep -R -E ":\s+\{\{\w+[^\}]$" .
```
shows the missing braces (and no others), but is not an exhaustive set.

An open question is how to test these kinds of changes? Any thoughts?


benclarkwood:
I think bootstrap should validate that something is valid YAML after it's done templating it.

PR #1498, opened 2017-08-25
[INFINITY-2308] Fix flaky tests - helloworld placement

paegun:
Changed the tests to assert what the feature states, max_per and group_by placement rules provide an at most guarantee, not an exact match guarantee for the cardinality per host.

### DIFF
--- a/frameworks/helloworld/tests/test_placement.py
+++ b/frameworks/helloworld/tests/test_placement.py
@@ -119,7 +119,7 @@ def test_max_per_hostname():
     }
 
     sdk_install.install(config.PACKAGE_NAME, num_private_agents * 5, additional_options=options)
-    ensure_count_per_agent(hello_count=2, world_count=3)
+    ensure_max_count_per_agent(hello_count=2, world_count=3)
 
 
 @pytest.mark.sanity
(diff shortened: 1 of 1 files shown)


frameworks/helloworld/tests/test_placement.py
```diff
@@ -141,7 +141,7 @@ def test_rr_by_hostname():
     }
 
     sdk_install.install(config.PACKAGE_NAME, num_private_agents * 4, additional_options=options)
-    ensure_count_per_agent(hello_count=2, world_count=2)
+    ensure_max_count_per_agent(hello_count=2, world_count=2)
```
elezar: Doesn't round-robin also imply some ordering in the tasks? That is to say that the order of the tasks (if this is deterministic) should not differ by less than the number of agents?

paegun: @elezar great point regarding round-robin (rr), but TBH I don't see (read: it is not inherently apparent from this test) how the configuration options specified, which are mostly a placement rule using GROUP_BY is supposed to guarantee a round-robin ordering/distribution.

elezar: Yes, it's not really clear. For the recored, I don't think this is a blocker for merging this. But we could merge this as is.

benclarkwood: Merged. Please follow up with any necessary JIRAs / PRs.

elezar:
I have created https://jira.mesosphere.com/browse/INFINITY-2323 to track the round-robin investigation.

elezar:
Can we delete this branch?

PR #1503, opened 2017-08-25
INFINITY-1975 Tweaks to Cassandra endpoints: rename node=>native-client, remove VIP, show thrift-client when enabled

nickbp:
- The cassandra VIP was previously only enabled because VIP presence == shown in `endpoints`. Now that the two are decoupled, the cassandra VIP should be removed.
- Renames the 'node' endpoint 'native-client'
- Enables display of a 'thrift-client' endpoint when thrift is enabled
- Refrains from reserving the thrift port when thrift isn't enabled


keithchambers:
Looks great in terms the endpoints exposed and the naming.

benclarkwood:
@keithchambers do you want this in GA?

keithchambers:
Yes in GA.

PR #1504, opened 2017-08-25
[INFINITY-2311][HDFS] Document how to replace a Journal Node

NimaVaziri:


sascala:
I made some small edits, otherwise, LGTM.

PR #1505, opened 2017-08-25
[INFINITY-2306] Correct closing brace in cassandra mustache (#1497)

benclarkwood:


benclarkwood:
@elezar, nope. Looking at the test history, the test has a success rate of 88%. Which bears investigation.

benclarkwood:
Especially given that it failed in both the PRs into this branch.

PR #1510, opened 2017-08-28
[INFINITY-2316] Check for a null pod and handle the case of null prereserved roles

elezar:
This PR addresses a NPE when calling `PreReservedRoleCannotChange` in the case where:
1. a pod is removed in the new configuration
2. the new configuration has no `PreReservedRole`.

I have added `null` checks and used `StringUtils` for the role comparison to better handle nulls. Unit tests have also been added.

@benclarkwood it may be required to backport this too.

**UPDATE:** Looking at the use of `ConfigValidator`, the following need to be checked:
- [x] `PodSpecsCannotChangeNetworkRegime` (The relevant null checks are in place)
- [x] `PodSpecsCannotShrink` (also added more unit tests)
- [x] `PreReservationCannotChange` (part of this PR)
- [x] `ServiceNameCannotContainDoubleUnderscores`
- [x] `TaskEnvCannotChange` (The coverage of for this could be improved quite a bit, but I don't want to address this in this PR)
- [x] `TaskVolumesCannotChange`
- [x] `TLSRequiresServiceAccount` (added some more unit tests)
- [x] `UserCannotChange`


benclarkwood:
I could go either way on backport. @gabrielhartmann do you think it's necessary?

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/PreReservationCannotChange.java
```diff
@@ -10,27 +11,36 @@
...
 
-        List<ConfigValidationError> errors = new ArrayList<>();
-        for (PodSpec oldPod : oldConfig.get().getPods()) {
-            PodSpec newPod = newPods.get(oldPod.getType());
-            if (!newPod.getPreReservedRole().equals(oldPod.getPreReservedRole())) {
+        final List<ConfigValidationError> errors = new ArrayList<>();
+        for (final PodSpec oldPod : oldConfig.get().getPods()) {
+            final PodSpec newPod = newPods.get(oldPod.getType());
+
+            if (newPod == null) {
```
gabrielhartmann: Is this check needed in other validators?

gabrielhartmann:
I spoke with Chris.  We should backport.

elezar:
@gabrielhartmann while looking into this, I did check some of the other and they all had not null checks. I will confirm tomorrow that this really is the case.

elezar:
@gabrielhartmann I have looked at all the implementations of `ConfigValidator`, and it seems as if the null condition that was present in the the original bug.

I have added some other tests and a minor fix for string equality to `TLSRequiresServiceAccount`. I can also revert these if you or @benclarkwood feel that this is out of scope for this PR.

One thing that I should add is that where I have added new tests, I have used `Hamcrest` matchers and `assertThat()` instead of `Assert.assertEquals()` as I do find the conditions generated using these are easier to read. This is something that we could discuss from a style perspective.

There are some other comments that I would like to make about some of the Validators, but will do so as comments on https://jira.mesosphere.com/browse/INFINITY-2319.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/PreReservationCannotChange.java
elezar: I have looked through the other validators, and they explicitly handle this case -- either by first checking if the map contains the key or checking for null on the returned value.

PR #1511, opened 2017-08-28
[INFINITY-2317] Pass version in check of minimum dcos version

elezar:
The `version` parameter should be passed to the `dcos_version_less_than()` function.

This would most likely affect the disabling of tests for 1.8.


elezar:
This is a duplicate of #1488. Closing this.

PR #1515, opened 2017-08-28
[INFINITY-23222] Fix bootstrap networking shim.

benclarkwood:
1. Fix logic in container IP lookup.
2. Enhance logging in container IP lookup and assignment.

### DIFF
--- a/sdk/bootstrap/ip.go
+++ b/sdk/bootstrap/ip.go
@@ -6,38 +6,42 @@ import (
 	"os"
 )
 
-// This helper function gives the current container's IP address, when run in a
-// given network, mnt and UTS namespace.  The function first checks the presence of
-// `LIBPROCESS_IP` if present it will return this as the container's IP address
-// else it will look at the `hostname` setup for the container to pick out the
-// IP address associated with the hostname.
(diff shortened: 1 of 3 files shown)


sdk/bootstrap/ip.go
```diff
@@ -6,38 +6,42 @@ import (
...
-			return
-		}
+var libprocessIP = "LIBPROCESS_IP"
+var mesosContainerIP = "MESOS_CONTAINER_IP"
+
+// ContainerIP This helper function gives the current container's IP address, when run in a
+// given network, mnt and UTS namespace.
+func ContainerIP() (net.IP, error) {
+	// First, try MESOS_CONTAINER_IP (DC/OS 1.10+, when running the default executor)
+	ip, err := validateIP(mesosContainerIP)
```
asridharan: Just a thought, might be easier to write:
envVar := []string{"MESOS_CONTAINER_IP", "LIBPROCESS_IP"}
for var in range(envVar) {
   ip, err := {
     //validate retrieval of IP based on the envVar.
   }()
}

benclarkwood: Not a bad idea. Didn't want to re-roll CI.

PR #1516, opened 2017-08-29
Graceful Shutdown Test Change (#1465) - backport

benclarkwood:
* Fix up import, make test a sanity test

* CLI call changed

* Move setup so service is installed with correct kill policy


elezar:
Could this be the fix for the problems we now se in 1.9 after #1488?

benclarkwood:
I'm not sure what you mean. This is the backport. This was already applied to `master`.

elezar:
My bad. Was just thinking aloud.

PR #1517, opened 2017-08-29
[INFINITY-23222] Fix bootstrap networking shim - backport

benclarkwood:
1. Fix logic in container IP lookup. (#1515)
2. Enhance logging in container IP lookup and assignment.

### DIFF
--- a/sdk/bootstrap/ip.go
+++ b/sdk/bootstrap/ip.go
@@ -6,38 +6,42 @@ import (
 	"os"
 )
 
-// This helper function gives the current container's IP address, when run in a
-// given network, mnt and UTS namespace.  The function first checks the presence of
-// `LIBPROCESS_IP` if present it will return this as the container's IP address
-// else it will look at the `hostname` setup for the container to pick out the
-// IP address associated with the hostname.
-func ContainerIP() (containerIP net.IP, err error) {
-	ip := os.Getenv("LIBPROCESS_IP")
-	if ip != "" {
-		containerIP = net.ParseIP(ip)
-		if containerIP == nil {
-			err = fmt.Errorf("Invalid LIBPROCESS_IP found: %s", ip)
-			return
-		}
+var libprocessIP = "LIBPROCESS_IP"
+var mesosContainerIP = "MESOS_CONTAINER_IP"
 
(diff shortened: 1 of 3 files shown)


elezar:
This is the backport of #1515, right, so this isn't the place to make style comments.

sdk/bootstrap/ip.go
```diff
@@ -6,38 +6,42 @@ import (
...
+	validIP := net.ParseIP(ip)
+	if validIP == nil {
+		return nil, fmt.Errorf("Invalid %s found: %s", envvar, ip)
 	}
 
-	ipAddr, err := net.ResolveIPAddr("ip", hostName)
-	if err != nil {
-		err = fmt.Errorf("Unable to resolve hostname(%s): %s", hostName, err)
-		return
+	if validIP.String() == "0.0.0.0" || validIP.String() == "::" {
```
elezar: This check here implies support for IPv6 addresses, but the tests cases below don't exercise this.

sdk/bootstrap/main.go
```diff
@@ -336,52 +336,43 @@ func isFile(path string) (bool, error) {
...
 
-	pod_ip, err := GetLocalIP()
+	log.Printf("Looking up container IP address...")
+	containerIP, err := getContainerIPAddress()
 	if err != nil {
-		log.Fatalf("Cannot find the container's IP address: ", err)
+		log.Fatalf("Cannot look up container's IP address: %s", err)
+	}
+	log.Printf("Container IP address was found to be: %s", containerIP)
+	// Set both LIBPROCESS_IP and MESOS_CONTAINER_IP so that SDK developers
```
elezar: Thise could also be done in a loop like @asridharan suggested in #1515.

sdk/bootstrap/ip.go
benclarkwood: That is true. I was lazy.

sdk/bootstrap/main.go
benclarkwood: I personally don't like the loop style, as someone can easily swap the order of an array.

benclarkwood: (it also IMO sacrifices readability for saving a handful of lines)

PR #1521, opened 2017-08-29
[INFINITY-2301] Verify service-specific metrics are being reported

NimaVaziri:
This checks each service for reporting a set of specific metrics, the logic being that if this set is reported by the service then the greater set of all metrics is being properly reported as well.

Some autopep8 cosmetic edits are also in this change set.

### DIFF
--- a/frameworks/cassandra/tests/config.py
+++ b/frameworks/cassandra/tests/config.py
@@ -15,8 +15,15 @@ SERVICE_NAME = 'cassandra'
 DEFAULT_TASK_COUNT = 3
 DEFAULT_CASSANDRA_TIMEOUT = 600
 
-DEFAULT_NODE_ADDRESS = os.getenv('CASSANDRA_NODE_ADDRESS', sdk_hosts.autoip_host(SERVICE_NAME, 'node-0-server'))
+DEFAULT_NODE_ADDRESS = os.getenv(
+    'CASSANDRA_NODE_ADDRESS', sdk_hosts.autoip_host(SERVICE_NAME, 'node-0-server'))
 DEFAULT_NODE_PORT = os.getenv('CASSANDRA_NODE_PORT', '9042')
+
+EXPECTED_METRICS = [
+    "org.apache.cassandra.metrics.Table.CoordinatorReadLatency.system.hints.p999",
+    "org.apache.cassandra.metrics.Table.CompressionRatio.system_schema.indexes",
+    "org.apache.cassandra.metrics.ThreadPools.ActiveTasks.internal.MemtableReclaimMemory"
+]
 
 
 def _get_test_job(name, cmd, restart_policy='ON_FAILURE'):
@@ -25,11 +32,11 @@ def _get_test_job(name, cmd, restart_policy='ON_FAILURE'):
         'id': 'test.cassandra.' + name,
         'run': {
             'cmd': cmd,
-            'docker': { 'image': 'cassandra:3.0.13' },
+            'docker': {'image': 'cassandra:3.0.13'},
             'cpus': 1,
             'mem': 512,
             'user': 'nobody',
-            'restart': { 'policy': restart_policy }
+            'restart': {'policy': restart_policy}
         }
     }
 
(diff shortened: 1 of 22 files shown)


frameworks/cassandra/tests/config.py
```diff
@@ -105,7 +112,8 @@ def run_backup_and_restore(
     write_data_job = get_write_data_job(node_address=job_node_address)
     verify_data_job = get_verify_data_job(node_address=job_node_address)
     delete_data_job = get_delete_data_job(node_address=job_node_address)
-    verify_deletion_job = get_verify_deletion_job(node_address=job_node_address)
+    verify_deletion_job = get_verify_deletion_job(
```
nickbp: Looks like your col width is set to 80 and creating a lot of noise in the PR. How about 100?

frameworks/kafka/tests/config.py
```diff
@@ -9,3 +9,8 @@
 DEFAULT_TASK_NAME = 'broker'
 DEFAULT_KAFKA_TIMEOUT = 10 * 60
 DEFAULT_TOPIC_NAME = 'topic1'
+EXPECTED_METRICS = [
```
nickbp: given that this is only used in test_sanity, may as well just put it there and reduce scope

testing/sdk_metrics.py
```diff
@@ -1,6 +1,11 @@
 '''
 Utilities relating to verifying the metrics functionality as reported
 by the DC/OS metrics component.
+
+************************************************************************
+FOR THE TIME BEING WHATEVER MODIFICATIONS ARE APPLIED TO THIS FILE
```
nickbp: This is the case with all of `testing/` though? Is there anything special with this case?

testing/sdk_metrics.py
```diff
@@ -53,15 +48,33 @@ def get_metrics(package_name, service_name, task_name):
...
         return []
 
-    container_id = task_info["status"]["containerStatus"]["containerId"]["value"]
+    task_container_id = task_info["status"]["containerStatus"]["containerId"]["value"]
+
+    # Not related to functionality but consuming this
+    # endpoint to verify downstream integrity
+    containers_url = "{}/system/v1/agent/{}/metrics/v0/containers".format(
+        shakedown.dcos_url(), agent_id)
+    containers_response = sdk_cmd.request("GET", containers_url, retry=False)
```
nickbp: looks like `containers_response` is only used for debugging missing container ids. maybe move this retrieval into the failure case at L77? in other words only get/print the list of containers on the agent if the one we expected to be there was missing

testing/sdk_metrics.py
```diff
@@ -70,11 +83,61 @@ def get_metrics(package_name, service_name, task_name):
     raise Exception("No metrics found")
 
 
-def wait_for_any_metrics(package_name, service_name, task_name, timeout):
-    def metrics_exist():
-        log.info("verifying metrics exist for {}".format(service_name))
-        service_metrics = get_metrics(package_name, service_name, task_name)
-        # there are 2 generic metrics that are always emitted
-        return len(service_metrics) > 2
+def extract_metric_names(service_name, service_metrics):
```
nickbp: Rather than putting per-service cases here, how about having the per-service tests define a callback function that they pass in to `wait_for_service_metrics`? Maybe a system like:
- `wait_for_service_metrics`'s `expected_metrics` list parameter could be replaced with a list of callbacks (one per expected metric).
- `wait_for_service_metrics` would then pass the list of names it gets to each of the callbacks, waiting for each of them to return true _at least once_.
- The callbacks themselves would internally do whatever service-specific mangling is needed on a given metric name

Or alternately a simpler solution could be:
- The `expected_metrics` parameter could be a list of regexes rather than plain strings

testing/sdk_metrics.py
NimaVaziri: Wouldn't we only know that the expected container is missing from the list by comparing against the containers list, thereby making the retrieval necessary anyway?

nickbp: I'd assume the flow would look like:
1. test does retrieval against expected container id
2. retrieval fails with 404 or similar error
3. test returns failure: retrieves list of known container ids to be included in failure message

This assumes that the metrics API explicitly returns an error in step 2. However, if that's wrong and dcos-metrics returns e.g. an empty list, then we wouldn't know the difference between "container id isn't known" and "container with that id hasn't started sending metrics yet". In that case then the current behavior would make more sense. Any idea which is the case?

NimaVaziri: Ah, yeah it's expected that the metrics API return an error if it can't handle the given request, in which case we don't distinguish why there's problem and we do a retry from one layer higher. 

In the case the request succeeds but the list is empty, we still fail and do a retry as we expect some container id to be reported.

testing/sdk_metrics.py
NimaVaziri: Included this in all of `/testing/`

PR #1522, opened 2017-08-30
sleep_seconds=10, ignore_exceptions=False

falseperfection:
INFINITY-2324 - Add ignore_exceptions=False to shakedown.wait_for() calls

`shakedown.wait_for()` is hiding exceptions. All we see in the logs is a loop followed by a timeout, but not enough to diagnose the underlying issues. (Also increased sleep_seconds while touching all those lines to help with all the noise in the logs. Could probably do a much longer sleep...)

### DIFF
--- a/frameworks/elastic/tests/test_overlay.py
+++ b/frameworks/elastic/tests/test_overlay.py
@@ -55,7 +55,8 @@ def test_indexing(default_populated_index):
         observed_name = doc["_source"]["name"]
         return observed_name == "Loren"
 
-    return shakedown.wait_for(fun, timeout_seconds=config.DEFAULT_ELASTIC_TIMEOUT)
+    return shakedown.wait_for(fun, timeout_seconds=config.DEFAULT_ELASTIC_TIMEOUT,
+        sleep_seconds=10, ignore_exceptions=False)
 
 
 @pytest.mark.sanity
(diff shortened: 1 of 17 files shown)


elezar:
Looking at the changes, it seems as if all of the calls to `wait_for` have been modified. Would it not be cleaner to wrap the call in a separate function (e.g. in `sdk_utils`) and specify these defaults there?

falseperfection:
Yeah, that's all this is. Since reading your comment I've been looking at how to best clean everything up. What we really should do is replace the `wait_for` calls with a decorator and get rid of all the `lambda` and `fn` calls. I'm going to do that so we can see how it looks.

frameworks/elastic/tests/test_overlay.py
```diff
@@ -55,7 +55,8 @@ def fun():
         observed_name = doc["_source"]["name"]
         return observed_name == "Loren"
 
-    return shakedown.wait_for(fun, timeout_seconds=config.DEFAULT_ELASTIC_TIMEOUT)
+    return shakedown.wait_for(fun, timeout_seconds=config.DEFAULT_ELASTIC_TIMEOUT,
```
mellenburg: This test actually does nothing at the moment.... It returns True or False...  which means that failures are a success.

falseperfection:
Closing. This work is irrelevant given the `retrying` work @mellenburg is doing in #1534.

PR #1524, opened 2017-08-30
Default to attached cluster in test.sh

elezar:
This PR removes the need for `CLUSTER_URL` to be set when running `test.sh` if `dcos` has already been attached to a cluster. If if `dcos` does not exist, then this should still raise an error indicating that the environment variable should be set.

I should add that this is more in keeping with my own way of working, and if there are objections, then I don't mind pulling this back.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -55,7 +55,7 @@ fi
 
 if [ -z "$CLUSTER_URL" ]; then
     echo "CLUSTER_URL not set. Trying to determine attached cluster"
-    CLUSTER_URL=$(dcos cluster list --attached --json | grep "url" | sed -e 's/.*"url": "\(.*\)".*/\1/')
+    CLUSTER_URL=$(dcos cluster list --attached --json | grep "url" | sed -e 's/.*"url": "\(http.*\)".*/\1/')
     if [ -z "$CLUSTER_URL" ]; then
         echo "Cluster not found. Create and configure one then set \$CLUSTER_URL."
         exit 1
--- a/test.sh
+++ b/test.sh
@@ -27,7 +27,8 @@ function usage()
     echo "-k passed to pytest directly [default NONE]"
     echo "-p PATH to cluster SSH key [default ${ssh_path}]"
     echo "-s run in strict mode (sets \$SECURITY=\"strict\")"
-    echo "Cluster must be created and \$CLUSTER_URL set"
+    echo "Cluster must be created and \$CLUSTER_URL set."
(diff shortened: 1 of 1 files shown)


benclarkwood:
Makes sense to me! One note: is the output of this a http or https url? Strict mode requires the latter.

elezar:
The output is the url that dcos is connected to. I don't assume either `http` or `https`, so both should work and the strict mode check is unaffected.

falseperfection:
I want to make sure I understand (it's helpful to learn how folks use all these tools). You have `dcos` configured locally for your regular use. This will pull out `CLUSTER_URL` from that local config and pass it into the test runner. I think that makes sense.

Does that mean you have very long-running clusters that you re-test on repeatedly? Right now the test teardown functionality is only partially implemented. (Eg, the stub universes are removed, but accounts and permissions are not.) So as you re-use a cluster it's not in a completely clean state. Is that a concern?

Also, if you wanted to set the security mode based on your configured cluster:

```
token=$(dcos config show core.dcos_acs_token)
SECURITY=$(curl "$CLUSTER_URL/dcos-metadata/bootstrap-config.json" --silent --insecure --header "Authorization: token=$token" | jq -r .security)
```

elezar:
@falseperfection my workflow is generally as follows:
1. Start a CCM cluster
2. Connect to the cluster using `dcos cluster setup`
3. Start `test.sh` to check if I can reproduce a failing test in one of the test suites.

I agree that this is not ideal in terms of reproducing what gets done in TC, and as such it may be better to drop this PR and I can just define a bash alias for getting the current cluster. Your thoughts?

I should note, that I also sometimes _abuse_ `test.sh` by changing the command to run in the container to just `bash` to give me a quick docker-based dev environment.

With regards to:
> Also, if you wanted to set the security mode based on your configured cluster:

That's great, thanks! If this PR is to continue, do you think that should be added as part of this PR.

falseperfection:
@elezar Thanks for the insight. My workflow is mostly manual: I don't use `test.sh`, though I'm not debugging actual framework code either.

I'll let the decision on how much to add/remove in `test.sh` stay with @benclarkwood and your team. I don't have a strong opinion. In my mind, the ideal situation is to have a single set of scripts that controlled test execution locally and on CI -- so we don't have to maintain parallel implementations of the same thing -- but my primary focus is CI.

elezar:
Thinking about this again, I'm going to close this PR and implement this is some dotfiles. Thanks for the review efforts.

PR #1531, opened 2017-08-30
Remove unused tools

mellenburg:
There are no longer downstream projects which check out these tools and rely on them. Projects that previously did this have either pinned local copies of these utilities or use the new test running idiom, which is:
- Grab dev docker container or create a virtual env. See test_requirements.txt
- Create a cluster with dcos-launch or provide a URL from one generated elsewhere
- Call pytest


mellenburg:
@benclarkwood There are no longer downstream users that I know of and we are not using this code. Can you verify this?  If so, best to remove it if its not used and not maintained.

tools/configure_test_cluster.py
```diff
@@ -1,170 +0,0 @@
...
-            self.dcos_url = 'http://%s' % dns_address
-        else:
-            self.dcos_url = 'https://%s' % dns_address
-        self._install_cli()
-
-
-    def create_mount_volumes(self):
-        fmt = 'Enabling mount volumes for cluster {} (stack id {})'
-        logger.info(fmt.format(self.dns_address, self.aws_stack_id))
-        import enable_mount_volumes
```
mellenburg: @gabrielhartmann were you referring to this? 

This just calls main of this script: https://github.com/mesosphere/dcos-commons/blob/master/tools/enable_mount_volumes.py

And I did most definitely not remove that script in this change.

mellenburg:
So, @gabrielhartmann if you want to make CI tests or modify a test script to include this capability, it should be quite easy. The AWS cluster ID is on hand in the launch config so we just pass that to this script directly

gabrielhartmann:
@mellenburg: As long as I can still add MOUNT volumes to a cluster via a command I can run from a shell, I'm happy.  I think you're saying that's still possible.

mellenburg:
@gabrielhartmann Correct, you still can. However, I do not know of any automated testing we have for that. If you point me at test case(s), I'd gladly setup at least a nightly test to verify it

elezar:
@benclarkwood @mellenburg would it make sense to backport this to 0.30.X?

I have created #1616 to do this if this is desired.

PR #1533, opened 2017-08-31
Added example for placement constraint in config.

joerg84:


triclambert:
@joerg84 can you back port this to the `0.30.X` branch?

PR #1534, opened 2017-08-31
Refactor shakedown.wait_for calls into retrying

mellenburg:
`shakedown.wait_for`:
retries all exceptions by default (and suppresses exceptions)
retries for 2 minutes by default
retries with 1 second polling by default
evaluates the function to only return if predicate() by default. This quite confusing as there is no strong typing so this can be True/False, empty string, empty list, empty set, etc
prints spinning...

`retrying.retry`:
retries all exceptions by default (does not suppress unless you implement your own suppression)
retries with infinite timeout by default
retries with 0 second polling by default
does not retry on ANY return value by default. If you want retrying.retry to consider the result, then you must supply retry_on_result

Rather than trying to adjust the shakedown polling function to log more useful information, its more direct if individual tests and invocations of a retry loop can define their own logging or print statements

### DIFF
--- a/frameworks/elastic/tests/test_overlay.py
+++ b/frameworks/elastic/tests/test_overlay.py
@@ -1,4 +1,5 @@
 import pytest
+import retrying
 import sdk_install
 import sdk_networks
 import sdk_tasks
@@ -48,16 +49,16 @@ def test_service_health():
 @pytest.mark.sanity
 @pytest.mark.overlay
 @sdk_utils.dcos_1_9_or_higher
+@retrying.retry(
+    wait_fixed=10000,
+    stop_max_delay=config.DEFAULT_ELASTIC_TIMEOUT*1000)
 def test_indexing(default_populated_index):
-    def fun():
-        indices_stats = config.get_elasticsearch_indices_stats(config.DEFAULT_INDEX_NAME)
-        observed_count = indices_stats["_all"]["primaries"]["docs"]["count"]
-        assert observed_count == 1, "Indices has incorrect count: should be 1, got {}".format(observed_count)
-        doc = config.get_document(config.DEFAULT_INDEX_NAME, config.DEFAULT_INDEX_TYPE, 1)
-        observed_name = doc["_source"]["name"]
-        return observed_name == "Loren"
-
-    return shakedown.wait_for(fun, timeout_seconds=config.DEFAULT_ELASTIC_TIMEOUT)
+    indices_stats = config.get_elasticsearch_indices_stats(config.DEFAULT_INDEX_NAME)
+    observed_count = indices_stats["_all"]["primaries"]["docs"]["count"]
+    assert observed_count == 1, "Indices has incorrect count: should be 1, got {}".format(observed_count)
+    doc = config.get_document(config.DEFAULT_INDEX_NAME, config.DEFAULT_INDEX_TYPE, 1)
+    observed_name = doc["_source"]["name"]
+    assert observed_name == "Loren"
 
 
 @pytest.mark.sanity
--- a/frameworks/hdfs/tests/config.py
+++ b/frameworks/hdfs/tests/config.py
@@ -58,7 +58,8 @@ def get_active_name_node(service_name):
 
 @retrying.retry(
     wait_fixed=10000,
-    stop_max_delay=DEFAULT_HDFS_TIMEOUT*1000)
+    stop_max_delay=DEFAULT_HDFS_TIMEOUT*1000,
+    retry_on_results=lambda res: not res)
 def get_name_node_status(service_name, name_node):
     rc, output = run_hdfs_command(service_name, "./bin/hdfs haadmin -getServiceState {}".format(name_node))
     if not rc:
--- a/frameworks/hdfs/tests/config.py
+++ b/frameworks/hdfs/tests/config.py
@@ -59,7 +59,7 @@ def get_active_name_node(service_name):
 @retrying.retry(
     wait_fixed=10000,
     stop_max_delay=DEFAULT_HDFS_TIMEOUT*1000,
-    retry_on_results=lambda res: not res)
+    retry_on_result=lambda res: not res)
 def get_name_node_status(service_name, name_node):
     rc, output = run_hdfs_command(service_name, "./bin/hdfs haadmin -getServiceState {}".format(name_node))
     if not rc:
--- a/frameworks/hdfs/tests/config.py
+++ b/frameworks/hdfs/tests/config.py
@@ -1,8 +1,8 @@
+import retrying
 import shakedown
 
 import sdk_cmd
 import sdk_plan
-import sdk_utils
 import sdk_tasks
 
 PACKAGE_NAME = 'beta-hdfs'
@@ -56,15 +56,14 @@ def get_active_name_node(service_name):
     raise Exception("Failed to determine active name node")
 
 
+@retrying.retry(
+    wait_fixed=10000,
+    stop_max_delay=DEFAULT_HDFS_TIMEOUT*1000)
 def get_name_node_status(service_name, name_node):
-    def get_status():
-        rc, output = run_hdfs_command(service_name, "./bin/hdfs haadmin -getServiceState {}".format(name_node))
-        if not rc:
-            return rc
-
-        return output.strip()
-
-    return shakedown.wait_for(lambda: get_status(), timeout_seconds=DEFAULT_HDFS_TIMEOUT)
+    rc, output = run_hdfs_command(service_name, "./bin/hdfs haadmin -getServiceState {}".format(name_node))
+    if not rc:
+        return rc
+    return output.strip()
 
 
 def run_hdfs_command(service_name, command):
--- a/frameworks/hdfs/tests/test_overlay.py
+++ b/frameworks/hdfs/tests/test_overlay.py
@@ -1,5 +1,6 @@
 from xml.etree import ElementTree
 
+import retrying
 import pytest
 import sdk_cmd
 import sdk_hosts
@@ -105,13 +106,13 @@ def test_integrity_on_name_node_failure():
     config.check_healthy(service_name=config.SERVICE_NAME)
 
 
+@retrying.retry(
+    wait_fixed=10000,
+    stop_max_delay=config.DEFAULT_HDFS_TIMEOUT*1000)
 def wait_for_failover_to_complete(namenode):
     """
     Inspects the name node logs to make sure ZK signals a complete failover.
     The given namenode is the one to become active after the failover is complete.
     """
-    def failover_detection():
-        status = config.get_name_node_status(config.SERVICE_NAME, namenode)
-        return status == "active"
-
-    shakedown.wait_for(lambda: failover_detection(), timeout_seconds=config.DEFAULT_HDFS_TIMEOUT)
+    status = config.get_name_node_status(config.SERVICE_NAME, namenode)
+    return status == "active"
--- a/frameworks/helloworld/tests/test_sanity.py
+++ b/frameworks/helloworld/tests/test_sanity.py
@@ -3,6 +3,8 @@ import re
 
 import dcos.marathon
 import pytest
+import retrying
+
 import sdk_cmd
 import sdk_install
 import sdk_marathon
@@ -182,9 +184,20 @@ def test_pod_info():
     assert task['status']['state'] == 'TASK_RUNNING'
 
 
+@retrying.retry(
+    wait_fixed=10000,
+    stop_max_delay=30000)
+def wait_for_nonempty_properties():
+    """'suppressed' could be missing if the scheduler recently started,
+    loop for a bit just in case
+    """
+        jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.FOLDERED_SERVICE_NAME, 'state properties', json=True)
+    assert len(jsonobj) > 0
+
+
 @pytest.mark.sanity
 def test_state_properties_get():
-    foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
+    wait_for_nonempty_properties()
 
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, foldered_name, 'state properties', json=True)
     assert len(jsonobj) == 5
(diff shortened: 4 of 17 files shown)


frameworks/elastic/tests/test_overlay.py
```diff
@@ -48,16 +49,16 @@ def test_service_health():
...
-        observed_name = doc["_source"]["name"]
-        return observed_name == "Loren"
-
-    return shakedown.wait_for(fun, timeout_seconds=config.DEFAULT_ELASTIC_TIMEOUT)
+    indices_stats = config.get_elasticsearch_indices_stats(config.DEFAULT_INDEX_NAME)
+    observed_count = indices_stats["_all"]["primaries"]["docs"]["count"]
+    assert observed_count == 1, "Indices has incorrect count: should be 1, got {}".format(observed_count)
+    doc = config.get_document(config.DEFAULT_INDEX_NAME, config.DEFAULT_INDEX_TYPE, 1)
+    observed_name = doc["_source"]["name"]
+    assert observed_name == "Loren"
```
elezar: Should the function not return a non-falsey value for `retry` to stop?

frameworks/hdfs/tests/config.py
```diff
@@ -56,15 +56,14 @@ def get_active_name_node(service_name):
...
-        rc, output = run_hdfs_command(service_name, "./bin/hdfs haadmin -getServiceState {}".format(name_node))
-        if not rc:
-            return rc
-
-        return output.strip()
-
-    return shakedown.wait_for(lambda: get_status(), timeout_seconds=DEFAULT_HDFS_TIMEOUT)
+    rc, output = run_hdfs_command(service_name, "./bin/hdfs haadmin -getServiceState {}".format(name_node))
+    if not rc:
+        return rc
```
elezar: This is not clear that `rc` is returning a falsey value. Rather raise or return `None` or `False` explicitly.

frameworks/hdfs/tests/config.py
```diff
@@ -56,15 +56,14 @@ def get_active_name_node(service_name):
...
+    return output.strip()
```
elezar: Is a blank string successful output? Would retry not treat this as failure?

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -173,14 +175,19 @@ def test_pod_info():
...
 
+@retrying.retry(
+    wait_fixed=10000,
+    stop_max_delay=30000)
+def wait_for_nonempty_properties():
+    """'suppressed' could be missing if the scheduler recently started,
+    loop for a bit just in case
+    """
+    jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, FOLDERED_SERVICE_NAME, 'state properties', json=True)
+    assert len(jsonobj) > 0
```
elezar: Don't we need a return statement for `retry` to succeed?

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -196,6 +203,26 @@ def check_for_nonempty_properties():
...
+    wait_fixed=10000,
+    stop_max_delay=120000,
+    retry_on_result=lambda res: res is False,
+    retry_on_exception=lambda ex: False if "failed: 409 Conflict" in ex.args[0] else True)
+def wait_for_refresh_cache_fails_409conflict():
+    """caching disabled, refresh_cache should fail with a
+    409 error (eventually, once scheduler is up)
+    """
+    sdk_cmd.svc_cli(config.PACKAGE_NAME, FOLDERED_SERVICE_NAME, 'state refresh_cache')
+    return False
```
elezar: Is this not always going to retry? (I may be missing something with the lambdas)

frameworks/helloworld/tests/test_sidecar.py
```diff
@@ -54,34 +54,30 @@ def test_sidecar_parameterized():
...
 
-    def __call__(self):
-        cmd, expected_output = self.get_cmd_output_pair()
-        output = sdk_cmd.run_cli(cmd).strip()
-        logging.info("Checking for toxic output returned: %s", output)
+    output = sdk_cmd.run_cli(cmd).strip()
+    logging.info("Checking for toxic output returned: %s", output)
 
-        return output == expected_output
+    assert output == expected_output
```
elezar: There is no return statement. Does this not mean continued retries?

frameworks/helloworld/tests/test_taskcfg.py
```diff
@@ -31,21 +33,27 @@ def test_deploy():
 
     # we can get brief blips of TASK_RUNNING but they shouldnt last more than 2-3s:
     consecutive_task_running = 0
-    def fn():
+
+    @retrying.retry(
+        wait_fixed=1000,
+        stop_max_delay=wait_time*1000,
+        retry_on_exception=lambda ex: False,
+        retry_on_result=lambda res: res is False)
```
elezar: Does returning `False` mean we should `retry` in this case? Would removing the `return` statement not have the same effect?

frameworks/kafka/tests/test_sanity.py
```diff
@@ -341,6 +347,14 @@ def test_metrics():
 
 # --------- Suppressed -------------
 
+@retrying.retry(
+    wait_fixed=10000,
+    stop_max_delay=120000)
+def wait_for_response(suppressed_url):
+    response = dcos.http.get(suppressed_url)
+    response.raise_for_status()
+    assert response.text == "true"
```
elezar: There is no return statement. Does this not imply infinite retry?

frameworks/kafka/tests/test_sanity.py
```diff
@@ -349,9 +363,4 @@ def test_suppress():
     suppressed_url = urllib.parse.urljoin(
         dcos_url, 'service/{}/v1/state/properties/suppressed'.format(FOLDERED_SERVICE_NAME))
 
-    def fun():
-        response = dcos.http.get(suppressed_url)
-        response.raise_for_status()
-        return response.text == "true"
-
-    shakedown.wait_for(fun)
+    wait_for_response(suppressed_url)
```
elezar: An out of scope comment. The suppressed url does not exist in 1.8.

testing/sdk_jobs.py
```diff
@@ -68,14 +69,25 @@ def run_job(job_dict, timeout_seconds=600, raise_on_failure=True):
...
+    @retrying.retry(
+        wait_fixed=10000,
+        stop_max_delay=timeout_seconds*1000,
+        retry_on_exception=lambda ex: False,
+        retry_on_result=lambda res: not res)
     def wait_for_run_id():
         runs = json.loads(sdk_cmd.run_cli('job show runs {} --json'.format(job_name)))
         if len(runs) > 0:
             return runs[0]['id']
         return ''
```
elezar: Returning nothing should also trigger a retry.

testing/sdk_metrics.py
```diff
@@ -71,10 +72,13 @@ def get_metrics(package_name, service_name, task_name):
...
-    def metrics_exist():
+    @retrying.retry(
+        wait_fixed=5000,
+        stop_max_delay=timeout*1000)
+    def wait_for_metrics_to_exist():
         log.info("verifying metrics exist for {}".format(service_name))
         service_metrics = get_metrics(package_name, service_name, task_name)
         # there are 2 generic metrics that are always emitted
-        return len(service_metrics) > 2
+        assert len(service_metrics) > 2
```
elezar: Will this not retry infinitely?

testing/sdk_plan.py
```diff
@@ -85,6 +83,10 @@ def wait_for_plan_status(service_name, plan_name, status, timeout_seconds=TIMEOU
     else:
         statuses = status
 
+    @retrying.retry(
+        wait_fixed=5000,
+        stop_max_delay=timeout_seconds*1000,
+        retry_on_result=lambda res: res is False)
     def fn():
```
elezar: We could rename `fn` to make it clearer what we are waiting for.

testing/sdk_upgrade.py
```diff
@@ -196,5 +197,8 @@ def _add_last_repo(repo_name, repo_url, prev_version, default_repo_package_name)
     _wait_for_new_default_version(prev_version, default_repo_package_name)
 
 
+@retrying.retry(
+    wait_fixed=5000,
+    stop_max_delay=120000)
 def _wait_for_new_default_version(prev_version, default_repo_package_name):
-    shakedown.wait_for(lambda: _get_pkg_version(default_repo_package_name) != prev_version, noisy=True)
+    assert _get_pkg_version(default_repo_package_name) != prev_version
```
elezar: Infinite retry due to no return.

falseperfection:
Apart from typos and misc bugs, this looks very clean. Definitely worth the effort, IMO.  I **love** the inner functions going away as well as having control managed by the retrying lib (`retry_on_*` params).

I am curious if we lose any hidden behavior that `shakedown.wait_for` provides, and how to reproduce that behavior if it still makes sense. (I haven't yet looked closely enough to form an opinion.)

frameworks/kafka/tests/test_sanity.py
nickbp: Strictly speaking `suppressed` will exist if the build available on 1.8 is new enough. It's served by the scheduler.

frameworks/elastic/tests/test_overlay.py
mellenburg: This test actually did nothing. `fun()` can raise an assertion or return True/False. This test ends with a returns are just successful exits in pyest i.e. returning false in a test is NOT equivalent to failing a test.

Now, consider that shakedown.wait_for will retry all exceptions and runs until True is returned at which point no assertion is evaluated.

So, return value is meaningless. If we are just trying to cause a restart, use an assert because retrying.rety by default retries all exceptions

frameworks/hdfs/tests/config.py
mellenburg: Good catch! this is an inconsistency in the refactor. See above for fix

frameworks/hdfs/tests/config.py
mellenburg: shakedown.wait_for works this way where it is fine with empty variables or False which I think has created a ton of confusion in the code. Here, we definitely need to add `retry_on_result=lambda res: not res`

frameworks/hdfs/tests/test_overlay.py
```diff
@@ -105,13 +106,13 @@ def test_integrity_on_name_node_failure():
...
     Inspects the name node logs to make sure ZK signals a complete failover.
     The given namenode is the one to become active after the failover is complete.
     """
-    def failover_detection():
-        status = config.get_name_node_status(config.SERVICE_NAME, namenode)
-        return status == "active"
-
-    shakedown.wait_for(lambda: failover_detection(), timeout_seconds=config.DEFAULT_HDFS_TIMEOUT)
+    status = config.get_name_node_status(config.SERVICE_NAME, namenode)
+    return status == "active"
```
mellenburg: This should be an assert

frameworks/helloworld/tests/test_sanity.py
mellenburg: by default retrying retries all exceptions so using assert is the generally recommended way to trigger a retry unless you specify `retry_on_exception=lambda ex: False`

frameworks/helloworld/tests/test_sanity.py
mellenburg: The intention of the code here is a little confusing:
* check the state refresh_cache command
* if it raises an exception (ANY exception), then we go on and check if its a 409. If it is, we exit the loop
* if its not a 409 exception then we continue

so how do we do that here? 
* We say `retry_on_result=lambda res: res is False,` so that non-exception raising runs simply retry (same as before)
* We say `retry_on_exception=lambda ex: False if "failed: 409 Conflict" in ex.args[0] else True` so that we always retry exceptions that are not the exected one.

However, there is a failure here and that is that once the expected 409 is raised, retrying.retry will raise it as an exception instead of suppressing it. So, we should either do:
```
try:
    wait_for_refresh_cache_fails_409conflict()
except retrying.RetryError:
    pass
```
However, this will also consider a timeout to be valid, so I think I need to take another look at the refactor. Most likely I need to add that `try:except` inside the retrying like the original

frameworks/helloworld/tests/test_sidecar.py
mellenburg: all return statements are good in retrying.retry unless specific with a `retry_on_result` kwargs. assertions are the idiomatic way to trigger a retry in tests

frameworks/helloworld/tests/test_taskcfg.py
mellenburg: That is assuming that retrying.retry works like shakedown.wait_for. Be default, all returns are valid; you must specific that return False is a retry.

frameworks/kafka/tests/test_sanity.py
mellenburg: it stops retrying once the assertion is not raised. See other comments on explaining this API vs retrying.retry

testing/sdk_jobs.py
mellenburg: This is already covered in the code: ` retry_on_result=lambda res: not res`

testing/sdk_metrics.py
mellenburg: See other comments above explaining this API

testing/sdk_upgrade.py
mellenburg: nope. see above explanation on API

mellenburg:
@elezar Thanks for the feedback, it think it might help to go over the API of retrying vs shakedown.wait_for:

`shakedown.wait_for`:
* retries all exceptions by default (and suppresses exceptions)
* retries for 2 minutes by default
* retries with 1 second polling by default
* evaluates the function to only return `if predicate()` by default. This quite confusing as there is no strong typing so this can be True/False, empty string, empty list, empty set, etc
* prints `spinning...`

`retrying.retry`:
* retries all exceptions by default (does not suppress unless you implement your own suppression)
* retries with infinite timeout by default
* retries with 0 second polling by default
* does not retry on ANY return value by default. If you want retrying.retry to consider the result, then you must supply `retry_on_result`

So hopefully that explains why there are now asserts and fewer returns

frameworks/hdfs/tests/config.py
mellenburg: `rc` is an extremely confusing variable. Its not the return code, but rather its a boolean representing if the return code was 0 which behaves the opposite of how return codes would behave in this statement. not sure what this is actually supposed to be doing

mellenburg: https://github.com/dcos/shakedown/blob/master/shakedown/dcos/command.py#L72

mellenburg: Ok, so after talking to Nick, it seems the usage is correct, its just `rc` is very misleading and should be fixed in another PR

frameworks/kafka/tests/test_sanity.py
elezar: My bad. It was my understanding that this was the reason that #1398 was needed (including skipping the suppressed check in 1.8 -- where I checked and this call would return 404).

elezar:
@mellenburg thanks for the clarifications. I think many of my comments were due to misinterpreting the docs. What had me confused with regards to:
> * does not retry on ANY return value by default. If you want retrying.retry to consider the result, then you must supply `retry_on_result`

is that https://pypi.python.org/pypi/retrying states that the following will retry infinitely.
```python
@retry
def never_give_up_never_surrender():
    print "Retry forever ignoring Exceptions, don't wait between retries"
```
Note that there is only an implicit `None` returned.

mellenburg:
@elezar Woah, I am so surprised to see that the retrying documentation is flat-out wrong. I can't believe that all of those examples just IMPLY that an exception is raised. It really isnt hard to make a perfectly correct explicit example:
```
import retrying
  
@retrying.retry
def foo():
      print('foo')
  
@retrying.retry(wait_fixed=2000)
def bar():
      print('bar')
      raise Exception
  
foo()
bar()
```
returns:
```
michael@BlackMagik:~/dcos-launch$ python test_foo.py 
foo
bar
bar
bar
[... bar forever ...]
```

mellenburg:
This work is ready to go, just waiting for a go ahead instead of fixing merge conflicts everyday

gabrielhartmann:
@mellenburg: This is very old.  If still needed please re-open / rebase

PR #1537, opened 2017-08-31
Allow comma separated URIs in the URI list for pods

gabrielhartmann:
This change also causes YAML rendering to reject blank URIs.


sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/yaml/YamlToInternalMappersTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Test(expected = URISyntaxException.class)
+    public void testEmptyUriString() throws URISyntaxException {
+        YAMLToInternalMappers.getUris("");
+    }
+
+    @Test
+    public void testSeparateSingleUriString() throws URISyntaxException {
+        URI uri = YAMLToInternalMappers.getUris(firstUriStr).stream().findAny().get();
+        Assert.assertEquals(firstUriStr, uri.toString());
```
elezar: Since this is a new file, I also just want to ask what the view is on the Hamcrest matchers? We could write this as:
```java
@Test
    public void testSeparateSingleUriString() throws URISyntaxException {
        final Collection<URI> uris = YAMLToInternalMappers.getUris(firstUriStr);
        assertThat(uris, equalTo(ImmutableList.of(new URI(firstUriStr))));
    }
```

and one of the following as:
```java
    @Test
    public void testSeparateCommaWithSpace() throws URISyntaxException {
        final Collection<URI> uris = YAMLToInternalMappers.getUris(multipleSpacedUriStr);
        assertThat(uris, equalTo(ImmutableList.of(new URI(firstUriStr), new URI(secondUriStr))));
    }
```

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/yaml/YamlToInternalMappersTest.java
```diff
@@ -0,0 +1,44 @@
...
+ * This class tests {@link YAMLToInternalMappers}.
+ */
+public class YamlToInternalMappersTest {
+    private static final String firstUriStr = "https://github.com/mesosphere/dcos-commons/blob/master/README.md";
+    private static final String secondUriStr = "https://github.com/mesosphere/dcos-commons/blob/master/docs/generate.sh";
+    private static final String multipleSpacedUriStr = firstUriStr + ", " + secondUriStr;
+    private static final String multipleNoSpaceUriStr = firstUriStr + "," + secondUriStr;
+
+    @Test(expected = URISyntaxException.class)
+    public void testEmptyUriString() throws URISyntaxException {
```
elezar: Probably overkill for this kind of method, but we could also test the following cases:
- `" "`
- `","`
- `firstString + ","`
- `"," + secondString`

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -294,7 +295,7 @@ private static PodSpec convertPod(
 
         Collection<URI> podUris = new ArrayList<>();
         for (String uriStr : rawPod.getUris()) {
-            podUris.add(new URI(uriStr));
+            podUris.addAll(getUris(uriStr));
```
elezar: If we pass the return of `rawPod.getUris()` to `getUris`, we could use a `flatMap` to handle them all as a single list:
```java
        Collection<String> allUris = rawPod.getUris().stream()
                .flatMap(uris -> Arrays.stream(uris.split(",")))
                .map(u -> u.trim())
                .filter(StringUtils::isNotBlank)
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -306,6 +307,27 @@ private static PodSpec convertPod(
         return builder.build();
     }
 
+    /**
+     * Converts a comma separated list of URIs to URI objects.
+     */
+    @VisibleForTesting
+    static Collection<URI> getUris(String uriStr) throws URISyntaxException {
+        // Separate comma separated list
+        List<String> uriStrs = Arrays.stream(uriStr.split(","))
```
elezar: Should we extract the `","` to a constant? Probably not since it's only used here.

frameworks/helloworld/src/main/dist/examples/uri.yml
```diff
@@ -35,7 +36,7 @@ pods:
           type: ROOT
           size: {{WORLD_DISK}}
     uris:
-      - "https://github.com/mesosphere/dcos-commons/blob/master/frameworks/helloworld/README.md"
+      - https://github.com/mesosphere/dcos-commons/blob/master/frameworks/helloworld/README.md
```
elezar: Out-of-scope: Would it make sense to also strip these kinds of characters from URIs when constructing our list, or is that already done in some way?

frameworks/helloworld/src/main/dist/examples/uri.yml
```diff
@@ -6,8 +6,9 @@ pods:
   hello:
     count: {{HELLO_COUNT}}
     uris:
-      - "https://github.com/mesosphere/dcos-commons/blob/master/README.md"
-      - "https://github.com/mesosphere/dcos-commons/blob/master/docs/generate.sh"
+      - https://github.com/mesosphere/dcos-commons/blob/master/README.md
```
elezar: Do we exercise these changes in a test of some sort?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
gabrielhartmann: Tried. Streaming method doesn't handle exceptions well.

frameworks/helloworld/src/main/dist/examples/uri.yml
gabrielhartmann: Yes. See ServiceSpecTest in hello-world 's unit tests.

frameworks/helloworld/src/main/dist/examples/uri.yml
gabrielhartmann: That's already done by the YAML reading library.

frameworks/helloworld/src/main/dist/examples/uri.yml
elezar: I ask because they have changed, and didn't see a corresponding change here (but I may have just missed it). OK. Found id. We basically just check that we're able to build the `DefaultScheduler` with the specification and don't check any specific values, right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
elezar: Yes, I also saw that w.r.t the URI constructor, but in this case, we're just getting a list of all non-blank URI strings. But this is definitely out of scope here, and was just me thinking aloud.

PR #1539, opened 2017-09-01
INFINITY-2207 Exercise universe packaging in per-service unit tests

nickbp:
- `ServiceSpecTest`s are now triggering a full rendering of `config.json`/`resource.json` onto `marathon.json.mustache`. The resulting env (including e.g. automatically generated `PORT_X` envvars to mimic real Marathon) is then passed to a `DefaultScheduler` object which renders that env against the service's `svc.yml`. In effect the full package rendering flow is now exercised in unit tests.
- Tests may provide a `Map<String,String>` of any `options.json` options to be exercised as well, but this feature isn't currently being used in any tests. Easy way to e.g. check that enabling an option doesn't trigger basic errors in `marathon.json` or `svc.yml`.
- In order to support rendering `svc.yml` from the unit tests, the path handling for config templates needed some reworking. As a result the previous `CONFIG_TEMPLATE_PATH` envvar hack has been removed in favor of expecting config templates to be in the same directory as `svc.yml` (or optionally in a custom directory provided during `DefaultScheduler` construction)


benclarkwood:
@nickbp is the loss of CONFIG_TEMPLATE_PATH a breaking change?

docs/pages/_layouts/gh-basic.html
```diff
@@ -91,9 +91,12 @@
     <span>Reference</span>
```
benclarkwood: What does this change?

frameworks/cassandra/src/test/java/com/mesosphere/sdk/cassandra/scheduler/CassandraRecoveryPlanOverriderTest.java
```diff
@@ -11,64 +11,45 @@
...
+import java.util.Collections;
+import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 
 /**
  * This class tests Cassandra's custom replacement of nodes.
  */
-public class CassandraRecoveryPlanOverriderTest extends BaseServiceSpecTest {
+public class CassandraRecoveryPlanOverriderTest {
```
benclarkwood: If we kept a base class, it seems like we could lose all of the constructor to a super, no?

nickbp:
Individual services were manually setting `CONFIG_TEMPLATE_PATH` in their `marathon.json` and then using it in their `svc.yml`. As of this PR it was effectively just removed from the current services.

However there _is_ a breaking change with the sdk-level path handling of config templates: Relative paths for config templates in `svc.yml` used to be relative to `pwd` as of when the scheduler process was launched. Now they default to being relative to the location of `svc.yml`.

frameworks/cassandra/src/test/java/com/mesosphere/sdk/cassandra/scheduler/CassandraRecoveryPlanOverriderTest.java
benclarkwood: Ah, I see that the ServiceSpecTests all are very simple.

I just worry about having this much setup for other tests.

benclarkwood:
We're still publishing as 0.20.X: https://github.com/mesosphere/dcos-commons/blob/master/build.gradle#L48

Can we up that to... 0.40 I guess? Since 0.30.X is stable.

benclarkwood:
(I'd like that change to come in this PR since it's a breaking change)

nickbp:
How about `0.31.0-SNAPSHOT`?

docs/pages/_layouts/gh-basic.html
nickbp: This section is creating the docs page menu. This change updates it to automatically populate entries against whatever's in the `reference/` directory, mirroring the `tutorials/` behavior a few lines up. At the moment the only thing in `reference/` is `yaml-reference`.

frameworks/cassandra/src/test/java/com/mesosphere/sdk/cassandra/scheduler/CassandraRecoveryPlanOverriderTest.java
nickbp: Redid the testing API, now there's much less boilerplate here.

elezar:
Just a quick question that I thought of. Do we use strict Mustache rendering that fails if the template variables are not present? (see, for example: https://github.com/cbroglie/mustache/blob/master/mustache.go#L20)

nickbp:
Yes, see calls to `TemplateUtils.renderMustacheThrowIfMissing` which internally throws a descriptive `MustacheException` if params are unset*

\* Doesn't throw if params are inside a conditional which isn't being rendered, like `{{#foo}} ... {{/foo}}` where `foo` is `false` or empty/unset, but this would be expected in practice

elezar:
OK. Thanks, so we handle this on the Java side, but this is not done in bootstrap for templated config files (e.g. `cassandra.yaml`)?

nickbp:
Not sure what the mustache lib that bootstrap uses does. The library doesn't have the best documentation.

benclarkwood:
[Universe package testing](https://trello.com/c/UgsMl0fI/7-universe-package-testing)

frameworks/cassandra/src/test/java/com/mesosphere/sdk/cassandra/scheduler/CassandraRecoveryPlanOverriderTest.java
benclarkwood: 👍

PR #1541, opened 2017-09-01
hdfs - replace import-time cluster actions

falseperfection:
INFINITY-2275: hdfs - replace import-time cluster actions

### DIFF
--- a/frameworks/hdfs/tests/test_sanity.py
+++ b/frameworks/hdfs/tests/test_sanity.py
@@ -16,64 +16,67 @@ from tests import config
 
 log = logging.getLogger(__name__)
 
-FOLDERED_SERVICE_NAME = sdk_utils.get_foldered_name(config.SERVICE_NAME)
-
 
 @pytest.fixture(scope='module', autouse=True)
 def configure_package(configure_security):
     try:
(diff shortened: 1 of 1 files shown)


falseperfection:
You're right, but I'm avoiding touching any more than I have to in these refactors. I don't want to introduce any new bugs, and I don't own the tests themselves so hesitate to do any deeper refactoring. Maybe if all these get done I can go back and streamline? (If that's a high priority.)

elezar:
If we introduce a local variable called: `FOLDERED_SERVICE_NAME`, then the diff will be significantly smaller. A follow-up PR can then be used to rename this (e.g. change the case).

That is to say, the simplest change to achieve what this PR does is to move the global variables to function-local variables (with no renaming), and then proceed from there. (this holds for the other related PRs #1540 #1542 #1543 #1544 as well.

falseperfection:
A smaller diff perhaps, but I'd argue not a simpler change. It's hard to get simpler than search/replace and updating some indention. It also matches the pattern used in cassandra (#1466).

Are you saying the issue is worth blocking these tickets?

falseperfection:
Ok. I'll do that work next.

PR #1545, opened 2017-09-01
[INFINITY-2353] Fix IP lookup fail on overlay

NimaVaziri:
On 1.9 overlay we're currently getting the following when the bootstrap tries to get the container IP
```
2017/09/01 09:08:28 Looking up container IP address...
2017/09/01 09:08:28 Cannot look up container's IP address: LIBPROCESS_IP is INADDR_ANY: 0.0.0.0
```

This change falls through to resolve the IP from the hostname.

Have confirmed functionality via HDFS deployment that failed before this change set.


NimaVaziri:
Closing this for now, will re-open upon re-assessment.

PR #1547, opened 2017-09-01
INFINITY-2326/2327 Cleanup of some startup warnings, and other minor logging improvements

nickbp:
- Annotations and generic type on ConfigResource: Fixing some jersey warnings emitted on scheduler startup.
- Downgrade of EndpointResource logs: Produces noise on every query, which is only really useful for service developers anyway
- Other log fixes: Nitpicks.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/ConfigResource.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/ConfigResource.java
@@ -3,13 +3,13 @@ package com.mesosphere.sdk.api;
 import java.util.Arrays;
 import java.util.UUID;
 
+import javax.inject.Singleton;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.core.Response;
 
 import com.mesosphere.sdk.api.types.PrettyJsonResource;
-import com.mesosphere.sdk.config.Configuration;
 import com.mesosphere.sdk.state.ConfigStore;
 import com.mesosphere.sdk.state.ConfigStoreException;
 import com.mesosphere.sdk.storage.StorageError.Reason;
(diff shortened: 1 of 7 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/api/ConfigResource.java
```diff
@@ -26,14 +26,15 @@
  *
  * @param <T> The configuration type which is being stored by the framework.
  */
+@Singleton
```
benclarkwood: Why's this needed?

benclarkwood: (More out of curiosity than anything else. I haven't ever needed this annotation on resources.)

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -209,15 +219,14 @@ public void slaveLost(SchedulerDriver driver, Protos.SlaveID agentId) {
     }
 
     @Override
-    public void executorLost(SchedulerDriver driver, Protos.ExecutorID executorId, Protos.SlaveID slaveId, int status) {
+    public void executorLost(SchedulerDriver driver, Protos.ExecutorID executorId, Protos.SlaveID agentId, int status) {
```
benclarkwood: :) 👍

benclarkwood:
(I already kicked the C* job)

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/ConfigResource.java
nickbp: solely to fix `WARNING: A resource, class com.mesosphere.sdk.api.ConfigResource, implements provider interface [interface javax.ws.rs.ext.ContextResolver] but does not explicitly define the scope (@Singleton, @PerLookup). The resource class will be managed as singleton.`

PR #1552, opened 2017-09-05
replace dcos version markers

falseperfection:


conftest.py
```diff
@@ -109,3 +109,12 @@ def get_mesos_state_on_failure(test_name: str):
...
         with open(log_name, 'w') as f:
             f.write(r.text)
+
+
+def pytest_runtest_setup(item):
+    if isinstance(item, item.Function):
+        min_version_mark = item.get_marker('dcos_min_version')
+        if min_version_mark:
+            min_version = min_version_mark.args[0]
+            if sdk_utils.dcos_version_less_than(min_version):
```
mellenburg: I don't know if this will work where we are going with the scripting. These plugins are loaded in first with the conftest.py files before any dependencies are actually loaded or test fixtures evaluated. Thus, this might only work now because we authenticate with the cluster outside the test run, but when we go to authenticate in ONLY the fixture, this might explode because it requires talking to a setup CLI

falseperfection: Good point. If we were going to do our authentication in a fixture you're right. We could easily plan to do the authentication in a hook function as well. Possibly even right here so we auth and then kick out any invalid tests. Keeping those close together doesn't seem bad. (And I like how following the `@pytest.mark.dcos_min_version(V)` pattern keeps things pretty standard.)

I've started liking plugins more and more as we've moved through this refactoring. The idea of providing a pytest plugin to manage tests feels like a simple, _installable_ way to share a testing framework -- instead of documenting how to configure a bunch of dependent fixtures. I'm not ready to argue for that yet.... but that's where my mind has been going lately.

frameworks/cassandra/tests/test_recovery.py
```diff
@@ -28,7 +28,7 @@ def configure_package(configure_security):
 
 
 @pytest.mark.sanity
-@sdk_utils.dcos_1_9_or_higher  # dcos task exec not supported < 1.9
+@pytest.mark.dcos_min_version('1.9')  # dcos task exec not supported < 1.9
```
elezar: Would it make sense to add a message as a second argument to `mark.dcos_min_version` (is this even possible?) so that we can provide a reason instead of having to add this in a comment? This can then be printed in the `skip` output.

testing/sdk_utils.py
```diff
@@ -43,7 +43,7 @@ def get_zk_path(service_name):
 
 @functools.lru_cache()
 def dcos_version_less_than(version):
-    return shakedown.dcos_version_less_than("1.10")
+    return shakedown.dcos_version_less_than(version)
```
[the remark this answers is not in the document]
falseperfection: I hadn't been following that thread. So we've had stable builds for a while now because I made a typo when I created `dcos_version_less_than`? Ha!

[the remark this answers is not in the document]
elezar: I think @benclarkwood has a fix pending (I haven't had a chance to look at it again). Just wanted to give you a heads-up in case you see hello-world-19 tests failing.

frameworks/cassandra/tests/test_recovery.py
falseperfection: I'll add it. Good idea. Much better than comments.

conftest.py
falseperfection: A decorator version I could add to `sdk_utils.py`:

```
def dcos_min_version(f, min_version)
    @functools.wraps(f)
    def dcos_min_version_wrapper(*args):
        if dcos_version_less_than(min_version):
            pytest.skip('Feature only supported in DC/OS {} and up'.format(min_version))
        else:
            return f(*args)
    return dcos_min_version_wrapper
```

That would run after fixtures. Because it uses `skip` we should get an accurate report at the end of the run. Cluster auth could be done in a fixture or hook function. I don't like it, but I don't know why.

testing/sdk_utils.py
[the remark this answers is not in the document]
falseperfection: Appreciated! I was hitting the `run` button in TeamCity to see if hello-world-19 was a flake. (It worked for me locally so I assumed so.)

benclarkwood:
#1488 will solve this.

benclarkwood:
(It's just waiting on CI then it will be merged)

conftest.py
mellenburg: I actually really like the idea of doing the auth (and perhaps getting some cluster attributes) in a plugin. I thought the plugins were a little hard to work with because its not super clear where everything fits in the pytest framework. However, for things like this, I think most devs never need to touch it, so its alright to be harder to work with.

That being said, its always easier said than done so we just gotta dive in and see...but you are about to be out for a while. For the moment,  I prefer the decorator greatly as its the same deal we have now and we can at least finish the auth-in-fixture work while you are off

falseperfection: Ok. I'll get the decorator version working today, and we can revisit the plugin idea. Or you can go crazy and implement the whole thing on a bender weekend while I'm gone. 😆

mellenburg:
Manually rebased kevin's work to resolve the conflicts. Once those tests are through this should be good to merge

falseperfection:
I put back the mark-and-scan method of version skipping. All tests went green (until I rebased, pushing that through now). As our previous discussion pointed out, this will require that auth be done in a plugin.

@mellenburg - re-review please?

testing/sdk_utils.py
```diff
@@ -68,14 +68,30 @@ def is_open_dcos():
...
 
 
-# WARNING: Any file that uses these must also "import shakedown" in the same file.
-dcos_1_9_or_higher = pytest.mark.skipif(
-    'sdk_utils.dcos_version_less_than("1.9")',
-    reason="Feature only supported in DC/OS 1.9 and up")
-dcos_1_10_or_higher = pytest.mark.skipif(
-    'sdk_utils.dcos_version_less_than("1.10")',
-    reason="Feature only supported in DC/OS 1.10 and up")
+# def dcos_min_version(min_version, reason=None):
```
mellenburg: What is the motivation for checking this in? Do we plan or foresee needing to use this again?

falseperfection: Failure to clean up after myself. :(

PR #1529, opened 2017-08-30
[INFINITY-2236] Verify Kafka upgrade instructions from 1.1.26-beta 

mbdcos:
This needs to be verified manually to ensure all these instructions are up to date and correct.

->

Originally  a **generic** document was prepared,  explaining upgrade from _0.10.2 Kafka_ to _0.11.1 Kafka_, and explaining upgrade in strict mode with service account change. It was referring versions as **old version** (with x) and **new version** (with y). 

Then based on reviewers, _version numbers_ in the current document were added. Now the existing document is not a generic document. It became version specific. 
See the following PRs:
#1449
#1429

2.0.0 GA release does not have any `upgradesFrom` field, so this document will be misleading
 We do not support upgrade from betas to GA for the moment.

This PR is to modify the upgrade document from 1.1.26-beta to 1.1.27-beta (these are the only ones with `upgradesFrom` in universe at this moment for Kafka).

**Note:** There is no upgrade from DC/OS Kafka 1.1.19 to 2.0.0 GA (manual marathon upgrade). 2.0.1 is not released yet.

**Note:** This PR is only for DC/OS 1.10 upgrade. DC/OS 1.9 needs manual marathon upgrade

PR #1553, opened 2017-09-05
[INFINITY-2236] Remove Kafka upgrade.doc (There is no upgradesFrom field in 2.0.0 GA)

mbdcos:
Remove the upgrade.md document for the moment (2.0.0 GA release does not have any `upgradesFrom` field, so this document will be misleading) till it is fixed.

#1529 
#1449
#1429


triclambert:
This document is intended for use by people consuming the beta releases.  None of those people are on 1.10, and therefore cannot use the new upgrade capabilities.  We need to make sure people using the beta-kafka or beta-confluent-kafka packages know they need to closely review their config options.

mbdcos:
@triclambert instructions for upgrade in DC/OS 1.9 are not done yet.

PR #1554, opened 2017-09-05
[INFINITY-2359] Disable kafka tls in stable

benclarkwood:
This removes the Kafka TLS settings from the stable branch. I've simply removed them from config.json and the docs. The overall plumbing still exists.


benclarkwood:
(Customers who want to try TLS will consume the beta packages from `master`.

benclarkwood:
Yes, I omitted the guides as the functionality is indeed present and would be hard to rip out.

PR #1562, opened 2017-09-06
[INFINITY-2223] Determine upgradesFrom automatically

elezar:
This PR adds `{{upgrades-from}}` (and `{{downgrades-to}}`) templating to the `UniversePackageBuilder`.

This is done through a simple `PackageManager` wrapper around universe (supporting both `curl` and `requests`). In order to provide the correct sorting of the versions returned by universe, a `Version` class is used which takes both the `releaseVersion` (effectively the folder number in the universe repository) and the `version` (a string version specified by us) into consideration. The latest version is the one with the **highest** `releaseVersion`.

Currently, I have a `VersionResolver` class which processes the output of the `PackageManager.get_packages()` method to determine the available package versions (similar to the `dcos package describe <package-name> --package-versions` command, and also provides the latest version.

In order to integrate this more cleanly into the `UniversePackageBuilder`, I also introduced a `Package` class which encapsulates the package specifics.

In the case where the latest version cannot be determined, `{{upgrades-from}}` (and `{{downgrades-to}}`) is set to `"*"` so that all upgrades are allowed.

# TODO:
- [x] Add templating to other frameworks
- [x] Ensure that `requests` times out too


tools/universe_builder.py
```diff
@@ -1,237 +1,43 @@
...
+                                                 "$ %(prog)s kafka 1.2.3-4.5.6 /path/to/template/jsons/ "
+                                                 "https://example.com/path/to/kafka-artifacts /path/to/artifact1.zip "
+                                                 "/path/to/artifact2.zip /path/to/artifact3.zip",
+                                     epilog="In addition, environment variables named \'TEMPLATE_SOME_PARAMETER\' will "
+                                            "be inserted against the provided package template (with params of the "
+                                            " form '{{some-parameter}}')")
+    parser.add_argument('package_name', type=str, help='The package name')
+    parser.add_argument('package_version', type=str, help='The package version string')
+    parser.add_argument('package_dir_path', type=str, help='The local path where the package template is located')
+    parser.add_argument('upload_dir_url', type=str, help='url of the directory where artifacts are located (S3, etc):')
```
nickbp: Is the trailing colon in the help string intentional?

tools/universe/package.py
```diff
@@ -0,0 +1,34 @@
...
+from .version_resolver import VersionResolver
+
+
+class Package:
+
+    def __init__(self, name, version, version_resolver=None):
+        self._name = name
+        self._version = version
+
+        self._set_version_resolver(version_resolver)
```
nickbp: Having this broken out in a separate function feels a little overkill, would it make sense to have the contained logic directly under `__init__`?

tools/universe/package.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def _set_version_resolver(self, version_resolver):
```
nickbp: Python likes double newlines between functions (here and below)

tools/universe/package_builder.py
```diff
@@ -0,0 +1,210 @@
...
+                    artifact_path, artifact_paths))
+            prior_path = self._artifact_file_paths.get(
+                os.path.basename(artifact_path), '')
+            if prior_path:
+                raise Exception('Duplicate filename between "{}" and "{}". Artifact filenames must be unique.'.format(
+                    prior_path, artifact_path))
+            self._artifact_file_paths[os.path.basename(
+                artifact_path)] = artifact_path
+
+    def set_input_dir_path(self, input_dir_path):
```
nickbp: IIRC, python likes double newlines between functions (here and below). Looks like the previous version didn't have them, but may as well fix it while here...

tools/universe/package_manager.py
```diff
@@ -0,0 +1,87 @@
...
+        self.universe_url = "https://universe.mesosphere.com/repo"
+        self._headers = {
+            "User-Agent": "dcos/{}".format(dcos_version),
+            "Accept": "application/vnd.dcos.universe.repo+json;"
+                      "charset=utf-8;version=v{}".format(package_version),
+        }
+
+        self.__package_cache = None
+
+    def get_packages(self):
```
nickbp: IIRC, python likes double newlines between functions (here and below). Looks like the previous version didn't have them, but may as well fix it while here...

tools/universe/version_resolver.py
```diff
@@ -0,0 +1,63 @@
...
+        if self.release_version > other.release_version:
+            return True
+
+        return self.package_version > other.package_version
+
+    def __str__(self):
+        return str(self.package_version)
+
+
+class VersionResolver:
```
nickbp: This feels pretty tied to the `PackageManager` implementation, in terms of expecting the Universe response data to match a certain format. It may make things a little simpler to just combine the two into a single class that does any version extraction internally

tools/universe_builder.py
```diff
@@ -1,237 +1,43 @@
-#!/usr/bin/env python3
```
nickbp: Given that this has a main() function it may make sense to retain the executable bit + crunchbang so that it's directly executable. IIRC some of the bash tooling may call directly into `universe_builder.py`.

tools/universe/package.py
```diff
@@ -0,0 +1,34 @@
...
+        if version_resolver is None:
+            self._version_resolver = VersionResolver(PackageManager())
+        else:
+            self._version_resolver = version_resolver
+
+    def get_upgrades_from(self):
+        if self._version_resolver is None:
```
nickbp: Given the logic in _set_version_resolver(), wouldn't this be unreachable?

tools/universe/package.py
```diff
@@ -0,0 +1,34 @@
+from .package_manager import PackageManager
+from .version_resolver import VersionResolver
+
+
+class Package:
```
nickbp: Now that I think about it, how about moving the version lookup outside of the Package class, and having the Package class function as a data-only struct/container for a name+version?

- `PackageManager`/`VersionResolver` could accept a package name as a parameter and return a Package object.
- `_get_template_mapping_for_content()` could deal with lookup and `"*"` fallback internally
- `PackageManager`'s cache would be more useful since it would survive across package names by default. It could hold a pre-parsed dict of `name => Package` too. Debug logs could print the list of package versions that it found by dumping that dict.

(This comment was written after the following comments)

nickbp: Also, I could see us wanting to have deeper relationships between `Package` objects at some point, for example having a `Package` object graph that follows the `upgradesFrom`/`downgradesTo` paths. Having `PackageManager` construct the `Package`s directly could make constructing this graph a little easier.

benclarkwood:
[Determine upgradeFrom](https://trello.com/c/wa4TRfim/26-determine-upgradefrom)

tools/universe/package.py
elezar: As far as I am aware, PEP8 only requires the double newlines between module-level functions or classes, for class methods, or other whitespace, a single blank line is required. I fill run `flake8` on the files here again to be sure.

elezar:
Thanks for the review @nickbp. You made some good points. I have made some changes, most notably the following:

* `VersionResolver` and `PackageManager` are no longer separate classes, but a single `PackageManager` class.
* The `PackageManger` methods now return `Package` objects.
* The package cache is now a dictionary keyed on the package name with values all the versioned `Package` objects.
* The handling of the fallback to `"*"` is handled in `UniversePackageBuilder`
* The formatting of the files has been addressed.

tools/universe/package.py
```diff
@@ -0,0 +1,82 @@
...
+
+        return self.get_version() > other.get_version()
+
+    def __str__(self):
+        return json.dumps({
+            'name': self.get_name(),
+            'version': str(self.get_version())
+         })
+
+    # def get_upgrades_from(self):
```
nickbp: remove dead code imo

tools/universe/test_package.py
```diff
@@ -0,0 +1,56 @@
+from .package import Package
+from .package import Version
+
+
+# def test_no_version_resolver_upgrades_from_all(mocker):
+
```
nickbp: dead code

tools/universe_builder.py
```diff
@@ -1,237 +1,62 @@
...
-
-
-def print_help(argv):
-    logger.info(
-        'Syntax: {} <package-name> <package-version> <template-package-dir> <artifact-base-path> [artifact files ...]'.format(argv[0]))
-    logger.info(
-        '  Example: $ {} kafka 1.2.3-4.5.6 /path/to/template/jsons/ https://example.com/path/to/kafka-artifacts /path/to/artifact1.zip /path/to/artifact2.zip /path/to/artifact3.zip'.format(argv[0]))
-    logger.info(
-        'In addition, environment variables named \'TEMPLATE_SOME_PARAMETER\' will be inserted against the provided package template (with params of the form \'{{some-parameter}}\')')
+EPILOG_STRING = "In addition, environment variables named 'TEMPLATE_SOME_PARAMETER' " \
```
nickbp: `EPILOGUE`?

tools/universe/package_manager.py
```diff
@@ -0,0 +1,127 @@
...
+            cmd.extend(["-H", "{}: {}".format(k, header)])
+
+        cmd.append(universe_url)
+
+        try:
+            output = subprocess.check_output(cmd)
+            status_code = int(output)
+
+            if status_code != 200:
+                raise Exception("Curl returned status code %s", status_code)
```
nickbp: the constructor to Exception doesn't appear to handle formatting like this automatically (tried manually on py3.5)

tools/universe/package_builder.py
```diff
@@ -0,0 +1,236 @@
...
+        # read files into memory and apply templating to files:
+        updated_package_files = {}
+        for filename, content in self._iterate_package_files():
+            updated_package_files[filename] = self._apply_templating_to_file(
+                filename, content)
+        scratchdir = tempfile.mkdtemp(prefix='stub-universe-tmp')
+        jsonpath = os.path.join(
+            scratchdir,
+            'stub-universe-{}.json'.format(self._package.get_name()))
+        jsonfile = open(jsonpath, 'w')
```
nickbp: while we're here... this could be `with open(jsonpath, 'w') as jsonfile: ...`

tools/universe_builder.py
elezar: Was basing the name on the fact that `argparse` refers to this as the `epilog`: https://docs.python.org/3/library/argparse.html#epilog, but will change it.

elezar:
Thanks @nickbp. I think I addressed your comments.

There was also an issue with the way we sort versions that was causing elastic to fail, but this should also have been addressed.

tools/universe/package_builder.py
nickbp: looks like this is still the same?

elezar:
Reverted a change that must have included the json file fix. Will reapply.

elezar:
The passing of the HDFS tests would require that #1619 be merged.

PR #1563, opened 2017-09-06
[DOCS-2104] Remove 'beta-' references w/i stable docs

paegun:
* Remove 'beta-' references w/i stable docs, ensuring that the removals
  are contextually correct (mostly package name and DC/OS CLI service
  name).


paegun:
This change was performed incorrectly w/ `master` as the base, should've been `0.30.X`

PR #1564, opened 2017-09-06
[DOCS-2104] Remove 'beta-' references w/i stable docs

paegun:
* Remove 'beta-' references w/i stable docs, ensuring that the removals
  are contextually correct (mostly package name and DC/OS CLI service
  name).


sascala:
Thank you, @paegun

elezar:
Just a comment on this. In #1526 a rename of `node` to `native-client` is mentioned. Is this also applicable for the `0.30.X` docs?

PR #1565, opened 2017-09-06
Decouple candidate generation from the offer cycle

gabrielhartmann:
This is a pure refactor with no functional changes expected.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultPlanCoordinator.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultPlanCoordinator.java
@@ -31,18 +31,13 @@ public class DefaultPlanCoordinator implements PlanCoordinator {
         this.planScheduler = planScheduler;
     }
 
-    @Override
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultPlanCoordinator.java
```diff
@@ -31,19 +31,10 @@ public DefaultPlanCoordinator(
...
     }
 
-    @Override
-    public Collection<OfferID> processOffers(
-            final SchedulerDriver driver,
-            final List<Offer> offersToProcess) {
-        // Offers that have already been used
-        final Set<OfferID> dirtiedOffers = new HashSet<>();
-
+    private List<Step> getCandidates() {
```
nickbp: may as well throw in a javadoc IMO

gabrielhartmann: Will do.

gabrielhartmann:
Cassandra test failed when deleting cluster.  All tests passed.

PR #1570, opened 2017-09-07
[INFINITY-2379] Extend state store API to get/set files

NimaVaziri:
This adds to the `StateStore` endpoints the ability to upload files, download them, and list the files the `StateStore` holds at any given time.

- [x] add tests

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/StateResource.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/api/StateResource.java
@@ -9,17 +9,18 @@ import com.mesosphere.sdk.storage.PersisterCache;
 import com.mesosphere.sdk.storage.PersisterException;
 import com.mesosphere.sdk.storage.StorageError.Reason;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.mesos.Protos;
+import org.glassfish.jersey.media.multipart.FormDataParam;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.ws.rs.GET;
-import javax.ws.rs.PUT;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import java.io.*;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Optional;
@@ -32,6 +33,8 @@ import java.util.Optional;
 public class StateResource {
 
     private static final Logger logger = LoggerFactory.getLogger(StateResource.class);
+    private static final String VERSION_FILE = "versionFile";
+    private static final String VERSION_FILE_ENCODING = "UTF-8";
 
     private final StateStore stateStore;
     private final PropertyDeserializer propertyDeserializer;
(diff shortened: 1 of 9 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/api/StateResource.java
```diff
@@ -94,6 +97,70 @@ public Response getPropertyKeys() {
...
+        String fileName = fileDetails.getFileName();
+        if (uploadedInputStream == null || fileName == null) {
+            return ResponseUtils.plainResponse(
+                    "Specify 'file' as a parameter in your request",
+                    Response.Status.BAD_REQUEST
+            );
+        }
+
+        try {
+            if (stateStore.hasProperty(StateStoreUtils.FILE_NAME_PREFIX + fileName)) {
```
nickbp: Should the prior content be overwritten in this case? If so, then may as well omit the new StateStore.hasProperty() call as this is the only place using it at the moment.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -452,6 +452,22 @@ public void clearAllData() throws StateStoreException {
...
+     */
+    public boolean hasProperty(String key) throws StateStoreException {
+        try {
+            final String path = PersisterUtils.join(PROPERTIES_PATH_NAME, key);
+            logger.debug("Determining whether key: {} exists at path: {}", key, path);
+            // persister throws an exception if the path (i.e. key) doesn't exist.
+            persister.get(path);
+            return true;
+        } catch (PersisterException e) {
+            logger.info("Property {} doesn't exist in the state store", key);
```
nickbp: Should check for e.getReason() == Reason.NOT_FOUND. see example at L446

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/StateResource.java
```diff
@@ -94,6 +97,70 @@ public Response getPropertyKeys() {
...
+                String msg = String.format("File %s is already in the state store", fileName);
+                logger.info(msg);
+                return ResponseUtils.plainOkResponse(msg);
+            }
+            logger.info("Storing {}", fileName);
```
nickbp: Should have a sanity check against the metadata that the advertised size isn't too big (maybe a field in fileDetails has this?)

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreUtils.java
```diff
@@ -304,4 +310,45 @@ public static void setLastCompletedUpdateType(
             return ConfigurationUpdater.UpdateResult.DeploymentType.valueOf(value);
         }
     }
+
```
nickbp: Looks like these methods are only used in StateResource. Can just move them to private methods in StateResource.

sdk/scheduler/src/test/java/com/mesosphere/sdk/state/StateStoreUtilsTest.java
```diff
@@ -32,6 +33,8 @@
  */
 public class StateStoreUtilsTest {
```
nickbp: Missing unit tests against the new StateResource calls?

Should also exercise the malicious large content cases described above:
- `Content-Length` too big: fail fast.
- `Content-Length` small but actual content too big: fail during streaming.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreUtils.java
```diff
@@ -304,4 +310,45 @@ public static void setLastCompletedUpdateType(
...
+    /**
+     * Stores the file in the state store.
+     * @param stateStore The state store to store the file in.
+     * @param fileName The name of the file to store.
+     * @param uploadedInputStream The input stream holding the content of the file.
+     */
+    public static void storeFile(StateStore stateStore, String fileName, InputStream uploadedInputStream)
+            throws StateStoreException, IOException {
+        StringWriter stringWriter = new StringWriter();
+        IOUtils.copy(uploadedInputStream, stringWriter, FILE_ENCODING);
```
nickbp: Need to limit the size of this copy. This check is needed in addition to any size check against the metadata above.
 Malicious input could have a `Content-Length` of 1024byte and an actual content length of infinity. E.g. piping /dev/zero into HTTP.

NimaVaziri:
These commits were separated into their own PR but originally it was for a journal node case with HDFS where the journal nodes have to upload a file as part of handling a replace operation (which is coming up in another PR).

sdk/scheduler/src/test/java/com/mesosphere/sdk/state/StateStoreUtilsTest.java
nickbp: (These cases could probably be done by mocking InputStream and/or FormDataContentDisposition)

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/StateResource.java
NimaVaziri: Yeah, this restricts the more general use case...will omit.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreUtils.java
NimaVaziri: Ah, in memory of heartbleed.

NimaVaziri:
@nickbp addressed your comments, any other ones?

nickbp:
Nope!

PR #1573, opened 2017-09-07
DCOS-2106: Update post install documentation link.

elezar:
* Add templates for documentation-path
* Add templates for issues-path
* Update packaging and build scripts

### DIFF
--- a/tools/universe_builder.py
+++ b/tools/universe_builder.py
@@ -85,7 +85,10 @@ class UniversePackageBuilder(object):
         return hasher.hexdigest()
 
     def _get_documentation_path(self):
-        return "{}/service-docs/{}/".format(_docs_root, self._pkg_name)
+        documentation_path = "{}/service-docs/{}/".format(_docs_root, self._pkg_name)
+        if self._pkg_version != "stub-universe":
+            documentation_path = "{}v{}/".format(documentation_path, self._pkg_version)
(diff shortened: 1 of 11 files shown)


nickbp:
DCOS-2106 appears to be unrelated. Different ticket?

elezar:
@nickbp yes, the number of places we end up making the same change is a bit much.

tools/universe_builder.py
```diff
@@ -20,6 +20,8 @@
 _jre_jce_unlimited_url = 'https://downloads.mesosphere.com/java/jre-8u131-linux-x64-jce-unlimited.tar.gz'
 _libmesos_bundle_url = 'https://downloads.mesosphere.io/libmesos-bundle/libmesos-bundle-1.10-1.4-63e0814.tar.gz'
 
+_docs_root = "https://docs.mesosphere.com"
```
triclambert: In the future we'll probably want this overridable for partners.

triclambert:
Awesome.

tools/universe_builder.py
elezar: Yes, definitely. We can already override the final templating values by setting a `TEMPLATE_DOCUMENTATION_PATH` environment variable (in keeping with the other template variables), but this may not be sufficient.

PR #1577, opened 2017-09-07
[DOCS-2105] Split out the template service guide into appropriate docs

paegun:
* Aggregate and reconcile differences between `frameworks/*/docs/*.md` and
  `README.md` .
* Change file names and menu_order to match new (1.10) layout.
* Reduce `README.md` to provide instructions and the layout.


sascala:
many thanks, @paegun

PR #1579, opened 2017-09-07
[INFINITY-2374] Disable tests that will not pass in open (#1567) (backport)

benclarkwood:
* Detect DC/OS Open and disable tests for features that do not work there.

* Bad python

* Disable C* TLS test in open


benclarkwood:
[Disable things that will never run in open](https://trello.com/c/y0IfFhXW/1-disable-things-that-will-never-run-in-open)

PR #1583, opened 2017-09-07
INFINITY-2385 Clean up/unify Limitations: Include common bits, leave service-specific bits where relevant

nickbp:


nickbp:
Removed automatic recovery throughout, and additionally applied the union of fixes that had been mentioned elsewhere throughout

PR #1588, opened 2017-09-07
Set Mag7 as selected in stable branch

benclarkwood:
Elastic and Kibana are already marked as selected.


benclarkwood:
[Mark as selected in stable](https://trello.com/c/rDzKIT0p/14-mark-as-selected-in-stable)

PR #1594, opened 2017-09-08
Decouple plan execution from the offer cycle

gabrielhartmann:
The work represented by plans is not solely accomplished by the consumption of offers.  In particular the leaf element of plans, a step, can return an `Optional<PodInstanceRequirement>` when it is started.

The execution of plans is today driven by the presence or absence of offers, however logically when no work requiring offers is present the scheduler should suppress offers.  However steps which do not require offers should still be processed.

In this change, we now no longer block plan execution on waiting for offers.  If offers are present they are consumed otherwise plan execution continues without them.

This is a pre-requisite for #1578

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
@@ -43,6 +43,7 @@ public abstract class AbstractScheduler implements Scheduler {
 
     private Object inProgressLock = new Object();
     private Set<Protos.OfferID> offersInProgress = new HashSet<>();
+    private AtomicBoolean processingOffers = new AtomicBoolean(false);
 
     /**
      * Executor for handling TaskStatus updates in {@link #statusUpdate(SchedulerDriver, Protos.TaskStatus)}.
@@ -50,7 +51,7 @@ public abstract class AbstractScheduler implements Scheduler {
     protected final ExecutorService statusExecutor = Executors.newSingleThreadExecutor();
 
     /**
-     * Executor for processing offers off the queue in {@code processOffers()}.
+     * Executor for processing offers off the queue in {@link #executePlansLoop()}.
      */
     private final ExecutorService offerExecutor = Executors.newSingleThreadExecutor();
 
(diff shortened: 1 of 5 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -43,14 +43,15 @@
...
+    private AtomicBoolean processingOffers = new AtomicBoolean(false);
 
     /**
      * Executor for handling TaskStatus updates in {@link #statusUpdate(SchedulerDriver, Protos.TaskStatus)}.
      */
     protected final ExecutorService statusExecutor = Executors.newSingleThreadExecutor();
 
     /**
-     * Executor for processing offers off the queue in {@code processOffers()}.
+     * Executor for processing offers off the queue in {@code executePlans()}.
```
nickbp: could do `{@link #executePlans()}` (or `{@link #executePlansLoop()}`) here, and then it should update automatically in any future IDE renames

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -93,16 +93,19 @@ public void registered(SchedulerDriver driver, Protos.FrameworkID frameworkId, P
         postRegister();
     }
 
-    private void processOffers() {
+    /**
+     * This method starts the main execution thread for the scheduler.  It drives the execution of plans including
+     * providing offers.
+     */
+    private void executePlans() {
```
nickbp: `executePlansLoop`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/OfferQueue.java
```diff
@@ -29,15 +32,19 @@ public OfferQueue(int capacity) {
     }
 
     /**
-     * Calling this method is a blocking Operation. It returns all Offers currently in the queue. It always returns at
-     * least one Offer.
+     * Calling this method will wait for Offers for the provided duration.
+     * It returns all Offers currently in the queue if any are present and none otherwise.
      */
-    public List<Protos.Offer> takeAll() {
+    public List<Protos.Offer> takeAll(Duration duration) {
```
nickbp: `... and an empty list if the duration is reached`?

PR #1595, opened 2017-09-08
[DOCS-2106] Update post-install links for packages

elezar:
* backport changes from #1573 to 0.30.X branch


elezar:
Note, I have an open question. What are the correct `Issues` links for the various packages. I've taken the default as `https://docs.mesosphere.com/support`, but some of the projects refer to JIRA.

PR #1599, opened 2017-09-08
Fix EE check

benclarkwood:
pytest skipif is not correctly setup.

Add function invocation.


benclarkwood:
[Fix EE check](https://trello.com/c/riHy9Nex/20-fix-ee-check)

elezar:
@benclarkwood @mhrabovcin there was an error in the parameter order for the cassandra install step. I have pushed a change to correct this.

elezar:
@benclarkwood @mhrabovcin The Cassandra TLS test now fails. I see an error in the logs regarding the `.ssl` directory in the `MESOS_SANDBOX` not being found. Is there some PR out that should fix this?

benclarkwood:
I hadn't had a chance to circle back around to this.

@mhrabovcin, I saw some rumblings somewhere that this directory has disappeared? Don't we rely on it for all sorts of things?

elezar:
Just remembered that Martin is on leave this week.

PR #1601, opened 2017-09-08
Fix disabling framework bits, and support parameters in create_service_account

nickbp:
(Backporting of fixes from k8s)

### DIFF
--- a/tools/build_framework.sh
+++ b/tools/build_framework.sh
@@ -13,8 +13,9 @@ usage() {
 # Optional envvars:
 #   REPO_ROOT_DIR: path to root of source repository (default: parent directory of this file)
 #   REPO_NAME: name of the source repository (default: directory name of REPO_ROOT_DIR)
-#   BOOTSTRAP_DIR: path to bootstrap tool, or an empty string to disable bootstrap build (default: <REPO_ROOT_DIR>/sdk/bootstrap)
-#   CLI_DIR: path to the CLI directory, or an empty string to disable CLI build (default: </absolute/framework/path>/cli/)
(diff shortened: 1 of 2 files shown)


tools/build_framework.sh
```diff
@@ -13,8 +13,9 @@ usage() {
 # Optional envvars:
 #   REPO_ROOT_DIR: path to root of source repository (default: parent directory of this file)
 #   REPO_NAME: name of the source repository (default: directory name of REPO_ROOT_DIR)
-#   BOOTSTRAP_DIR: path to bootstrap tool, or an empty string to disable bootstrap build (default: <REPO_ROOT_DIR>/sdk/bootstrap)
-#   CLI_DIR: path to the CLI directory, or an empty string to disable CLI build (default: </absolute/framework/path>/cli/)
+#   BOOTSTRAP_DIR: path to bootstrap tool, or "disable" to disable bootstrap build (default: <REPO_ROOT_DIR>/sdk/bootstrap)
```
nickbp: (The issue encountered here: treating empty string as 'disabled' just leads to the following code using the default path -- because empty string is equivalent to unset)

nickbp:
Gonna merge this as it only affects service builds, and those all worked.

tools/build_framework.sh
```diff
@@ -94,12 +96,19 @@ if [ x"$cli_only" = xtrue ]; then
     exit
 fi
 
-if [ -n "$BOOTSTRAP_DIR" ]; then
+DISABLED_VALUE="disable"
+
+if [ "$BOOTSTRAP_DIR" != $DISABLED_VALUE ]; then
```
elezar: I know this is merged, but just wanted to point out that these conditional blocks here are redundant now as the output for both branches will be the same if the value is set to `"disabled"`.

PR #1608, opened 2017-09-11
[CASSANDRA-647] Fix injection of configuration on Cassandra's restore sidecar task

gabrielhartmann:
The root cause of restore not working without default ports is that each TaskGroup get its own sandbox, so the restore tasks were using default `cassandra.yml` which always has default setting for everything.

### DIFF
--- a/frameworks/cassandra/src/main/dist/svc.yml
+++ b/frameworks/cassandra/src/main/dist/svc.yml
@@ -185,6 +185,10 @@ pods:
         {{/TASKCFG_ALL_CASSANDRA_ENABLE_TLS}}
       restore-snapshot:
         goal: FINISHED
+        configs:
+          cassandra:
+            template: {{CONFIG_TEMPLATE_PATH}}/cassandra.yaml
+            dest: apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra.yaml
         cmd: >
                 ./bootstrap --resolve=false ;
                 export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/) ;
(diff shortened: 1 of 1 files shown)


frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -185,6 +185,10 @@ pods:
         {{/TASKCFG_ALL_CASSANDRA_ENABLE_TLS}}
       restore-snapshot:
         goal: FINISHED
+        configs:
```
elezar: There are some other tasks that also don't explicitly specify this file in the `configs`. Is this file implicitly used in any of these tasks?

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -202,15 +206,17 @@ pods:
                             -f ./apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra.yaml \
                             -d ${LOCAL_SEEDS} \
                             -p ${CASSANDRA_NATIVE_TRANSPORT_PORT} "$t" ;
+                    result=$?
```
nickbp: - can `set -e` be called at the start of the `cmd`?
- the block's indent is off

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -202,15 +206,17 @@ pods:
...
+                    echo sstableloader exited with code: $result
+                    if [ $result != 0 ]
```
nickbp: `if [ $result != 0 ]; then` imo

frameworks/cassandra/src/main/dist/svc.yml
nickbp: Did moving the config up here change anything? The PR description implies that it wasn't listed at all before, looks like the effective change is just the error checking in `cmd`

PR #1610, opened 2017-09-11
[KAF-374] Fix test_availability env variable consistency

paegun:
* Make env variable name consistent BROKER_KILL_GRACE_PERIOD.
* Remove errant personal pytest mark.

backport PR #1611

### DIFF
--- a/frameworks/kafka/tests/test_availability.py
+++ b/frameworks/kafka/tests/test_availability.py
@@ -22,7 +22,7 @@ STARTUP_POLL_DELAY_SECONDS = os.environ.get('STARTUP_LOG_POLL_DELAY', 2)
 def setup_module(module):
     options = {
         "brokers": {
-            "kill_grace_period": BROKERS_KILL_GRACE_PERIOD
+            "kill_grace_period": BROKER_KILL_GRACE_PERIOD
         }
     }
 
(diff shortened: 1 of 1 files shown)


frameworks/kafka/tests/test_availability.py
```diff
@@ -40,7 +40,6 @@ def teardown_module(module):
     sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
 
 
-@pytest.mark.paegun
 @pytest.mark.availability
```
elezar: @paegun the test below fails due to the use of `pods restart` instead of `pod restart`. An additional point is that the `retries` variable is not decremented when running `topic producer_tests test`. Would make sense to rather use the `retrying` module here if this is what we want? This would allow us to, for example, raise an exception in the failure case.

paegun: Yes, it would be better to use `retrying`, but IMO we should apply this throughout, not be inconsistent.

paegun:
@elezar we already have a backport PR for this, so I'll create similar fixes for Confluent. Less paperwork :)

elezar:
With regards to:
> @elezar we already have a backport PR for this, so I'll create similar fixes for Confluent. Less paperwork :)

The `pod` fix is already in Confluent, so there's no rush there.

PR #1615, opened 2017-09-12
Dont collect completed logs

mellenburg:
also try/catch all to avoid raising an
error within pytest

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -31,7 +31,7 @@ def get_task_ids(user: str=None):
     """ This function uses dcos task WITHOUT the JSON options because
     that can return the wrong user for schedulers
     """
-    tasks = subprocess.check_output(['dcos', 'task', '--completed']).decode().split('\n')
+    tasks = subprocess.check_output(['dcos', 'task']).decode().split('\n')
     for task_str in tasks[1:]: # First line is the header line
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -42,7 +42,7 @@ def get_task_ids(user: str=None):
 
 def get_task_logs_for_id(task_id: str,  task_file: str='stdout', lines: int=1000000):
```
paegun: "Naming things is hard"? I'd expect a method name "get_X" to return X or throw. The only reason I'm commenting on this is b/c removing `--completed` task ids above and task logs w/i this method are changes to the semantic meaning of the method, BUT given that the method is already:
1. not named very well according to its semantic
1. localized in effect b/c the "got" thing is actually emitted to stdout

The change in semantic to not include completed items is good.

mellenburg: Totally agree. will consider that the next time something near this gets touched

mellenburg:
HDFS failed so you can see this in action: https://teamcity.mesosphere.io/viewLog.html?buildId=781471&buildTypeId=DcosIo_DcosCommons_Hdfs_BuildAndTestPr&tab=buildLog&_focus=1728#_state=1727

PR #1617, opened 2017-09-13
[HDFS-455] Support TLS in HDFS

mhrabovcin:


frameworks/hdfs/tests/test_tls.py
```diff
@@ -0,0 +1,112 @@
...
+    sdk_cmd.run_cli(
+        "security org groups add_user superusers {name}".format(name=name))
+    yield name
+    sdk_security.delete_service_account(
+        service_account_name=name, service_account_secret=name)
+
+
+@pytest.fixture(scope='module')
+def hdfs_service_tls(service_account):
+    sdk_install.install(
```
NimaVaziri: Maybe wrap this install/uninstall in a try/finally block to ensure cleanup?

frameworks/hdfs/tests/test_tls.py
```diff
@@ -0,0 +1,112 @@
...
+@pytest.mark.parametrize("node_type,port", [
+    ('journal', DEFAULT_JOURNAL_NODE_TLS_PORT),
+    ('name', DEFAULT_NAME_NODE_TLS_PORT),
+    ('data', DEFAULT_DATA_NODE_TLS_PORT),
+])
+def test_verify_https_ports(node_type, port, hdfs_service_tls):
+    """
+    Verify that HTTPS port is open name, journal and data node types.
+    """
+    journal_host = sdk_hosts.autoip_host(
```
NimaVaziri: I'm assuming `journal_host` is just a stale variable name?

NimaVaziri:
@mhrabovcin thanks for this! Overall lgtm, just a couple minor comments.

frameworks/hdfs/src/main/dist/hdfs-site.xml
```diff
@@ -20,7 +20,7 @@
     {{#TLS_ENABLED}}
     <property>
         <name>dfs.http.policy</name>
-        <value>HTTP_AND_HTTPS</value>
+        <value>HTTPS_ONLY</value>
```
NimaVaziri: Perhaps value should be templated in case TLS is not enabled? This looks like it might break if it's using HTTPS by default and the setup doesn't include TLS.

NimaVaziri: Ah, never mind. This entire property is templated, meaning if TLS isn't enabled it'll use HTTP by default 👍

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -53,6 +57,9 @@ pods:
           ssl-server:
             template: ssl-server.xml
             dest: {{HDFS_VERSION}}/etc/hadoop/ssl-server.xml
+          ssl-client:
+            template: {{CONFIG_TEMPLATE_PATH}}/ssl-client.xml
```
elezar: @nickbp made some changes w.r.t `{{CONFIG_TEMPLATE_PATH}}` in #1539. It seems as if this is not needed anymore, but I would check with him to be sure.

elezar: (Looking at the `ssl-server` section just above this seems to confirm that this is no longer needed).

frameworks/hdfs/tests/test_tls.py
```diff
@@ -0,0 +1,112 @@
...
+
+@pytest.fixture(scope='module')
+def service_account():
+    """
+    Creates service account with `hdfs` name and yields the name.
+    """
+    name = SERVICE_NAME
+    sdk_security.create_service_account(
+        service_account_name=name, service_account_secret=name)
+     # TODO(mh): Fine grained permissions needs to be addressed in DCOS-16475
```
elezar: Since this function is used in all the TLS tests, we could consider moving it to a function in `sdk_security` (perhaps in a different PR).

mhrabovcin: https://jira.mesosphere.com/browse/INFINITY-2419

frameworks/hdfs/src/main/dist/svc.yml
nickbp: Yes, relative paths can be provided now, and they will default to the directory that the svc.yml is in (can be customized in the java API)

frameworks/hdfs/universe/config.json
```diff
@@ -261,6 +261,11 @@
           "description": "The HTTP port for HDFS Name Nodes. ",
           "default": 9002
         },
+        "name_node_https_port": {
```
nickbp: Could you add a unit test to HDFS' ServiceTest.java for enabling TLS and/or customizing these ports? Unit tests can now exercise universe configs. See Cassandra's ServiceTest.java for an example of enabling tls in config.json and then checking that the resulting per-task cassandra.yaml config is correctly generated:

https://github.com/mesosphere/dcos-commons/blob/2ef89c3fd66d01280ae722932d73f84d046aa712/frameworks/cassandra/src/test/java/com/mesosphere/sdk/cassandra/scheduler/ServiceTest.java#L28

frameworks/hdfs/src/main/dist/svc.yml
elezar: @nickbp  the template is still used jn Cassandra. Should this be the case? If not, I can create a PR tomorrow to remove it too.

nickbp: Oh interesting, would've expected that to not work. It's not being set by cassandra's marathon.json.mustache anymore... (yes it can be removed from cassandra)

frameworks/hdfs/docs/install.md
```diff
@@ -256,3 +256,29 @@ HDFS supports deployment on virtual networks on DC/OS (including the `dcos` over
...
+	}
+  }
+}
+```
+
+For more information about TLS in the SDK see [the TLS documentation](https://mesosphere.github.io/dcos-commons/developer-guide.html#tls).
+
+### Clients
+
+Clients connecting to HDFS over a TLS connection must connect to HTTPS specific port. Each node type (`journal`, `name` and `data`) can be configured with different port number for TLS connections.
```
benclarkwood: `connect to an HTTPS specific port`

benclarkwood: `can be configured with different port numbers`

frameworks/hdfs/tests/test_tls.py
```diff
@@ -0,0 +1,119 @@
...
+        raise error
+
+    yield
+
+    sdk_install.uninstall(PACKAGE_NAME, SERVICE_NAME)
+
+
+@pytest.mark.tls
+@pytest.mark.sanity
+@sdk_utils.dcos_1_10_or_higher
```
benclarkwood: These need the ee only pytest mark

frameworks/hdfs/src/main/dist/hdfs-site.xml
benclarkwood: This setting implies to me that HDFS can do HTTP, HTTP + HTTPS, or just HTTPS. Is that true? If true, I think we should support that as we've been doing so for the other services.

NimaVaziri: Yeah, so setting this to HTTPS_ONLY in case of enabled TLS seems to be correct, as not setting it defaults to HTTP. Having `HTTP_AND_HTTPS` seems ambiguously too generous, as we either have TLS, or we don't, in which case this setting (and the default) provide sufficient coverage.

NimaVaziri:
@mhrabovcin I just noticed we're not including `dfs.client.https.need-auth` in `hdfs-default.xml` - this is needed for 2-way SSL, right?

frameworks/hdfs/tests/test_tls.py
mhrabovcin: Added.

mhrabovcin:
@elezar thanks for feedback. I've removed http addresses when TLS is enabled.

mhrabovcin:
@NimaVaziri I think that enabling authentication is a separate story.

PR #1618, opened 2017-09-13
[INFINITY-2409] Get all log files

NimaVaziri:
Instead of just getting the first 10 log files (even in times of having fewer than 10 and handling a log's non-existence's exception), get all the log files present in a task's sandbox.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -83,8 +83,14 @@ def pytest_runtest_makereport(item, call):
 
 
 def get_rotating_task_log_lines(task_id: str, task_file: str):
-    rotated_filenames = [task_file,]
-    rotated_filenames.extend(['{}.{}'.format(task_file, i) for i in range(1, 10)])
+    result = subprocess.run(['dcos', 'task', 'ls', task_id], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if result.returncode:
+        log.error(result.stderr.decode())
+        return []
+
+    sandbox_files = result.stdout.decode().split()
+    rotated_file_names = [log_file_name for log_file_name in sandbox_files if task_file in log_file_name]
+
     for filename in rotated_filenames:
         lines = get_task_logs_for_id(task_id, filename)
         if not lines:


conftest.py
```diff
@@ -83,8 +83,14 @@ def pytest_runtest_makereport(item, call):
...
 def get_rotating_task_log_lines(task_id: str, task_file: str):
-    rotated_filenames = [task_file,]
-    rotated_filenames.extend(['{}.{}'.format(task_file, i) for i in range(1, 10)])
+    result = subprocess.run(['dcos', 'task', 'ls', task_id], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if result.returncode:
+        log.error(result.stderr.decode())
+        return []
+
+    sandbox_files = result.stdout.decode().split()
+    rotated_file_names = [log_file_name for log_file_name in sandbox_files if task_file in log_file_name]
```
jsrodman: I'm pretty sure we only have 10 log files.  Why this change?

NimaVaziri: Hmm I don't recall seeing more than 10 log files either though the mesos docs don't seem to indicate that 10 is a default number.

https://teamcity.mesosphere.io/viewLog.html?buildId=776397&buildTypeId=DcosIo_DcosCommons_Hdfs_BuildAndTestPr&tab=artifacts

The above build has schedulers which launched a while before the oldest respective recorded task logs, and I'm interested in the logs of the first few task log files recorded along the same time as the scheduler logs.

Either way, this change doesn't actually break existing functionality anyway. If there's another way to catch these missing logs that I've overlooked, I'm all ears.

jsrodman: logrotate.conf in our environment ships with a rotation value of 9, so we won't ever have more unless this is changed.

I asked because I believed you were making the change based on something you encountered, and because my knowledge of the system meant this shouldn't be necessary, I wondered what happened.  Contrarily, all change is risk, so I tend to make no changes unless they're needed.

This change will handle scenarios where the logrotate is adjusted or weird behavior occurs, though it could collect an infinite number of files which I'm not sure is a good idea (alllowing test cluster breakage to spread to teamcity breakage).  

----

If you want to make this change anyway, the test logic should probably be changed to

`... if log_file_name.startswith(task_file)`

conftest.py
```diff
@@ -83,8 +83,14 @@ def pytest_runtest_makereport(item, call):
 
 
 def get_rotating_task_log_lines(task_id: str, task_file: str):
-    rotated_filenames = [task_file,]
-    rotated_filenames.extend(['{}.{}'.format(task_file, i) for i in range(1, 10)])
+    result = subprocess.run(['dcos', 'task', 'ls', task_id], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if result.returncode:
+        log.error(result.stderr.decode())
```
mellenburg: This log statement could definitely use some context

conftest.py
mellenburg: This seems a little more prone to error but I like getting the exact set of current files. I think I'm cool with it

NimaVaziri:
On second thought, this happens rare enough to dissuade any change here.

PR #1619, opened 2017-09-13
[INFINITY-2408] Fix HDFS test_sanity - tasks updated detection

paegun:
* Fix HDFS test_sanity.test_modify_app_config, send proper task
  prefixes.
* Improve tasks updated detection, not yet reconciled log line to
  use set methods to report the intersection and difference instead
  of iteration.

Unrelated, but discovered:
* Fix .gitignore for Kafka to match pattern used in other services.

### DIFF
--- a/frameworks/hdfs/tests/test_sanity.py
+++ b/frameworks/hdfs/tests/test_sanity.py
@@ -284,6 +284,7 @@ def test_modify_app_config():
     app_config_field = 'TASKCFG_ALL_CLIENT_READ_SHORTCIRCUIT_STREAMS_CACHE_SIZE_EXPIRY_MS'
     journal_ids = sdk_tasks.get_task_ids(foldered_name, 'journal')
     name_ids = sdk_tasks.get_task_ids(foldered_name, 'name')
+    data_ids = sdk_tasks.get_task_ids(foldered_name, 'data')
 
     marathon_config = sdk_marathon.get_config(foldered_name)
     log.info('marathon config: ')
@@ -296,7 +297,7 @@ def test_modify_app_config():
     config.check_healthy(service_name=foldered_name)
     sdk_tasks.check_tasks_updated(foldered_name, 'journal', journal_ids)
     sdk_tasks.check_tasks_updated(foldered_name, 'name', name_ids)
-    sdk_tasks.check_tasks_updated(foldered_name, 'data', journal_ids)
+    sdk_tasks.check_tasks_updated(foldered_name, 'data', data_ids)
 
     sdk_plan.wait_for_completed_recovery(foldered_name)
     new_recovery_plan = sdk_plan.get_plan(foldered_name, "recovery")
--- a/testing/sdk_tasks.py
+++ b/testing/sdk_tasks.py
@@ -47,6 +47,9 @@ def get_task_ids(service_name, task_prefix):
 
 
 def check_tasks_updated(service_name, prefix, old_task_ids, timeout_seconds=DEFAULT_TIMEOUT_SECONDS):
+    # TODO: strongly consider merging the use of checking that tasks have been replaced (this method)
+    # and checking that the deploy/upgrade/repair plan has completed. Each serves a part in the bigger
+    # atomic test, that the plan completed properly where properly includes that no old tasks remain.
     def fn():
         try:
             task_ids = get_task_ids(service_name, prefix)
@@ -66,7 +69,7 @@ def check_tasks_updated(service_name, prefix, old_task_ids, timeout_seconds=DEFA
         # deploy/recovery/whatever plan, not task cardinality, but some uses of this method are not
         # using the plan, so not the definitive source, so will fail when the finished state of a
         # plan yields more or less tasks per pod.
-        all_updated = len(newly_launched_set) == len(new_set) and len(old_remaining_set) == 0 and len(old_set) == len(new_set)
+        all_updated = len(newly_launched_set) == len(new_set) and len(old_remaining_set) == 0 and len(new_set) >= len(old_set)
         if all_updated:
             log.info('All of the tasks{} have updated\n- Old tasks: {}\n- New tasks: {}'.format(
                 prefix_clause,
--- a/testing/sdk_tasks.py
+++ b/testing/sdk_tasks.py
@@ -57,16 +57,21 @@ def check_tasks_updated(service_name, prefix, old_task_ids, timeout_seconds=DEFA
         prefix_clause = ''
         if prefix:
             prefix_clause = ' starting with "{}"'.format(prefix)
-        log.info('Waiting for {} tasks{} to have updated ids:\n- Old tasks: {}\n- Current tasks: {}'.format(
-            len(old_task_ids), prefix_clause, sorted(old_task_ids), sorted(task_ids)))
 
-        all_updated = True
-        for id in task_ids:
-            if id in old_task_ids:
-                all_updated = False
-        if len(task_ids) < len(old_task_ids):
-            all_updated = False
-        return all_updated
+        old_set = set(old_task_ids)
+        new_set = set(task_ids)
+        newly_launched_set = (new_set - old_set)
+        all_updated = len(newly_launched_set) == len(new_set)
+        if all_updated:
+            return all_updated
+
+        # forgive the language a bit, but len('remained') == len('launched'),
+        # and similar for the rest of the label for task ids in the log line,
+        # so makes for easier reading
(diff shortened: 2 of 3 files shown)


testing/sdk_tasks.py
```diff
@@ -57,16 +57,21 @@ def fn():
...
-            if id in old_task_ids:
-                all_updated = False
-        if len(task_ids) < len(old_task_ids):
-            all_updated = False
-        return all_updated
+
+        old_set = set(old_task_ids)
+        new_set = set(task_ids)
+        newly_launched_set = (new_set - old_set)
+        all_updated = len(newly_launched_set) == len(new_set)
```
jsrodman: This is validating that there are all new tasks, and that there are the same number?
Minor preference for something like

```python
task_overlap = new_set.intersection(old_set)
if len(new_set) == len(old_set) and len(task_overlap) == 0:
    return True
```

If that's the set of conditions that are being validated.

testing/sdk_tasks.py
```diff
@@ -57,16 +57,21 @@ def fn():
...
+        if all_updated:
+            return all_updated
+
+        # forgive the language a bit, but len('remained') == len('launched'),
+        # and similar for the rest of the label for task ids in the log line,
+        # so makes for easier reading
+        log.info('Waiting for tasks{} to have updated ids:\n- Old tasks (remained): {}\n- New tasks (launched): {}'.format(
+            prefix_clause,
+            old_set & new_set,
```
jsrodman: I find old_set.intersection(new_set) much clearer, but ignore if you don't.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -296,7 +297,7 @@ def test_modify_app_config():
     config.check_healthy(service_name=foldered_name)
     sdk_tasks.check_tasks_updated(foldered_name, 'journal', journal_ids)
     sdk_tasks.check_tasks_updated(foldered_name, 'name', name_ids)
-    sdk_tasks.check_tasks_updated(foldered_name, 'data', journal_ids)
+    sdk_tasks.check_tasks_updated(foldered_name, 'data', data_ids)
```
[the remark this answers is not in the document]
paegun: I don't believe it is the job here in PR review to perform a git spelunking expedition. I looked back:
1. one commit and this line changed, but already had incorrectly sent in `journal_ids`.
1. two commits and this line wasn't in the delta.
1. three commits and this line changed, but already had incorrectly sent in `journal_ids`.
1. four commits and this line wasn't in the delta.
1. five...not in delta.
1. six...not in delta.
...

Point being, it's been long enough wrong AND the line touched more than once, so should've been corrected long ago. It's not encouraging to put the spelunking expedition on the person who fixed the problem.

testing/sdk_tasks.py
paegun: 1. Set operations via the set symbols should be clear.
  * new_set - old_set yields the difference, so newly_launched_set.
2. `newly_launched_set` should provide a clear intention of the result.

I believe (rather strongly) that such a name is much clearer than:
1. using a method call versus using the set symbols is possibly easier to read, but then we should also be using 2.plus(3) to yield 5.
2. `task_overlap` is an abstract name, so loses meaning as the value is used later in the method.

testing/sdk_tasks.py
paegun: See above, set math should be known. To further argue the point, Python didn't provide `&&` for something so commonly used `if A && B:`, but did provide these set symbols.

frameworks/hdfs/tests/test_sanity.py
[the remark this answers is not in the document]
jsrodman: I wasn't trying to imply that answering this was your responsibility.  I was curious how long we'd left it broken and thought this might already be known. It was an attempt to learn.

testing/sdk_tasks.py
jsrodman: I don't agree that the symbolic set operators are well-known.  set1 - set2 is clear enough.  Set1 & set2 is not, imo.

Regardless, this commentary was about making explicit the idea that we're checking for size equality and non-overlap, if that's what we're doing.  It wasn't about the name.

testing/sdk_tasks.py
jsrodman: I do not know what `old_set & new_set` means.  I had to look it up three times while reading this code, because I repeatedly forgot.   If you feel my knowledge is insufficient, then that's OK, but I didn't provide this feedback just to pick nits.

testing/sdk_tasks.py
elezar: @paegun the check for the lengths above is not sufficient to ensure that all tasks have been updated. We would also need to check that the size of the new set is the same as the old set. An empty new set, for example, would pass the check and return `all_updated`.

paegun: True, I'll add a check to ensure that the old tasks are all gone.

testing/sdk_tasks.py
paegun: Understood and will change. I don't set the bar, and your feedback is greatly appreciated. If set math isn't clear, set methods will be used.

testing/sdk_tasks.py
```diff
@@ -60,8 +60,9 @@ def fn():
 
         old_set = set(old_task_ids)
         new_set = set(task_ids)
-        newly_launched_set = (new_set - old_set)
-        all_updated = len(newly_launched_set) == len(new_set)
+        newly_launched_set = new_set.difference(old_set)
+        old_remaining_set = old_set.intersection(new_set)
+        all_updated = len(newly_launched_set) == len(new_set) and len(old_remaining_set) == 0
```
elezar: Should we not also check that the new set has the same length as the old set? If we had (t1, t2) and then launch (t3) these checks will pass?

Would:

 len(new_set) == len(old_set) and len(old_remaining_set)

Not be the correct condition?

Sorry for the formatting, I'm on my phone.

paegun: @elezar the condition implied by all_updated is that there are no old tasks present, not that the new set matches the old set. To translate the condition as all old tasks have a corresponding new task would fail if a plan launched multiple tasks where previously only a single task was launched (or the inverse).

The follow on condition check that the plan completed covers the overarching concern, that the plan completed.

elezar: OK. Thanks for the clarification. I understood it differently, and think the original implementation wasn't implemented this way either:
```python
        all_updated = True
        for id in task_ids:
            if id in old_task_ids:
                all_updated = False
        if len(task_ids) < len(old_task_ids):
            all_updated = False
        return all_updated
```
I'm not saying that this was correct for HDFS, but changing this does break the hello-world tests.

This may mean that we need two checks:
- `tasks_replaced`: Checks for 1-1 replacement of tasks
- `tasks_removed_or_updated`: Checks that the intersection of the new task list and the old task list is zero.

For what it's worth, it may be useful move the inner function to the module level so that we could exercise it with a couple of unit tests, but that may be out of scope of this PR.

testing/sdk_tasks.py
```diff
@@ -57,16 +57,22 @@ def fn():
...
+        if all_updated:
+            return all_updated
+
+        # forgive the language a bit, but len('remained') == len('launched'),
+        # and similar for the rest of the label for task ids in the log line,
+        # so makes for easier reading
+        log.info('Waiting for tasks{} to have updated ids:\n- Old tasks (remained): {}\n- New tasks (launched): {}'.format(
```
elezar: As it is now, we no longer log the success condition (since we return before this is the case).

paegun: True. The success condition was previously logged as 'Waiting for ...', so prone to be removed as a "success". That said, I removed it intentionally.

We are not consistently logging to feed a state machine like observer that requires Enter/Exit <State> messages AND if we do, we should feed it better.

The logging here and throughout should be consistently providing either:
* Context that is useful if the system under test (or operation) were to crash or incorrectly operate.
* Context for a long-running process that may be under (typically human, most typically developer, but more importantly operator) observation.

Since this is test code logging, the operator is the developer so generally these reasons fold, yielding log in tests what provides feedback that the system isn't stalled or otherwise operating in a way that will lead to a sure fail.

paegun: Fortunately, you were right @elezar . The method of expecting the resulting (new) task set to have an equal number of tasks to the previous (old) task set was actually being used (in one stray case, a check of placement constraints updated. Ideally, the tests would use a single method to ensure that the expected plan executed and completed, cycling the tasks in the process, but the tests are not at this point, so...re-instated the constraint.

PR #1620, opened 2017-09-14
Implement multi-disk sending support through the framework

adragomir:
Ping @triclambert, @gabrielhartmann 

Details on what this does are here: https://jira.mesosphere.com/projects/DCOS_HDFS/issues/DCOS_HDFS-4

Basically, what this work does is start implementation on supporting PATH volumes. The path volumes are not atomic in the sense that offers can be shared between frameworks. So, I added code that can handle this splitting of PATH volumes between frameworks, as well as some tests. 

The ugliest part is actually sending the needed resources through to the framework, since we only have environment variables, and actually using this array of objects inside the Mustache template. . Another way would be to encode a json or something in an environment variable value. 

This isn't by any means final, but let me know if this is a direction that would be acceptable, or any other hints on how to proceed. 

This patch has been tested with both the HDFS and Kafka frameworks (some more changes in the service yaml files needed, I can to a PR for those as well. Please see https://github.com/adragomir/dcos-commons

### DIFF
--- a/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
+++ b/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
@@ -42,7 +42,7 @@ public class Main {
                                             .id("hello-resources")
                                             .cpus(CPUS)
                                             .memory(256.0)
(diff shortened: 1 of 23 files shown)


kaiwalyajoshi:
This is a 3 year old PR, closing this as it is stale. Feel free to resurrect if there's still interest in this.

PR #1622, opened 2017-09-14
INFINITY-2418 Fix plan examples in dev guide

nickbp:
Checked all instances of `plans:` and cleaned up the ones that were out of date. Validated against yaml reference as well as relevant `Raw*.java` (yaml reference appears up to date)


elezar:
Should this be backported too?

nickbp:
Guess so! Will make another PR.

PR #1623, opened 2017-09-14
INFINITY-2399 INFINITY-2429 BACKPORT: Go vendoring fix + Cassandra config generation fix

nickbp:
- Backport of cassandra svc.yml fix which was included in #1426
- Backport of go vendoring fix to fix CI (see master version: #1630)


frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -164,6 +165,10 @@ pods:
 
                 rm -r container-path/snapshot
         resource-set: sidecar-resources
+        configs:
+          cassandra:
+            template: {{CONFIG_TEMPLATE_PATH}}/cassandra.yaml
```
elezar: To be clear, is `{{CONFIG_TEMPLATE_PATH}}` still required in `0.30.X`?

elezar: (just checked the code, and it's still used there).

For reference, this was removed in #1539.

nickbp:
All the builds are failing against CNI tests. Not seeing similar failures in master, so it appears to be something specific to 0.30.x. Another possibility is something relating to CNI happened to be broken as of when CI ran. Kicking builds to see whether that's the case..

nickbp:
The above overlay failures were actually tied to the CLI itself. It was doing this:

```
$ dcos kafka -v endpoints
Running DC/OS CLI command: dcos config show core.dcos_url
HTTP Query: GET http://nick-lv7v-elasticl-1odsognrwak9c-473658943.us-west-2.elb.amazonaws.com/service/v1/endpoints
Running DC/OS CLI command: dcos config show core.dcos_acs_token
Running DC/OS CLI command: dcos config show core.ssl_verify
Response: 404 Not Found (-1 bytes)
Could not reach the service scheduler with name ''.
Did you provide the correct service name? Specify a different name with '--name=<name>'.
Was the service recently installed or updated? It may still be initializing, wait a bit and try again.
```

Turns out a recent change in kingpin caused the problem, and our CLI vendoring wasn't doing anything to fix it. Lets see if this will fix the vendoring...

PR #1625, opened 2017-09-15
Remove CONFIG_TEMPLATE_PATH from cassandra

elezar:
Following #1539 and also https://github.com/mesosphere/dcos-commons/pull/1617#discussion_r138995536

I have removed `CONFIG_TEMPLATE_PATH` from the Cassandra service spec.


nickbp:
The change only touches cassandra, and cassandra is green.

PR #1626, opened 2017-09-15
[KAF-368] Make Kafka integration tests compatible with Confluent

elezar:
This PR makes minor changes to the Kafka integration tests to make it simpler to copy these to Confluent (see mesosphere/confluent#59)

The biggest change was around checking for created topics, where I added a function to filter out topics starting with `_` so as to gracefully handle the `_confluent-metrics` topic.

This also contains changes to `test.sh` and `test-runner.sh` to allow these to be used from the CI system.


benclarkwood:
[https://github.com/mesosphere/dcos-commons/pull/1626](https://trello.com/c/XS3BEVud/33-https-githubcom-mesosphere-dcos-commons-pull-1626)

test.sh
```diff
@@ -147,20 +150,34 @@ if [ "$framework" = "all" -a -n "$STUB_UNIVERSE_URL" ]; then
...
+    -e PYTEST_ARGS="$PYTEST_ARGS" \
     -e FRAMEWORK=$framework \
     -e STUB_UNIVERSE_URL="$STUB_UNIVERSE_URL" \
     -v $(pwd):/build \
     -v $ssh_path:/ssh/key \
     -w /build \
     -t \
     -i \
-    mesosphere/dcos-commons:latest \
+    mesosphere/dcos-commons:dev \
```
benclarkwood: Shouldn't we be using :latest? @mellenburg

elezar:
You're right that latest is probably beter. I made this change because we use :dev in TeamCity and I was striving for convergence. Will change it back because we should probably converge in the other direction anyway...

benclarkwood:
Cool. I'm good with this after we revert it to `:latest`.

elezar:
Thanks @benclarkwood. I have switched back to `:latest` and also created a PR to do the same in mesosphere/confluent#63. If i could get a review there too, that would be great.

PR #1628, opened 2017-09-15
[INFINITY-2299][HDFS] Automate replace of a journal node

NimaVaziri:
This makes sure that once a journal node is taken down (via a crash/replace such that its persistence volume is erased) it catches up properly with the other journal nodes and can talk to the name node instead of erring on a format exception.

Steps to ensure functionality:
 - deploy an HDFS instance
 - call `replace` on a journal pod
 - upon recovery, a `bootstrap` task runs, then the journal node starts as usual


frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -34,6 +29,20 @@ pods:
           path: journal-data
           size: {{JOURNAL_DISK}}
           type: {{JOURNAL_DISK_TYPE}}
+    tasks:
+      bootstrap:
+        goal: FINISHED
+        cmd: >
+          mkdir -p journal-data/hdfs/current &&
```
benclarkwood: Can you please wrap this with echoes so it's easy to tell from logs if any given piece failed?

Also, echo the content of the api response.

e.g. 

`echo "creating version directory" && ... && echo "fetching version file" && ... && echo "version file has content $CONTENT" && ...`

benclarkwood: If you want to do this as a follow on PR, that's a-ok with me (since that will just run HDFS)

NimaVaziri: Done

PR #1629, opened 2017-09-15
Backport INFINITY-2418 Dev guide plans fix

nickbp:
Backport of #1622

### DIFF
--- a/docs/pages/developer-guide.md
+++ b/docs/pages/developer-guide.md
@@ -38,7 +38,7 @@ Every DC/OS service must provide a package definition in the format expected by 
 
 ## ZooKeeper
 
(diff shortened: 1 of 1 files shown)


docs/pages/developer-guide.md
```diff
@@ -1550,7 +1558,7 @@ pods:
         memory: 256
         configs:
           config.xml:
-            template: "config.xml.mustache"
```
elezar: Recently there was the change w.r.t `{{CONFIG_TEMPLATE_PATH}}` (which was not removed in `0.30.X`) is the path here consistent?

nickbp: Including it in this documentation would require also explaining that CONFIG_TEMPLATE_PATH should be manually configured in marathon.json... At this point of the doc it'd be a bit overkill.

TBH I don't expect anyone to be using this version of the dev guide anyway. Ideally they'd be against the most recent version available.

nickbp:
Merging -- this is a docs fix and additionally the CI failures seen here were fixed in 0.30.x as of #1623

PR #1630, opened 2017-09-15
INFINITY-2429 Fix Go vendoring and use vendoring consistently

nickbp:
Go is confused by directories that reference other directories in the same repo, particularly when external repos are being pulled in as well. Adds a shared `govendor` dir to be symlinked by CLIs/bootstrap in dcos-commons. Replaces O(Nframeworks) vendor copies that are currently present. External repos wouldn't need to jump through these hoops.

Removes a `go get` command from `build_go_exe.sh`, as all libraries should be included in the `govendor` directory anyway. And this effectively enforces that libraries are vendored to avoid spontaneous build-time issues.


clivendor/README.md
```diff
@@ -0,0 +1,7 @@
+# CLI Vendor hack
+
+This is a workaround for the go tooling being confused by `frameworks/myframework/cli/` pointing into `cli/` when using vendoring.
+
+Frameworks in dcos-commons/frameworks/ should create a 'vendor' symlink which points to here. In other words, a symlink at `frameworks/yoursvc/cli/dcos-yoursvc/vendor` should point to `clivendor/`.
```
elezar: Should we add this to the `new-framework.sh` script?

clivendor/README.md
```diff
@@ -0,0 +1,7 @@
+# CLI Vendor hack
+
+This is a workaround for the go tooling being confused by `frameworks/myframework/cli/` pointing into `cli/` when using vendoring.
```
elezar: replace: "go" -> "Go"?

clivendor/vendor.json
```diff
@@ -21,6 +21,18 @@
...
 		},
 		{
+			"checksumSHA1": "ulFwBsxjpyu+bk44+LKDl8AKQiA=",
+			"path": "github.com/aryann/difflib",
+			"revision": "e206f873d14a916d3d26c40ab667bca123f365a3",
+			"revisionTime": "2017-07-10T04:42:30Z"
+		},
+		{
+			"checksumSHA1": "AIxWUvzOjHni8IwkElJw+b1IES8=",
+			"path": "github.com/nickbp/mustache",
```
elezar: Do the CLIs actually use `mustache` or `difflib`? Or is this vendor directory now more general than just CLI?

clivendor/README.md
nickbp: Open question: delete that script as of this PR?

clivendor/vendor.json
nickbp: This is the union of both CLI and bootstrap dependencies

elezar: Should we then change the folder name `clivendor` to `vendor` or `go-vendor`?

clivendor/README.md
elezar: What is the documented way to add a new framework and/or create a new standalone service. But yes, the scripts that are not relevant anymore should be removed.

clivendor/vendor.json
nickbp: I now realize that the name `clivendor` isn't quite accurate, I'll update this to be `govendor`, even though that's technically also the name of the tool used to manage vendoring.

nickbp: Actually I'll try `govendor` just to follow the example of e.g. `frameworks/helloworld` of avoiding punctuation

paegun:
infinity/hdfs build failure `test_sanity.test_modify_app_config` highly relates to https://github.com/mesosphere/dcos-commons/pull/1619

elezar:
Just looked at https://github.com/mesosphere/dcos-commons/blob/a39ce51bdff6d8e3ebf66968e7d13f078306c773/cli/README.md and it seems `govendor` is a go package/utility.

nickbp:
Missing approval

PR #1639, opened 2017-09-19
Fix unintended while(true) loop in plan execution

gabrielhartmann:


nickbp:
PS I have some changes to this handling in (just opened) PR #1640 which may resolve this issue anyway

gabrielhartmann:
@nickbp:  Should have an effect on re-register, when we go through reconciliation again.

PR #1640, opened 2017-09-19
INFINITY-2185 Java API cleanup: Part 1 (doesn't actually change APIs yet..)

nickbp:
PR is a subset of a larger change. For this PR, we're focusing on the internals of AbstractScheduler and UninstallScheduler/DefaultScheduler. The highlights are:

- The schedulers are now able to start the API server in the background, and then start accepting offers when it has finished started. This should lead to faster startup times.
- The Mesos "Scheduler" interface is now only implemented by a subclass of our AbstractScheduler. This decouples us from the Mesos interfaces, and removes prior duplicate logic across UninstallScheduler and DefaultScheduler.


gabrielhartmann:
@nickbp:  Wasn't the API scheduler always started in the background?

nickbp:
As of current `master`, the API server startup is only triggered after we get the registered callback (and the SchedulerDriver). As of this PR the API server startup itself is triggered around the same time as the framework register call, and then when we get SchedulerDriver back via the callback, it's passed to the API server's resources that use it.

nickbp:
Yes. Main changes will be:
- DefaultScheduler.Builder will be broken out into a separate class
- DefaultService construction will make more sense
- Some of the startup cross-dependencies will be cleaned up

In the process of getting the above (particularly the last item) I ended up doing a bunch of internal cleanup as in this PR.

PR #1632, opened 2017-09-18
[INFINITY-2353] Use dcos-cni package w/i bootstrap for ip.

paegun:
* Remove local implementatino of ip detection interface into bootstrap.
* Add package import, vendoring, and use of dcos-cni for ip detection
  interface into bootstrap.

Related:
1. https://github.com/mesosphere/dcos-commons/pull/1630
  * This PR contains vendoring of bootstrap's (golang) imported packages.
  * #1630 provides a general solution to golang subproject's imported packages.
  * To expedite this PR (not entangle), the current norm for vendoring w/i dcos-commons golang projects was used.

PR #1633, opened 2017-09-18
Rename 'clivendor' to 'govendor' in 0.30.X

nickbp:
Gets 0.30.X in sync with recent changes to PR #1630. The directory is also being used by bootstrap so 'clivendor' isn't quite accurate.

PR #1644, opened 2017-09-19
[INFINITY-2353] Use dcos-cni package w/i bootstrap for ip.

paegun:
* Backport #1632

* Remove local implementatino of ip detection interface into bootstrap.
* Add package import, vendoring, and use of dcos-cni for ip detection
  interface into bootstrap.

* Related:

1. #1630
  * This PR contains vendoring of bootstrap's (golang) imported packages.
  * #1630 provides a general solution to golang subproject's imported
    packages.
  * To expedite this PR (not entangle), the current norm for vendoring w/i
    dcos-commons golang projects was used.


nickbp:
Should be `govendor` instead of `clivendor`. See #1633 which fixes this for 0.30.X

PR #1645, opened 2017-09-19
[INFINITY-2382] Revive throttling

gabrielhartmann:
There are two major portions to this work.
1. Add a revive rate limiting mechanism.  See the addition of a [TokenBucket](https://github.com/mesosphere/dcos-commons/pull/1645/files#diff-73dd3107d636c34f68f190b11655d582R15)
1. Focus the PlanCoordinator on providing work candidates and [move the executor of that work (PlanScheduler) outside of the PlanCoordinator](https://github.com/mesosphere/dcos-commons/pull/1645/files#diff-73dd3107d636c34f68f190b11655d582R15).

There are two minor changes in this work.
1. There has been a refactoring of the ReviveManager as a consequence of the work above.  It no longer operates on its own polling thread but is [passed in the current work candidates from the main execution loop](https://github.com/mesosphere/dcos-commons/pull/1645/files#diff-0f50d17566fab38af4dee3c8360db86cR135).
1. All persistence of a suppressed status has been removed, as Schedulers never suppress offers now.

Much of this work is motivated by having a single source of truth for the current work being executed by the system.  The [work is determined here](https://github.com/mesosphere/dcos-commons/pull/1645/files#diff-0f50d17566fab38af4dee3c8360db86cR132).

I could break this up into a few different PRs, but I find all these topics related.  If anyone feels strongly that smaller PRs are desired, I'm happy to break this up abit.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TokenBucket.java
```diff
@@ -0,0 +1,148 @@
...
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
+    /**
+     * A TokenBucket acts as a rate limiting helper.  Clients should not perform rate limited work without acquiring a
+     * token from the {@link #tryAcquire()} method.
+     * @param initial The initial number of tokens available
+     * @param capacity The maximum number of tokens the bucket may hold
+     * @param incrementInterval The incrementInterval between adding a new token to the bucket
```
nickbp: Missing `minAcquireInterval`

Alternately feel free to just remove the param docs here, as the params are effectively being documented by the builder, arguably making these duplicates

Also may as well rename minAcquireInterval => acquireInterval to match naming elsewhere

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TokenBucket.java
```diff
@@ -0,0 +1,148 @@
...
+            count--;
+            lastRevive = System.currentTimeMillis();
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * This method adds a token to the bucket.
```
nickbp: Maybe mention that it does nothing if the token capacity has been reached

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/TokenBucketTest.java
```diff
@@ -0,0 +1,120 @@
...
+        Duration incrementInterval = Duration.ofMillis(100);
+        TokenBucket bucket = TokenBucket.newBuilder()
+                .initialTokenCount(1)
+                .acquireInterval(Duration.ZERO)
+                .incrementInterval(incrementInterval)
+                .build();
+
+        Assert.assertTrue(bucket.tryAcquire());
+        Assert.assertFalse(bucket.tryAcquire());
+        Thread.sleep(incrementInterval.toMillis() + 50); // Give 50ms of breathing room to the token incrementing thread
```
nickbp: Passage of time should be simulated by doing the following:
- Adding a `protected now()` function to `TokenBucket` which returns `System.currentTimeMillis()`
- Updating all internal calls to `System.currentTimeMillis()` in `TokenBucket` with calls to `now()`
- In these tests, have a `TestTokenBucket` implementation which overrides `now()`, and has a function to allow incrementing the value to be returned by `now()`.

nickbp:
FYI looks like `test_sanity.test_state_properties_get` in hello world needs an update to reflect the removal of the suppressed bit.

gabrielhartmann:
@nickbp:  Addressed comments.  We spoke offline regarding the `now()` method being undesirable in the case of bucket replenishing, but otherwise suggestions have been followed.

benclarkwood:
@gabrielhartmann what was the content of said discussion? I see a `now()` method, but no callers of said method.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/ReviveManager.java
```diff
@@ -1,81 +1,37 @@
...
-                pollDelay,
-                pollInterval,
-                TimeUnit.SECONDS);
-
-        logger.info(
-                "Monitoring these plans for suppress/revive: {}",
-                planCoordinator.getPlanManagers().stream()
-                        .map(planManager -> planManager.getPlan().getName())
-                        .collect(Collectors.toList()));
+    public ReviveManager(SchedulerDriver driver) {
```
benclarkwood: IMO, ReviveManager should take the TokenBuilder as a dependency, that would allow adding tests of the critical switch around token fulfillment.

benclarkwood: Alternately, you could just have a second constructor that takes it as a parameter, then only use that directly in tests (e.g. make it protected).

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/ReviveManager.java
```diff
@@ -87,44 +43,92 @@ public void run() {
...
+        Set<WorkItem> currCandidates = getCandidates(steps);
+        Set<WorkItem> newCandidates = new HashSet<>(currCandidates);
         newCandidates.removeAll(candidates);
 
-        logger.debug("Old candidates: {}", candidates);
-        logger.debug("New candidates: {}", newCandidates);
+        logger.info("Candidates, old: {}, current: {}, new:{}", candidates, currCandidates, newCandidates);
+
+        if (!newCandidates.isEmpty()) {
+            if (tokenBucket.tryAcquire()) {
```
benclarkwood: As I mentioned above, this code path isn't fully exercised by the tests (or at least, I don't see one that does since the defaults of TokenBucket are so high).

I definitely think the code is 100% correct, but I'd feel more comfortable with a test.

gabrielhartmann:
@benclarkwood: I'm confused by your comments.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/ReviveManager.java
gabrielhartmann: There is a second constructor immediately below this comment.  Am I missing something?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/ReviveManager.java
gabrielhartmann: This code path is exercised by several tests.  Namely every test that successfully acquires a token.

gabrielhartmann: Also it's tested extensively by the TokenBucket tests.

gabrielhartmann: Also, ReviveManager had total test coverage with the original ReviveManager test except for one line.  I've added a test so that line is now exercised.

PR #1646, opened 2017-09-19
[INFINITY-2408] Fix HDFS test_sanity - tasks updated detection

paegun:
* Backport of #1619.

* Fix HDFS test_sanity.test_modify_app_config, send proper task
  prefixes.
* Improve tasks updated detection, not yet reconciled log line to
  use set methods to report the intersection and difference instead
  of iteration.

Unrelated, but discovered:

* Fix .gitignore for Kafka to match pattern used in other services.

### DIFF
--- a/testing/sdk_tasks.py
+++ b/testing/sdk_tasks.py
@@ -79,8 +79,6 @@ def check_tasks_updated(service_name, prefix, old_task_ids, timeout_seconds=DEFA
             old_remaining_set,
             newly_launched_set))
 
-        log.info('Waiting for tasks{} to have updated ids:\n- Old tasks (remained): {}\n- New tasks (launched): {}'.format(
-
 
 def check_tasks_not_updated(service_name, prefix, old_task_ids):
     sdk_plan.wait_for_completed_deployment(service_name)
--- a/testing/sdk_tasks.py
+++ b/testing/sdk_tasks.py
@@ -41,6 +41,9 @@ def get_task_ids(service_name, task_prefix):
 
 
 def check_tasks_updated(service_name, prefix, old_task_ids, timeout_seconds=DEFAULT_TIMEOUT_SECONDS):
+    # TODO: strongly consider merging the use of checking that tasks have been replaced (this method)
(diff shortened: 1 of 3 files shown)


paegun:
This PR is running into unrelated flaky tests and there is no baseline proof that this PR has caused the flakes to arise.

testing/sdk_tasks.py
```diff
@@ -41,26 +41,46 @@ def get_task_ids(service_name, task_prefix):
...
+
+        old_set = set(old_task_ids)
+        new_set = set(task_ids)
+        newly_launched_set = new_set.difference(old_set)
+        old_remaining_set = old_set.intersection(new_set)
+        # the constrainst of old and new task cardinality match should be covered by completion of
+        # deploy/recovery/whatever plan, not task cardinality, but some uses of this method are not
+        # using the plan, so not the definitive source, so will fail when the finished state of a
+        # plan yields more or less tasks per pod.9
+        all_updated = len(newly_launched_set) == len(new_set) and len(old_remaining_set) == 0 and len(new_set) >= len(old_set)
```
NimaVaziri: Any edge case that I'm failing to see where `len(new_set) > len(old_set)` would be true? If the size of the sets do in fact differ shouldn't that be cause for failure since that's not what we expect?

paegun: See #1619 for fuller context of this wonkiness due to the overly large set of overloaded uses of this method.

The above was originally a strict definition in the set sense, but the consensus on review of #1619 is/was that the check here is now a good semantic check based on the SDK's meaning for all tasks updated.

I strongly recommend making this method private to sdk_plan as the "all" really means all of the tasks that should have been updated as the result of an action where the action is the execution of a plan. Encapsulating this properly has the positive effect of ensuring that the plan is checked for completion, which includes passing the readiness check, not only that the tasks have cycled.

NimaVaziri: Looks like this change is already in master, so perhaps a rebase is in order before we merge?

NimaVaziri: Oh never mind, I see.

PR #1649, opened 2017-09-20
Use real ServiceSpec objects for testing

gabrielhartmann:
This is at the root of my `.equals()` pain.


sdk/scheduler/src/test/java/com/mesosphere/sdk/testutils/PodTestUtils.java
```diff
@@ -1,202 +1,58 @@
...
 import com.mesosphere.sdk.scheduler.plan.PodInstanceRequirement;
 import com.mesosphere.sdk.specification.*;
-import com.mesosphere.sdk.specification.util.RLimit;
 import org.apache.mesos.Protos;
 
-import java.net.URI;
 import java.util.*;
 
 /**
- * Created by gabriel on 9/7/17.
```
benclarkwood: d'aww

PR #1650, opened 2017-09-20
x-pack link

sascala:


sascala:
@loren We're moving to a new docs site, so I'm waiting on making changes like that until I know what the structure will be.

PR #1653, opened 2017-09-20
[INFINITY-2443] Made the driver port dynamic

susanxhuynh:
- Made the driver port configurable in config.json, with default of 0
- The coordinator driver task reserves this port and uses it as the spark.driver.port
- The executor tasks uses this port in the `driver-url`


susanxhuynh:
This needs more work

PR #1654, opened 2017-09-20
[INFINITY-2422] Squelch common lib test coverage

paegun:
* Squelch common lib test coverage.
  * Benefit of common lib test coverage is low, coverage is including
    lines that cannot cause issue such as setting a const used as a key.
* Fix unrelated golang sub-project setup, closing out the great
  re-vendoring.

### DIFF
--- a/sdk/common/build.gradle
+++ b/sdk/common/build.gradle
@@ -96,12 +96,13 @@ dependencies {
     testCompile "org.awaitility:awaitility:${awaitilityVer}"
 }
 
-check.finalizedBy jacocoTestReport
-
-jacocoTestReport {
+// squelch code coverage report for this project
+// check.finalizedBy jacocoTestReport
+/*jacocoTestReport {
     reports {
         xml.enabled false
         csv.enabled false
         html.destination "${buildDir}/jacocoHtml"
     }
 }
+*/
--- a/build.sh
+++ b/build.sh
@@ -40,8 +40,13 @@ if [ x$PULLREQUEST = "xtrue" ]; then
     git pull origin $MERGE_FROM --no-commit --ff
 fi
 
-# Verify SDK CLI, run unit tests
-go test ./...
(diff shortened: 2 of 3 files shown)


sdk/common/build.gradle
```diff
@@ -96,12 +96,13 @@ dependencies {
...
+// squelch code coverage report for this project
+// check.finalizedBy jacocoTestReport
+/*jacocoTestReport {
     reports {
         xml.enabled false
         csv.enabled false
         html.destination "${buildDir}/jacocoHtml"
     }
 }
+*/
```
jsrodman: Are we going to re-enable this, later?

jsrodman:
Incidentally, not currently in a position to review the "Add symlink from GOPATH for the vendor deps that aren't under our repo org" section.

sdk/common/build.gradle
paegun: I considered either removing the coverage report configuration or commenting it out. I intentionally chose to clearly indicate that we have turned off the coverage report for this project. This is to stave off rework when someone re-enables the coverage report due to being unaware of the team's (contributor set's) consensus opinion that the coverage for this specific project should be squelched. Otherwise said, yes, project knowledge could be gleaned from other sources, but in this case the likelihood that the knowledge would be gleaned is low and the cost of repeating this work is high enough to merit IMO commenting here.

benclarkwood: Nick has a JIRA to see if it's worth adding test coverage to `common`. If it is, he'll reenable this.

benclarkwood: cc @nickbp

paegun:
This is not necessary to backport due to:
* the stable branch does not exhibit the `./build.sh` breakage that was on `master`.
* the feature of squelching the `common` test coverage is not a necessary feature for release.

PR #1659, opened 2017-09-21
Revert "Exit on restore-snapshot failure in Cassandra"

sascala:
Reverts mesosphere/dcos-commons#1658


sascala:
@ArtRand @gabrielhartmann Sorry, there, I merged before I saw that Art was also requested as a reviewer.

PR #1660, opened 2017-09-21
[CASSANDRA-647, CASSANDRA-648] Exit on restore-snapshot failure in Cassandra

gabrielhartmann:


frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -202,6 +202,12 @@ pods:
                             -f ./apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra.yaml \
                             -d ${LOCAL_SEEDS} \
                             -p ${CASSANDRA_NATIVE_TRANSPORT_PORT} "$t" ;
+                        result=$?
+                        if [ $result != 0 ]
```
nickbp: could do `if [ $result != 0 ]; then` to match with the for loop formatting above

nickbp: (we've done it everyone, we've reached peak nitpick)

PR #1643, opened 2017-09-19
[INFINITY-2375] Update postInstall notes

elezar:
This PR changes the `postInstallNotes` notes to:
```
"The {{long-service-name}} service is being uninstalled.

For DC/OS versions from 1.10 no further action is required. For older DC/OS versions follow the instructions at {{documentation-path}}uninstall to remove any persistent state if required."
```

Where `{{long-service-name}}` is just a placeholder here for:
- DC/OS Apache Cassandra
- DC/OS Apache HDFS
- DC/OS Apache Kafka
- DC/OS Elastic
- DC/OS hello-world
- DC/OS Kibana

The `{{documentation-path}}` is defined such that the link points to the latest uninstall documentation. (This required that the string for `helloworld` be edited accordingly).

I have not updated the `template` framework.

PR #1661, opened 2017-09-22
[INFINITY-2375] Backport update of postinstall notes.

elezar:
Merge pull request #1643 from mesosphere/INFINITY-2375-update-post-install-notes.

[INFINITY-2375] Update postInstall notes.

@loren I also removed a beta notice from elastic, I assume that this is in order for 0.30.X?


loren:
Yes, I believe it's safe to delete that dire warning.

elezar:
Note. 1.9 builds fail, but we're testing against: 
```
DCOS_CHANNEL_19 | testing/1.9.3
```

benclarkwood:
Smells like cluster issue to me. Re-running. The failed tests were the _last_ tests.

PR #1663, opened 2017-09-22
[INFINITY-2454] Properly handle deploy plans with errors

elezar:
In #1365 the plan endpoint was changed to return a 417 instead of 202
when the plan contained errors. This was not properly handled by the
cli.

This PR:
- Adds an explicit handler (and tests) for the case where the plan endpoint
returns 417.
- Terminates the CLI with a non-zero exit code (1) if such an error occurs.

I will make these changes to master too.

### DIFF
--- a/cli/client/print.go
+++ b/cli/client/print.go
@@ -18,13 +18,21 @@ var PrintMessage = printMessage
 // PrintMessage() before exiting to allow assertions against captured output.
 var PrintMessageAndExit = printMessageAndExit
 
+// Exit is a placeholder function that wraps a call to os.Exit to allow
+// assertions against exit with an exit code.
+var Exit = exit
+
+func exit(code int) {
+	os.Exit(code)
+}
+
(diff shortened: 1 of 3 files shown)


cli/client/print.go
```diff
@@ -18,13 +18,21 @@ var PrintMessage = printMessage
 // PrintMessage() before exiting to allow assertions against captured output.
 var PrintMessageAndExit = printMessageAndExit
 
+// Exit is a placeholder function that wraps a call to os.Exit to allow
+// assertions against exit with an exit code.
+var Exit = exit
```
benclarkwood: @elezar Why the indirection on this?

elezar: The unit tests now check for a non-zero status code, but there is some discussion in  #1664 on whether this is needed.

elezar:
I am closing this because it probably makes sense to first address all comments in #1664 and then backport (I will reopen this once it's is ready for comments/review).

PR #1664, opened 2017-09-22
INFINITY-2454: Handle a 417 return code on the plan endpoint.

elezar:
In #1365 the plan endpoint was changed to return a 417 instead of 202
when the plan contained errors. This was not properly handled by the
cli.

This PR (backported in #1663): 
- Adds an explicit handler (and tests) for the case where the plan endpoint
returns 417.
- Terminates the CLI with a non-zero exit code (1) if such an error occurs.

I will make these changes to master too.


cli/commands/plan.go
```diff
@@ -238,8 +240,13 @@ func (cmd *planHandler) handleStart(a *kingpin.Application, e *kingpin.ParseElem
 func printStatus(planName string, rawJSON bool) {
 	client.SetCustomResponseCheck(checkPlansResponse)
 	responseBytes, err := client.HTTPServiceGet(fmt.Sprintf("v1/plans/%s", planName))
+
 	if err != nil {
-		client.PrintMessageAndExit(err.Error())
+		if err.Error() == "plan has errors" {
```
gabrielhartmann: Why are comparing against the error message instead of the error code.  Does something else return 417?

cli/commands/plan.go
```diff
@@ -238,8 +240,13 @@ func (cmd *planHandler) handleStart(a *kingpin.Application, e *kingpin.ParseElem
...
+			defer client.Exit(1)
```
gabrielhartmann: Why an exit code of 1?  Client didn't have a problem..

Also we whould set constants for exit codes if we're going to have them be meaningful.

cli/commands/plan.go
elezar: At this point we only have an error object with the `Error` value `"plan has errors"` as set on https://github.com/mesosphere/dcos-commons/pull/1664/files#diff-f765f89176290c9ef7fa3ba2d0b3d0d5R94. It would probably make it clearer to set this to `"417"` instead.

cli/commands/plan.go
elezar: Looking at the code, we use `PrintMessageAndExit` if an error is returned (and always exit with an exit code of 1: https://github.com/mesosphere/dcos-commons/pull/1664/files#diff-fe7dfb33c57e6046f4d1d53c9af885c6R33).

We call this termination function for many error cases, some of which are not client specific. But you're right in that we probably shouldn't use it here. We have a success condition, and tooling can check the json output of the command to determine if the plan has errors and not have to rely on the (undefined) status code.

This does make me wonder why it was needed to change the plan endpoint to return a 4xx HTTP code, as the error part of the JSON object returned? Is there a case where returning a 202 (as was the case originally) is incorrect?

cli/commands/plan.go
```diff
@@ -91,6 +93,8 @@ func checkPlansResponse(response *http.Response, body []byte) error {
 		}
 	case response.StatusCode == http.StatusAlreadyReported:
 		return errors.New("Cannot execute command. Command has already been issued or the plan has completed.")
+	case response.StatusCode == http.StatusExpectationFailed:
+		return errPlanStatus417
```
nickbp: Would it make sense to just return nil in this case?

elezar: We need to return an error to not fall through to the default HTTP error handler.

nickbp: Ah got it, ok

PR #1666, opened 2017-09-22
[INFINITY-2462] Rely on Mesos offer accept/decline behavior

gabrielhartmann:
This is a small PR with a big effect because it relies on Mesos behavior.  The end result of this PR is that any given set of crash looping tasks generate at most `1` revive call.  This means that very scalable recovery is possible without any actions which impact other frameworks (i.e. `revive` calls) and without relying on our revive throttling mechanism.

## Big Change
No revive calls are necessary from a Mesos perspective when a Task crashes.  The task was originally launched by _accepting_ an offer.  So that offer was not declined forever (2 weeks).  When the task crashes those resources will be re-offered to the scheduler.  The re-offer is guaranteed in our case because we reserve resources.  The scheduler should accept them and recover the task.

From an SDK scheduler perspective we now treat work as falling into only 2 categories: `COMPLETE` or not.  When new _incomplete_ work is detected `revive` is called.  Otherwise it isn't.  When work is completed it leaves the set of `current` work, so if it becomes incomplete again it will be processed again.  We accomplish this categorization by leaving status out of the definition of a `WorkItem`.  Complete work is removed from the system, by being filtered out in the `getCandidates` call.

### Testing
I ran 13 scheduler's with one crash-looping task and one good task on each.  No starvation occurred for any scheduler including Marathon.  No `revive` calls were observed on any scheduler while Tasks were continually relaunched at great speed.  I'll be doing further larger scale testing later.

#### Small change
I moved the revive check after plan processing as it made more sense to me.  Shouldn't have any real effect, but basically it makes sense to evaluate the state of the system with regard to needing to revive after execution of plans.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
@@ -131,11 +131,11 @@ public abstract class AbstractScheduler implements Scheduler {
                 // Get the current work
                 Collection<Step> steps = getPlanCoordinator().getCandidates();
 
-                // Revive offers if necessary
-                reviveManager.revive(steps);
-
                 // Match offers with work
                 executePlans(offers, steps);
+
+                // Revive offers if necessary
+                reviveManager.revive(steps);
 
                 synchronized (inProgressLock) {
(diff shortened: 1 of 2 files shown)


gabrielhartmann:
Investigated C* flake: https://jira.mesosphere.com/browse/CASSANDRA-649

PR #1667, opened 2017-09-22
[HDFS-477] Expose HDFS configuration settings

NimaVaziri:
- [x] Expose config settings for `hdfs-site.xml`
- [x] Expose config settings for `core-default.xml`
- [x] Update documentation wrt additional config settings that are now exposed


NimaVaziri:
Waiting on @sascala to confirm docs changes.

PR #1669, opened 2017-09-22
Cache config objects in memory in the ConfigStore

gabrielhartmann:
Used VisualVM profiler to find what was slowing down DefaultSchedulerTests.  ConfigStore was caching the ServiceSpec's serialized format but still parsing it every time some component looked up a configuration.  The parsing was taking a long time.  So I've added some caching post-parsing.  This manifested in the Scheduler logs as an approximately `0.250s` delay on an empty iteration of the scheduling loop.  This delay has disappeared.


gabrielhartmann:
@nickbp Where's the close call?  Don't see it.

nickbp:
Way at the bottom of ConfigStore.java

PR #1673, opened 2017-09-25
INFINITY-2465 Consolidate TLS handling

nickbp:
- Reduces unused builder code, sticking with default values. Figure we can add customization back later as it's required.
- Consolidates TLS-specific handling a bit:
  - Have SchedulerFlags return the TokenProvider directly (instead of exposing several individual envvars)
  - Have UninstallPlanBuilder internally determine whether to add a TLS cleanup phase (rather than being determined upstream)


sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/ca/DefaultCAClient.java
```diff
@@ -110,8 +104,8 @@ public X509Certificate sign(byte[] csr) throws Exception {
         return certificates;
     }
 
-    private JSONObject doPostRequest(String path, JSONObject data) throws IOException, CAException {
-        Request request = Request.Post(URLUtils.addPathUnchecked(baseURL, path).toString())
+    private JSONObject doPostRequest(String path, JSONObject data) throws IOException, CAException, URISyntaxException {
+        Request request = Request.Post(new URI(DcosConstants.CA_BASE_URI + path))
```
benclarkwood: @mhrabovcin do we lose anything making this a constant? Given that unit tests pass, I assume we weren't doing any actual network tests of this?

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/secrets/DefaultSecretsClient.java
```diff
@@ -23,34 +22,22 @@
...
-        this.store = store;
-        this.httpExecutor = executor;
-    }
-
-    public DefaultSecretsClient(String store, Executor executor) {
-        this(URLUtils.fromUnchecked(DcosConstants.SECRETS_BASE_URI), store, executor);
-    }
-
     public DefaultSecretsClient(Executor executor) {
-        this("default", executor);
```
benclarkwood: I see this was the only constructor ever being called.

Are we storing all generated secrets as `secret/default/...`? Should we change that to something like "generated"?

benclarkwood: Oh, or is a "store" a silo within the secrets api?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
```diff
@@ -34,20 +34,19 @@
...
+     * {@link StateStore}. The {@link UninstallScheduler} builds an uninstall {@link Plan} which will clean up the
+     * service's reservations, TLS artifacts, zookeeper data, and any other artifacts from running the service.
      */
     public UninstallScheduler(
             ServiceSpec serviceSpec,
             StateStore stateStore,
             ConfigStore<ServiceSpec> configStore,
             SchedulerFlags schedulerFlags,
-            Optional<SecretsClient> secretsClient) {
+            Optional<SecretsClient> customSecretsClientForTests) {
```
benclarkwood: Could you just refactor to a second protected constructor that takes the secret client?

sdk/scheduler/src/test/java/com/mesosphere/sdk/testutils/OfferRequirementTestUtils.java
```diff
@@ -47,11 +51,13 @@ public static DcosCluster getTestCluster(String version) {
     }
 
     public static SchedulerFlags getTestSchedulerFlags() {
```
benclarkwood: `SchedulerFlags` doesn't seem like a totally accurate ClassName anymore. I think having a minimal interface to the Scheduler's system/environment is sensible (e.g. not exposing a bunch of envvars to then only use them one place).

What about calling it `SchedulerEnvironment` or `SchedulerSystem`?

nickbp: How about `SchedulerConfig`? `SchedulerSystem` feels like it'd be invoking ulimits settings or something, while `SchedulerEnvironment` is a bit further into being specific to just envvar fetching.

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/secrets/DefaultSecretsClient.java
nickbp: If it helps to give some context, the complete base URL ends up looking like: `http://master.mesos/secrets/v1/secret/default/`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
nickbp: Good point, done

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/secrets/DefaultSecretsClient.java
nickbp: If it helps to give some context, the complete base URL ends up looking like: `http://master.mesos/secrets/v1/secret/default/`

mhrabovcin: The DC/OS has concept of different secrets stores, where default has name `default`. Having a name as a constant makes sense here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -465,20 +465,12 @@ private static void logOutcome(StringBuilder stringBuilder, EvaluationOutcome ou
         return evaluationStages;
     }
 
-    private static Optional<TLSEvaluationStage.Builder> getTLSEvaluationStageBuilderFromEnvironment(
-            SchedulerFlags flags, PodInstanceRequirement podInstanceRequirement) {
-        Optional<TLSEvaluationStage.Builder> tlsBuilder = Optional.empty();
+    private static Optional<TLSEvaluationStage.Builder> getTLSEvaluationStageBuilder(
+            String serviceName, SchedulerConfig flags, PodInstanceRequirement podInstanceRequirement) {
```
mhrabovcin: Rename `flags` variable to `schedulerConfig` ?

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/ca/DefaultCAClient.java
mhrabovcin: I was using this originally to test this class against local cluster when I was developing it originally. It is not used anymore.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/TLSRequiresServiceAccount.java
```diff
@@ -12,37 +11,30 @@
 
 /**
  * A {@link TLSRequiresServiceAccount} checks whether the configuration contains provisioning of TLS artifacts
- * and whether the provided {@link SchedulerFlags} contains a service account.
+ * and whether the provided {@link SchedulerConfig} contains a service account.
  */
 public class TLSRequiresServiceAccount implements ConfigValidator<ServiceSpec> {
 
-    private final SchedulerFlags flags;
```
elezar: is `flags` now a representative name for this member variable?

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/secrets/DefaultSecretsClient.java
elezar: Yes, we could extract `default` to a constant (e.g. `DEFAULT_SECRET_STORE_NAME`) to make it clearer what is being set.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -113,23 +126,23 @@ public String getMessage() {
...
+     * Returns a new {@link SchedulerConfig} instance which is based off the provided custom environment map.
      */
-    public static SchedulerFlags fromMap(Map<String, String> map) {
-        return new SchedulerFlags(map);
+    public static SchedulerConfig fromMap(Map<String, String> map) {
+        return new SchedulerConfig(map);
     }
 
-    private final FlagStore flagStore;
+    private final EnvStore flagStore;
```
elezar: should the variable name also be changed?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/TLSCleanupStep.java
```diff
@@ -23,7 +23,18 @@
  */
 public class TLSCleanupStep extends AbstractStep {
 
-    private static final Pattern PATTERN = createSecretNamePattern();
+    private static final Pattern PATTERN;
+    static {
+        List<String> possibleSecretNames = Arrays.asList(
+                SecretNameGenerator.SECRET_NAME_CERTIFICATE,
```
elezar: Shoudl the list of possible secret names not be the responsibility of the `SecretNameGenerator`? If it were an `enum`, for example, we could iterate over the values.

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/secrets/DefaultSecretsClient.java
benclarkwood: seconded

benclarkwood: (Also a comment saying what Martin said would be 👍 )

nickbp:
Any other comments? Last call...

PR #1681, opened 2017-09-26
Revert "INFINITY-1975 Tweaks to Cassandra endpoints: rename node=>nat…

benclarkwood:
…ive-client, remove VIP, show thrift-client when enabled (#1503) (#1512)"

This reverts commit 168401552315b2b908cbcacf4d600460a498344f.

### DIFF
--- a/frameworks/cassandra/src/main/dist/svc.yml
+++ b/frameworks/cassandra/src/main/dist/svc.yml
@@ -32,14 +32,14 @@ pods:
             port: {{TASKCFG_ALL_CASSANDRA_STORAGE_PORT}}
           ssl:
             port: {{TASKCFG_ALL_CASSANDRA_SSL_STORAGE_PORT}}
-          native-client:
+          node:
             port: {{TASKCFG_ALL_CASSANDRA_NATIVE_TRANSPORT_PORT}}
             advertise: true
-          {{#TASKCFG_ALL_CASSANDRA_START_RPC}}
-          thrift-client:
+            vip:
+              prefix: node
+              port: 9042
+          rpc:
             port: {{TASKCFG_ALL_CASSANDRA_RPC_PORT}}
-            advertise: true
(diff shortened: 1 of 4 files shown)


frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -32,14 +32,14 @@ pods:
...
-          native-client:
+          node:
             port: {{TASKCFG_ALL_CASSANDRA_NATIVE_TRANSPORT_PORT}}
             advertise: true
-          {{#TASKCFG_ALL_CASSANDRA_START_RPC}}
-          thrift-client:
+            vip:
+              prefix: node
+              port: 9042
+          rpc:
```
nickbp: this endpoint will only be available if START_RPC is enabled. its arguably a bug to display it when the feature is disabled

frameworks/cassandra/universe/config.json
```diff
@@ -285,7 +285,7 @@
         },
         "start_rpc": {
           "type": "boolean",
-          "description": "If true Thrift RPC is enabled. This is deprecated but may be necessary for legacy applications.",
+          "description": "If true Thrift RPC is enable. This is deprecated by may be necessary for legacy applications.",
```
nickbp: I think this can stay as-is

frameworks/cassandra/src/main/dist/svc.yml
benclarkwood: Yep. @nickbp can you create a ticket to make just this fix in 0.30.X?

elezar:
Sorry for getting on this so late. I know that DSE also recently had similar changes made to the endpoint names. Should these changes be made there too?

benclarkwood:
@elezar Have those changes already leaked out? If they have, let's not change them back.

elezar:
From what I can tell, the changes were merged as part of mesosphere/dse-private#218.

PR #1692, opened 2017-09-28
Fix EE check (#1599)

benclarkwood:
* Tweak evaluation function

* Missing function invocation

* Update sdk_install.install parameter ordering

* Also set service name in install fixture.

* Fix "test", but then run into another issue and disable it for now. Ticket opened.


benclarkwood:
@elezar 1 was a bad cluster bring up, other 2 were known flapping tests. Looking into those next week.

PR #1694, opened 2017-09-28
[ELASTIC-108] Specify YAML block to configure cluster-wide settings

loren:
- WIP to see how bootstrap (and specifically, github.com/nickbp/mustache) handles rendering newlines and spaces in a shell env var


grillorafael:
That would be good to cover extra scenarios. Also is this per machine type or applied to all types (Master/Data/Coordinator/Ingest) ?

loren:
it would be applied to all types.

gabrielhartmann:
@loren: Is it possible to write a couple test cases that validate that indents are maintained from input to mustache injection?

frameworks/elastic/universe/config.json
```diff
@@ -315,6 +315,11 @@
           "type": "string",
           "default": ""
         },
+        "custom_yaml_block_base64": {
```
triclambert: What is the future plan for integration with the DC/OS UI?

loren: Pretty sure I won't know until _after_ you know. See [this comment](https://jira.mesosphere.com/browse/ELASTIC-108?focusedCommentId=121970&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-121970) in Jira.

gabrielhartmann: This same field will be used when the UI is ready. It will handle encoding and decoding from base64. @loren we should probably rename this field. It's supposed to be only for plugin so maybe we should call it 'plugin_config'.  We should ask Keith.  Otherwise good to go.

gabrielhartmann: @loren: Please go with `custom_yaml` as the field name.  In the description say, "This field must be base64 encoded...".  Then go ahead and merge once something reasonable has passed in CI. /cc @triclambert

gabrielhartmann: @loren:  Actually, just heard back from Keith.  Config value should be named `custom_elasticsearch_yml`.

keithchambers: Spoke with @gabrielhartmann about this. We landed on `custom_elasticsearch_yml` in case we need to support other plugins that use files other than elasticsearch.yml.

grillorafael:
Hey guys, 

Quick question as I'm not used to the release cycles. When is this due to go to `universal` registry?

PR #1697, opened 2017-09-29
[INFINITY-2446] Update docs for changed option for service account.

jsrodman:
We changed the option name.  Docs should follow.

I'm assuming the "principal" that appears in resource sets, reservations, and log messages about offers are all unchanged.

### DIFF
--- a/docs/pages/developer-guide.md
+++ b/docs/pages/developer-guide.md
@@ -134,7 +134,7 @@ This simple YAML definition of a DC/OS service that prints "hello world" to stdo
 ```yaml
 name: "hello-world"
 scheduler:
-  principal: "hello-world-principal"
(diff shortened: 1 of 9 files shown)


jsrodman:
I'd love it if someone (nick, james, or otherwise) could confirm that reservations, logs, and resource sets are still all "princpal", though no-change isn't a disaster even if they're stale.

jsrodman:
@benclarkwood seems like the docs-only pull requests dont' fire the tests, but also don't check them as green.

benclarkwood:
@jsrodman We did not change the "yaml" field(s) because they are in fact "mesos" principals. We just make the customer facing name `service_account` to match the rest of DC/OS.

Can you revert the changes you made to the dev guide and the yaml-reference?

PR #1698, opened 2017-10-02
[INFINITY-2455] Properly parse ints in update errors

elezar:
This PR properly handles integers (or non-string data type) in update specification errors.

The main problem was the the `Found` field in `cosmosError` was specified as a `string`, meaning that `int` data (e.g. the number of nodes) could not be properly unmarshalled.

I changed the field type (and that of `Expected`) to `interface{}` allowing all data types to be handed.

I also added two simple unit tests for two cases.

### DIFF
--- a/cli/client/cosmos.go
+++ b/cli/client/cosmos.go
@@ -40,8 +40,8 @@ type cosmosErrorInstance struct {
 type cosmosError struct {
 	Keyword  string
 	Message  string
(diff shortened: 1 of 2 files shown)


elezar:
Thanks for the review and pushing this through @benclarkwood and @nickbp. I have created #1708 to backport these changes.

PR #1701, opened 2017-10-02
[ELASTIC-139] Add additional config options to elastic

mbdcos:


protetore:
Hey @mbdcos , do you think you can add these config options too??
```
threadpool.bulk.queue_size
indices.query.bool.max_clause_count
```

mbdcos:
Added `indices.query.bool.max_clause_count`. 
`threadpool.bulk.queue_size` should be already there...

grillorafael:
:tada:

PR #1702, opened 2017-10-02
[INFINITY-2498] Drop sticky dynamic ports on replacement

benclarkwood:
If a task is being replaced, drop its previous dynamic ports from consideration.

Add test for this behavior.


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -87,7 +85,12 @@ public PodInfoBuilder(
             // Just store against the full TaskInfo name ala 'broker-0-node'. The task spec name will be mapped to the
             // TaskInfo name in the getter function below. This is easier than extracting the task spec name from the
             // TaskInfo name.
-            portsByTask.put(currentTask.getName(), new TaskPortLookup(currentTask));
+
+            // If the pod was replaced, discard any previously used ports. We want dynamic ports
+            // to re-roll.
```
nickbp: FYI lines can go to 120char (linter limit)

PR #1703, opened 2017-10-03
[INFINITY-2446] Correct infinity-2446

jsrodman:


docs/pages/developer-guide.md
```diff
@@ -134,7 +134,7 @@ This simple YAML definition of a DC/OS service that prints "hello world" to stdo
 ```yaml
 name: "hello-world"
 scheduler:
-  service_account: "hello-world-principal"
+  principal: "hello-world-principal"
   api-port: {{PORT_API}}
```
nickbp: looks like `api-port` can also be removed. it's not listed in `RawScheduler`. probably got removed from the spec but forgot to remove it here/below..

PR #1710, opened 2017-10-04
[INFINITY-2491] Remove Elastic zero ingest nodes tests

paegun:
* Remove Elastic zero ingest nodes tests, the default has changed.
* Fix Elastic tests to consistently use default node count.
  * But also propose (TODO) a test improvement to determine the expected
    node count based on the configuration options.


paegun:
@benclarkwood we do not have a tests in place that test that "base tech" ingest functionality works properly when there are both zero and non-zero ingest nodes.

What we do have is a test that demonstrates that we can set the ingest node count (via Marathon, not via install using config), see https://github.com/mesosphere/dcos-commons/blob/3195b8b9054d555cceb119f93f7183a80bb593ba/frameworks/elastic/tests/test_sanity.py#L246

I feel that this is approaching "in for a penny, in for a pound". The initial foray into this work was to remove test redundancy, where a test suite marked for removal is an obvious set of tests that should be removed.

paegun:
Again, local tests are working, CI is failing.

elezar:
Local tests use mesosphere/dcos-commons:latest image. CI uses :dev and
possibly different instance types.

Evan Lezar

paegun:
Closing, this is well beyond the scope of simply following what was intended as a TODO: or HACK: comment to cleanup tests that are no longer applicable.

PR #1712, opened 2017-10-05
INFINITY-2512 Resource builder cleanup

nickbp:
- Be more stringent/explicit in ResourceBuilder construction: reduce the risk of mixing up basic resource vs root volume vs mount volume.
- Remove unread MesosResource field from EvaluationOutcome.
- Avoid building the Resource object twice when evaluating resources (once in `OfferEvaluationUtils`, then again upstream).
- Silence Mesos deprecation warnings by wrapping access to the `role` field. We still need it for compatibility with 1.9 and earlier.


nickbp:
Gonna mark this WIP until I have time to debug the integration failures

PR #1713, opened 2017-10-06
[INFINITY-2514] Fix automatic login regression

benclarkwood:
Docker passes in the envvars added in https://github.com/mesosphere/dcos-commons/commit/15fb4dcd6a402c48184fb40581fb47ae1c32f961 as `" "` in the case that the user does not have them set.

I've patched this in the Python, rather than trying figure out how to get Docker to NOT do this. And also because this was the fastest way I thought of to fix it.


nickbp:
`helloworld-19` and `template` are successfully running, so should be all set.

PR #1715, opened 2017-10-06
[KAF-379] Update Kafka app version (from 0.11.0.0 0.11.0.1)

mbdcos:


mbdcos:
Can not download new artifact in TC???

PR #1717, opened 2017-10-07
INFINITY-2513 Stop pods prelude: State store support

nickbp:
- Defines the `GoalStateOverride` enum/class
- Adds calls for setting/getting overrides in the StateStore (writes to Task/foo/Metadata/...)
- Updates `Persister` transaction code to support mixed writes+deletes
- And about 600 lines of new unit tests


sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -313,61 +313,76 @@ public void setMany(Map<String, byte[]> unprefixedPathBytesMap) throws Persister
     }
 
     @Override
+    public Map<String, byte[]> getMany(Collection<String> unprefixedPaths) throws PersisterException {
+        if (unprefixedPaths.isEmpty()) {
+            return Collections.emptyMap();
+        }
+        logger.debug("Getting many entries: {}", unprefixedPaths);
```
mrbrowning: For debugging it'd probably be useful to actually say how many here? Similar to above log like this.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -313,61 +313,76 @@ public void setMany(Map<String, byte[]> unprefixedPathBytesMap) throws Persister
...
+
+        Map<String, byte[]> result = new TreeMap<>();
+        // Unlike with writes, there is not an atomic read operation. Therefore we wing it with a series of plain reads.
```
mrbrowning: Now that I think of it, we could take a lock in ZK itself to this end? It's not essential now, but I think we should keep it in our back pocket if we see synchronization issues here at some point. Maybe an RW lock with two semaphores?

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -266,30 +254,42 @@ public void setMany(Map<String, byte[]> unprefixedPathBytesMap) throws Persister
...
+                Set<String> existingAndPendingCreatePaths = new HashSet<>();
+                // List of paths which are about to be deleted by the transaction
+                Set<String> pendingDeletePaths = new HashSet<>();
+
+                CuratorTransactionFinal transaction = client.inTransaction().check().forPath(serviceRootPath).and();
+                for (Map.Entry<String, byte[]> entry : pathBytesMap.entrySet()) {
+                    String path = entry.getKey();
+                    if (entry.getValue() == null) {
+                        // null value: if present, delete path and any children (unless already being deleted)
+                        if (!pendingDeletePaths.contains(path)
```
mrbrowning: The logic here gets pretty hairy at this level of nesting, and considering that it concerns something as fundamental as ZK, I think it'd be worth breaking apart for readability.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
```diff
@@ -0,0 +1,61 @@
...
+    /** The definition of the "STOPPED" override state, where commands are replaced with sleep()s.*/
+    STOPPED("stop", "STOPPED", "STOPPING");
+
+    private final String command;
+    private final String serializedName;
+    private final String pendingName;
+
+    private GoalStateOverride(String command, String serializedName, String pendingName) {
+        // Validation: Overrides may not overlap with existing GoalStates
+        for (GoalState goalState : GoalState.values()) {
```
mrbrowning: Hm, since this check is static, I'm loath to spend runtime evaluating it, is there another way to achieve this?

nickbp: FWIW this code is effectively run statically at VM initialization via the "constructor" calls for each of the enum values. So any validation problems should be caught quickly via e.g. a unit test.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
nickbp: In practice it shouldn't be an issue assuming the PersisterCache is enabled, as in that situation this function wouldn't be getting called anyway (the cache just fetches all data in the tree up-front then serves from that). Added a comment to explain/justify the current winging-of-it.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
nickbp: Good point, added to both.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
nickbp: Good point, broke out the transaction creation itself into a separate function. Much cleaner now.

nickbp:
I tested this by hand by running `debug pod stop`/`debug pod start` in the CLI against a pod:
- `stop`: Pod was correctly restarted with a stopped+inprogress metadata state in ZK
- `start`: Pod was restarted again with a none+inprogress metadata state in ZK

The operations are effectively a noop as the pod commands themselves aren't actually being changed yet, but the glue surrounding that seems to be working.

frameworks/cassandra/universe/resource.json
```diff
@@ -7,12 +7,13 @@
     },
     "uris": {
       "jre-tar-gz": "{{jre-url}}",
-      "cassandra-jre-tar-gz": "{{jre-url}}",
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
-      "scheduler-zip": "{{artifact-dir}}/cassandra-scheduler.zip",
-      "cassandra-tar-gz": "https://downloads.mesosphere.com/cassandra/assets/apache-cassandra-3.0.14-bin-dcos.tar.gz",
+      "executor-zip": "{{artifact-dir}}/executor.zip",
```
gabrielhartmann: This file shouldn't change.

frameworks/elastic/universe/resource.json
```diff
@@ -6,8 +6,9 @@
       "executor-zip": "{{artifact-dir}}/executor.zip",
       "bootstrap-zip": "{{artifact-dir}}/bootstrap.zip",
       "scheduler-zip": "{{artifact-dir}}/elastic-scheduler.zip",
-      "statsd-plugin-zip": "https://github.com/mesosphere/elasticsearch-statsd-plugin/releases/download/{{elastic-version}}.0/elasticsearch-statsd-{{elastic-version}}.0.zip",
+
```
gabrielhartmann: This file shouldn't change.

frameworks/hdfs/universe/resource.json
```diff
@@ -3,12 +3,13 @@
     "uris": {
       "jre-tar-gz": "{{jre-jce-unlimited-url}}",
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
-      "hdfs-tar-gz": "https://downloads.mesosphere.com/hdfs/assets/hadoop-2.6.0-cdh5.11.0.tar.gz",
-      "hdfs-bin-tar-gz": "https://downloads.mesosphere.com/hdfs/assets/hdfs-bin.tar.gz",
-      "hdfs-jre-tar-gz": "{{jre-jce-unlimited-url}}",
+      "executor-zip": "{{artifact-dir}}/executor.zip",
```
gabrielhartmann: This file shouldn't change.

frameworks/helloworld/universe/resource.json
```diff
@@ -3,9 +3,10 @@
     "uris": {
       "jre-tar-gz": "{{jre-url}}",
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
+      "executor-zip": "{{artifact-dir}}/executor.zip",
```
gabrielhartmann: This file shouldn't change.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PodResource.java
```diff
@@ -165,6 +172,107 @@ public Response getPodInfo(@PathParam("name") String name) {
...
+        }
+        try {
+            return overrideGoalState(name, taskFilter, GoalStateOverride.NONE);
+        } catch (Exception e) {
+            LOGGER.error(String.format("Failed to start pod '%s' with task filter '%s'", name, taskFilter), e);
+            return Response.serverError().build();
+        }
+    }
+
+    private static Set<String> parseJsonList(String payload) throws JSONException {
```
gabrielhartmann: I think these internal methods should get out of the `PodResource` class and go somewhere else.  I prefer these classes having very little logic in them.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PodResource.java
```diff
@@ -165,6 +172,107 @@ public Response getPodInfo(@PathParam("name") String name) {
...
+        Iterator<Object> iter = new JSONArray(payload).iterator();
+        while (iter.hasNext()) {
+            strings.add(iter.next().toString());
+        }
+        return strings;
+    }
+
+    private Response overrideGoalState(String podName, Set<String> taskNameFilter, GoalStateOverride override) {
+        // look up all tasks in the provided pod name:
+        List<TaskInfoAndStatus> podTasks = GroupedTasks.create(stateStore).byPod.get(podName);
```
gabrielhartmann: Can we factor out the fetching/filtering here into a separate method?

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PodResource.java
```diff
@@ -277,22 +391,48 @@ public int compare(TaskInfoAndStatus a, TaskInfoAndStatus b) {
...
+                jsonTask.put("type", new TaskLabelReader(task.getInfo()).getType());
+            } catch (TaskException e) {
+                // no type found, just omit type field from response
             }
             jsonPod.put(jsonTask);
         }
         return jsonPod;
     }
 
+    private static String getSDKTaskState(StateStore stateStore, String taskName, Protos.TaskState mesosState) {
```
gabrielhartmann: Javadoc please.  Comment on line 417 should probably move up.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
```diff
@@ -0,0 +1,144 @@
...
+            return serializedName;
+        }
+    }
+
+    /**
+     * Describes the current state of an override.
+     *
+     * The state of the override itself. Sample flow for enabling and disabling an override:
+     *
+     * <ol><li>NONE + COMPLETE (!isActive())</li>
```
gabrielhartmann: I'm confused by this table and would like to understand it. Schema is `override-goal + override-status`?  Can we write more explanatory comments here?

gabrielhartmann: I understand it after reading it 3 times.  Maybe could be clearer.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
```diff
@@ -0,0 +1,144 @@
...
+     * <li>STOPPED + IN_PROGRESS</li>
+     * <li>STOPPED + COMPLETE</li>
+     * <li>NONE + PENDING</li>
+     * <li>NONE + IN_PROGRESS</li>
+     * <li>NONE + COMPLETE (!isActive())</li></ol>
+     */
+    public static class Status {
+
+        /**
+         * The status of a task for which no overrides are applicable.
```
gabrielhartmann: `no overrides are applicable, and it has reached its goal state`

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
```diff
@@ -0,0 +1,144 @@
...
+        public static final Status INACTIVE = new Status(GoalStateOverride.NONE, Progress.COMPLETE);
+
+        /**
+         * The target override state for this task. May be {@link GoalStateOverride#NONE} in the case of no override.
+         */
+        public final GoalStateOverride target;
+
+        /**
+         * The current state for transitioning to the {@link #target} in question. May be {@link Progress#NONE} in the
+         * case of no override being applicable.
```
gabrielhartmann: Can progress be `NONE` when the task hasn't been killed yet?

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
```diff
@@ -0,0 +1,144 @@
...
+
+    public Status newStatus(Progress progress) {
+        return new Status(this, progress);
+    }
+
+    private final String serializedName;
+    private final String pendingName;
+
+    private GoalStateOverride(String serializedName, String pendingName) {
+        // Validation: Overrides may not overlap with existing GoalStates
```
gabrielhartmann: Require all caps, no spaces for names as well?

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
```diff
@@ -0,0 +1,144 @@
...
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+import com.mesosphere.sdk.specification.GoalState;
+
+/**
+ * The definition of a goal state override for a task. These are custom states, other than {@link GoalState}s, which a
+ * task may enter. The main difference between this and {@link GoalState} is that {@link GoalState}s are configured by
+ * the service developer, whereas {@link GoalStateOverride}s are applied by the operator.
+ */
+public enum GoalStateOverride {
```
gabrielhartmann: What's the value of making this an `enum`?  Not saying it's a bad idea, just looks a whole lot like a class to me.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -433,6 +440,92 @@ public void clearProperty(final String key) throws StateStoreException {
...
+     * @throws StateStoreException in the event of a storage error
+     */
+    public void storeGoalOverrideStatus(String taskName, GoalStateOverride.Status status)
+            throws StateStoreException {
+        try {
+            Map<String, byte[]> values = new TreeMap<>();
+
+            if (GoalStateOverride.Status.INACTIVE.equals(status)) {
+                // Mark inactive state by clearing any override bits.
+                values.put(getGoalOverridePath(taskName), null);
```
gabrielhartmann: What's more fault-tolerant do you think?  Attempting to keep `empty` meaning no overrides, or having an explicit value for no overrides.  `setter` and `getter` both have branches for present and not-present.  I wonder if we can simplify this by removing a branch.  It's definitely fine as is.

One other thing.  It seems a little strange to me to put `null` values for deleting a node.  Maybe the persister should have a `clearMany` method.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -433,6 +440,92 @@ public void clearProperty(final String key) throws StateStoreException {
...
+            String goalOverridePath = getGoalOverridePath(taskName);
+            String goalOverrideStatusPath = getGoalOverrideStatusPath(taskName);
+            Map<String, byte[]> values = persister.getMany(Arrays.asList(goalOverridePath, goalOverrideStatusPath));
+            byte[] nameBytes = values.get(goalOverridePath);
+            byte[] statusBytes = values.get(goalOverrideStatusPath);
+            if (nameBytes == null && statusBytes == null) {
+                // Cleared override bits => Inactive state
+                return GoalStateOverride.Status.INACTIVE;
+            } else if (nameBytes == null || statusBytes == null) {
+                throw new StateStoreException(Reason.LOGIC_ERROR,
```
gabrielhartmann: Do we definitely want to throw here?  Maybe we should return `INACTIVE` and so fallback to original behavior.  Then a user could retry the override, I think.  It would be nice for the end-user to be able to clear/retry overrides.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -433,6 +440,92 @@ public void clearProperty(final String key) throws StateStoreException {
...
+    }
+
+    private static GoalStateOverride parseOverrideName(byte[] nameBytes) throws StateStoreException {
+        String name = new String(nameBytes, StandardCharsets.UTF_8);
+        for (GoalStateOverride override : GoalStateOverride.values()) {
+            if (override.getSerializedName().equals(name)) {
+                return override;
+            }
+        }
+        throw new StateStoreException(Reason.SERIALIZATION_ERROR,
```
gabrielhartmann: Again, I'm not sure we want to throw and fail in these scenarios.  Willing to be convinced otherwise, but maybe logging errors is enough with fall back to `INACTIVE` for safety sake.  How would we get out of this situation?  `podtool`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -433,6 +440,92 @@ public void clearProperty(final String key) throws StateStoreException {
...
+    }
+
+    private static GoalStateOverride.Progress parseOverrideProgress(byte[] statusBytes) throws StateStoreException {
+        String status = new String(statusBytes, StandardCharsets.UTF_8);
+        for (GoalStateOverride.Progress state : GoalStateOverride.Progress.values()) {
+            if (state.getSerializedName().equals(status)) {
+                return state;
+            }
+        }
+        throw new StateStoreException(Reason.SERIALIZATION_ERROR,
```
gabrielhartmann: same

frameworks/cassandra/universe/resource.json
nickbp: Removed changes (here and below)

frameworks/elastic/universe/resource.json
nickbp: Fixed

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PodResource.java
nickbp: Added a `RequestUtils` class with this method as well as a method to filter tasks

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PodResource.java
nickbp: The filtering by task name has been moved into `RequestUtils`, and the `GroupedTasks` class has been broken out into the `api.types` package.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PodResource.java
nickbp: Removed this, it was temporary and is being replaced again in the following PR (currently the `stop-pods-podstatus` branch).

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
nickbp: Rewrote this as a table with columns for the step taken and the resulting state.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
nickbp: The javadoc was out of date, there is no `NONE` progress state (only `PENDING`/`IN_PROGRESS`/`COMPLETE`). When the override "bit" is initially set, the state will be set to `STOPPED`+`PENDING` and then the kill command will be issued for the tasks. After the `TASK_KILLED` TaskStatus has been received, the state will be updated to `STOPPED`+`IN_PROGRESS`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
nickbp: Good point, added a regex

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
nickbp: Mainly just for the freebie identity handling. An object-based system would involve likely involve a registry or something. Doing that would be more flexible in that developers could specify their own overrides, but I don't think we're to that point yet (would imply a well-defined set of 'hooks' where an override would take effect).

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
nickbp: Switched to having a new separate `deleteMany` method to be called here. Originally there had been a possibility of mixing writes and deletes in the same call, but that's not really needed now so having two separate calls makes things a lot simpler and/or less error-prone.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
nickbp: Sounds reasonable, in theory shouldn't happen but doesn't hurt to have a reasonable fallback here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
nickbp: Very good point. This should fall back to inactive when the name isn't known, to avoid causing undue problems during a downgrade. Switched to logging a warning and returning the inactive state.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
nickbp: Done

PR #1683, opened 2017-09-26
Use custom test options for a smaller Elastic footprint during testing.

benclarkwood:
Elastic's defaults are too large to fit on a m4.large. Set smaller defaults during tests.

PR #1719, opened 2017-10-09
Reduce Kafka test settings

benclarkwood:


testing/sdk_install.py
```diff
@@ -181,7 +181,7 @@ def get_package_options(additional_options={}):
     if os.environ.get('SECURITY', '') == 'strict':
         # strict mode requires correct principal and secret to perform install.
         # see also: tools/setup_permissions.sh and tools/create_service_account.sh
-        return _merge_dictionaries({
+        return merge_dictionaries({
```
NimaVaziri: Perhaps better to move this method to a utility module? Otherwise seems like we'd just be going against convention of having a single underscore to indicate private methods?

NimaVaziri:
lgtm overall besides a minor comment.

testing/sdk_install.py
paegun: The prefix was rightly applied. This is a private method by all accounts. I recommend restoring it.

benclarkwood: #1683 also requires this change. I'll make the tweak there (as this is about to pass CI)

benclarkwood: (The tweak == move to utility function)

PR #1724, opened 2017-10-11
Backport - Reduce Kafka test settings (#1719)

benclarkwood:
Backport of changes to make Kafka fit on m4.large.

* Reduce Kafka test settings

* Missed overlay install


frameworks/kafka/tests/test_availability.py
```diff
@@ -29,10 +29,10 @@ def setup_module(module):
 
     sdk_install.uninstall(SERVICE_NAME, PACKAGE_NAME)
 
-    sdk_install.install(
-        PACKAGE_NAME,
-        DEFAULT_BROKER_COUNT,
-        service_name=SERVICE_NAME,
+    config.install(
+        package_name=config.PACKAGE_NAME,
```
NimaVaziri: Perhaps we should retain consistency in the `config.install` calls by either specifying argument names or just dropping them (as we've done in most calls) ?

benclarkwood: There is a difference between argument order in 0.30.X and `master` so I added argument names where necessary. I'll clean it up when I backport the similar changes for Elastic.

PR #1727, opened 2017-10-11
Ss/packagename variables

sascala:
I replaced the package names in code blocks with `$packagename` and wrote an inelegant little script to populate.


elezar:
@sascala, it seems as all parts of commands starting with `$` have been removed.

sascala:
thanks for noticing that, @elezar. Will be fixed in the next commit

gabrielhartmann:
@sascala I'm going to close this.  It's huge and didn't land since October.  Please re-open if it's important.  If it's not could you delete the branch please?

PR #1733, opened 2017-10-13
INFINITY-2539 Install scheduler certs as early as possible

nickbp:
Fixes regression from #1704


nickbp:
Added some boilerplate around ensuring that the scheduler consistently throws up if the cert init call isn't called before the scheduler starts querying the cluster. This should ensure that future regressions are caught without requiring a strict mode cluster.

elezar:
Thanks a lot @nickbp! Strict mode on HW is looking a lot better now.

PR #1737, opened 2017-10-16
Skip the v1 API test in strict mode

elezar:
This PR adds:
- A utility function to check for strict mode
- Skip the v1 Mesos API test in strict mode:
  - An alternative would be to mark it as failure expected for strict mode and then properly cleanup the installation of the framework on failure. Currently this test failing causes all following tests to fail because the framework is not uninstalled on failure.
- Also remove the `nobody` permissions in strict cleanup:
  - I'm not 100% sure about this, but the asymmetry between the addition and removal of permissions bothered me a little. A further note is that we are adding roles which we may not need for particular tests, and also making assumptions about the name of the framework.

I also noted that the reservation refinement test fails consistently here. Is this expected to pass at all?


benclarkwood:
@elezar RR test failing is expected. It's blocked by upstream mesos work.

PR #1736, opened 2017-10-14
Retry login for up to 2 minutes

mellenburg:
Login sporadically flakes out if attempted directly after deployment time

PR #1738, opened 2017-10-16
Skip the v1 API test in strict mode (#1737)

elezar:
* Skip v1 api in strict

* Use mark.skipif

Backport of #1737


benclarkwood:
#1736 merged, which fixed the HW test that failed here. Re-running.

PR #1743, opened 2017-10-17
Expose evaluation outcomes as a debug endpoint

benclarkwood:


sdk/scheduler/src/main/java/com/mesosphere/sdk/api/OfferOutcomeResource.java
```diff
@@ -0,0 +1,34 @@
...
+    public OfferOutcomeResource(OfferOutcomeTracker offerOutcomeTracker) {
+        this.offerOutcomeTracker = offerOutcomeTracker;
+    }
+
+    /**
+     * Renders the current set of offer outcomes as an HTML <table>.
+     * @return HTML response of the table.
+     */
+    @GET
+    public Response getOfferOutcomes(@QueryParam("json") boolean json) {
```
elezar: Would it not be better to use the Content-Type header field than a query parameter here? Or is this not in keeping with how our APIs currently work?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/history/OfferOutcome.java
```diff
@@ -0,0 +1,39 @@
...
+    private final long timestamp;
+    private final String podInstanceName;
+    private final boolean pass;
+    private final Protos.Offer offer;
+    private final String outcomeDetails;
+
+    public OfferOutcome(String podInstanceName, boolean pass, Protos.Offer offer, String outcomeDetails) {
+        this.timestamp = System.currentTimeMillis();
+        this.podInstanceName = podInstanceName;
+        this.pass = pass;
```
elezar: Would it make sense to rather store the `failedOutcomeCount`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/history/OfferOutcomeTracker.java
```diff
@@ -0,0 +1,99 @@
...
+    public OfferOutcomeTracker(int capacity) {
+        this.outcomes = EvictingQueue.create(capacity);
+    }
+
+    public void track(OfferOutcome outcome) {
+        outcomes.add(outcome);
+    }
+
+    public void track(OfferOutcome... outcomes) {
+        Arrays.stream(outcomes).forEach(this::track);
```
elezar: Is there any advantage to calling `this.outcomes.addAll()` here directly?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/history/OfferOutcomeTracker.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private List<OfferOutcome> reverseList() {
```
elezar: If only there was an `EvictingStack`, then this would not be required.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/OfferOutcomeResource.java
```diff
@@ -0,0 +1,34 @@
...
+        if (json) {
+            return ResponseUtils.jsonOkResponse(offerOutcomeTracker.toJson());
+        } else {
+            return ResponseUtils.htmlOkResponse(offerOutcomeTracker.toHtml());
```
elezar: I may have missed the discussion as to why the JSON output is not sufficient?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -121,13 +127,26 @@ public OfferEvaluator(
                         failedOutcomeCount,
                         evaluationStages.size(),
                         outcomeDetails.toString());
+
+                offerOutcomeTracker.track(new OfferOutcome(
```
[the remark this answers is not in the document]
benclarkwood: I dirtily threw it into the conditionals since one of them does a return. I'll refactor it a bit.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/history/OfferOutcomeTracker.java
benclarkwood: That would be better. Swapping it for `this.outcomes.addAll(Arrays.asList(outcomes));`

benclarkwood: (I wish I knew enough Java to know which of these is "faster"

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/history/OfferOutcomeTracker.java
benclarkwood: Just be glad I didn't write my own ;)

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/OfferOutcomeResource.java
benclarkwood: I wanted a way to get the JSON in a browser for testing. Content-type would be better for actually production. Will make that change.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/OfferOutcomeResource.java
benclarkwood: Support + others have indicated that having a human consumable (e.g. HTML) endpoint is useful to them. I know that some users wont have the CLI Installed, so there's some utility in having it available directly from the service endpoint. If the CLI were to dump this out, it will obviously just fetch and parse the JSON.

benclarkwood:
@elezar thanks for the review. This is just an idea that was put forth at a meeting about testing improvements. Namely, that it would be great to be able to ask the scheduler "Hey! What were the last N offers you took / declined". To some extent this is a MVP that we can choose to discard and never merge if we want. I think it has some utility however, and having focus tested it against @jeid64 , support finds utility in it.

benclarkwood:
In terms of overall instrumentation my view is:

High level abstractions: This, other things that answer _very_ specific questions

Metrics: Rolling aggregates of timers, counters, etc

Logging: fine-grained events

We need all three to be successful.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/history/OfferOutcome.java
benclarkwood: I went with the boolean for now.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/OfferOutcomeResource.java
elezar: 👍

elezar:
Thanks @benclarkwood, you're absolutely right that a multi-level approach is the way to go. This looks good to me then, pending a green SDK build.

nickbp:
Bump plz would be nice to have this

benclarkwood:
It's on my list. I'll probably get to it later this week.

gabrielhartmann:
@benclarkwood This is too old.  I'm closing it.  Please re-open if you still want to do this.

benclarkwood:
@nickbp @elezar: I'd like to get this in so that we at least have it. We can then add in the things we wanted to have / refactor a bit.

PR #1747, opened 2017-10-17
[INFINITY-2552] Add basic logging to get_framework_srv_records

elezar:
The test `test_overlay.test_srv_records()` fails on 1.9. This PR adds some basic logging to make it simpler to debug this failure.


benclarkwood:
@elezar see https://jira.mesosphere.com/browse/INFINITY-2552 for what I believe is the fix to this.

PR #1751, opened 2017-10-17
[ELASTIC-149] Fix flaky pod replace tests

loren:
- Wait for recovery to begin and then wait for it to end
- Also moved `foldered_name` outside `try` in `configure_package` as it might get referenced in `finally` clause before it's initialized

### DIFF
--- a/frameworks/elastic/tests/test_sanity.py
+++ b/frameworks/elastic/tests/test_sanity.py
@@ -20,8 +20,8 @@ current_expected_task_count = config.DEFAULT_TASK_COUNT
 
 @pytest.fixture(scope='module', autouse=True)
 def configure_package(configure_security):
+    foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
     try:
-        foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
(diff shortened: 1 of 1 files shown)


frameworks/elastic/tests/test_sanity.py
```diff
@@ -221,28 +223,27 @@ def test_master_node_replace():
...
 
 
 @pytest.mark.recovery
 @pytest.mark.sanity
 def test_data_node_replace():
-    data_ids = sdk_tasks.get_task_ids(sdk_utils.get_foldered_name(config.SERVICE_NAME), 'data-0')
-    sdk_cmd.svc_cli(config.PACKAGE_NAME, sdk_utils.get_foldered_name(config.SERVICE_NAME), 'pod replace data-0')
-    sdk_tasks.check_tasks_updated(sdk_utils.get_foldered_name(config.SERVICE_NAME), 'data-0', data_ids)
-    # pre_test_setup will verify that the cluster becomes healthy again.
+    foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
```
falseperfection: Good catch. I thought I'd gotten all of these.

PR #1755, opened 2017-10-18
Add the v1 api test as a smoke test.

benclarkwood:


frameworks/helloworld/tests/test_sanity.py
```diff
@@ -181,13 +183,18 @@ def test_state_properties_get():
...
+    # should be in alphabetical order:
+    expected = [
+        "hello-0-server:task-status",
+        "hello-1-server:task-status",
+        "last-completed-update-type",
+        "world-0-server:task-status",
+        "world-1-server:task-status"]
+    # the properties list may also have a 'suppressed' bit, which would have been left behind by the
+    # prior version when upgrades were being tested during suite setup
+    expected_with_suppressed = list(expected)
```
elezar: Would it not be better to remove "suppressed" from the result set and test for equality then?

nickbp: That would work too. With this method we at least verify that the suppressed bit, if present, is at the expected index in the returned list.

nickbp:
Merging this now as it fixes a couple more blocking CI issues

PR #1756, opened 2017-10-18
Never set the principal in 'FrameworkInfo'.

greggomann:


gabrielhartmann:
@greggomann We can't accept this change.

greggomann:
@gabrielhartmann why not? Mesos master will store your authenticated principal for you if you don't supply one in the `FrameworkInfo`.

PR #1757, opened 2017-10-19
Fixes for scheduler test API before presentation

nickbp:
Adds Expect implementation for unreserves, adds some javadocs, uses better function names


nickbp:
Going to just merge this if/when `sdk` passes

PR #1762, opened 2017-10-21
[HDFS-483] Weird parsing workaround by number->string type conversion

NimaVaziri:


elezar:
Nima, this seems like a pretty large change, especially considering that we
now lose all validation that happens when values are supposed to be
integers.

I have not had a look at the ticket for more context.

Evan Lezar

frameworks/hdfs/universe/config.json
```diff
@@ -1523,24 +1522,24 @@
...
         "fs_s3n_multipart_uploads_block_size": {
-          "type": "integer",
+          "type": "string",
           "description": "The block size for multipart uploads to native S3 filesystem. Default size is 64MB.",
-          "default": 67108864
+          "default": "67108864"
         },
         "fs_s3n_multipart_copy_block_size": {
-          "type": "integer",
+          "type": "string",
```
elezar: Looking at https://developers.google.com/discovery/v1/type-format it seems that adding `"format": "int64"` here may be an option.

frameworks/hdfs/universe/config.json
```diff
@@ -424,34 +423,34 @@
           "default": 0.25
```
elezar: Any reason that this one was skipped? Or did we just replace `"integer"`?

frameworks/hdfs/universe/config.json
NimaVaziri: That seemed reasonable, though I tried it and it didn't actually do anything (for eg, I made the format be `uint64` for a config field but the UI validation didn't actually complain when I put a negative number for the value so this extra attribute wouldn't actually be useful)

frameworks/hdfs/universe/config.json
NimaVaziri: Just oversight, should not have been skipped.

NimaVaziri:
> I would postulate that we could get more crashes now because of non-numeric strings being passed.

There is truth to that, although we're deciding to pass integers as strings only for settings that we're passing through to the base tech. For other settings, we're still preserving our validation (i.e. `number` for cpus, mem, etc)

PR #1763, opened 2017-10-21
INFINITY-2538,INFINITY-2554 Update plan and pod APIs to reflect stopped pods

nickbp:
- Scheduler:
  - Adds the concept of a 'display status' to `Step`s in plans: A status label for operators which doesn't affect plan logic. (For example `STOPPED` instead of `RUNNING`: Both are "running" from a deployment/plan completion perspective). For the moment this is only used for stopped pods, but it's intended to be used for other custom states as they're introduced.
  - Renames the previous `Element.isInProgress()` helper function (which confusingly returned true if a step was `IN_PROGRESS`, or was in several other states) with `Element.isRunning()` which avoids the confusion with status enums (there is no `RUNNING` element status to conflict with). There was previously a _different_ `Element.isRunning()` but it wasn't actually being used.
- Testing: The service CLI module wasn't being upgraded as a part of the new `dcos svcname update` flow, so the wrong CLI module was being tested. Adds a command to update the CLI when in that flow.
- Plans API: Now shows the 'display status'. In practice this is the same as it was before, except for the case where a pod or task is `STOPPING`/`STOPPED`. This leaves open the window for other custom display statuses later.
- Pod status API:
  - This API hadn't really been getting much use, but it's about to be used much more shortly. This PR reformats the response tree to be more flexible long-term before it's effectively locked-in, while updating the CLI to support the new format.
  - The CLI can now also show a pretty tree of pod status, similar to what it currently can do for plan status. This is the new default, with a `--json` flag added for viewing the raw response (matching plan status behavior)


cli/commands/pods.go
```diff
@@ -24,29 +23,44 @@ func (cmd *podHandler) handleList(a *kingpin.Application, e *kingpin.ParseElemen
 }
 
 func (cmd *podHandler) handleStatus(a *kingpin.Application, e *kingpin.ParseElement, c *kingpin.ParseContext) error {
-	endpointPath := "v1/pod/status"
+	var endpointPath string
 	if len(cmd.PodName) > 0 {
```
elezar: Would it make sense to introduce a local variable here that captures this condition since it is used again later?

cli/commands/pods.go
```diff
@@ -24,29 +23,44 @@ func (cmd *podHandler) handleList(a *kingpin.Application, e *kingpin.ParseElemen
...
+	buf.WriteString(fmt.Sprintf("%s\n", response["service"]))
 
 	i := 0
-	for podName, rawPod := range podsJSON {
-		appendPod(&buf, podName, rawPod, i == len(podsJSON)-1)
+	rawPods, ok := response["pods"].([]interface{})
+	if !ok {
+		client.PrintMessageAndExit("Failed to parse JSON pods in response")
+	}
+	for _, rawPod := range rawPods {
```
elezar: Why not replace the `_` with `i` locally? Then there would be no need for the post-increment.

elezar: It is also not 100% on first reading that we are calculating `isLast` as the final parameter.

cli/commands/pods.go
```diff
@@ -56,55 +70,157 @@ func toPodTree(podsJSONBytes []byte) string {
...
+	rawPodInstances, ok := pod["instances"].([]interface{})
+	if !ok {
+		return
+	}
+	for i, rawPodInstance := range rawPodInstances {
+		appendServicePodInstance(buf, rawPodInstance, lastPod, i == len(rawPodInstances)-1)
+	}
+}
+
+func appendServicePodInstance(buf *bytes.Buffer, rawPodInstance interface{}, lastPod bool, lastPodInstance bool) {
```
elezar: Probably out of scope of this PR, but there has to be a neater way of doing this nesting. It would be good to be able to use the same function here. Especially considering how nested the "if last" structures get.

cli/commands/pods.go
```diff
@@ -56,55 +70,157 @@ func toPodTree(podsJSONBytes []byte) string {
...
+func toPodTree(podsJSONBytes []byte) string {
+	response, err := client.UnmarshalJSON(podsJSONBytes)
+	if err != nil {
+		client.PrintMessageAndExit(fmt.Sprintf("Failed to parse JSON in pod response: %s", err))
+	}
+
+	var buf bytes.Buffer
+	buf.WriteString(fmt.Sprintf("%s\n", response["name"]))
+
+	i := 0
```
elezar: Same comment regarding `i` and `_` here.

cli/commands/pods.go
```diff
@@ -56,55 +70,157 @@ func toPodTree(podsJSONBytes []byte) string {
...
+	}
+	for _, rawPod := range rawTasks {
+		appendPodTask(&buf, rawPod, i == len(rawTasks)-1)
+		i++
+	}
+
+	// Trim extra newline from end:
+	buf.Truncate(buf.Len() - 1)
+
+	return buf.String()
```
elezar: Instead of manipulating the buffer directly, could we call a trim-right method on the string representation? The trimming *assumes* that we have a trailing newline, and this may be easy to miss in future refactorings.

cli/commands/pods.go
nickbp: Cleaned this up somewhat: the parent is now passing down a prefix to be used in the output.

cli/commands/pods.go
nickbp: Good point, fixed here and elsewhere

cli/commands/pods.go
nickbp: Done

cli/commands/pods.go
nickbp: Switched to `strings.TrimRight()`, a bit more succinct and arguably a bit easier to understand

cli/commands/pods.go
elezar: Already looking a lot better. I found: 
https://github.com/campoy/tools/blob/master/tree/main.go
and
https://github.com/a8m/tree/blob/master/node.go
which may allow for a more general implementation. As I say -- really outside of the scope of this PR. Thanks for the effort.

docs/pages/operations-guide.md
```diff
@@ -851,9 +851,54 @@ $ curl -k -X POST -H "Authorization: token=$(dcos config show core.dcos_acs_toke
...
+## Stop a pod
+
+Stopping a pod relaunches it in an idle command state. This allows the operator to debug the contents of the pod, possibly making changes to fix problems. While these problems are often fixed by just replacing the pod, there may be cases where an in-place repair or other operation is needed. For example:
+- A pod which crash-loops as soon as it's started may need additional work to be performed.
+- Some services may _require_ that certain repair operations be performed manually when the task itself isn't running.
+Being able to put the pod in an offline but accessible state makes it easier to resolve these situations.
+
+After the pod has been stopped, it may be started again, at which point it will be restarted again and will resume running task(s) where it left off.
+
+Here is an example session where an `index-1` node is crash looping due to some corrupted data in a persistent volume. The operator stops the `index-1` pod, then uses `task exec` to repair the index. Following this, the operator starts the pod and it resumes normal operation:
```
gabrielhartmann: `node` --> `pod`

docs/pages/operations-guide.md
```diff
@@ -851,9 +851,54 @@ $ curl -k -X POST -H "Authorization: token=$(dcos config show core.dcos_acs_toke
...
+```bash
+$ dcos myservice debug pod stop index-1
+{
+  "pod": "index-1",
+  "tasks": [
+    "index-1-agent",
+    "index-1-node"
+  ]
+}
+$ dcos myservice plan status deploy
```
gabrielhartmann: This example only shows plan status.  What about the pod topology endpoint?

docs/pages/operations-guide.md
gabrielhartmann: The example session below makes it look like commands take effect instantaneously.  Maybe put some `STOPPING` and `STARTING` status messages in there.

docs/pages/operations-guide.md
nickbp: That's also available, would arguably be better to show that here than the plan status actually, in terms of being more firmly tied to pod/task structure

nickbp:
Will have a followup PR in a couple minutes with the docs fixes and anything else that should come up

docs/pages/developer-guide.md
```diff
@@ -462,7 +462,7 @@ For a fully detailed explanation of service packaging [see here](https://dcos.io
 
 * `marathon.json.mustache` - A mustache-templated file that provides a Marathon application definition.  Its mustache elements are rendered by the values present in the config.json and resource.json files.
 
-* `resource.json` - A list of URIs of all downloaded elements.This list allows your service to be deployed in datacenters without Internet access.
+* `resource.json` - A list of URIs of all downloaded elements. Any artifacts needed by your service _must_ be listed here, or else your service will fail to install to airgapped clusters. When an airgapped cluster installs your package, only the files listed here will be available. This list contains some required items in order for the service to run, such as `bootstrap.zip` (bootstrap utility described elsewhere in this guide) and `executor.zip` (custom executor for DC/OS 1.9 compatibility).
```
elezar: --> "This list contains some items required in order for the service to run,".

docs/pages/operations-guide.md
```diff
@@ -851,9 +851,54 @@ $ curl -k -X POST -H "Authorization: token=$(dcos config show core.dcos_acs_toke
 }
 ```
 
+## Stop a pod
+
+Stopping a pod relaunches it in an idle command state. This allows the operator to debug the contents of the pod, possibly making changes to fix problems. While these problems are often fixed by just replacing the pod, there may be cases where an in-place repair or other operation is needed. For example:
```
elezar: Not sure about the formatting, but we could start "For example:" as a new paragraph here.

docs/pages/operations-guide.md
```diff
@@ -851,9 +851,54 @@ $ curl -k -X POST -H "Authorization: token=$(dcos config show core.dcos_acs_toke
...
+- A pod which crash-loops as soon as it's started may need additional work to be performed.
+- Some services may _require_ that certain repair operations be performed manually when the task itself isn't running.
+Being able to put the pod in an offline but accessible state makes it easier to resolve these situations.
+
+After the pod has been stopped, it may be started again, at which point it will be restarted again and will resume running task(s) where it left off.
```
elezar: "restarted again" --> "restarted"

docs/pages/operations-guide.md
elezar: "crash-looping" (or change it above to "crash loops").

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -142,28 +142,30 @@ def test_pod_list():
...
-    for k, v in jsonobj.items():
-        assert re.match('(hello|world)-[0-9]+', k)
-        assert len(v) == 1
-        task = v[0]
-        assert len(task) == 3
-        assert re.match('(hello|world)-[0-9]+-server__[0-9a-f-]+', task['id'])
-        assert re.match('(hello|world)-[0-9]+-server', task['name'])
-        assert task['state'] == 'TASK_RUNNING'
+    jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, foldered_name, 'pod status --json', json=True)
+    assert jsonobj['service'] == foldered_name
```
elezar: Looking at how we compare these outputs, could we rather specify a dict to which we want to compare?  That makes it a lot simpler to parse and extend?

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PlansResource.java
```diff
@@ -167,7 +163,7 @@ public Response continueCommand(
             }
 
             boolean allInProgress = phases.stream()
-                    .filter(phz -> phz.isInProgress())
+                    .filter(phz -> phz.isRunning())
```
elezar: Minor: `phz` -> `phase`.

nickbp: I had meant to change this as well, but it conflicts with the `String phase` method param

frameworks/helloworld/tests/test_sanity.py
nickbp: An issue is that other tests could add/remove pods, or the ordering may not be fixed. This is just checking that the resulting data looks valid rather than matching an exact value.

docs/pages/operations-guide.md
nickbp: Updated the earlier mention to 'immediately crashes', in PR #1771

docs/pages/operations-guide.md
nickbp: Done. #1771

docs/pages/operations-guide.md
nickbp: Added a paragraph break in #1771

docs/pages/developer-guide.md
nickbp: Done. #1771

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PlansResource.java
elezar: OK. Didn't see that. In this case one could use:
```java
boolean allInProgress = phases.stream()
                                                    .filter(Phase::isRunning)
```
For that matter, looking at the code around the diff again, one could also use `allMatch` here instead of comparing the size of the filtered stream to the original list.

PR #1764, opened 2017-10-23
Added KDC marathon app def

NimaVaziri:
@benclarkwood This is taken from the old kerberos HDFS branch.

### DIFF
--- a/frameworks/hdfs/tools/kdc.json
+++ b/frameworks/hdfs/tools/kdc.json
@@ -1,0 +1,40 @@
+{
+  "id": "/kerberos",
+  "instances": 1,
+  "cpus": 1,
+  "mem": 512,
+  "labels": {
+    "DCOS_SERVICE_NAME": "kerberos"
+  },
+  "env": {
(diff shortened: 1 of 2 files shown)


frameworks/hdfs/tools/kdc.json
```diff
@@ -0,0 +1,40 @@
+{
+  "id": "/kerberos",
+  "instances": 1,
+  "cpus": 1,
+  "mem": 512,
+  "labels": {
+    "DCOS_SERVICE_NAME": "kerberos"
+  },
+  "env": {
+    "PRINCIPAL_hdfs": "hdfs@LOCAL:secret",
```
benclarkwood: What do these envvars do?

NimaVaziri: Nothing of significance AFAICT.

Just committed the file as it was.

benclarkwood: Kk. At least gives us all a KDC to mess around with.

PR #1767, opened 2017-10-24
Remove unused permissions

nickbp:
If spark CI needs these again someday, they can be added by spark CI.


benclarkwood:
Trying a strict run here: https://teamcity.mesosphere.io/viewQueued.html?itemId=831604&tab=queuedBuildOverviewTab

PR #1768, opened 2017-10-25
[INFINITY-2515]: Only check for updated tasks if there is a config change

elezar:
Backport of #1754


benclarkwood:
I didn't look into why this C* test is always failing, but it did have lots of logs. So it may finally be possible to investigate it.

elezar:
Were the Cassandra tests fixed?

PR #1771, opened 2017-10-25
INFINITY-2584/2596: Rename stop to pause, and followup fixes from PR 1763

nickbp:
Docs fixes and whatever else comes up following PR #1763


nickbp:
TODO: need to also rename `start` to `resume`

elezar:
> TODO: need to also rename start to resume

Is still think `||` and `|>` are the way to go. Or even: ⏯

PR #1779, opened 2017-10-27
INFINITY-2573 Add opt-in bit to pod specs allowing truncation of pods

nickbp:
Just updates the spec and config validation to implement the new bit. The actual shrink flow will be implemented separately. The YAML reference has also been updated a bit to reflect current reality following some recent prior changes. Going to leave this unmentioned in the ops guide until it's actually usable (and until the actual decommission behavior is more quantifiable).

### DIFF
--- a/docs/pages/reference/yaml-reference.md
+++ b/docs/pages/reference/yaml-reference.md
@@ -32,6 +32,10 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
     Custom zookeeper URL for storing scheduler state. Defaults to `master.mesos:2181`.
 
+  * `user`
+
+    The system user to run the service's scheduler and pods as. Availability of usernames depends on the cluster. If DC/OS Security is enabled, this may need to be set to `nobody`, or the administrator may be required to configure custom permissions allowing the service to run as a different user.
+
 * `pods`
 
   This section contains a listing of all pod types managed by the service.
@@ -50,7 +54,11 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
   * `count`
 
-    The number of pods of this type to be deployed. This may either be hardcoded or exposed to end users via mustache templating.
+    The number of pods of this type to be deployed. This may either be hardcoded or exposed to end users via mustache templating. This value may be always increased after the service has been deployed, but it can only be decreased if `allow-decomission` is `true`.
+
+  * `allow-decommission`
+
+    Whether to allow the `count` to be decreased by an operator in a configuration update. For safety reasons this defaults to `false`, but the service developer may set this field to `true` to explicitly allow scale-down on a per-pod basis.
 
   * `container`
 
@@ -125,6 +133,14 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
   * `volume`/`volumes`
 
     One or more persistent volumes to be mounted into the pod environment. These behave the same as volumes on a task or resource set, but are guaranteed to be shared between tasks in a pod. Although volumes defined on a task currently behave the same way, individual tasks will not be able to access volumes defined by another task in the future.
+
+  * `pre-reserved-role`
+
+    TODO(nickbp): Explain Mesos pre-reserved roles.
+
+  * `share-pid-namespace`
+
+    Whether the tasks within this pod should share the same process id namespace (`true`), or whether pid namespaces should be distinct for every task in the pod (`false`).
 
   * `tasks`
 
@@ -374,10 +390,6 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
         This can be set either to `TLS` for PEM encoded private key file, certificate and CA bundle or `KEYSTORE` for certificate and private key to be delivered in a separate keystore file and CA bundle in other truststore file.
 
       For detailed information see the [SDK Developer Guide](developer-guide.html#tls).
-
-  * `user`
-
-    The system user to run this pod as. The available users depend on the administrator's cluster. If clusters are using DC/OS Security enabled, this may need to be set to `nobody`.
 
 * `plans`
 
--- a/docs/pages/reference/yaml-reference.md
+++ b/docs/pages/reference/yaml-reference.md
@@ -58,7 +58,7 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
   * `allow-decommission`
 
-    Whether to allow the `count` to be decreased by an operator in a configuration update. For safety reasons this defaults to `false`, but the service developer may set this field to `true` to explicitly allow scale-down on a per-pod basis.
+    Whether to allow this pod's `count` to be decreased by an operator in a configuration update. For safety reasons this defaults to `false`, but the service developer may set this field to `true` to explicitly allow scale-down on a per-pod basis.
 
   * `container`
 
@@ -136,7 +136,7 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
   * `pre-reserved-role`
 
-    TODO(nickbp): Explain Mesos pre-reserved roles.
+    Ensures that this pod only consumes resources against a role which has already been statically assigned within the cluster. This is mainly useful for placing pods within a predefined quota, or otherwise assigning them a specific set of resources. For example, DC/OS clusters have a convention of using the `slave_public` role for machines which are not firewalled. Pods which have their `pre-reserved-role` set to `slave_public` will be placed on those machines so that they are visible outside the cluster.
 
   * `share-pid-namespace`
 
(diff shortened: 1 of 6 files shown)


docs/pages/reference/yaml-reference.md
```diff
@@ -32,6 +32,10 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
     Custom zookeeper URL for storing scheduler state. Defaults to `master.mesos:2181`.
 
+  * `user`
+
+    The system user to run the service's scheduler and pods as. Availability of usernames depends on the cluster. If DC/OS Security is enabled, this may need to be set to `nobody`, or the administrator may be required to configure custom permissions allowing the service to run as a different user.
```
elezar: I thought the scheduler runs as `root`?

docs/pages/reference/yaml-reference.md
```diff
@@ -50,7 +54,11 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
   * `count`
 
-    The number of pods of this type to be deployed. This may either be hardcoded or exposed to end users via mustache templating.
+    The number of pods of this type to be deployed. This may either be hardcoded or exposed to end users via mustache templating. This value may be always increased after the service has been deployed, but it can only be decreased if `allow-decomission` is `true`.
+
+  * `allow-decommission`
+
+    Whether to allow the `count` to be decreased by an operator in a configuration update. For safety reasons this defaults to `false`, but the service developer may set this field to `true` to explicitly allow scale-down on a per-pod basis.
```
elezar: "the pod `count`" or just "~the~ `count`"?

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/PodSpecsCannotShrink.java
```diff
@@ -32,16 +32,24 @@
         Collection<ConfigValidationError> errors = new ArrayList<>();
         for (PodSpec oldPod : oldConfig.get().getPods()) {
             PodSpec newPod = newPods.get(oldPod.getType());
-            if (newPod == null) {
+            if (newPod == null && !oldPod.getAllowDecommission()) {
```
elezar: It seems as if we could just `continue` on `oldPod.getAllowDecommission()` instead of adding this to each of the conditionals?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
```diff
@@ -28,6 +28,8 @@
     @NotNull
     @Min(0)
     private final Integer count;
+    @NotNull
+    private Boolean allowDecommission;
```
elezar: Does this still default to `false` if we add the `@NotNull` annotation?

elezar: (do we ever instantiate this class through something other than a builder?)

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
```diff
@@ -65,7 +67,8 @@ public DefaultPodSpec(
             @JsonProperty("volumes") Collection<VolumeSpec> volumes,
             @JsonProperty("pre-reserved-role") String preReservedRole,
             @JsonProperty("secrets") Collection<SecretSpec> secrets,
-            @JsonProperty("share-pid-namespace") Boolean sharePidNamespace) {
+            @JsonProperty("share-pid-namespace") Boolean sharePidNamespace,
+            @JsonProperty("allow-decommission") Boolean allowDecommission) {
```
elezar: **Out of scope of this PR:** but it seems as if this is duplicated in `RawPod`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawPod.java
```diff
@@ -29,6 +29,7 @@
     private final String preReservedRole;
     private final WriteOnceLinkedHashMap<String, RawSecret> secrets;
     private final Boolean sharePidNamespace;
+    private final Boolean allowDecommission;
```
elezar: Should we initialise this to `false` explicitly?

docs/pages/reference/yaml-reference.md
nickbp: No it runs as the user specified here -- i.e. as the same user the service itself runs as. (had checked this via the code)

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/PodSpecsCannotShrink.java
nickbp: That would indeed be better -- the catch is that it's checking oldPod only when newPod is null. Otherwise it checks newPod. However these clauses are indeed pretty dense now. Broke them up a bit, adding a continue in the null case. It should hopefully be a bit less confusing

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
nickbp: `Raw*` are the direct YAML schema, while this layer is what's used internally. `YAMLToInternalMappers` handles translating from the prior to the latter

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
nickbp: These classes are serialized into ZK, I think this gets triggered during deserialization from ZK.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawPod.java
nickbp: Eh I was just reusing the current `sharePidNamespace` behavior here

docs/pages/reference/yaml-reference.md
nickbp: Went with "this pod's `count`"

docs/pages/reference/yaml-reference.md
benclarkwood: That's theoretically true with the caveat that the user specified in the Marathon app definition much be the same as the one set in the YAML.

docs/pages/reference/yaml-reference.md
```diff
@@ -50,7 +54,11 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
   * `count`
 
-    The number of pods of this type to be deployed. This may either be hardcoded or exposed to end users via mustache templating.
+    The number of pods of this type to be deployed. This may either be hardcoded or exposed to end users via mustache templating. This value may be always increased after the service has been deployed, but it can only be decreased if `allow-decomission` is `true`.
```
benclarkwood: s/may be always/may always be

docs/pages/reference/yaml-reference.md
```diff
@@ -126,6 +134,14 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
     One or more persistent volumes to be mounted into the pod environment. These behave the same as volumes on a task or resource set, but are guaranteed to be shared between tasks in a pod. Although volumes defined on a task currently behave the same way, individual tasks will not be able to access volumes defined by another task in the future.
 
+  * `pre-reserved-role`
+
+    Ensures that this pod only consumes resources against a role which has already been statically assigned within the cluster. This is mainly useful for placing pods within a predefined quota, or otherwise assigning them a specific set of resources. For example, DC/OS clusters have a convention of using the `slave_public` role for machines which are not firewalled. Pods which have their `pre-reserved-role` set to `slave_public` will be placed on those machines so that they are visible outside the cluster.
+
+  * `share-pid-namespace`
+
+    Whether the tasks within this pod should share the same process id namespace (`true`), or whether pid namespaces should be distinct for every task in the pod (`false`).
```
benclarkwood: What is the default behavior?

docs/pages/reference/yaml-reference.md
benclarkwood: The second sentence should read:

"In DC/OS Enterprise, if the security mode is set to `strict`, the service account used by the service must have permissions to launch tasks as the assigned user."

Also, is there a default? Or is it a required field?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultPodSpec.java
benclarkwood: From a quick peek at the code, we run the validation either in the builder or in the constructor that deserialization would use (or both). Should be a-ok.

benclarkwood: The only instantiation paths are via builders or the `@JsonCreator` constructor. And the builders use the latter.

nickbp:
Going to do a bit of CI trickery to avoid waiting for another CI run: Going to merge this and then make the docs changes in a separate docs-only PR, which will then be merged as well. Avoids triggering CI again despite only docs being changed.

nickbp:
(done via #1783)

PR #1783, opened 2017-10-27
Followup changes to YAML reference from #1779

nickbp:


nickbp:
Just implementing the docs-specific requests from approved #1779

PR #1784, opened 2017-10-27
[INFINITY-2521] Kerberos testing infra

benclarkwood:


testing/sdk_tasks.py
```diff
@@ -116,3 +117,8 @@ def fn():
 
     if exit_status != 0:
         raise RuntimeError('Failed to kill task with pattern "{}", exit status: {}'.format(pattern, exit_status))
+
+
+def task_exec(task_name, command, **kwargs):
```
elezar: Does this replace the separate implementations in some of the `test_*` files?

tools/kdc.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Exit immediately on errors
+set -e
+
+# default values
+ssh_path="${HOME}/.ssh/ccm.pem"
+PRINCIPALS_FILE_NAME=""
+PRINCIPALS_FILE_DIR=""
+
+function usage()
```
elezar: This duplicates the functionality in `kdc.py`. Could we rather only maintain one of them?

tools/kdc.sh
```diff
@@ -0,0 +1,49 @@
...
+  docker run --rm \
+    -e CLUSTER_URL="$CLUSTER_URL" \
+    -e PRINCIPALS_FILE_NAME="$PRINCIPALS_FILE_NAME" \
+    -v $(pwd):/build \
+    -v $ssh_path:/ssh/key \
+    -w /build \
+    -t \
+    -i \
+    nvaziri/kdc:dev_1 \
+    bash tools/kdc_controller.sh $1 $2
```
elezar: Should one not also mount `PRINCIPALS_FILE_DIR`?

There is also not a real reason that these two commands could not be the same. They could both be replaced by:
```
docker run --rm \
     -e CLUSTER_URL="$CLUSTER_URL" \
     -e PRINCIPALS_FILE_NAME="$PRINCIPALS_FILE_NAME" \
     -v $(pwd):/build \
     -v $ssh_path:/ssh/key \
     -w /build \
     -t \
     -i \
     nvaziri/kdc:dev_1 \
     bash tools/kdc_controller.sh $*
```

testing/sdk_marathon.py
```diff
@@ -32,6 +34,55 @@ def fn():
     return config
 
 
+def install_app(app_def_path: str) -> (bool, str):
```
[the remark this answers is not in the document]
benclarkwood: Yes, please do. We should also add a non-aggressive linter/static checker (No full Pep8 please)

benclarkwood:
@nima Please work with @elezar to get these comments addressed.

Let's err on the side of "get the common tooling checked in" during that process. Please focus on getting the python code in over the "nice to have" scripting. We need the tooling now.

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -12,7 +12,6 @@
 import shakedown
 from tests import config
 
-
```
elezar: Let's remove this unrelated change.

testing/sdk_auth.py
```diff
@@ -0,0 +1,383 @@
...
+        host_id=self.kdc_host_id,
+        framework_id=self.framework_id,
+        task_id=self.task_id,
+        keytab_file=self.keytab_file_name,
+        linux_user=LINUX_USER
+    )
+    cmd = "node ssh --master-proxy --mesos-id={host_id} --option StrictHostKeyChecking=no '{copy_cmd}'".format(
+        host_id=self.kdc_host_id, copy_cmd=copy_cmd)
+    try:
+        sdk_cmd.run_cli(cmd)
```
benclarkwood: @NimaVaziri did this work locally? It failed in CI with:
```
There is no SSH_AUTH_SOCK env variable, which likely means you aren't running `ssh-agent
```

The existing pattern we use is to use the Shakedown SSH utilities. @mellenburg given that Paramiko has issues often, is it possible to make a direct SSH work?

elezar: In other tests we have: `shakedown.run_command_on_agent` which I assume encapsulates this to some extent.

benclarkwood: Yep, that runs via Paramiko ("native" python ssh)

testing/sdk_cmd.py
```diff
@@ -14,9 +14,9 @@
 log = logging.getLogger(__name__)
 
 
-def request(method, url, retry=True, log_args=True, **kwargs):
+def request(method, url, retry=True, verify=None, log_args=True, **kwargs):
```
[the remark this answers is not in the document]
elezar: Looking at the call in `sdk_auth._get_master_public_ip` there might be an issue there.

testing/sdk_auth.py
```diff
@@ -0,0 +1,383 @@
...
+def _get_master_public_ip() -> str:
+    """
+    :return (str): The public IP of the master node in the DC/OS cluster.
+    """
+    dcos_url, headers = sdk_security.get_dcos_credentials()
+    cluster_metadata_url = "{cluster_url}/metadata".format(cluster_url=dcos_url)
+    request_metadata = {
+            "--header": "Authorization: {token}".format(token=headers["Authorization"])
+    }
+    raw_response = sdk_cmd.request("GET", cluster_metadata_url, request_metadata, verify=False)
```
elezar: I think `request_metadata` should be a `kwarg` here. Otherwise it will be mapped to `retry` in `sdk_cmd.request`.

testing/sdk_auth.py
NimaVaziri: Yeah this worked locally, but saw that it failed CI. Will need to address it.

testing/sdk_marathon.py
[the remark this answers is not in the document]
NimaVaziri: https://jira.mesosphere.com/browse/INFINITY-2535

testing/sdk_tasks.py
NimaVaziri: There was only one instance of that, in `test_soak.py` for the helloworld tests, which was updated.

testing/sdk_auth.py
elezar: I was able to run this locally in a docker container:
```bash
$ eval "$(ssh-agent -s)"
$ ssh-add -k /ssh/key
```
(we could try adding these ... note the `:dev` image has a different ssh key)

elezar: The `:dev` image has a ket at `/root/.ssh/id_rsa`, which one could try adding directly to the ssh commands or having the agent running as above. Ideally, we need to "play nice" with `dcos-launch` clusters that generate the keys on the fly, but we could address this as a next step.

testing/sdk_tasks.py
elezar: (sorry, I was probably looking at outdated code)

elezar:
Thanks @NimaVaziri for addressing the comments so quickly. There is still an open question with regards to the `sdk_cmd.request()` call in `sdk_auth`, but other than this it looks pretty good now. 

Do we expect the hdfs tests to pass with these changes in?

testing/sdk_cmd.py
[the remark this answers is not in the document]
NimaVaziri: All the existing calls on this function explicitly use the argument names to set them and don't rely on argument ordering. Nonetheless, I've moved the `verify` parameter to the end regardless.

NimaVaziri:
@elezar Removed the call to `sdk_cmd.request` as it wasn't actually needed.

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+    :return (str): The ID of the framework running the KDC app.
+    """
+    log.info("Getting framework id")
+    raw_tasks = sdk_cmd.run_cli("task --json")
+    if raw_tasks:
+        tasks = json.loads(raw_tasks)
+        for task in tasks:
+            if task["name"] == KERBEROS_APP_ID:
+                log.info("Framework id is {}".format(task["framework_id"]))
+                return task["framework_id"]
```
elezar: This code is the same as for `_get_task_id()` above. It doesn't need to be addressed now, but we could consider getting the entire task object, and then returning the id or framework id as required.

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+def _get_host_id() -> str:
+    """
+    :return (str): The ID of the node running the KDC app.
+    """
+    log.info("Getting host id")
+    raw_tasks = sdk_cmd.run_cli("task --json")
+    if raw_tasks:
+        tasks = json.loads(raw_tasks)
+        for task in tasks:
+            if task["name"] == KERBEROS_APP_ID:
```
elezar: Same here (see comment w.r.t framework id and task id).

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+    :param host_id (str): The ID of the host, used to look up the appropriate node.
+    :return (str): Name of the host running the KDC app.
+    """
+    log.info("Getting host name")
+    raw_nodes = sdk_cmd.run_cli("node --json")
+    if raw_nodes:
+        nodes = json.loads(raw_nodes)
+        for node in nodes:
+            if node["id"] == host_id:
+                log.info("Host name is {host_name}".format(host_name=node["hostname"]))
```
elezar: Same here (see comment w.r.t framework id and task id).

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+    """
+    :return (str): The public IP of the master node in the DC/OS cluster.
+    """
+    dcos_url, headers = sdk_security.get_dcos_credentials()
+    cluster_metadata_url = "{cluster_url}/metadata".format(cluster_url=dcos_url)
+    raw_response = sdk_cmd.request("GET", cluster_metadata_url, verify=False)
+    if not raw_response.ok:
+        raise RuntimeError("Unable to get the master node's public IP address: {err}".format(err=repr(raw_response)))
+
+    response = json.loads(raw_response.text)
```
elezar: Is `raw_response` a wrapper around a `request`  response object? If so, then it may have a `.json()` method that one could use directly. (as a future note)

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+    log.info("Master public ip is {public_ip}".format(public_ip=public_ip))
+    return public_ip
+
+
+def _create_temp_working_dir() -> str:
+    """
+    Creates a temporary working directory to enable setup of the Kerberos environment.
+    :return (str): The path of the working directory.
+    """
+    dir_name = "/tmp/{kerberos}_{current_time}".format(kerberos=KERBEROS_APP_ID, current_time=str(time.time()))
```
elezar: We could consider using https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory here instead. Not critical though.

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+    Copies the keytab that was generated inside the container running the KDC server to the localhost
+    so it can be uploaded to the secret store later. This must be done in multiple steps as we have to jump through
+    hoops.
+
+    The keytab will end up in path: <temp_working_dir>/<keytab_file>
+    """
+    log.info("Copying {} to the temp working directory".format(self.keytab_file_name))
+
+    # 1. copy from within container to private agent
+    copy_cmd = "sudo cp /var/lib/mesos/slave/slaves/{host_id}/frameworks/{framework_id}/executors/{task_id}/runs/latest/{keytab_file} /home/{linux_user}/{keytab_file}".format(
```
elezar: Could we split out the source and destination paths here giving them descriptive names?

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+    except dcos.errors.DCOSException as e:
+        raise RuntimeError("Failed to chown keytab file: {}".format(repr(e)))
+
+    # 3. Copy from private agent to leader
+    scp_cmd = "scp {linux_user}@{kdc_host_name}:/home/{linux_user}/{keytab_file} /home/{linux_user}/{keytab_file}".format(
+        linux_user=LINUX_USER,
+        kdc_host_name=self.kdc_host_name,
+        keytab_file=self.keytab_file_name
+    )
+    cmd = "eval $(ssh-agent -s) && ssh-add /ssh/key && dcos node ssh --master-proxy --leader --option StrictHostKeyChecking=no '{}'".format(scp_cmd)
```
elezar: Should the key be configurable? This will, for example, not work in our local `test.sh` runs, ad we there mount the key file at a different path.

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+        Eg. hdfs/name-0-node.hdfs.autoip.dcos.thisdcos.directory@LOCAL
+
+        :param principals: The list of principals to be added to KDC.
+        """
+        # TODO: Perform sanitation check against validity of format for all given principals and raise an
+        # exception when the format of a principal is invalid.
+        self.principals = principals
+
+        log.info("Adding the following list of principals to the KDC: {principals}".format(principals=principals))
+        add_principal_cmd = "task exec {task_id} /usr/sbin/kadmin -l add --use-defaults --random-password {principal}"
```
elezar: Later: We could have a `run_kadmin` method to wrap this.

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+        _copy_file_to_localhost(self)
+
+    def __create_and_upload_secret(self):
+        """
+        This method base64 encodes the keytab file and creates a secret with this encoded content so the
+        tasks can fetch it.
+        """
+        log.info("Creating and uploading the keytab file to the secret store")
+
+        cmd = "package install --yes --cli dcos-enterprise-cli"
```
elezar: Question: is there a benefit in doing this here, instead of as part of the test setup or earlier in the process? Here we may install the enterprise CLI multiple times.

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+    def get_port(self):
+        return str(self.kdc_port)
+
+    def get_keytab_path(self):
+        return self.keytab_secret_path
+
+    def cleanup(self):
+        sdk_marathon.destroy_app(KERBEROS_APP_ID)
+        if os.path.exists(self.temp_working_dir):
+            shutil.rmtree(self.temp_working_dir)
```
elezar: This is pretty dangerous if we set the wrong folder here. If we switch to https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory, this should not be required, as the directory is removed as soon as the context is destroyed.

testing/sdk_auth.py
```diff
@@ -0,0 +1,376 @@
...
+
+        delete_secret_cmd = "security secrets delete {}".format(self.keytab_secret_path)
+        try:
+            sdk_cmd.run_cli(delete_secret_cmd)
+        except RuntimeError as e:
+            raise RuntimeError("Failed to delete secret for the base64-encoded keytab file: {}".format(repr(e)))
```
elezar: As far as I remember, in the hello-world secret tests, we are a bit more flexible in terms of the success of this command. If, for example, the secret was not created due to an error, we would not expect this to pass.

elezar: For later: We could add an `sdk_secrets` module that wraps this kind of operation for us.

NimaVaziri:
@elezar thanks for the comments, I'll address them in an upcoming PR once this is merged.

testing/sdk_auth.py
NimaVaziri: Done.

testing/sdk_auth.py
NimaVaziri: Done.

testing/sdk_auth.py
NimaVaziri: Not quite the same here, as we're going through the nodes and not the tasks.

testing/sdk_auth.py
NimaVaziri: Yup, eradicated wrapper.

testing/sdk_auth.py
NimaVaziri: 👍

testing/sdk_auth.py
NimaVaziri: Done.

testing/sdk_auth.py
NimaVaziri: So it seems like the scripts use `/ssh/key` as the mount directory, in both local runs and CI. I mean, this worked for me locally, so I may be missing something you're referring to?

testing/sdk_auth.py
NimaVaziri: Good suggestion, semantically cleaner.

testing/sdk_auth.py
NimaVaziri: No benefit, just did it as part of the context of dealing with secrets.

Moved to end of constructor.

testing/sdk_auth.py
NimaVaziri: Although the `temp_working_dir` was set pretty explicitly and wasn't configurable to be changed, yes, this isn't good to do.

Now letting `TemporaryDirectory` handle the cleanup.

testing/sdk_auth.py
NimaVaziri: Yeah might be worth considering. Add this thought as a comment.

testing/sdk_auth.py
elezar: I checked again, and I stand corrected. I think there was some other reason that this was not working for me.

PR #1785, opened 2017-10-28
[INFINITY-2016] Override commands for pause a pod

gabrielhartmann:
This PR isn't perfect, but I wanted to get it out so people can comment.  It passes @nickbp 's integration tests.  I'll add some unit tests and rebase on master.

### DIFF
--- a/frameworks/helloworld/tests/test_recovery.py
+++ b/frameworks/helloworld/tests/test_recovery.py
@@ -90,7 +90,11 @@ def test_pod_stop():
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod info hello-0', json=True)
     assert len(jsonobj) == 1
     assert old_agent == jsonobj[0]['info']['slaveId']['value']
-    assert 'echo This task is PAUSED, sleeping ... && ./bootstrap --resolve=false && while true; do sleep 1209600; done' == jsonobj[0]['info']['command']['value']
+    cmd = jsonobj[0]['info']['command']['value']
+    assert 'This task is PAUSED' in cmd
+
+    readiness_check = jsonobj[0]['info']['check']['command']['command']['value']
+    assert 'exit 1' == readiness_check
 
     # check STOPPED state in plan and in pod status:
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod status hello-0 --json', json=True)
--- a/frameworks/helloworld/tests/test_recovery.py
+++ b/frameworks/helloworld/tests/test_recovery.py
@@ -57,7 +57,17 @@ def test_pod_restart():
 
 @pytest.mark.sanity
 @pytest.mark.recovery
+@pytest.mark.dcos_min_version('1.10')
 def test_pod_pause_resume():
+    test_pod_pause_resume_internal()
+
+@pytest.mark.sanity
+@pytest.mark.recovery
+@pytest.mark.dcos_min_version('1.9')
+def test_pod_pause_resume():
+    test_pod_pause_resume_internal(False)
+
+def test_pod_pause_resume_internal(validateReadinessCheck=True):
     '''Tests pausing and resuming a pod. Similar to pod restart, except the task is marked with a PAUSED state'''
 
     # get current agent id:
@@ -93,8 +103,9 @@ def test_pod_pause_resume():
     cmd = jsonobj[0]['info']['command']['value']
     assert 'This task is PAUSED' in cmd
 
-    readiness_check = jsonobj[0]['info']['check']['command']['command']['value']
-    assert 'exit 1' == readiness_check
+    if validateReadinessCheck:
+        readiness_check = jsonobj[0]['info']['check']['command']['command']['value']
+        assert 'exit 1' == readiness_check
 
     # check PAUSED state in plan and in pod status:
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod status hello-0 --json', json=True)
--- a/frameworks/helloworld/tests/test_recovery.py
+++ b/frameworks/helloworld/tests/test_recovery.py
@@ -90,19 +90,13 @@ def test_pod_stop():
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod info hello-0', json=True)
     assert len(jsonobj) == 1
     assert old_agent == jsonobj[0]['info']['slaveId']['value']
-    # TODO: this check should fail once command updates for stopped pods is implemented:
-    #       the task command should be updated to some new value that all stopped tasks would have
-    #assert 'stopped-task-cmd-here?' == jsonobj[0]['info']['command']['value']
-    assert old_cmd == jsonobj[0]['info']['command']['value'] # remove this and uncomment/update above
+    assert 'echo This task is PAUSED, sleeping ... && ./bootstrap --resolve=false && while true; do sleep 1209600; done' == jsonobj[0]['info']['command']['value']
 
     # check STOPPED state in plan and in pod status:
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod status hello-0 --json', json=True)
     assert len(jsonobj['tasks']) == 1
     assert jsonobj['tasks'][0]['name'] == 'hello-0-server'
-    # TODO: this check should fail once command updates for stopped pods is implemented:
-    #       the command updater should set the override status to COMPLETE as the task is relaunched
-    #assert jsonobj['tasks'][0]['status'] == 'STOPPED'
-    assert jsonobj['tasks'][0]['status'] == 'STOPPING' # remove this and uncomment above
+    assert jsonobj['tasks'][0]['status'] == 'STOPPED'
     phase = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'plan status deploy --json', json=True)['phases'][0]
     assert phase['name'] == 'hello'
     assert phase['status'] == 'COMPLETE'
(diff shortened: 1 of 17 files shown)


frameworks/helloworld/tests/test_recovery.py
```diff
@@ -90,19 +90,13 @@ def test_pod_stop():
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod info hello-0', json=True)
     assert len(jsonobj) == 1
     assert old_agent == jsonobj[0]['info']['slaveId']['value']
-    # TODO: this check should fail once command updates for stopped pods is implemented:
-    #       the task command should be updated to some new value that all stopped tasks would have
-    #assert 'stopped-task-cmd-here?' == jsonobj[0]['info']['command']['value']
-    assert old_cmd == jsonobj[0]['info']['command']['value'] # remove this and uncomment/update above
+    assert 'echo This task is PAUSED, sleeping ... && ./bootstrap --resolve=false && while true; do sleep 1209600; done' == jsonobj[0]['info']['command']['value']
```
nickbp: Feel free to just do a contains() or similar if this is too onerous. Just trying to validate the command has indeed been changed.

If you want, could also check that the health/readiness checks were updated as expected.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
```diff
@@ -20,6 +21,15 @@
...
     STOPPED("STOPPED", "STOPPING");
 
+    // Sleep for 2 weeks when pausing
+    public static final String PAUSE_COMMAND =
+            String.format(
+                    "echo This task is PAUSED, sleeping ... && " +
+                            "./bootstrap --resolve=false && while true; do sleep %d; done",
+                    Constants.LONG_DECLINE_SECONDS);
+
+    public static final String PAUSE_READINESS_COMMAND = "exit 1";
```
nickbp: Maybe leave a comment explaining the reasoning for doing this in the readiness command.

frameworks/helloworld/tests/test_recovery.py
elezar: Yes, this type of comparison seem pretty brittle and dependent on whitespace and general formatting.

One suggestion would be to split on `&&` and use `contains()` on each of the three expected parts. The idea of checking for a changed command is also a good option.

frameworks/helloworld/tests/test_recovery.py
```diff
@@ -90,19 +90,13 @@ def test_pod_stop():
...
 
     # check STOPPED state in plan and in pod status:
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod status hello-0 --json', json=True)
     assert len(jsonobj['tasks']) == 1
     assert jsonobj['tasks'][0]['name'] == 'hello-0-server'
-    # TODO: this check should fail once command updates for stopped pods is implemented:
-    #       the command updater should set the override status to COMPLETE as the task is relaunched
-    #assert jsonobj['tasks'][0]['status'] == 'STOPPED'
-    assert jsonobj['tasks'][0]['status'] == 'STOPPING' # remove this and uncomment above
+    assert jsonobj['tasks'][0]['status'] == 'STOPPED'
```
elezar: Should this now be `PAUSED`?

frameworks/helloworld/tests/test_recovery.py
```diff
@@ -90,19 +90,13 @@ def test_pod_stop():
     jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod info hello-0', json=True)
     assert len(jsonobj) == 1
     assert old_agent == jsonobj[0]['info']['slaveId']['value']
-    # TODO: this check should fail once command updates for stopped pods is implemented:
-    #       the task command should be updated to some new value that all stopped tasks would have
-    #assert 'stopped-task-cmd-here?' == jsonobj[0]['info']['command']['value']
-    assert old_cmd == jsonobj[0]['info']['command']['value'] # remove this and uncomment/update above
+    assert 'echo This task is PAUSED, sleeping ... && ./bootstrap --resolve=false && while true; do sleep 1209600; done' == jsonobj[0]['info']['command']['value']
 
     # check STOPPED state in plan and in pod status:
```
elezar: PAUSED state?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -86,14 +86,26 @@ public OfferEvaluator(
...
+
             PodInfoBuilder podInfoBuilder = new PodInfoBuilder(
                     podInstanceRequirement,
                     serviceName,
                     getTargetConfig(podInstanceRequirement, thisPodTasks.values()),
                     schedulerConfig,
                     thisPodTasks.values(),
                     stateStore.fetchFrameworkId().get(),
-                    useDefaultExecutor);
+                    useDefaultExecutor,
```
elezar: Since this is a builder, could we not construct it earlier and then call an `addOverride()` method in the task loop above? It seems as if extending the argument list for the construction of a builder goes against the builder pattern a little. (or is this not that kind of Builder?)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -432,12 +452,18 @@ private void setHealthCheck(
...
             PodInstance podInstance,
-            TaskSpec taskSpec) {
+            TaskSpec taskSpec,
+            GoalStateOverride override) {
         if (!taskSpec.getHealthCheck().isPresent()) {
             LOGGER.debug("No health check defined for taskSpec: {}", taskSpec.getName());
             return;
         }
 
+        if (override.equals(GoalStateOverride.STOPPED)) {
```
elezar: Should this be `GoalStateOverride.PAUSED`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -460,13 +486,21 @@ private void setReadinessCheck(
...
         if (!taskSpec.getReadinessCheck().isPresent()) {
             LOGGER.debug("No readiness check defined for taskSpec: {}", taskSpec.getName());
             return;
         }
 
+
         ReadinessCheckSpec readinessCheckSpec = taskSpec.getReadinessCheck().get();
+        String readinessCheckCmd = readinessCheckSpec.getCommand();
+
+        if (override.equals(GoalStateOverride.STOPPED)) {
```
elezar: `GoalStateOverride.PAUSED`?

elezar: Reordering the conditional as:
```java
 if (GoalStateOverride.STOPPED.equals(override)) {

}
```
will also handle `null` values for `override` correctly.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
```diff
@@ -215,7 +215,25 @@ public synchronized void update(Protos.TaskStatus status) {
...
+    /**
+     * WARNING: This is a private method for a reason.  Callers must validate that the taskID is already present in the
+     * tasks map.
+     */
+    private String getTaskName(Protos.TaskID taskID) {
+        return tasks.get(taskID).getTaskInfo().getName();
+    }
+
     private void setTaskStatus(Protos.TaskID taskID, Status status) {
+        GoalStateOverride.Status overrideStatus = stateStore.fetchGoalOverrideStatus(getTaskName(taskID));
```
elezar: This method now sets both the regular task status and the overridden task status. Does this mean that they need to be kept in sync in some way? If this is the case, how do we ensure this?

Would it make sense to break the override-specific code below into its own method to improve the readability of this method a little. (`updateOverrideStatusIfComplete`?)

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
```diff
@@ -20,6 +21,15 @@
     /** The definition of the "STOPPED" override state, where commands are replaced with sleep()s.*/
     STOPPED("STOPPED", "STOPPING");
```
elezar: Is it in the scope of this PR to rename these to `PAUSE*`?

frameworks/helloworld/tests/test_recovery.py
gabrielhartmann: I'll rebase on Nick's PR once it's merged.  He's doing the move from `STOPPED` to `PAUSED`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
gabrielhartmann: ditto

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
gabrielhartmann: ditto

gabrielhartmann: `override` should never be null.  I did this on purpose.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/GoalStateOverride.java
gabrielhartmann: no

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
gabrielhartmann: Not that kind of builder.  As currently structured the constructor needs the override information, so a `.add` style call would be too late.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
gabrielhartmann: Factored it out into a different method.  This method is the mechanism of synchronization for the two statuses.  Plan execution drives tasks to their goal state (override or otherwise).  The setting of a Step's status is the canonical source of truth for plan execution.  Plan execution is therefore the correct place to record override status.

gabrielhartmann:
@elezar:  We probably shouldn't call PodInfoBuilder a Builder.  It may or may not be the right place to **always** pass in the overrides.  I'm not sure yet as we have a sample size of one.  I suspect it _isn't_ the perfect place for it.  We're going to be doing a second goal state override of `decommission` shortly after this goes in.  Then we'll have a sample size of two and refactor according to the generalizations that become obvious once the work arrives.

In the `decommission` case, the override will affect not the content of the PodInfo, but the generation of a new `decomission` plan which will remove some pods from a service.  With this addition we may some more satisfying generalization of the pattern.  In the mean time this feature needs to ship.

PR #1786, opened 2017-10-30
Use argparse for kdc.py. Minor changes to sdk_cmd and sdk_marathon.

elezar:
@NimaVaziri here is a quick PR that adds `argparse` support to `kdc.py`. I ran through the `deploy` and `teardown` until I also got the same SSL error that @benclarkwood observed in the CI. This is secondary though. We could remove these additional files, and get the test tooling in first.

### DIFF
--- a/testing/kdc.py
+++ b/testing/kdc.py
@@ -8,26 +8,20 @@ This tool expects as its arguments:
     - subcommand for setup or teardown
     - path to file holding principals as newline-separated strings
 """
+import argparse
+import logging
 import os
-import sys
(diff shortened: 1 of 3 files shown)


NimaVaziri:
@elezar thanks for the comments and for pulling `kdc.py` over - the initial work for the ad-hoc stuff was rather rough, as I wanted to just get the bare minimum working to see if it was useful to begin with, and if so we could iterate on it.

elezar:
I am closing this, as the changes are no longer relevant to the target branch.

PR #1789, opened 2017-10-31
[INFINITY-2566] Initial implementation for kerberized HDFS

NimaVaziri:


frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -361,9 +381,17 @@ pods:
         env:
           DATANODE: true
         # If HDFS does not report a data node to be live, we don't consider the data node to be ready yet
-        # TODO: LIBPROCESS_IP gets overriden to 0.0.0.0 in overlay networking so this readiness-check needs to be fixed
         readiness-check:
+          {{#KERBEROS_ENABLED}}
+          cmd: >
+            export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/) &&
```
elezar: Any reason why we don't just add the `KERBEROS_ENABLED` section after the `JAVA_HOME` and before `TASK_IP`?

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -361,9 +381,17 @@ pods:
...
+            KRB5_CONFIG={{JRE_VERSION}}/lib/security/krb5.conf kinit -k -t ${MESOS_SANDBOX}/hdfs.keytab ${KERBEROS_PRIMARY}/data-${POD_INSTANCE_INDEX}-node.${FRAMEWORK_NAME}.autoip.dcos.thisdcos.directory@${KERBEROS_REALM} &&
```
elezar: Use `{{FRAMEWORK_HOST}}` and `{{TASK_NAME}}` for the FQDN.

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -251,11 +266,11 @@ pods:
               ./bootstrap --resolve-hosts name-$POD_INSTANCE_INDEX-node.$FRAMEWORK_HOST && ./{{HDFS_VERSION}}/bin/hdfs zkfc
         resource-set: zkfc-resources
         configs:
-          {{#KRB5_CONF_URI}}
+          {{#KERBEROS_ENABLED}}
           krb5-conf:
-            template: {{MESOS_SANDBOX}}/krb5.conf
+            template: krb5.conf
```
elezar: Is this config file required for this task? It's not explicitly used in the command?

frameworks/hdfs/universe/config.json
```diff
@@ -83,14 +83,20 @@
...
+              "default": "kdc.marathon.l4lb.thisdcos.directory"
             },
-            "krb5_conf_uri": {
+            "kdc_host_port": {
               "type": "string",
-              "description": "The URI of the krb5.conf file.  This file will be downloaded and used by each HDFS task."
+              "description": "The port of the host running a KDC for that realm.",
+              "default": "88"
+            },
+            "keytab_secret_path": {
```
elezar: For the service account secret, we just use `service_account_secret`. Is this also a path, or is there something different here?

frameworks/hdfs/universe/marathon.json.mustache
```diff
@@ -85,9 +86,11 @@
 
 
     {{#service.kerberos.enabled}}
-    "KEYTABS_URI": "{{service.kerberos.keytabs_uri}}",
-    "KRB5_CONF_URI": "{{service.kerberos.krb5_conf_uri}}",
-    "TASKCFG_ALL_SECURE_MODE": "{{service.kerberos.enabled}}",
+    "KERBEROS_ENABLED": "{{service.kerberos.enabled}}",
```
elezar: Are both of these required?

frameworks/hdfs/universe/marathon.json.mustache
```diff
@@ -29,6 +29,7 @@
     "FRAMEWORK_NAME": "{{service.name}}",
     "FRAMEWORK_LOG_LEVEL": "{{service.log_level}}",
     "HDFS_VERSION": "hadoop-2.6.0-cdh5.11.0",
+    "TASKCFG_ALL_HDFS_VERSION": "hadoop-2.6.0-cdh5.11.0",
```
elezar: In which cases do we need both the regular and `TASKCFG_ALL_*` variants of an environment variable?

frameworks/hdfs/src/main/dist/svc.yml
NimaVaziri: Yeah it's still required because zkfc still needs to interact with the name node as its own component of HDFS and needs to be auth'ed for that communication to proceed successfully. krb5, as a result, needs to be in its JRE so it can talk to the KDC.

frameworks/hdfs/src/main/dist/svc.yml
NimaVaziri: Yup, we could do that.

frameworks/hdfs/src/main/dist/svc.yml
NimaVaziri: Good catch

frameworks/hdfs/universe/config.json
NimaVaziri: Ah in that case, I'll drop the `_path` suffix, though it could be more clear from a UX perspective that `service_account_secret` is a field expecting a path to a secret. From an initial impression, it could be interpreted as that field wanting the secret itself, and not the path.

frameworks/hdfs/universe/marathon.json.mustache
NimaVaziri: The env vars don't need to be prefixed if they're not sent to any tasks essentially. So `HDFS_VERSION` will only be rendered in the scheduler's environment, whereas `TASKCFG_ALL_HDFS_VERSION` will get pushed through to all the tasks' environments (say to render a template, like `hdfs-site.xml`)

In this case, `TASKCFG_ALL_HDFS_VERSION` seems like doesn't actually need to be here as it's not used in any task environments, so will remove.

frameworks/hdfs/universe/marathon.json.mustache
NimaVaziri: So `TASKCFG_ALL_*` isn't generally used in the service spec, since any env vars that are prefixed with this are meant for tasks.

In this case, `KERBEROS_ENABLED` is used in the service spec, and `TASKCFG_ALL_KERBEROS_ENABLED` is used to properly render template configs on the tasks' environments.

Same things applies for TLS configs.

frameworks/hdfs/src/main/dist/core-site.xml
```diff
@@ -517,34 +519,6 @@
...
-    <property>
-        <name>hadoop.http.authentication.simple.anonymous.allowed</name>
-        <value>{{HADOOP_HTTP_AUTHENTICATION_SIMPLE_ANONYMOUS_ALLOWED}}</value>
-    </property>
-    <property>
-        <name>hadoop.http.authentication.kerberos.principal</name>
-        <value>{{HADOOP_HTTP_AUTHENTICATION_KERBEROS_PRINCIPAL}}</value>
-    </property>
-    <property>
-        <name>hadoop.http.authentication.kerberos.keytab</name>
```
imaxxs: Is this the path/url to download the keytab? or where the keytab gets placed after downloaded from secret store?

frameworks/hdfs/src/main/dist/core-site.xml
```diff
@@ -656,5 +630,41 @@
...
     </property>
-    {{/SECURE_MODE}}
+    <property>
+        <name>hadoop.rpc.protection</name>
+        <value>privacy</value>
+    </property>
+    <property>
+        <name>hadoop.security.auth_to_local</name>
+        <value>
+          RULE:[2:$1@$0](.*)s/.*/{{TASK_USER}}/
```
imaxxs: {TASK_USER} does this resolves to "hdfs" user?

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -146,16 +156,21 @@ pods:
               ./bootstrap && ./{{HDFS_VERSION}}/bin/hdfs namenode
         resource-set: name-resources
         readiness-check:
-          cmd: export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/) && ./{{HDFS_VERSION}}/bin/hdfs haadmin -getServiceState name-$POD_INSTANCE_INDEX-node
+          cmd: >
+            export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/) &&
+          {{#KERBEROS_ENABLED}}
+            KRB5_CONFIG={{JRE_VERSION}}/lib/security/krb5.conf kinit -k -t $MESOS_SANDBOX/hdfs.keytab $KERBEROS_PRIMARY/$TASK_NAME.$FRAMEWORK_HOST@$KERBEROS_REALM &&
```
imaxxs: How do we handle the failure scenarios? For example, a case when for some reason, knit - k -t ... command fails and we do not obtain credentials?

frameworks/hdfs/src/main/dist/hdfs-site.xml
```diff
@@ -853,93 +841,102 @@
...
+            <value>*</value>
         </property>
         <property>
-            <name>dfs.datanode.kerberos.principal</name>
-            <value>{{KERBEROS_PRIMARY}}/data-{{POD_INSTANCE_INDEX}}-node.{{FRAMEWORK_NAME}}.mesos@{{KERBEROS_REALM}}</value>
+            <name>dfs.web.authentication.kerberos.principal</name>
+            <value>*</value>
         </property>
         <property>
-            <name>dfs.datanode.data.dir.perm</name>
```
imaxxs: Don't we need the datanode.data.dir.perm property?

NimaVaziri: Yeah, it was just moved.

frameworks/hdfs/src/main/dist/core-site.xml
NimaVaziri: No, this resolves to `nobody` - it was just a placeholder so we can discuss it.

frameworks/hdfs/src/main/dist/core-site.xml
NimaVaziri: It's where the keytab gets downloaded to.

NimaVaziri:
> 1. Do we assume all the permissions are set correctly for the hdfs user for the root / directory?
Haven't gotten around to setting up authz yet, but the operator can configure the permissions for the hdfs user.

> 2. Keytab file path - should it on local file system or on hdfs file system? the default like /etc/hadoop/conf/hdfs.keytab? for data node, journal node etc.
This should be on the local file system but there's no specific place this needs to be, as long as HDFS can read it.

3. Can we enable debugging output from the Sun kerberos classes using HADOOP_OPTS="-Dsun.security.krb5.debug=true"?
Yes, HADOOP_OPTS will be included with this.

frameworks/hdfs/src/main/dist/svc.yml
NimaVaziri: When we publish the finalized package, we'll make sure any non-base-tech failure scenarios are covered. Otherwise, any other error will be logged and we can further diagnose based on said logs.

For example, readiness checks will keep retrying and keep failing under a failure scenario, but their failures will be logged, which can then be inspected.

frameworks/hdfs/universe/marathon.json.mustache
elezar: Thanks for the clarification.

imaxxs:
I understand there is little more work to be done here.  
We can iterate on this further and create another PR and improve.

frameworks/hdfs/universe/marathon.json.mustache
benclarkwood: There is no real moral purity around not using TASKCFG_ALL_* in svc.yml. We have done it both ways.

Arguably using just the TASKCFG_ALL makes it slightly easier to maintain.

PR #1792, opened 2017-10-31
Refactor offer evaluation to use visitors over specs

mrbrowning:


gabrielhartmann:
@mrbrowning 
1. This PR is pretty large in aggregate although it reduces overall lines of code.  Merging it at this size would be challenging from a review perspective.
1. We are at the point where we are about to release a major stable version of the SDK.  Digesting the changes here and re-stabilizing CI at this point would be expensive from an engineer/hours perspective.
1. The core motivation for this work was I believe to separate the Custom/Default executor paths.  Support for the custom executor is going to go away altogether with this upcoming release, as 1.10 and 1.11 will be the last 2 releases.   So in large part I think this work is no longer necessary, correct me if I'm wrong.

/cc @triclambert

PR #1793, opened 2017-11-01
Improve revive calculation

gabrielhartmann:
When doing scale tests I noticed occasional revives where I expected none.  The fact that we were getting the work candidates, then processing offers then deciding whether we needed to revive was causing a situation where the Steps changed status.  So...
* Move revive calculation right next to candidate selection

There was also the fact that things move in and out of the candidate set, but shouldn't have revives called on them.  Once something moves into the active workset it should only leave upon completion.  So...
* We merge the steps which are in progress (not complete and not pending) into the active work set

N.B. All tasks will be detected as PENDING and in need of revive as part of the getCandidates call.

screen shot 2017-11-01 at 4 04 10 pm

Scale testing confirms no unexpected revives.  20 schedulers with 80 tasks each crash looping have only a single revive (when they start and go from no work to some work).

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/ReviveManager.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/ReviveManager.java
@@ -64,17 +64,20 @@ public class ReviveManager {
      *     kafka-0-broker fails    @ 11:00, it's new work!
      *     ...
      */
-    public void revive(Collection<Step> steps) {
-        Set<WorkItem> currCandidates = getCandidates(steps);
+    public void revive(Collection<Step> activeWorkSet) {
+        Set<WorkItem> currCandidates = getCandidates(activeWorkSet);
         Set<WorkItem> newCandidates = new HashSet<>(currCandidates);
-        newCandidates.removeAll(candidates);
+        newCandidates.removeAll(this.candidates);
 
-        logger.info("Candidates, old: {}, current: {}, new:{}", candidates, currCandidates, newCandidates);
+        logger.info("Candidates, old: {}, current: {}, new:{}", this.candidates, currCandidates, newCandidates);
 
         if (!newCandidates.isEmpty()) {
             if (tokenBucket.tryAcquire()) {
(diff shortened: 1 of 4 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/ReviveManager.java
```diff
@@ -64,17 +64,20 @@ public ReviveManager(SchedulerDriver driver, TokenBucket tokenBucket) {
...
 
         if (!newCandidates.isEmpty()) {
             if (tokenBucket.tryAcquire()) {
-                logger.info("Reviving offers.");
-                driver.reviveOffers();
+                logger.info(
+                        "Reviving offers with candidates, old: {}, current: {}, new:{}",
+                        this.candidates,
+                        currCandidates,
+                        newCandidates); driver.reviveOffers();
```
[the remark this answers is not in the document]
gabrielhartmann: oops, thanks

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/ReviveManager.java
```diff
@@ -83,15 +86,15 @@ public void revive(Collection<Step> steps) {
...
 
-        candidates = currCandidates;
+        this.candidates = currCandidates;
     }
 
+
     /**
      * Returns candidates which potentially need new offers.
      */
     private Set<WorkItem> getCandidates(Collection<Step> steps) {
```
benclarkwood: I'd call this `toWorkItems(Collection<Step> steps)` now that it doesn't have any conditional logic.

benclarkwood: (Or just do the conversion inline in `revive` and drop this method)

benclarkwood:
@gabrielhartmann what you just pushed should fix those Canary strategy failures, yah?

gabrielhartmann:
@benclarkwood: yes

PR #1794, opened 2017-11-01
[INFINITY-2521] Auth testing module refactor & usage

NimaVaziri:
This PR does 2 things:
- addresses comments in the [already-merged PR](https://github.com/mesosphere/dcos-commons/pull/1784)
- does further refactoring (eg use of files API, change networking mode)


testing/sdk_auth.py
```diff
@@ -244,20 +211,49 @@ def __init__(self):
...
+        """
+        kadmin_cmd = "task exec {task_id} /usr/sbin/kadmin {options} {cmd} {args}".format(
+            task_id=self.task_id,
+            options=' '.join(options),
+            cmd=cmd,
+            args=' '.join(args)
+        )
+        log.info("Running kadmin: {}".format(kadmin_cmd))
+        try:
+            sdk_cmd.run_cli(kadmin_cmd).strip()
```
elezar: Was there a reason that we chose `run_cli` here instead of `sdk_tasks.task_exec`?

testing/sdk_auth.py
```diff
@@ -283,11 +279,14 @@ def add_principals(self, principals: list):
...
-        add_principal_cmd = "task exec {task_id} /usr/sbin/kadmin -l add --use-defaults --random-password {principal}"
+        kadmin_options = ["-l"]
+        kadmin_cmd = "add"
+        kadmin_args = ["--use-defaults", "--random-password"]
+
         for principal in principals:
             try:
-                sdk_cmd.run_cli(add_principal_cmd.format(task_id=KERBEROS_APP_ID, principal=principal))
-            except dcos.errors.DCOSException as e:
+                self.__run_kadmin(kadmin_options, kadmin_cmd, kadmin_args + [principal])
```
elezar: As far as I know, `kadmin` supports adding the principals in a list. So one could use:
```bash
kadmin_args.extend(principals)
self.__run_kadmin(kadmin_options, kadmin_cmd, kadmin_args)
```
I'm not sure how it handles, for example, existing principals in this case.

testing/sdk_auth.py
```diff
@@ -364,10 +353,14 @@ def get_keytab_path(self):
...
         sdk_marathon.destroy_app(KERBEROS_APP_ID)
-        if os.path.exists(self.temp_working_dir):
-            shutil.rmtree(self.temp_working_dir)
 
+        log.info("Deleting temporary working directory")
+        self.temp_working_dir.cleanup()
+
+        #TODO: separate secrets handling into another module
+        log.info("Deleting keytab secret")
         delete_secret_cmd = "security secrets delete {}".format(self.keytab_secret_path)
```
elezar: We could use the `sdk_security.delete_secret` function that you just added?

tools/kdc.json
```diff
@@ -29,7 +29,8 @@
   },
   "networks": [
     {
-      "mode": "container/bridge"
+      "name": "dcos",
```
elezar: Nice! Does this mean that it works on non-overlay networks too?

frameworks/elastic/tests/test_tls.py
```diff
@@ -24,8 +24,8 @@ def service_account():
     sdk_cmd.run_cli(
         "security org groups add_user superusers {name}".format(name=name))
     yield name
-    sdk_security.delete_service_account(
-        service_account_name=name, service_account_secret=name)
+    sdk_security.delete_service_account(name)
+    sdk_security.delete_secret(name)
```
elezar: This is an unexpected change. Would it make sense to call `delete_secret` from within `delete_service_account` instead of relying on the user to do both? Do we ever have a service account without a secret?

frameworks/hdfs/tests/test_auth.py
```diff
@@ -41,39 +42,57 @@ def configure_package(configure_universe):
...
+# 6. use service client binary to interact with client once authed
 
 @pytest.mark.skipif(os.environ.get("SECURITY") == "strict", reason="auth tests currently broken in strict")
 @pytest.mark.auth
 @pytest.mark.sanity
-def test_user_can_write():
+def test_user_can_write_and_read(kerberos):
+    #try:
+    #    client_app_def = config.kerberized_hdfs_client_marathon_app
+    #    client_app_def["secrets"]["hdfs_keytab"]["source"] = kerberos.get_keytab_path()
```
NimaVaziri: Need to add `REALM` and `KDC_ADDRESS` as env vars for the client container

testing/sdk_auth.py
```diff
@@ -142,35 +97,31 @@ def _get_master_public_ip() -> str:
...
-    except OSError as e:
-        raise OSError("Can't create temp working directory: {}".format(repr(e)))
-
-    return dir_name
+    tmp_dir = TemporaryDirectory()
+    log.info("Created temp working directory {}".format(tmp_dir.name))
+    return tmp_dir
 
 
+#TODO: make this generic and put in sdk_utils.py
```
benclarkwood: I did a little spelunking in the API docs for the Mesos file api as well as the DC/OS CLI code that uses it to do task logging.

I think we can actually pretty easily swap this to the file api:
<cluster-url>/slave/<agent-id>/files/read.json?path=<url-encoded-absolute-path>

I tested this against your cluster and was able to download the `keytab` file from the KDC with no problems.

frameworks/hdfs/tests/test_auth.py
```diff
@@ -41,39 +42,57 @@ def configure_package(configure_universe):
...
+        #kerberos_env.cleanup()
+        pass
 
+# General process for each test
+# 1. Will use a keytab local to the client binary (key will either be a generic one or one made within the test context)
+# 2. Launch marathon app referencing said keytab via secret store
+# 3. container environment will have kerberos CLI and conf file setup
+# 4. task exec to app's container
+# 5. within each test, depending on the keytab that's pulled, one should kinit with appropriate principal
+# 6. use service client binary to interact with client once authed
```
NimaVaziri: Write smoke test to make sure kerberized service is healthy.

testing/sdk_auth.py
```diff
@@ -42,9 +41,9 @@
 # resiliency towards possible intermittent network failures.
 
 
-def _launch_marathon_app(app_definition):
+def launch_marathon_app(app_definition):
```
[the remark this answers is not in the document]
NimaVaziri: 👍

testing/sdk_auth.py
```diff
@@ -204,33 +160,44 @@ def _copy_file_to_localhost(self):
...
         kdc_host_name=self.kdc_host_name,
         keytab_file=self.keytab_file_name
     )
-    cmd = "eval $(ssh-agent -s) && ssh-add /ssh/key && dcos node ssh --master-proxy --leader --option StrictHostKeyChecking=no '{}'".format(scp_cmd)
+    dest = "/home/{linux_user}/{keytab_file}".format(
+        linux_user=LINUX_USER,
+        keytab_file=self.keytab_file_name
+    )
+    scp_cmd = "scp -o StrictHostKeyChecking=no {src} {dest}".format(src=src, dest=dest)
+    ssh_context = "eval $(ssh-agent -s) && ssh-add /ssh/key"
```
elezar: When I tried this locally, this failed in the first call to `run_command_on_agent` due to my SSH key not having been added. We could run this earlier in this method.

frameworks/elastic/tests/test_tls.py
NimaVaziri: There are cases where secrets aren't tied to service accounts, but usually a service account has a secret with it, so I opted to separate the functions but call `delete_secret` from within `delete_service_account`

testing/sdk_auth.py
NimaVaziri: Nope, updated.

testing/sdk_auth.py
NimaVaziri: Yeah, all principals can be added simultaneously.

tools/kdc.json
NimaVaziri: Managed to make it simply work on `host`.

testing/sdk_auth.py
NimaVaziri: Changed to use the files API.

testing/sdk_auth.py
NimaVaziri: Don't need SSH anymore as we just fetch the file. Removed all of this.

NimaVaziri:
As for the networking problem, since the KDC marathon app is using UCR, we can't use `bridge` networking / containerPorts.

And using VIPs on `host` networking doesn't currently allow us to allocate the same port number to both UDP and TCP traffic. The working solution that gets around this which is passable since it's only in our testing environment: we map the same port number from the service to the host (in this case `2500` but we can change it, it was arbitrary), then use app's DNS address to resolve the hostname + statically assigned port number to  pass both UDP and TCP traffic through.

NimaVaziri:
I'll remove the first test implementation out of the scope of this PR and move it to a follow up PR so we can get these changes in.

frameworks/hdfs/universe/config.json
```diff
@@ -1210,7 +1210,7 @@
         "client_read_shortcircuit": {
           "type": "boolean",
           "description": "This configuration parameter turns on short-circuit local reads.",
-          "default": true
```
elezar: @NimaVaziri this is an API change, but is not mentioned in the PR comments. Was this intentional? If so, we should be sure to add this to the release notes.

benclarkwood: my understanding is this corrects an edge case. Is that right, Nima?

NimaVaziri: It's actually not even an edge case. The effective functionality doesn't change, save for some performance boost. It's used when a client wants to read some data - if it's enabled, it'll try to read the file directly; if it's disabled, the read goes through the Data Node via a TCP socket.

Turning this off by default just prevents an error from being thrown because if it's on and the client isn't co-located with the data, HDFS will throw a warning but it'll just fall back to reading through the Data Node.

NimaVaziri: But yes, we should also add this change to the release notes regardless.

PR #1795, opened 2017-11-02
Temporarily skips strict testing for auth tests

NimaVaziri:


NimaVaziri:
@benclarkwood the KDC marathon app needs to be configured to work with strict apparently - it's on the radar (i.e. JIRA), just temporarily suspended

frameworks/hdfs/tests/test_auth.py
```diff
@@ -71,6 +72,7 @@ def configure_package(configure_security):
         kerberos.cleanup()
 
 
+@pytest.mark.skipif(os.environ.get("SECURITY") == "strict", reason="auth tests currently broken in strict")
```
elezar: There is an `sdk_utils.is_strict_mode()` function that can be used here instead.

PR #1801, opened 2017-11-03
Add SIGTERM handler to release ZK lock

mrbrowning:


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerRunner.java
```diff
@@ -98,22 +98,24 @@ private SchedulerRunner(SchedulerBuilder schedulerBuilder) {
...
-                        mesosScheduler.get(),
-                        schedulerBuilder.getServiceSpec(),
-                        schedulerBuilder.getSchedulerConfig(),
-                        schedulerBuilder.getStateStore());
-            }
-        } finally {
+
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            LOGGER.info("Shutdown initiated, releasing curator lock");
             locker.unlock();
```
benclarkwood: This won't actually work.

`A NON re-entrant mutex that works across JVMs. Uses Zookeeper to hold the lock. All processes in all JVMs that use the same lock path will achieve an inter-process critical section.`

Only the originally locking thread can do the unlock. 

In the exploratory PR Gabriel put together awhile ago, he found that doing a driver stop was one way to have the original thread do the cleanup (https://github.com/mesosphere/dcos-commons/pull/1804).

mrbrowning: It ought to, and I observed it to; the reason I changed the type of the mutex to `InterProcessSemaphoreMutex` is specifically because it permits other threads to unlock (because it's just backed by a semaphore with one slot, which incidentally also explains why it's not reentrant, unlike `InterProcessMutex`). It's not well-explained in the javadocs, but note that `InterProcessSemaphoreMutex` lacks a `isOwnedByCurrentThread()` method, because it doesn't know or care about thread-level ownership.

benclarkwood: Ah ok, its docs read similarly enough that I thought it wouldn't. 👍

PR #1811, opened 2017-11-07
[BACKPORT] Don't set user=root when running in strict (#1802)

benclarkwood:


benclarkwood:
I found that we've been testing upgrades in stable against the beta packages. Whoops. Fixed.

benclarkwood:
Leaving the branch around in case we wind up needing to resurrect this.

PR #1812, opened 2017-11-07
Update KDC app's networking type

NimaVaziri:
This changes 3 things:  
- the KDC container's networking type to `host` mode
- forces KDC to handle all traffic through TCP
- allows configuration of the TCP port it listens on

### DIFF
--- a/tools/kdc.json
+++ b/tools/kdc.json
@@ -6,30 +6,24 @@
   "container": {
     "type": "MESOS",
     "docker": {
-      "image": "mesosphere/kdc:latest",
+      "image": "nvaziri/kdc:dev",
       "forcePullImage": true
-    },
-    "portMappings": [
-      {
-        "containerPort": 88,
-        "protocol": "tcp",
-        "labels": {"VIP_0": "kdc:88"}
-      },
-      {
-        "containerPort": 88,
-        "protocol": "udp",
-        "labels": {"VIP_0": "kdc:88"}
-      },
-      {
-        "containerPort": 8000,
-        "protocol": "tcp",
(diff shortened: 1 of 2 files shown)


tools/kdc.json
```diff
@@ -6,30 +6,24 @@
...
   },
   "networks": [
     {
-      "mode": "container/bridge"
+      "mode": "host"
+    }
+  ],
+  "portDefinitions": [
+    {
+      "port": 2500,
```
NimaVaziri: Note that this isn't the actual port KDC listens on as the application port, 2500 just maps to a dynamically assigned port at runtime.

NimaVaziri:
Turns out KDC doesn't force all traffic through one protocol. Tested this with HDFS and the service comes up fine, but the clients fail to connect as they use UDP.

On the other hand, turning only UDP on for KDC prevents the service (HDFS) from coming up in the first place...so, still investigating.

tools/kdc.json
```diff
@@ -6,30 +6,24 @@
   "container": {
     "type": "MESOS",
     "docker": {
-      "image": "mesosphere/kdc:latest",
+      "image": "nvaziri/kdc:dev",
```
benclarkwood: @NimaVaziri where does the Dockerfile for this live?

NimaVaziri:
I've addressed this issue in PR #1794 as the working version of this had a couple dependencies that were in that change set.

tools/kdc.json
NimaVaziri: I included it in `tools/` (perhaps there's a better place), but it's literally just a `RUN` instruction that invokes a shell script.

PR #1810, opened 2017-11-07
[INFINITY-2632] Simplify readiness check.

benclarkwood:

PR #1813, opened 2017-11-07
[INFINITY-2632][BACKPORT] Simplify readiness check. (#1810)

benclarkwood:


frameworks/kafka/src/main/dist/svc.yml
```diff
@@ -81,16 +81,28 @@ pods:
...
+            grep -q "INFO \[Kafka Server $POD_INSTANCE_INDEX\], started (kafka.server.KafkaServer)" $kafka_server_log_files
+            if [ $? -eq 0 ] ; then
+              echo "Found started log line."
+            else
+              echo "started log line not found. Exiting."
+              exit 1
+            fi
+
+            echo "Both required log lines found."
+            exit 0
```
elezar: Does this work across pod restarts? Was this not the point of the date magic being done in the existing healthcheck?

benclarkwood: The file in question, server.log, lives in the sandbox. It will not persist across restarts.

PR #1815, opened 2017-11-07
Just scan examples dir for yaml files to test

nickbp:
Ended up also catching a (presumably mistakenly?) empty `cni.yml` in the process.


frameworks/helloworld/src/main/dist/examples/cni.yml
```diff
@@ -0,0 +1,55 @@
+name: {{FRAMEWORK_NAME}}
```
nickbp: @ArtRand was this intended to be deleted or should it be brought back as is being done here?

benclarkwood: I think the yaml used is overlay.yml

nickbp: Yeah that looks right, they aren't the same but `overlay.yml` appears to be a superset of `cni.yml`. I'll just remove the empty `cni.yml` instead of trying to repopulate it.

PR #1818, opened 2017-11-08
[INFINITY-2528] Basic auth test against kerberized HDFS

NimaVaziri:
This change set is rebased on top of `kerberos_test_module` as it depends on updated `sdk_auth.py` functionality.

It implements the most basic auth test by doing a write and read after authenticating the client.

It also adds a marathon app def which launches a container containing a kerberized HDFS client setup to interact with a kerberized HDFS cluster, and its only dependencies are
- the secret path to pull down the keytab (needed to auth)
- the realm in which the KDC and the HDFS cluster is operating in
- the KDC address to perform auth calls against

It optionally allows configuration of the JVM heap size which the HDFS client uses.

### DIFF
--- a/frameworks/hdfs/tests/config.py
+++ b/frameworks/hdfs/tests/config.py
@@ -2,6 +2,7 @@ import json
 import os
 import shakedown
 
+import sdk_auth
 import sdk_cmd
 import sdk_plan
 import sdk_utils
@@ -19,10 +20,9 @@ TEST_FILE_1_NAME = "test_1"
 TEST_FILE_2_NAME = "test_2"
 DEFAULT_HDFS_TIMEOUT = 5 * 60
 HDFS_POD_TYPES = {"journal", "name", "data"}
-DOCKER_IMAGE_NAME = "mesosphere/hdfs-client:2.6.4"
-KERBERIZED_CLIENT_IMAGE_NAME = "nvaziri/kerberized-hdfs-client:dev"
-KERBERIZED_CLIENT_MEM_SIZE = 512
+DOCKER_IMAGE_NAME = "nvaziri/hdfs-client:dev"
 KEYTAB = "hdfs.keytab"
+GENERIC_HDFS_USER_PRINCIPAL = "hdfs@{realm}".format(realm=sdk_auth.REALM)
 
 
 def get_kerberized_hdfs_client_app():
@@ -33,9 +33,6 @@ def get_kerberized_hdfs_client_app():
     with open(app_def_path) as f:
         app_def = json.load(f)
 
-    app_def["mem"] = KERBERIZED_CLIENT_MEM_SIZE
-    app_def["container"]["docker"]["image"] = KERBERIZED_CLIENT_IMAGE_NAME
-    app_def["container"]["volumes"][0]["containerPath"] = "/hadoop-2.6.0-cdh5.9.1/{}".format(KEYTAB)
     return app_def
 
 
--- a/frameworks/hdfs/tests/config.py
+++ b/frameworks/hdfs/tests/config.py
@@ -1,3 +1,5 @@
+import json
+import os
 import shakedown
 
 import sdk_cmd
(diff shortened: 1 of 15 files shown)


frameworks/hdfs/tests/config.py
```diff
@@ -20,58 +22,48 @@
 DOCKER_IMAGE_NAME = "mesosphere/hdfs-client:2.6.4"
 KERBERIZED_CLIENT_IMAGE_NAME = "nvaziri/kerberized-hdfs-client:dev"
```
elezar: Would it make sense to only specify this in the `json` file?

frameworks/hdfs/tests/config.py
```diff
@@ -20,58 +22,48 @@
...
 KERBERIZED_CLIENT_MEM_SIZE = 512
```
elezar: Should we have to override the memory here? It is already specified in the `json` file?

frameworks/hdfs/tests/config.py
```diff
@@ -20,58 +22,48 @@
...
+KEYTAB = "hdfs.keytab"
```
elezar: Let's just use the value as specified in the `json` file.

frameworks/hdfs/tests/config.py
```diff
@@ -20,58 +22,48 @@
...
+    with open(app_def_path) as f:
+        app_def = json.load(f)
+
+    app_def["mem"] = KERBERIZED_CLIENT_MEM_SIZE
+    app_def["container"]["docker"]["image"] = KERBERIZED_CLIENT_IMAGE_NAME
+    app_def["container"]["volumes"][0]["containerPath"] = "/hadoop-2.6.0-cdh5.9.1/{}".format(KEYTAB)
+    return app_def
+
+
+def hdfs_write_command(filename, content_to_write):
```
elezar: I generally prefer a `source, destination` in these kinds of arguments, but since utilities like `tar` swap these around this is also fine.

frameworks/hdfs/tests/config.py
```diff
@@ -20,58 +22,48 @@
...
 
+def hdfs_read_command(filename):
+    return "./bin/hdfs dfs -cat /{}".format(filename)
+
+
 def read_data_from_hdfs(service_name, filename):
-    read_command = "./bin/hdfs dfs -cat /{}".format(filename)
-    rc, output = run_hdfs_command(service_name, read_command)
+    rc, output = run_hdfs_command(service_name, hdfs_read_command(filename))
     return rc and output.rstrip() == TEST_CONTENT_SMALL
```
elezar: I know that this is outside of the scope of this PR, but it seems strange that the read command doesn't return the content read. Even if we only return `True` or `False`, it seems strange to only compare to `TEST_CONTENT_SMALL`. Also, why do we only use `rstrip()`?

frameworks/hdfs/tests/config.py
```diff
@@ -20,58 +22,48 @@
...
 
+def hdfs_delete_file_command(filename):
+    return "./bin/hdfs dfs -rm /{}".format(filename)
+
+
 def delete_data_from_hdfs(service_name, filename):
-    delete_command = "./bin/hdfs dfs -rm /{}".format(filename)
-    rc, output = run_hdfs_command(service_name, delete_command)
+    rc, output = run_hdfs_command(service_name, hdfs_delete_file_command(filename))
     return rc
```
elezar: *Also out of scope:* Could we not just generate a random (with fixed seed) byte-string to write to HDFS instead of relying on `wget` here?

frameworks/hdfs/tests/test_auth.py
```diff
@@ -84,38 +85,35 @@ def kerberos(configure_universe):
...
+    try:
+        client_app_def = config.get_kerberized_hdfs_client_app()
+        client_app_def["secrets"]["hdfs_keytab"]["source"] = kerberos.get_keytab_path()
+        client_app_def["env"]["REALM"] = kerberos.get_realm()
+        client_app_def["env"]["KDC_ADDRESS"] = kerberos.get_kdc_address()
+        sdk_marathon.install_app(client_app_def)
+        client_task_id = client_app_def["id"]
+
+        # authenticate client as any generic user
+        principal = "hdfs/name-0-node.hdfs.autoip.dcos.thisdcos.directory@LOCAL"
```
elezar: I would add a client principal to the list of principals created in setup and use this instead.

frameworks/hdfs/tests/test_auth.py
```diff
@@ -84,38 +85,35 @@ def kerberos(configure_universe):
...
+        kinit = sdk_auth.KINIT.format(keytab=config.KEYTAB, principal=principal)
```
elezar: Could we rather have a `get_kinit_command` function here instead of calling format on `KINIT` directly?

testing/sdk_tasks.py
```diff
@@ -119,6 +119,22 @@ def fn():
...
-        "task exec {} {}".format(task_name, command), **kwargs)
+def task_exec(task_name : str, command : str) -> tuple:
+    """
+    Invokes the given command on the task via spawning `dcos task exec` in a subprocess.
+    :param task_name: Name of task to run command on.
+    :param command: The command to execute.
+    :return: a tuple consisting of the task exec's return code, stdout, and stderr
+    """
+    exec_cmd = "dcos task exec {} {}".format(task_name, command)
+    result = subprocess.run([exec_cmd], shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
elezar: What was the motivation for this change? Are we not now duplicating the code in `sdk_cmd.run_cli` here? It also seems unrelated to the changes required for HDFS, as I only found calls that check `stdout`.

frameworks/hdfs/tests/config.py
NimaVaziri: Yeah it probably does, as this shouldn't need to be modified anyway.

frameworks/hdfs/tests/config.py
NimaVaziri: Yeah, can also be removed.

frameworks/hdfs/tests/config.py
NimaVaziri: The only this is here is because it helps avoid parsing the path of the keytab when a client wants to do `kinit` (https://github.com/mesosphere/dcos-commons/blob/auth_test/frameworks/hdfs/tests/test_auth.py#L107)

frameworks/hdfs/tests/config.py
NimaVaziri: Yeah, it's weird that we can modify the filename but not the content of the file. Let's address this in the following set of test-related PRs.

frameworks/hdfs/tests/config.py
NimaVaziri: Did you mean to comment on another line? I'm failing to see any `wget`s here

frameworks/hdfs/tests/test_auth.py
NimaVaziri: Yeah, that's the intention and will do that, just wanted to push this PR out as soon as I had a working version.

frameworks/hdfs/tests/test_auth.py
NimaVaziri: Sounds better to me.

testing/sdk_tasks.py
NimaVaziri: I tried using `sdk_cmd.run_cli`, however the command string that it takes as input is then manipulated via [shakedown's `run_dcos_command`](https://github.com/dcos/shakedown/blob/2cfb691cffc9a76e8e4ba629bbedd463fc755221/shakedown/dcos/command.py#L103) and even after wrestling with putting the right quotation marks and escapes in the right places (which made things ugly), the command was still failing. However, running the exact same command simply via the shell worked perfectly as expected. @nickbp ran into this issue as well recently.

Even though shakedown is essentially [performing a subprocess call as well](https://github.com/dcos/shakedown/blob/2cfb691cffc9a76e8e4ba629bbedd463fc755221/shakedown/dcos/command.py#L122) it seems its internal formatting is somehow messing things up.

As for getting stderr, that's just a good practice to pass it along upstream.

nickbp: Could you 'upstream' this into sdk_cmd?

frameworks/hdfs/tests/test_auth.py
benclarkwood: Glad we all agree on this point.

We should always use client principals.

frameworks/hdfs/tests/config.py
elezar: I was talking about the `write_lots_of_data_to_hdfs()` function here: https://github.com/mesosphere/dcos-commons/pull/1818/files#diff-8e4f6dcc5298350f4a9043a17c80d3b1R70

GitHub isn't as friendly about making arbitrary comments.

testing/sdk_tasks.py
elezar: Yes, then we would remove the shakedown dependency there too.

NimaVaziri:
@benclarkwood @imaxxs any other comments on this change set?

nickbp:
The command fixes LGTM, I'll abstain on the kerberos bits

benclarkwood:
LGTM. I'm sure we'll shake anything loose in testing.

benclarkwood:
@NimaVaziri I think maybe your changes had a side effect on the TLS testing?
```
frameworks/helloworld/tests/test_tls.py:223 (test_changing_discovery_replaces_certificate_sans)
hello_world_service = 'hello-world'

    @pytest.mark.tls
    @pytest.mark.sanity
    @sdk_utils.dcos_ee_only
    @pytest.mark.dcos_min_version('1.10')
    def test_changing_discovery_replaces_certificate_sans(hello_world_service):
        """
        Update service configuration to change discovery prefix of a task.
        Scheduler should update task and new SANs should be generated.
        """
        original_tasks = sdk_tasks.get_task_ids(config.PACKAGE_NAME, 'discovery')
        assert len(original_tasks) == 1, 'Expecting exactly one task ID'
    
        task_id = original_tasks[0]
        assert task_id
    
        # Load end-entity certificate from PEM encoded file
>       _, stdout, _ = sdk_tasks.task_exec(task_id, 'cat server.crt').encode('ascii'),
E       AttributeError: 'tuple' object has no attribute 'encode'

```

benclarkwood:
And test_sanity.test_state_refresh_disable_cache is failing for Hello World

NimaVaziri:
yeah, looks like something is wonky, the same set of tests are failing. I'll run the tests locally to isolate them.

As for the encoding issue, looks like it's just a refactor.

PR #1819, opened 2017-11-08
Refactor the StringMatcher rules

gabrielhartmann:
This PR reduces substantially the amount of boilerplate needed to write a StringMatcher based Rule.  It also now enforces interface requirements instead of relying on convention.

I'm going to need this to add Zone/Region rules.


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/MarathonConstraintParser.java
```diff
@@ -247,9 +248,9 @@ public PlacementRule run(StringMatcher taskFilter, String fieldName, String oper
                                  Optional<String> requiredParameter) throws IOException {
             String parameter = validateRequiredParameter(operatorName, requiredParameter);
             if (isHostname(fieldName)) {
```
benclarkwood: There's probably some kind of clean subtype thing to do here, but I don't know what it is off the top of my head.

gabrielhartmann: More PRs coming.  Don't worry.

PR #1830, opened 2017-11-13
[INFINITY-2640, INFINITY-2650] Add zone/region placement rules to the Marathon constraint parser

gabrielhartmann:


frameworks/helloworld/tests/test_placement.py
```diff
@@ -73,6 +73,153 @@ def test_rack_not_found():
...
+        'world': {
+            'placement': '@zone:GROUP_BY:2'
+        }
+    }
+
+    fail_placement(options)
+
+
+def succeed_placement(options):
+    """
```
nickbp: These docs look wrong (here and below), copied from one of the test cases?

frameworks/helloworld/tests/test_placement.py
```diff
@@ -73,6 +73,153 @@ def test_rack_not_found():
...
+        assert False, "Should have failed to deploy world-1"
+    except AssertionError as arg:
+        raise arg
+    except:
+        pass # expected to fail
+
+    sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+
```
nickbp: extra newline

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/MarathonConstraintParser.java
```diff
@@ -231,6 +231,10 @@ public PlacementRule run(
             switch (PlacementUtils.getField(fieldName)) {
                 case HOSTNAME:
                     return new MaxPerHostnameRule(1, taskFilter);
+                case ZONE:
+                    return new MaxPerZoneRule(1, taskFilter);
+                case REGION:
+                    return new MaxPerRegionRule(1, taskFilter);
                 case ATTRIBUTE:
                 default:
```
nickbp: Would it make sense to throw `UnsupportedOperationException` in the `default` case? It looks like that'd ensure things visibly blow up if a fifth field type is added, rather than silently treating it as an attribute

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/RegionRuleFactory.java
```diff
@@ -0,0 +1,17 @@
+package com.mesosphere.sdk.offer.evaluate.placement;
+
+/**
+ * This class generates {@link RegionRule}s.
+ */
+public class RegionRuleFactory implements RuleFactory {
+    private static final RegionRuleFactory factory = new RegionRuleFactory();
+
+    public static RegionRuleFactory getInstance() {
```
nickbp: private constructor as below?

frameworks/helloworld/tests/test_placement.py
gabrielhartmann: These docs are correct.

gabrielhartmann:
@nickbp:  Thanks. Comments addressed.

frameworks/helloworld/tests/test_placement.py
nickbp: Oh I get it now.

PR #1832, opened 2017-11-13
[INFINITY-2641] Inject region/zone envvars into tasks

gabrielhartmann:


gabrielhartmann:
@nickbp @benclarkwood: Added integration test.

PR #1834, opened 2017-11-13
Report strategy as serial when undefined

mrbrowning:


cli/commands/plan.go
```diff
@@ -323,7 +324,7 @@ func toPlanStatusTree(planName string, planJSONBytes []byte) string {
 	}
 	planStrategy, ok := planJSON["strategy"]
 	if !ok {
-		planStatus = UNKNOWN_VALUE
+		planStatus = DEFAULT_PLAN_STRATEGY
```
benclarkwood: Won't this always be incorrect in the case of running against an old scheduler that doesn't have the API addition?

mrbrowning: Yes, fair point. This was written on the assumption that the scheduler version would be at least as current as the CLI version, but that may not be valid. How do we want to handle compatibility guarantees between the two, given that they might diverge in either direction (i.e. CLI being newer than scheduler, or scheduler being newer than CLI)?

benclarkwood: I would just report it as "unknown" as we would just be guessing.

What was the reason for doing the serial fallback?

mrbrowning: @gabrielhartmann now that I'm looking at `DefaultPhaseFactory`, this shouldn't happen due to the case of the default strategy fallthrough to serial. Can you verify the issue you saw wasn't down to mismatched CLI/scheduler versions?

gabrielhartmann: @mrbrowning:  That's what confused me.  A strategy is always present.  I'll give it a try on something which doesn't have a plan set at all off of master.

gabrielhartmann: I verified that `hello-world` which doesn't have any custom plans specified behaves correctly.  I probably screwed something up when I was using a mashup of different versions.  Thanks for looking into this @mrbrowning

PR #1843, opened 2017-11-15
[INFINITY-2611] Enable tests to work correctly in strict mode clusters

NimaVaziri:
The main issue with this not working initially was performing operations in the `kdc` (and `hdfsclient` app) as `root` user, hence it working in permissive. When the same set of steps were performed, it obviously failed because the images (and the files/executables being accessed) weren't given the appropriate permissions/ownership for user `nobody`.

Since this now works in both strict and permissive, it's safe to say these pieces of the auth tooling are relatively stable now, so we should probably move the image to the `mesosphere` docker hub  account.

### DIFF
--- a/frameworks/hdfs/tests/test_auth.py
+++ b/frameworks/hdfs/tests/test_auth.py
@@ -80,11 +80,15 @@ def kerberos(configure_security):
 
 
 @pytest.fixture(autouse=True)
+@pytest.mark.dcos_min_version('1.10')
+@sdk_utils.dcos_ee_only
 @pytest.mark.smoke
 def test_health_of_kerberized_hdfs():
     config.check_healthy(service_name=config.SERVICE_NAME)
 
 
+@pytest.mark.dcos_min_version('1.10')
+@sdk_utils.dcos_ee_only
 @pytest.mark.auth
 @pytest.mark.sanity
 def test_user_can_write_and_read(kerberos):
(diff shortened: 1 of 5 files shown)


NimaVaziri:
Real test of functionality is against this TC job using a strict cluster: https://teamcity.mesosphere.io/viewQueued.html?itemId=859672

elezar:
@NimaVaziri were any changes to the client docker image required?

benclarkwood:
@NimaVaziri Just realized (sorry): this test or the suite should be marked such that it's skipped in 1.9.

NimaVaziri:
@elezar if by client docker image you're referring to `hdfs-client:dev` then yeah, there was a script placed in `/` that needed to be modified. Just making sure the paths and executables are accessible by `nobody` should be sufficient for the Kafka client image as well, I'd suppose.

@benclarkwood good catch, forgot about that

frameworks/hdfs/tests/test_auth.py
```diff
@@ -78,21 +78,17 @@ def kerberos(configure_universe):
...
+@pytest.mark.dcos_min_version('1.10')
+@sdk_utils.dcos_ee_only
 @pytest.mark.smoke
 def test_health_of_kerberized_hdfs():
     config.check_healthy(service_name=config.SERVICE_NAME)
 
 
-sdk_utils.is_strict_mode()
+@pytest.mark.dcos_min_version('1.10')
+@sdk_utils.dcos_ee_only
```
elezar: Thanks for adding these to my PR too.

PR #1845, opened 2017-11-15
Suppress offers when there's no work to do

gabrielhartmann:


gabrielhartmann:
@benclarkwood: Breaks uninstall... :-)  Not worth it right now.

PR #1769, opened 2017-10-25
[INFINITY-2517]: Kafka SASL authn

elezar:
This PR adds SASL authentication for Kafka using Kerberos.

For instructions for kerberos see the file [./frameworks/kafka/docs/kerberos.md]

**UPDATE:** @benclarkwood I have removed the wip label here. Please feel free to review. We can then iterate and add things like automated tests and more configuration options.

PR #1847, opened 2017-11-15
[INFINITY-2687] Fix the authn test for Kafka in strict

elezar:
As @NimaVaziri pointed out for #1843, the auth tests don't work as expected on Strict mode clusters.

The tests were merged in #1769


NimaVaziri:
LGTM

@elezar is the kafka client image configured to work with user `nobody`? No point in kicking off strict tests if they're not configured yet, since this is marked as WIP.

elezar:
I tested with a 1.10 strict mode cluster, and this seemed to work.

PR #1850, opened 2017-11-16
[INFINITY-2673] kerberos for zookeeper

elezar:
This PR enables Kerberos support for ZooKeeper from Kafka and fixes some minor issues with the KDC utility.

See mesosphere/dcos-zookeeper#38

@benclarkwood @NimaVaziri what do you think of where the ZooKeeper setting is currently enabled? I'm not 100% happy with it, and feel that a `kafka.zookeeper.uri` and `kafka.zookeeper.kerberos.enabled` may be better. Thoughts?


benclarkwood:
@elezar I prefer it under the single security section.

PR #1852, opened 2017-11-17
[KAF-383] Update to 1.0.0.

benclarkwood:
Update to 1.0.0.
Update readiness check for new log line format.

### DIFF
--- a/frameworks/kafka/src/main/dist/svc.yml
+++ b/frameworks/kafka/src/main/dist/svc.yml
@@ -95,7 +95,7 @@ pods:
             kafka_server_log_files=${kafka_dir}/logs/server.log*
 
             echo "Checking for started log line in $kafka_server_log_files."
-            grep -q "INFO \[Kafka Server $POD_INSTANCE_INDEX\], started (kafka.server.KafkaServer)" $kafka_server_log_files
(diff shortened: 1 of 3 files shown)


triclambert:
What about the protocol/log version upgrades?  What is the plan for that?

frameworks/kafka/src/main/dist/svc.yml
```diff
@@ -95,7 +95,7 @@ pods:
             kafka_server_log_files=${kafka_dir}/logs/server.log*
```
elezar: Just a thought. Would it make sense to explicitly set `kafka_dir=$MESOS_SANDBOX/{{KAFKA_VERSION_PATH}}` instead of relying on the `ls` magic?

elezar: (This would be one less change required for updates to confluent)

benclarkwood:
@clambert See the jira ticket but: we make a second package release that updates the protocol and log versions.

frameworks/kafka/src/main/dist/svc.yml
benclarkwood: That would certainly work, yeah.

PR #1853, opened 2017-11-17
[KAF-384] Make the advertised listeners consistent.

benclarkwood:
0. Drop the legacy `advertised.host.name` parameter.
1. Make all `listeners` bind to `MESOS_CONTAINER_IP`.
2. Make all `advertised.listeners` bind to `<broker-dns>`.

### DIFF
--- a/frameworks/kafka/src/main/dist/server.properties.mustache
+++ b/frameworks/kafka/src/main/dist/server.properties.mustache
@@ -27,10 +27,18 @@ broker.id={{POD_INSTANCE_INDEX}}
 #     listeners = security_protocol://host_name:port
 #   EXAMPLE:
 #     listeners = PLAINTEXT://your.host.name:9092
+#
+# Hostname and port the broker will advertise to producers and consumers. If not set,
+# it uses the value for "listeners" if configured.  Otherwise, it will use the value
+# returned from java.net.InetAddress.getCanonicalHostName().
+#
+# advertised.listeners=PLAINTEXT://your.host.name:9092
+
 {{^SECURITY_KERBEROS_ENABLED}}
 {{^KAFKA_ENABLE_TLS}}
-listeners=PLAINTEXT://:{{KAFKA_BROKER_PORT}}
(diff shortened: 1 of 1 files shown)


benclarkwood:
Also, holy moly I want to not have such a spaghetti nest of mustache conditionals.

frameworks/kafka/src/main/dist/server.properties.mustache
```diff
@@ -81,11 +89,6 @@ sasl.kerberos.service.name={{SECURITY_KERBEROS_PRIMARY}}
 
 {{/SECURITY_KERBEROS_ENABLED}}
 
-# Hostname and port the broker will advertise to producers and consumers. If not set,
```
elezar: Thanks for removing this!

frameworks/kafka/src/main/dist/server.properties.mustache
```diff
@@ -27,21 +27,29 @@ broker.id={{POD_INSTANCE_INDEX}}
...
+# returned from java.net.InetAddress.getCanonicalHostName().
+#
+# advertised.listeners=PLAINTEXT://your.host.name:9092
+
 {{^SECURITY_KERBEROS_ENABLED}}
 {{^KAFKA_ENABLE_TLS}}
-listeners=PLAINTEXT://:{{KAFKA_BROKER_PORT}}
+listeners=PLAINTEXT://{{MESOS_CONTAINER_IP}}:{{KAFKA_BROKER_PORT}}
 port={{KAFKA_BROKER_PORT}}
+advertised.listeners=PLAINTEXT://{{TASK_NAME}}.{{FRAMEWORK_HOST}}:{{KAFKA_BROKER_PORT}}
```
elezar: Would it make sense to have all the listeners defined here (instead of here *OR* in the TLS section *OR* in the SASL section)?

benclarkwood:
@elezar I'm going to go ahead and do what we discussed and move all this nasty templating into a go script.

PR #1854, opened 2017-11-17
INFINITY-2631 Implement pod decommissioning by truncation

nickbp:
Operators may now decrease the pod count for pods which have opted-in with `allow-decommission: true` in their svc.yml. The decommission itself will result in the following behaviors:
- A new `decommission` plan will appear when the scheduler is restarted with a smaller count, showing the tasks/pods to be removed. Otherwise no `decommission` plan is listed.
- The `decommission` plan will have phases for each pod instance to be decommissioned, in the order of highest index first (foo-3 before foo-2), hence 'truncation'. If multiple pod types are being decommissioned simultaneously, they will start at the bottom of the `ServiceSpec` and work up -- reverse of default deployment order of starting at top and working down.
- The `pod status` view will show affected pods with a state of `DECOMMISSIONING`. As the pod instances are decommissioned, they are removed from the `pod status` view, as everything about them will have been deleted at that point.

This PR tests the feature in both integration and service tests by (only) enabling decommission in hello-world's `world` pod type. Some additional changes/fixes were made to the service test libraries to support this sort of reconfiguration flow.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/decommission/DecommissionPlanFactory.java
```diff
@@ -0,0 +1,289 @@
...
+            this.plan = plan;
+            this.resourceSteps = resourceSteps;
+        }
+    }
+
+    /**
+     * Returns a {@link Plan} for decommissioning tasks, or an empty {@link Optional} if no decommission is necessary.
+     */
+    private static PlanInfo buildPlanInfo(ServiceSpec serviceSpec, StateStore stateStore, TaskKiller taskKiller) {
+        // Determine which tasks should be decommissioned (and which shouldn't)
```
benclarkwood: Note to self: Decommissioning is done at the pod level. So, a single task being decommissioned should result in the pod going down.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/decommission/DecommissionPlanFactory.java
```diff
@@ -0,0 +1,289 @@
...
+        private final String podType;
+        private final int podIndex;
+
+        /**
+         * @param taskInfo a taskinfo whose pod is to be sorted
+         * @param orderedPodTypes the list of known pod types in order of priority (reverse of ServiceSpec ordering)
+         */
+        PodKey(String podType, int podIndex, List<String> orderedPodTypes) throws TaskException {
+            this.podType = podType;
+            // If the pod is unknown, it gets a -1, placing it at higher priority than listed pods:
```
benclarkwood: When would the podtype be unknown? When the pod is dropped from the ServiceSpec during an upgrade?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/decommission/DecommissionPlanFactory.java
```diff
@@ -0,0 +1,289 @@
...
+                podKey = new PodKey(labelReader.getType(), labelReader.getIndex(), orderedPodTypes);
+            } catch (TaskException e) {
+                LOGGER.error(String.format(
+                        "Failed to retrieve task metadata. Omitting task from decommission: %s", task.getName()), e);
+                continue;
+            }
+
+            Integer expectedPodCount = expectedPodCounts.get(podKey.podType);
+            if (expectedPodCount == null) {
+                LOGGER.info("Scheduling '{}' for decommission: '{}' is not present in service spec: {}",
```
benclarkwood: This seems like very dangerous behavior. It could mean dropping an entire pod type on accident through a bad update.

What is the current behavior if an entire unknown pod type is found in the state store?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/decommission/DecommissionPlanFactory.java
```diff
@@ -0,0 +1,289 @@
...
+         * Orders the pod according to the following rules:
+         * <ol><li>Pod types at the bottom of the ServiceSpec (lower podTypeIndex) have priority over ones at the top.
+         * Effectively the reverse of the default deployment ordering based on ServiceSpec ordering.</li>
+         * <li>Pod types missing from the ServiceSpec have priority over listed pod types. If multiple of these unknown
+         * pod types are involved, they are decommissioned in alphabetical order ("unknownA" before "unknownB").</li>
+         * <li>Within a given pod type, pods with higher index are given priority ("pod-3" before "pod-2")</li></ol>
+         */
+        @Override
+        public int compareTo(PodKey other) {
+            if (podTypeIndex != other.podTypeIndex) {
```
benclarkwood: These are very small sets, but:
```
if (this.equals(other)) {
  return 0
}
```

if you're feeling optimizey.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/decommission/DecommissionPlanFactory.java
benclarkwood: If we do want this behavior, we should make it clear that any update dropping a pod type is irreversible.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallRecorder.java
```diff
@@ -51,46 +42,59 @@ public void record(OfferRecommendation offerRecommendation) throws Exception {
...
-        if (!tasksToUpdate.isEmpty()) {
-            stateStore.storeTasks(updateResources(resource, tasksToUpdate));
-
-            // broadcast uninstallRecommendation to each UninstallStep in resource phase
-            List<OfferRecommendation> uninstallRecommendations = Collections.singletonList(uninstallRecommendation);
-            resourceSteps.forEach(step -> step.updateOfferStatus(uninstallRecommendations));
+        if (tasksToUpdate.isEmpty()) {
+            return;
         }
+
```
benclarkwood: Seems like a lot of code duplication here with DecommissionRecorder. Can we factor that out into a base class?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/decommission/DecommissionPlanFactory.java
gabrielhartmann: Supporting dropping pod types is very desirable.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/decommission/DecommissionPlanFactory.java
gabrielhartmann: We definitely want the ability to drop pod-types.  Very good for schedulers which want to generate/remove pod types programmatically.

benclarkwood: I'm fine with clear documentation in the dev guide that old pod types will be garbage collected. Rather, that they create one-way upgrades if you do it in a package upgrade.

I hadn't thought of the generate/remove pod types use case. Neat.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -179,51 +175,25 @@ private PlanCoordinator buildPlanCoordinator() throws ConfigStoreException {
...
         planManagers.addAll(plans.stream()
                 .filter(plan -> !plan.isDeployPlan())
-                .map(DefaultPlanManager::new)
+                .map(DefaultPlanManager::createInterrupted)
                 .collect(Collectors.toList()));
 
         return new DefaultPlanCoordinator(planManagers);
     }
 
-    private static void killUnneededAndPendingOverrideTasks(
```
gabrielhartmann: I don't think we can get rid of this.

gabrielhartmann: Yep, the integration tests were not failing appropriately.  I've made them fail and have put this method mostly back to make them pass again.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallRecorder.java
benclarkwood: There is sort of code duplication, but it's not really worth trying to factor out due to the subtle differences between these two.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/decommission/DecommissionPlanFactory.java
gabrielhartmann: Updated #1858

nickbp:
Re leaving artifacts/hints behind following a decommission: At the moment the taskstate property is the only thing left behind. We could add something else more explicit, probably wouldn't want to leave the original task data for reasons described below. Time for an operations journal?

The scheduler could leave behind task entries with a DECOMMISSION+COMPLETE bit set and then just ignore them from them on, but this could have problems if the service is downgraded with decommissioned entries to a version that doesn't know about decommissioning. But that should only be a problem if the pod count is increased again after the downgrade, I think...

PR #1859, opened 2017-11-18
INFINITY-2711 Emulate main CLI's envvar support with config selection

nickbp:
Now that we're not piggybacking off the main CLI, we should add support for a couple additional envvars that were previously unnecessary: https://dcos.io/docs/1.10/cli/#environment-variables

- `DCOS_CLUSTER`: name of active cluster (0.5.x)
- `DCOS_CONFIG`: specific config file (0.4.x)
- `DCOS_SSL_VERIFY`: override of `core.ssl_verify` setting (0.???)


nickbp:
Added tests while also fixing existing tests to pass.

Ended up coming across a bit of a discovery when checking the main CLI's envvar support. It has a whole set of conventions for overriding config values that we'd want to also support for the values that we fetch. This ended up needing a fair amount of changes to support, but actually ended up removing the need for a lot of custom override handling that we'd had before. For example, we used to manually check for `DCOS_URL` in the arg handling, whereas now it's automatically handled via the new env override handling during the config retrieval. See also: https://github.com/dcos/dcos-cli/blob/master/dcos/config.py#L148

PR #1865, opened 2017-11-21
[INFINITY-2693] Kerberized HDFS authz test

NimaVaziri:
This PR cleans up some security configuration revolving Kerberos, in overlap with properly configuring the service for authz testing.

### DIFF
--- a/frameworks/hdfs/src/main/dist/core-site.xml
+++ b/frameworks/hdfs/src/main/dist/core-site.xml
@@ -502,6 +502,10 @@
         <name>rpc.metrics.percentiles.intervals</name>
         <value>{{RPC_METRICS_PERCENTILES_INTERVALS}}</value>
     </property>
+    <property>
+        <name>hadoop.http.authentication.simple.anonymous.allowed</name>
+        <value>{{HADOOP_HTTP_AUTHENTICATION_SIMPLE_ANONYMOUS_ALLOWED}}</value>
+    </property>
 
     {{#TLS_ENABLED}}
     <property>
@@ -550,10 +554,6 @@
         <value>{{FRAMEWORK_HOST}}</value>
     </property>
     <property>
-        <name>hadoop.http.authentication.simple.anonymous.allowed</name>
-        <value>false</value>
-    </property>
-    <property>
         <name>hadoop.http.authentication.kerberos.principal</name>
         <value>{{KERBEROS_PRIMARY_HTTP}}/{{FRAMEWORK_NAME}}.marathon.autoip.dcos.thisdcos.directory@{{KERBEROS_REALM}}</value>
     </property>
@@ -569,7 +569,7 @@
     </property>
     <property>
         <name>hadoop.security.instrumentation.requires.admin</name>
-        <value>true</value>
+        <value>{{HADOOP_SECURITY_INSTRUMENTATION_REQUIRES_ADMIN}}</value>
     </property>
     <property>
         <name>hadoop.security.group.mapping</name>
--- a/frameworks/hdfs/src/main/dist/core-site.xml
+++ b/frameworks/hdfs/src/main/dist/core-site.xml
@@ -48,90 +48,6 @@
         <name>hadoop.http.filter.initializers</name>
         <value>{{HADOOP_HTTP_FILTER_INITIALIZER}}</value>
     </property>
-    <property>
-        <name>hadoop.security.authorization</name>
-        <value>{{HADOOP_SECURITY_AUTHORIZATION}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.instrumentation.requires.admin</name>
-        <value>{{HADOOP_SECURITY_INSTRUMENTATION_REQUIRES_ADMIN}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.authentication</name>
-        <value>{{HADOOP_SECURITY_AUTHENTICATION}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.group.mapping</name>
-        <value>{{HADOOP_SECURITY_GROUP_MAPPING}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.groups.cache.secs</name>
-        <value>{{HADOOP_SECURITY_GROUPS_CACHE_SECS}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.groups.negative-cache.secs</name>
-        <value>{{HADOOP_SECURITY_GROUPS_NEGATIVE_CACHE_SECS}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.groups.cache.warn.after.ms</name>
-        <value>{{HADOOP_SECURITY_GROUPS_CACHE_WARN_AFTER_MS}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.group.mapping.ldap.url</name>
-        <value>{{HADOOP_SECURITY_GROUP_MAPPING_LDAP_URL}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.group.mapping.ldap.ssl</name>
-        <value>{{HADOOP_SECURITY_GROUP_MAPPING_LDAP_SSL}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.group.mapping.ldap.ssl.keystore</name>
-        <value>{{HADOOP_SECURITY_GROUP_MAPPING_LDAP_SSL_KEYSTORE}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.group.mapping.ldap.ssl.keystore.password.file</name>
-        <value>{{HADOOP_SECURITY_GROUP_MAPPING_LDAP_SSL_KEYSTORE_PASSWORD_FILE}}</value>
-    </property>
-    <property>
-        <name>hadoop.security.group.mapping.ldap.bind.user</name>
-        <value>{{HADOOP_SECURITY_GROUP_MAPPING_LDAP_BIND_USER}}</value>
-    </property>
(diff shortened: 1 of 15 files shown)


frameworks/hdfs/universe/config.json
```diff
@@ -63,55 +63,51 @@
...
-            "keytab_secret": {
-              "type": "string",
-              "description": "A DC/OS Secret Store path.  The contents of this secret should be a keytab containing the credentials for all HDFS servers.",
-              "default": "__dcos_base64__hdfs_keytab"
+            "tls": {
+              "type": "object",
+              "description": "TLS configuration properties.",
+              "properties": {
+                "enabled": {
+                  "description": "Enable TLS support (requires Strict mode).",
```
elezar: Note that this only requires enterprise, and not strict. I was looking at INFINITY-2583 and noted that Cassandra and Elastic have the "correct" description, but Kafka doesn't.

frameworks/hdfs/src/main/dist/core-site.xml
```diff
@@ -652,19 +547,109 @@
...
     </property>
     <property>
         <name>hadoop.http.authentication.simple.anonymous.allowed</name>
-        <value>{{HADOOP_HTTP_AUTHENTICATION_SIMPLE_ANONYMOUS_ALLOWED}}</value>
+        <value>false</value>
     </property>
     <property>
         <name>hadoop.http.authentication.kerberos.principal</name>
-        <value>{{HADOOP_HTTP_AUTHENTICATION_KERBEROS_PRINCIPAL}}</value>
+        <value>{{KERBEROS_PRIMARY_HTTP}}/{{FRAMEWORK_NAME}}.marathon.autoip.dcos.thisdcos.directory@{{KERBEROS_REALM}}</value>
```
elezar: Should this refer to the `marathon` subdomain here?

NimaVaziri: No other dns endpoint resolves to a service-specific address, as everything else is task-specific.

frameworks/hdfs/tests/test_auth.py
```diff
@@ -15,51 +14,80 @@
...
+    for (primary, instance) in itertools.product(primaries, instances):
+        principals.append(
+            "{primary}/{instance}.{fqdn}@{REALM}".format(
+                primary=primary,
+                instance=instance,
+                fqdn=fqdn,
+                REALM=sdk_auth.REALM
+            )
+        )
+    for _, principal in config.CLIENT_PRINCIPALS.items():
```
elezar: You should be able to use:
```python
principals.extend(config.CLIENT_PRINCIPALS.values())
```

frameworks/hdfs/tests/test_auth.py
```diff
@@ -15,51 +14,80 @@
...
+    Kerberized HDFS maps the primary component of a principal to local users, so
+    we need to create an appropriate mapping to test authorization functionality.
+    :return: A base64-encoded string of principal->user mappings
+    """
+    rules = [
+        "RULE:[2:$1@$0](^hdfs@.*$)s/.*/hdfs/",
+        "RULE:[1:$1@$0](^nobody@.*$)s/.*/nobody/"
+    ]
+
+    for user, _ in config.CLIENT_PRINCIPALS.items():
```
elezar: We could iterate over `config.CLIENT_PRINCIPALS.keys()` here instead?

frameworks/hdfs/tests/test_auth.py
```diff
@@ -87,28 +115,90 @@ def test_health_of_kerberized_hdfs():
...
 
+@pytest.mark.dcos_min_version('1.10')
+@sdk_utils.dcos_ee_only
+@pytest.mark.auth
+@pytest.mark.sanity
+def test_users_have_appropriate_permissions(kerberized_hdfs_client):
+    # "hdfs" is a superuser
+    sdk_auth.kinit(kerberized_hdfs_client, keytab=config.KEYTAB, principal=config.CLIENT_PRINCIPALS["hdfs"])
+
+    for user, _ in config.CLIENT_PRINCIPALS.items():
```
elezar: We could loop over `config.CLIENT_PRINCIPALS.keys()`.

frameworks/hdfs/tests/test_auth.py
```diff
@@ -87,28 +115,90 @@ def test_health_of_kerberized_hdfs():
...
-        assert stdout == config.TEST_CONTENT_SMALL
 
-    finally:
-        sdk_marathon.destroy_app(client_task_id)
 
+@pytest.mark.dcos_min_version('1.10')
+@sdk_utils.dcos_ee_only
+@pytest.mark.auth
+@pytest.mark.sanity
+def test_users_have_appropriate_permissions(kerberized_hdfs_client):
```
elezar: Some thoughts about this test (could all be addressed in a follow-up PR):
* We only create folders with `700` permissions. Should we also test other permissions (e.g. where `bob` can read, but not write)?
* Should we test that the superuser can access all folders (after they are created)?
* It would be nice to split of the parts of this function into separate functions to make it a little cleaner. We could even split it into two separate tests:
  * User can read/write their own folder with `x00` permissions
  * User can't read/write someone else's folders with `x00` permissions
  * User can read to folder with `002` permissions
  * User can write to folder with `004` permissions

frameworks/hdfs/tests/test_auth.py
```diff
@@ -87,28 +115,90 @@ def test_health_of_kerberized_hdfs():
...
+
+    write_access_cmd = "/bin/bash -c \"{}\"".format(config.hdfs_write_command(
+        config.TEST_CONTENT_SMALL,
+        "/users/alice/{}".format(config.TEST_FILE_1_NAME)))
+    log.info("Bob tries to wrtie to alice's directory: {}".format(write_access_cmd))
+    _, _, stderr = sdk_tasks.task_exec(kerberized_hdfs_client, write_access_cmd)
+    log.info("Bob can't write to alice's directory: {}".format(write_access_cmd))
+    assert "put: Permission denied: user=bob" in stderr
+
+    read_access_cmd = config.hdfs_read_command("/users/alice/{}".format(config.TEST_FILE_1_NAME))
```
elezar: This is the same read command as above.

frameworks/hdfs/tests/test_auth.py
elezar: Note that we actually only use the `/users/alice` folder in the tests, so it is not required to create all of them.

frameworks/hdfs/universe/config.json
```diff
@@ -1423,8 +1414,12 @@
         },
         "security_auth_to_local": {
           "type": "string",
-          "description": "Maps kerberos principals to local user names",
-          "default": ""
+          "description": "Maps kerberos principals to local user names. This field must be base64 encoded.",
```
elezar: Is there another alternative other than base64 encoding? Something like the JSON placement constraints?

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/TaskEnvRouter.java
```diff
@@ -21,7 +21,7 @@
 public class TaskEnvRouter {
 
     private static final String TASKCFG_PREFIX = "TASKCFG_";
-    private static final String TASKCFG_GLOBAL_ENV_PREFIX = TASKCFG_PREFIX + "ALL_";
+    public static final String TASKCFG_GLOBAL_ENV_PREFIX = TASKCFG_PREFIX + "ALL_";
```
elezar: I'm not sure about making this static public just to use in the HDFS scheduler. Could we not use the methods of the `TaskEnvRouter` to get done what we want in this case?

@benclarkwood I would love to hear your thoughts here.

frameworks/hdfs/src/main/dist/core-site.xml
```diff
@@ -652,19 +547,109 @@
     </property>
     <property>
         <name>hadoop.http.authentication.cookie.domain</name>
-        <value>{{HADOOP_HTTP_AUTHENTICATION_COOKIE_DOMAIN}}</value>
+        <value>{{FRAMEWORK_HOST}}</value>
     </property>
     <property>
         <name>hadoop.http.authentication.simple.anonymous.allowed</name>
-        <value>{{HADOOP_HTTP_AUTHENTICATION_SIMPLE_ANONYMOUS_ALLOWED}}</value>
+        <value>false</value>
```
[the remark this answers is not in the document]
NimaVaziri: Technically this field is only applicable to `simple` authentication, not kerberos, so yeah, we should let this field be configurable outside of having kerberos enabled.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/TaskEnvRouter.java
NimaVaziri: No method in the class provides the functionality of prepending a given env var name with the desired prefix, unless you're suggesting we write one?

frameworks/hdfs/universe/config.json
NimaVaziri: Not that I'm aware of. This is consistent with how Elastic achieves this functionality.

frameworks/hdfs/tests/test_auth.py
NimaVaziri: Removed both latter `write_access_cmd` and `read_access_cmd`.

frameworks/hdfs/tests/test_auth.py
NimaVaziri: Yeah, 👍  to more extensive testing, objective of this PR was to introduce a working authz test asap. Let's address these points in a follow-up PR after this one gets in.

PR #1867, opened 2017-11-21
[INFINITY-2620] Add metadata for UI to render zone and advanced constraint fields.

mpereira:


mpereira:
@elezar good point about Kibana. Even though it might not end up being used too much I think it does make sense allowing users to configure placement constraints for it. I'll create a ticket for that.

mpereira:
@elezar https://jira.mesosphere.com/browse/ELASTIC-157

PR #1868, opened 2017-11-21
[INFINITY-2657] Add IS operator

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/ExactMatcher.java
```diff
@@ -20,15 +21,28 @@ public static StringMatcher createAttribute(String name, String value) {
...
+        }
     }
 
     @Override
     public boolean matches(String value) {
+        try {
+            Double in = Double.valueOf(value);
+            value = format.format(in);
+        } catch (NumberFormatException ex) {
+            // Continue with original value of `value`
```
mrbrowning: Wouldn't this let you set a bogus value and then keep it?

mrbrowning: Nevermind, read further. Only question is what use cases involve double-valued attributes, it scares me a little given floating point behavior.

gabrielhartmann: I don't know of any use case.  It's just part of the spec.  I'll send you a link elsewhere.

PR #1879, opened 2017-11-26
Add the FINISH goal state and rename the FINISHED goal state to ONCE

mrbrowning:
The goal state formerly known as `FINISHED` had the behavior that, even on a configuration update, a task with that goal state would not be relaunched. A helpful compromise between this and the  `RUNNING` goal state is one called `FINISH` that restarts a task with state `TASK_FINISHED` only if its configuration has been changed. To avoid ambiguity, the former `FINISHED` state has been renamed as `ONCE`.


frameworks/helloworld/src/main/dist/examples/executor_volume.yml
```diff
@@ -38,9 +38,9 @@ pods:
         env:
           SLEEP_DURATION: 1000
       once:
-        goal: FINISHED
+        goal: FINISH
```
gabrielhartmann: This name conflict looks strange.  task name is `once`, goal is `FINISH`?

sdk/common/src/main/java/com/mesosphere/sdk/specification/GoalState.java
```diff
@@ -17,5 +17,10 @@
     /**
      * Running and then exiting successfully. Should not be restarted upon a successful exit.
      */
-    FINISHED
+    FINISH,
+
+    /**
+     * Running only once over the lifetime of a service. Should not be run again for the duration of this install.
```
gabrielhartmann: `Should not be run again ... unless indicated by an external agent e.g. end-user.`  Something like that.

mrbrowning: Doesn't that describe the `FINISH` state? What input would I provide to re-run a `ONCE` task?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -369,6 +379,38 @@ public ServiceSpec parse(byte[] bytes) throws ConfigStoreException {
...
+                    JsonParser p, DeserializationContext ctxt) throws IOException, JsonParseException {
+                String value = ((TextNode) p.getCodec().readTree(p)).textValue();
+
+                if (value.equals("FINISHED") || value.equals("ONCE")) {
+                    return GoalState.ONCE;
+                } else if (value.equals("FINISH")) {
+                    return GoalState.FINISH;
+                } else if (value.equals("RUNNING")) {
+                    return GoalState.RUNNING;
+                } else if (value.equals("UNKNOWN")) {
```
nickbp: Would it make sense to do something like this, where any unrecognized value maps to UNKNOWN? I don't know one way or the other. Could help with downgrade compatibility later but may lead to other problems...:
```
} else {
  logger.warn(some warning here);
  return GoalState.UNKNOWN;
}
```

hectorj2f:
@mrbrowning , do you have any ETA for this PR ? We'd like to use the code of this PR for the Kubernetes upgrades.

sdk/common/src/main/java/com/mesosphere/sdk/specification/GoalState.java
gabrielhartmann: `dcos <service> plan restart <deploy> <phase> <step>`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DefaultStepFactory.java
```diff
@@ -162,7 +162,7 @@ private boolean hasReachedGoalState(PodInstance podInstance, Protos.TaskInfo tas
                 default:
                     return false;
             }
-        } else if (goalState.equals(GoalState.FINISHED)) {
+        } else if (goalState.equals(GoalState.ONCE) || goalState.equals(GoalState.FINISH)) {
```
gabrielhartmann: This logic is actually incorrect, but no more incorrect than it already is.  See: https://jira.mesosphere.com/browse/INFINITY-2837

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
gabrielhartmann: +1

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -370,6 +380,38 @@ public ServiceSpec parse(byte[] bytes) throws ConfigStoreException {
...
+                    return GoalState.UNKNOWN;
+                }
+
+                throw new JsonParseException(String.format("Unknown goal state: %s", value), p.getCurrentLocation());
```
gabrielhartmann: What do the logs look like when this is thrown?

mrbrowning: Very similar to a Python syntax error, e.g. ```JsonParseException, line 1, column 37: Unknown goal state: STATE```, followed by the usual stack trace.

gabrielhartmann: So no scheduler crash?

mrbrowning:
Sure, it just seemed as though there might be some semantics attached to `UNKNOWN` more specific than "I don't recognize this value at all." Will change.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
mrbrowning: Oh, it'll crash, since it ultimately results in an uncaught `ConfigStoreException` that will break during scheduler building.

hectorj2f:
Nice 🎉 . Thanks guys!

PR #1880, opened 2017-11-27
[INFINITY-2710] Fix TLS footprint evaluation

gabrielhartmann:


gabrielhartmann:
Moving here to rekick CI: https://github.com/mesosphere/dcos-commons/pull/1889

PR #1882, opened 2017-11-27
[INFINITY-2737] Pull down and use custom ZK lib

benclarkwood:
This PR updates ZK to pull down, and then include a custom zookeeper library jar. This library jar is then used by Kafka's ZK client.

That jar is based on this changeset on ZK 3.4.10: https://github.com/benclarkwood/zookeeper/compare/branch-3.4...benclarkwood:release-3.4.10-mesosphere?expand=1

This allows us to run Kerberized ZK without adding principals beyond those for the autoip DNS entries as Kafka will now use the autoip DNS as the principal.

### DIFF
--- a/frameworks/kafka/src/main/dist/svc.yml
+++ b/frameworks/kafka/src/main/dist/svc.yml
@@ -14,6 +14,7 @@ pods:
       - {{LIBMESOS_URI}}
       - {{KAFKA_STATSD_URI}}
       - {{CLIENT_STATSD_URI}}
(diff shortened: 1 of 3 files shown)


benclarkwood:
I am not quite sure what you mean, @NimaVaziri. Where do you want this reference?

NimaVaziri:
That's a good question...

right before we do the move? we do have clarifying comments in other places of the yaml spec, like readiness checks so seems to be consistent

benclarkwood:
I get it.

Good call, added.

PR #1884, opened 2017-11-28
[INFINITY-2736] Remove kinit dependency for kerberized HDFS service spec

NimaVaziri:
This PR moves away from using `kinit` (as it might not be in the task's environment) and relies on JAAS to auth.

### DIFF
--- a/frameworks/hdfs/src/main/dist/krb5.conf
+++ b/frameworks/hdfs/src/main/dist/krb5.conf
@@ -1,14 +1,14 @@
 [libdefaults]
-default_realm = LOCAL
-dns_lookup_realm = true
-dns_lookup_kdc = true
-udp_preference_limit = 1
+  default_realm = {{KERBEROS_REALM}}
+  dns_lookup_realm = true
+  dns_lookup_kdc = true
+  udp_preference_limit = 1
 
 [realms]
-  LOCAL = {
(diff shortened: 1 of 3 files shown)


frameworks/hdfs/src/main/dist/krb5.conf
```diff
@@ -1,14 +1,14 @@
 [libdefaults]
-default_realm = LOCAL
-dns_lookup_realm = true
-dns_lookup_kdc = true
-udp_preference_limit = 1
+  default_realm = {{KERBEROS_REALM}}
```
elezar: Don't we need to add references to this file in the environment too?
```
-Djava.security.krb5.conf=$MESOS_SANDBOX/{{KAFKA_VERSION_PATH}}/config/krb5.conf`
```
(taken from Kafka)

elezar: OK. Scratch that. THat's what's tracked in that other JIRA ticket, right? https://jira.mesosphere.com/browse/INFINITY-2726

benclarkwood:
Upon investigation, I believe the HDFS commands do not use the JAAS file.

We will defer fixing this for now.

benclarkwood:
Closing for now to unclutter PR list

PR #1887, opened 2017-11-28
[INFINITY-2700] Add CLI test step to build.sh

elezar:
This PR adds the running of the cli tests to `build.sh`. Note that this does not exercise them as part of the CI system.

I have noted the following:
- [x] `github.com/stretchr/testify` needs to be added to the vendored dependencies.
- [x] ~A number of tests in `--- FAIL: TestPlanTestSuite (4.55s)` are failing.~
- [ ] We need to consider how we trigger these tests. We could add this to the SDK PR job, for example.


build.sh
```diff
@@ -40,13 +40,23 @@ if [ x$PULLREQUEST = "xtrue" ]; then
...
+    ln -s $REPO_ROOT_DIR/govendor $GOPATH/src
+fi
+
 # Verify golang-based projects: SDK CLI (exercised via helloworld, our model service)
 # and SDK bootstrap, run unit tests
-for golang_sub_project in frameworks/helloworld/cli/dcos-hello-world sdk/bootstrap; do
+for golang_sub_project in cli sdk/bootstrap; do
     pushd $golang_sub_project
-    go test .
+    go test -v ./...
```
benclarkwood: I think this will trigger the vendor tests, no?

benclarkwood:
@elezar yep, vendoring should be fixed. We should also add some kind of check that fails if vendoring is incomplete.

The tests or CLI should be fixed before we put this in the path of successful builds.

IMO putting this into the SDK PR job makes sense to me.

build.sh
elezar: It doesn't seem like it ...

elezar:
I have added a (disabled) step to the SDK PR job: https://teamcity.mesosphere.io/admin/editBuildRunners.html?id=buildType:DcosIo_DcosCommons_Sdk_BuildAndTestPr

PR #1891, opened 2017-11-28
[INFINITY-2741] Document support for @zone placement constraints

gabrielhartmann:


docs/pages/developer-guide.md
```diff
@@ -1221,6 +1221,23 @@ pods:
...
+ZONE: us-west-2a
+```
+
+Services may choose to use this information to enable rack awareness. When doing so, they should use placement rules to ensure that their pods are appropriately placed withing regions and zones. One may apply placement constraints against regions and zones by referencing `@region` and `@zone` keys.  For example:
+
+```
+@zone:GROUP_BY:2
+```
+
+The placement rule above would apply the `GROUP_BY` operator to zones.  The SDK does not currently enable region awareness, so any given service will only receive offers in the local region.
```
gabrielhartmann: @triclambert and @crackcode123: Note the addition of the last sentence.  Otherwise I just moved the whole section to fix this section accidentally being between the TLS section and its children.

docs/pages/operations-guide.md
```diff
@@ -355,6 +355,10 @@ hostname:LIKE:10.0.0.159|10.0.1.202|10.0.3.3
 
 You must include spare capacity in this list, so that if one of the whitelisted systems goes down, there is still enough room to repair your service (via [`pod replace`](#replace-a-pod)) without requiring that system.
 
+### Zones
```
gabrielhartmann: I am intentionally not documenting Regions in the operations guide as the SDK is not yet region aware.

PR #1893, opened 2017-11-29
[INFINITY-2726] Move HDFS krb5.conf out of JRE folder

NimaVaziri:


frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -387,9 +408,9 @@ pods:
         readiness-check:
           cmd: >
             export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jre*/) &&
-          {{#KERBEROS_ENABLED}}
-            KRB5_CONFIG={{JRE_VERSION}}/lib/security/krb5.conf kinit -k -t $MESOS_SANDBOX/hdfs.keytab $KERBEROS_PRIMARY/$TASK_NAME.$FRAMEWORK_HOST@$KERBEROS_REALM &&
-          {{/KERBEROS_ENABLED}}
+            {{#KERBEROS_ENABLED}}
+            KRB5_CONFIG=$MESOS_SANDBOX/{{HDFS_VERSION}}/etc/hadoop/krb5.conf kinit -k -t $MESOS_SANDBOX/hdfs.keytab $KERBEROS_PRIMARY/$TASK_NAME.$FRAMEWORK_HOST@$KERBEROS_REALM &&
```
elezar: What is the reason for the readiness check not setting the `-Djava.security.krb5.conf` flag instead?

benclarkwood: As mentioned elsewhere, it appears that hdfs subcommands don't obey JAAS, so kinit is required for now.

NimaVaziri: Turns out that setting is actually needed for the JRE to know where to look for the krb5.conf file.

`JAVA_OPTS` isn't a recognized env var so it must be set via `HADOOP_OPTS`. The `KRB5_CONFIG` env var, on the other hand, is needed specifically for `kinit`.

PR #1896, opened 2017-11-29
[INFINITY-2709] Allow region awareness in 1.11+

gabrielhartmann:
[Region awareness](https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L384-L394) is enabled on DC/OS 1.11+ dependent upon a [feature flag](https://github.com/mesosphere/dcos-commons/pull/1896/files#diff-2a89a76888deeeec345fb8f34b418efdR340).

End-users may define placement rules referencing regions.  If they do this, then the scheduler [in no way modifies the user's intent](https://github.com/mesosphere/dcos-commons/pull/1896/files#diff-969d01b7dff081a309473a83ce816b10R349).  If an end-user writes a placement rule which does not reference a region it is assumed that the end-user intends for [placement to occur in the local region](https://github.com/mesosphere/dcos-commons/pull/1896/files#diff-969d01b7dff081a309473a83ce816b10R341).

If an end-user desires a pod to be spread across multiple regions they could indicate this with something like `@region:LIKE:*`


gabrielhartmann:
@benclarkwood https://jira.mesosphere.com/browse/INFINITY-2747 for docs

sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/Capabilities.java
```diff
@@ -100,6 +101,11 @@ public boolean supportsV1APIByDefault() {
         return hasOrExceedsVersion(1, 11);
     }
 
+    public boolean supportsRegionAwareness() {
+        // This feature is in BETA for 1.11, so requires explicit opt-in by end-users.
+        return SchedulerConfig.fromEnv().isregionAwarenessEnabled() && hasOrExceedsVersion(1, 11);
```
benclarkwood: Does it make more sense for this to be a setting on either the `scheduler` or on a per-pod basis in the service spec?

I don't know of any other service behavior that is controlled with envvars other than uninstall, and I'm not a huge fan of having to set a magic ENVVAR as a service developer.

benclarkwood: My stance is this is ok for beta, but we should determine how granular we want service developers to be able to go, and how we should officially expose it to them.

gabrielhartmann: 1. I think it makes sense at the scheduler layer.  You're opting into a behavior all up.  It's a feature flag.
1. If you want to use it for one pod and not another, then set your placement constraints appropriately.
1. Regarding the ENVVAR thing, it could be a first class citizen in the YAML (capabilities section), but that's not in scope here.

PR #1901, opened 2017-11-30
[INFINITY-2732] Allow programmatic plan customization

gabrielhartmann:


frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -23,21 +24,59 @@
     public static void main(String[] args) throws Exception {
         final SchedulerConfig schedulerConfig = SchedulerConfig.fromEnv();
         final SchedulerRunner runner;
+        File yamlSpecFile;
+
+        Scenario scenario = getScenario(args);
```
hectorj2f: Does this mean we have to choose between defining our plans in a programmatic manner or via YAML ? You can see in our PR https://github.com/mesosphere/dcos-kubernetes/pull/239/files#diff-965e73ab742e8a42e6874e28e5edda57R547 that we use a custom strategy for upgrades. But we keep using the YAML plans for deploy, replace,...

gabrielhartmann: No, this is just an artifact of HelloWorld being a demonstration framework which operates in many different modes.

gabrielhartmann: @hectorj2f In short, however you were modifying your strategy before, should now be done through this mechanism.  I'm planning to simplify the interface substantially.  End-users will have 3 interfaces for modifying the behavior of their service programmatically.

1. PlanCustomizer as presented here
1. ServiceSpecCustomer (coming soon)
1. RecoveryCustomizer (aka RecoveryPlanOverrider)

hectorj2f: @gabrielhartmann Alright! I was concerned about having to define all my Plans choosing a single scenario, (meaning all YAMLs or all customized).

PR #1903, opened 2017-12-01
Make default placement constraint values be valid stringified JSON.

mpereira:
Values should be a JSON-stringified array of constraints (which are expressed as arrays of strings).

Example: `"hostname:MAX_PER:1"` becomes `"[[\"hostname\", \"MAX_PER\", \"1\"]]"`

Empty strings become JSON-stringified empty arrays, `""` -> `"[]"`.

Chat with Andrew:
```
Murilo Pereira [12:14 PM]
for values that used to default to `"hostname:MAX_PER:1"`, does it matter
whether the JSON equivalent has the `1` as a string or number? e.g.
`"[[\"hostname\", \"MAX_PER\", \"1\"]]"` for string

Andrew Hoskins [12:21 PM]
Hey! Make it a string

Murilo Pereira [12:21 PM]
cool. and empty defaults just `""` right?

Andrew Hoskins [12:26 PM]
Empty state should actually be `"[]"`

[12:26]
Because the meme type is <…>+json, so it expects a valid JSON value
```


mpereira:
One of the things I was struggling with (after finding out about the need for the propagating mustache triple bracket on friday) was the need for double-escaping quotes in config.json files, e.g. https://github.com/mesosphere/dcos-commons/pull/1903/files#diff-56550f345e72e9ddaffbc774bb87a7a9R124

Another was missing test config files with hard-coded constraints like https://github.com/mesosphere/dcos-commons/pull/1903/files#diff-9c0a8da33f4e191b89245b7045401adfL4 

The other things was that making `[]` be the default value instead caused the constraint parser to treat it as a empty constraint instead of an empty list of constraints:

```
testcase com.mesosphere.sdk.elastic.scheduler.ServiceTest >
  testSpec: java.io.IOException: Invalid number of entries in rule. Expected 2 or 3, got 0: []
    at com.mesosphere.sdk.offer.evaluate.placement.MarathonConstraintParser.parseRow(MarathonConstraintParser.java:106)

```

which is addressed by https://github.com/mesosphere/dcos-commons/pull/1903/commits/b8662cfffe13ca8cf585f7b86a9cb8675bf2c017

I just noticed https://github.com/mesosphere/dcos-commons/blob/INFINITY-2620-placement-constraint-defaults/sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/placement/MarathonConstraintParserTest.java#L296-L299 so it's possible we don't need to double-escape double quotes in config.json files and use single quotes instead, I'll push a commit trying that after the checks for the current code pass.

mpereira:
To recap:

### `config.json` files need stringified JSON to double-escape double quotes

```json
"placement_constraint": {
  "type": "string",
  "description": "The Marathon-style placement constraint string controlling node placement",
  "default": "[[\\\"hostname\\\", \\\"UNIQUE\\\"]]",
  "media": {
    "type": "application/x-zone-constraints+json"
  }
}
```

### `marathon.json.mustache` files have triple-brace mustache templates
```json
"PLACEMENT_CONSTRAINT": "{{{nodes.placement_constraint}}}",
```
expands into:
```json
"PLACEMENT_CONSTRAINT": "[[\"hostname\", \"UNIQUE\"]]",
```

### `svc.yml` files need the single quote around the triple braces
```yaml
placement: '{{{PLACEMENT_CONSTRAINT}}}'
```
expands into:
```yaml
placement: '[["hostname", "UNIQUE"]]'
```

Maybe we could make it possible to just use a regular array value in config.json in the future, but for now this should work.

frameworks/elastic/universe/config.json
```diff
@@ -126,9 +126,9 @@
           "default": 9300
         },
         "placement": {
-          "description": "Marathon-style placement constraints (e.g., hostname:UNIQUE).",
+          "description": "Marathon-style placement constraints (e.g., [[\"hostname\", \"UNIQUE\"]]).",
           "type": "string",
-          "default": "",
+          "default": "[]",
```
benclarkwood: This default should be set. Use max per 1.

frameworks/hdfs/universe/config.json
```diff
@@ -152,9 +152,9 @@
           "default": "0"
         },
         "placement": {
-          "description": "Marathon-style placement constraints (e.g., hostname:UNIQUE).",
+          "description": "Marathon-style placement constraints (e.g., [[\"hostname\", \"UNIQUE\"]]).",
           "type": "string",
-          "default": "",
+          "default": "[]",
```
benclarkwood: Default should be hostname max_per 1

mpereira:
@benclarkwood https://jira.mesosphere.com/browse/INFINITY-2769

PR #1904, opened 2017-12-01
Initialize system keyspaces to replication factor 3 for Cassandra

gabrielhartmann:


frameworks/cassandra/tests/test_recovery.py
```diff
@@ -21,27 +21,10 @@
 @pytest.fixture(scope='module', autouse=True)
 def configure_package(configure_security):
     try:
-        test_jobs = config.get_all_jobs()
```
benclarkwood: Did you mean to drop this?

gabrielhartmann:
@elezar 
Q: Should any of the parameters be configurable? (replication_factor, class)
A: No.  When we have a goal state of `ONCE` v `FINISHED`, this may change.

Q: Does the init_system_keyspaces task need a readiness check?
A: No, it either succeeds or fails.

Q: Would we every want to run it on a node other than node-0?
A: Maybe.  We want to run it on any running node.  `node-0` is a safe bet.

PR #1905, opened 2017-12-04
Create compatibility layer in preparation for new goal states

mrbrowning:
This PR enables the scheduler to parse `ONCE` and `FINISHED` goal states from the config store, enabling it to be downgraded to from a later version making use of those new states.


gabrielhartmann:
LGTM.  Ok to merge once tests pass.

PR #1908, opened 2017-12-05
INFINITY-2663: Add fault domain support to Kafka.

mpereira:


frameworks/kafka/src/main/dist/server.properties.mustache
```diff
@@ -19,6 +19,10 @@
 # The id of the broker. This must be set to a unique integer for each broker.
 broker.id={{POD_INSTANCE_INDEX}}
 
+{{#DETECT_ZONES}}
+broker.rack={{ZONE}}
+{{/DETECT_ZONES}}
+
```
imaxxs: Is there any default value for `broker.rack`? that can be used when `detect_zones` is misconfigured or set to false or the underlying cluster does not have rack awareness support?

mpereira: https://github.com/mesosphere/dcos-commons/pull/1908/commits/07e6b7ad45dff168e87c305131a53a8e7b1a541a

imaxxs:
LGTM.

PR #1916, opened 2017-12-06
INFINITY-2796 Fix lint, refrain from passing base64 version of yaml block to tasks

nickbp:
Everything with TASKCFG_ALL_* gets forwarded to tasks automatically. In this case the BASE64 version is only used by the scheduler itself so avoid passing the BASE64 version to tasks.

Ultimately a nitpick but doesn't hurt to fix.

### DIFF
--- a/frameworks/elastic/src/main/java/com/mesosphere/sdk/elastic/scheduler/Main.java
+++ b/frameworks/elastic/src/main/java/com/mesosphere/sdk/elastic/scheduler/Main.java
@@ -5,6 +5,7 @@ import com.mesosphere.sdk.specification.DefaultServiceSpec;
 import com.mesosphere.sdk.specification.yaml.RawServiceSpec;
 
 import java.io.File;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Base64;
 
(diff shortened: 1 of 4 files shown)


frameworks/elastic/src/main/java/com/mesosphere/sdk/elastic/scheduler/Main.java
```diff
@@ -24,24 +25,25 @@ public static void main(String[] args) throws Exception {
...
-        String base64_yaml = System.getenv(TASKCFG_ALL_CUSTOM_YAML_BLOCK_BASE64_ENV);
-        byte[] es_yaml_bytes = decoder.decode(base64_yaml);
-        String es_yaml_block = new String(es_yaml_bytes, "UTF-8");
-
         RawServiceSpec rawServiceSpec = RawServiceSpec.newBuilder(yamlSpecFile).build();
         SchedulerConfig schedulerConfig = SchedulerConfig.fromEnv();
+
         // Modify pod environments in two ways:
         // 1) Elastic is unhappy if cluster.name contains slashes. Replace any slashes with double-underscores.
         // 2) Base64 decode the custom YAML block.
```
loren: can delete this

nickbp: It's still doing that though...

PR #1917, opened 2017-12-07
Update kafka-zookeeper stub to beta-1

elezar:
The zookeeper stub for Kafka has expired.

This changes the stub to that being tested in Soak.


nickbp:
Manually checked that the json and a couple resources it referenced were fetchable, gonna just merge to unblock CI.

PR #1919, opened 2017-12-07
[INFINITY-2794] Enforce user intent around zones in Cassandra

gabrielhartmann:

PR #1922, opened 2017-12-07
Remove detect zones stuff

mpereira:


benclarkwood:
I expect this to not work until #1919 merges

mpereira:
Right, this branch is to be merged onto #1919's branch.

benclarkwood:
Ah, I c

benclarkwood:
Please wait and merge into master after #1919 merges. Don't want to reset it part way through CI in case this is ready before 1919.

benclarkwood:
Please wait and merge into master after #1919 merges. Don't want to reset it part way through CI in case this is ready before 1919.

PR #1924, opened 2017-12-07
Create and wait for topic in all auth tests before trying to use it.

benclarkwood:


elezar:
I have created https://jira.mesosphere.com/browse/INFINITY-2801 to track cleaning this up a little going forward.

PR #1928, opened 2017-12-08
INFINITY-2806 Support "non-essential" tasks which are relaunched without affecting rest of pod

nickbp:
When multiple tasks are defined in a given pod, we currently require that if any individual task fails, all tasks in that pod are restarted as a group. This is convenient for developers because it ensures a consistent startup sequence for pods across initial deployment, upgrade/config changes, and task recovery.

However, there are cases where a service may require certain tasks within a pod instance to be restarted without disturbing other tasks in the same pod instance. To support this scenario we add an `essential` boolean flag to tasks which defaults to `true` (current behavior) but can be set `false` for any tasks that should be recovered without disturbing other tasks in the same pod instance.

This PR adds service unit tests verifying both the existing essential-relaunch-pod behavior and the new nonessential-relaunch-task behavior.


nickbp:
ping

docs/pages/reference/yaml-reference.md
```diff
@@ -146,6 +146,10 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
       <li><code>FINISHED</code>: The task should launch and exit successfully (zero exit code). If the task fails (nonzero exit code) then it is retried without relaunching the entire pod.</li>
       </ul></div>
 
+    * `essential`
+
+      Marks this task as either "Essential", where relaunching this task results in relaunching all `RUNNING` tasks in the pod instance, or "Non-essential", where relaunching it does not affect other `RUNNING` tasks in the pod instance. By default this value is `true`, such that if this task needs to be relaunched, then all other tasks sharing the same pod instance are relaunched alongside it. This is only a factor in cases where a given pod has multiple `RUNNING` tasks defined.
```
gabrielhartmann: I think the description here while technically accurate is a little confusing.  I would have written this form the perspective of failure instead of relaunch.  e.g. `When an essential task fails, all other tasks in the same pod will fail with it.`  The implementation detail that this is accomplished by the Scheduler on relaunch is maybe uninteresting in this document.

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
```diff
@@ -50,6 +52,133 @@ public void testDefaultDeployment() throws Exception {
         runDefaultDeployment();
     }
 
+    /**
+     * Checks that if an unessential task in a pod fails, that the other task in the same pod is unaffected.
+     */
+    @Test
+    public void testNonEssentialTaskFailure() throws Exception {
```
gabrielhartmann: NIT:  I've stopped prefixing all tests with `test`.  Things are verbose enough.  It says `@Test` right there.

docs/pages/reference/yaml-reference.md
nickbp: Done in #1953

PR #1941, opened 2017-12-12
INFINITY-2769: Default to unique hostname constraint.

mpereira:


gabrielhartmann:
Actually @mpereira this should be closed.  The empty defaults is the correct approach.  Anything here would be ANDed with the TaskType placement constraints in `Main.java`.  The default approach of not adding anything to the programmatic placement constraints is correct.  Talk to @NimaVaziri.  @benclarkwood, if there's some disagreement about consistency here, let me know.

mpereira:
@gabrielhartmann does the Elastic scheduler also have enforced constraints defined in java code?

gabrielhartmann:
@mpereira: Doesn't look like it: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/elastic/src/main/java/com/mesosphere/sdk/elastic/scheduler/Main.java#L16-L48

mpereira:
@gabrielhartmann so should it still not have default placement constraints in configuration?

benclarkwood:
I would think Elastic should have default constraints.

I always forget HDFS has java'd placement constraints

mpereira:
@gabrielhartmann @benclarkwood I opened another PR that only changes the Elastic placement constraints: https://github.com/mesosphere/dcos-commons/pull/1958

PR #1944, opened 2017-12-12
Add KDC docker configs

NimaVaziri:
Test to confirm functionality: built an image via the Dockerfile recipe and used said image in testing kerberized kafka tests and they passed as expected.

### DIFF
--- a/tools/kdc/Dockerfile
+++ b/tools/kdc/Dockerfile
@@ -1,0 +1,9 @@
+FROM nvaziri/kdc:dev
+
+RUN mkdir /kdc
+COPY run.sh /kdc/run.sh
+COPY kdc.conf /etc/heimdal-kdc/kdc.conf
+RUN chown -R nobody:nogroup /kdc
+RUN chmod -R 700 /kdc
+
+CMD /kdc/run.sh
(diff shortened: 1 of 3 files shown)


tools/kdc/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM nvaziri/kdc:dev
```
elezar: Is this correct? Is this not the image we are generating?

elezar: Was this the Docker file to start with? What was the original `FROM` statement?

tools/kdc.json
```diff
@@ -26,9 +26,5 @@
       "name": "http"
```
elezar: Since we no longer run the HTTP server, does this mean that this port is no longer required?

NimaVaziri: Yup...good catch

tools/kdc/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM mesosphere/kdc:latest
+
```
elezar: Should we add a `MAINTAINER` tag?

tools/kdc/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+RUN mkdir /kdc
+COPY run.sh /kdc/run.sh
+COPY kdc.conf /etc/heimdal-kdc/kdc.conf
+RUN chown -R nobody:nogroup /kdc
+RUN chmod -R 777 /var/lib/heimdal-kdc/
```
elezar: These permissions seem a little extreme, but I'm ok with them as is for the time being.

PR #1945, opened 2017-12-12
Upstream change from zk repo

benclarkwood:


testing/sdk_api.py
```diff
@@ -12,17 +12,29 @@
 log = logging.getLogger(__name__)
 
 
-def get(service_name, endpoint):
+def get(service_name, endpoint, **kwargs):
```
elezar: `kwargs` are not used.

PR #1871, opened 2017-11-22
[INFINITY-2583] ensure uniform security settings

elezar:
This PR unifies the security settings between all services. It does so by:

 * Move `service.tls` and `service.tls_allow_plaintext` to `service.security.transport_layer.enabled` and `service.security.transport_layer.allow_plaintext` in `config.json`
* Rename environment variables:
  * `*{{FRAMEWORK}}_ENABLE_TLS` => `*SECURITY_TRANSPORT_LAYER_ENABLED`
  * `*{{FRAMEWORK}}_ALLOW_PLAINTEXT` => `*SECURITY_TRANSPORT_LAYER_ALLOW_PLAINTEXT`

This has been done for:
- [ ] Kafka #1946 
- [ ] Cassandra #1947
- [ ] Elastic
- [ ] HDFS (rebased from #1873)
- [ ] Kafka ZooKeeper (mesosphere/dcos-zookeeper#47)

PR #1948, opened 2017-12-13
[INFINITY-2583][INFINITY-2831] Ensure uniform security settings for Elastic

elezar:
This PR splits #1871 into a separate PR for Elastic.


mpereira:
Should we do something similar for kibana too? https://github.com/mesosphere/dcos-commons/blob/master/frameworks/elastic/universe-kibana/config.json#L39-L43

elezar:
@mpereira what would you propose? Simply changing the property to `elasticsearch_transport_encryption`, or some more complex expansion? @loren @benclarkwood what would make sense in this case?

mpereira:
@elezar would the same shape make sense?
```
kibana config.json
{ "service": { "security": { "transport_encryption": { "enabled": true } } } }
```
> This is tracked separately

👍

elezar:
I've added a comment to the Kibana JIRA.

PR #1953, opened 2017-12-14
INFINITY-2816 Support non-essential tasks in 1.9: Custom executor changes

nickbp:
In order to support non-essential tasks on 1.9, several changes were needed to the custom executor's handling of exited tasks:
- Don't automatically terminate the executor, allow scheduler to handle killing other tasks in the pod if it wants
- Ensure that any health/readiness checks for the task are cleaned up when the task is no longer running
- Add a monitor which exits if no tasks are running (Note: in practice on 1.9 it looks like the Mesos agent immediately kills the container/executor after the last task exits, but it probably still can't hurt to leave in this check, just in case)

Other fixes:
- Some logging improvements in the custom executor
- YAML reference language clarified per review comment in #1928 
- Add integration tests which verify that ExecutorInfos match and that a shared volume behaves as expected
- Update SchedulerTest lib to support producing offers with any necessary executor volumes.


docs/pages/reference/yaml-reference.md
```diff
@@ -148,7 +148,7 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
     * `essential`
 
-      Marks this task as either "Essential", where relaunching this task results in relaunching all `RUNNING` tasks in the pod instance, or "Non-essential", where relaunching it does not affect other `RUNNING` tasks in the pod instance. By default this value is `true`, such that if this task needs to be relaunched, then all other tasks sharing the same pod instance are relaunched alongside it. This is only a factor in cases where a given pod has multiple `RUNNING` tasks defined.
+      Marks this task as either "Essential", where task failure results in relaunching all tasks in the pod as a unit, or "Non-essential", where task failure results in only relaunching this task and leaving other tasks in the pod unaffected. By default this value is `true`, such that the failure of the task will result in relaunching all tasks in the pod. This is only a factor in cases where a given pod has multiple `RUNNING` tasks defined.
```
gabrielhartmann: @sascala Do you want to look at this?

* `... multiple tasks with a goal state of RUNNING`
* `... only relaunching the failed task ...`

frameworks/helloworld/tests/test_nonessential_tasks.py
```diff
@@ -0,0 +1,109 @@
...
+        sdk_hosts.system_host(config.SERVICE_NAME, 'hello-0-essential'))
+
+    sdk_tasks.check_tasks_updated(config.SERVICE_NAME, 'hello-0', old_ids) # wait for ids to change...
+    sdk_plan.wait_for_completed_recovery(config.SERVICE_NAME) # ...and for tasks to be up and running
+
+    # the first verify_shared_executor call deleted the files. both should have come back via the relaunch.
+    verify_shared_executor('hello-0', delete_files=False) # leave files as-is for the next test
+
+
+@pytest.mark.nick
```
gabrielhartmann: You left some marks around.  Not blocking.  You can use `./test.sh -k test_kill_nonessential` of you want to run just this test, or `./test.sh -k <file_name>` if you want to run just one test file.

PR #1957, opened 2017-12-14
Don't escape Marathon config, just direct calls to Cosmos

benclarkwood:


benclarkwood:
Si

benclarkwood:
Wait, I lied

benclarkwood:
There was a direct marathon config udpate

benclarkwood:
I accidentally escaped that too

PR #1959, opened 2017-12-14
[INFINITY-2836] Fix scheduler DNS name in config file

NimaVaziri:


frameworks/hdfs/docs/kerberos.md
```diff
@@ -32,7 +32,7 @@ hdfs/data-1-node.hdfs.autoip.dcos.thisdcos.directory@LOCAL
 HTTP/data-1-node.hdfs.autoip.dcos.thisdcos.directory@LOCAL
 hdfs/data-2-node.hdfs.autoip.dcos.thisdcos.directory@LOCAL
 HTTP/data-2-node.hdfs.autoip.dcos.thisdcos.directory@LOCAL
-HTTP/hdfs.marathon.autoip.dcos.thisdcos.directory@LOCAL
```
elezar: Should we add a link as to how the `hdfs` part changes if we use the non-default name?

testing/sdk_utils.py
```diff
@@ -45,6 +45,11 @@ def get_foldered_name(service_name):
         return service_name
     return '/test/integration/' + service_name
 
+def get_foldered_dns_name(service_name):
```
elezar: Why is this in "utils" and not in something else like "hosts"?

frameworks/hdfs/tests/test_auth.py
```diff
@@ -13,6 +13,8 @@
 
 
 log = logging.getLogger(__name__)
+foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
```
elezar: I would suggest that we not use global variables, and instead use test fixtures if required. I don't feel too strongly about this though.

frameworks/hdfs/tests/test_auth.py
```diff
@@ -46,6 +48,9 @@ def get_principals() -> list:
             )
         )
     principals.extend(config.CLIENT_PRINCIPALS.values())
+
+    service_principal = "HTTP/api.{}.marathon.l4lb.thisdcos.directory".format(foldered_dns_name)
```
elezar: This is a bit more than a "service" principal though, right?

frameworks/hdfs/docs/kerberos.md
NimaVaziri: Perhaps this is addressed in https://github.com/mesosphere/dcos-commons/pull/1960

PR #1960, opened 2017-12-14
Clarify foldered service name for principals

NimaVaziri:


frameworks/hdfs/docs/kerberos.md
```diff
@@ -46,6 +46,12 @@ Given a service name of `hdfs-demo`, the principal for a name node becomes:
 hdfs/name-0-node.hdfs-demo.autoip.dcos.thisdcos.directory@LOCAL
 ```
 
+Also note that if the service name is foldered, such as `myfolder/hdfs`, then the FQDN in the instance of the principal
```
elezar: Should the example have a `-` or something in it? Or at least link docs on how the service name is mapped to a DNS name?

NimaVaziri: Where do we expect the `-` in the example? I looked at the marathon grouped applications docs but sadly didn't come across any that explained the DNS naming convention.

elezar: It's not important to have the `-`, I was just thinking that we should link to something indicating how the FDQN should be generated.  Not critical though.

PR #1966, opened 2017-12-15
[INFINITY-2837] Include the readiness check for goal state

elezar:
The readiness check should be used to determine whether a task is at its RUNNING goal state.

If this is not taken into account, a scheduler restart would result in a RUNNING task with a failed readiness check being marked as completed.

Ideally we should use similar logic for task updates and the initial determination of the steps.


elezar:
I'm closing this so that I can address it properly once I get time for the issue again.

PR #1967, opened 2017-12-16
CASSANDRA-659 Replace docker image with local python dist

nickbp:
Cassandra currently runs against a minimal docker image containing Python (needed for `cqlsh`, `aws`, and `az`), but the image creates problems with UIDs/permissions. This PR switches to using a local stock build of python2.7, removing the prior docker image.

Had previously looked into replacing `cqlsh` with a custom Go tool, but "DESCRIBE ..." commands are all implemented _in_ `cqlsh`, and as such cannot easily be implemented as passthrough queries to the server.

TODO: as of PR creation, the python package URL is in a dev bucket to see if things are green. need to move that to a prod bucket (and update `resource.json`) before merge


frameworks/cassandra/universe/resource.json
```diff
@@ -1,13 +1,9 @@
...
   "assets": {
-    "container": {
-      "docker": {
-        "cassandra": "mesosphere/cassandra:1.0.26-3.0.10"
-      }
-    },
     "uris": {
       "jre-tar-gz": "{{jre-url}}",
       "cassandra-jre-tar-gz": "{{jre-url}}",
+      "cassandra-python-tar-gz": "https://nick-dev.s3.amazonaws.com/python-2.7.14.tar.gz",
```
benclarkwood: You can use the Jenkins upload production asset job for this.

frameworks/cassandra/build-python.sh
```diff
@@ -0,0 +1,76 @@
...
+    # With this hack, invoking bin scripts (e.g. 'aws') should work as expected assuming "python"
+    # is in PATH. Note: this doesn't affect the 'az' command as that's technically a bash script.
+    sed -i "s:$(pwd)/${DIR_PREFIX}/bin/python:/usr/bin/env python:g" ${DIR_PREFIX}/bin/*
+fi
+
+rm -f $OUTPUT_TGZ_FILE
+tar czf $OUTPUT_TGZ_FILE $DIR_PREFIX
+
+echo "Build complete. Upload result, then update URL in resource.json:"
+echo "  aws s3 cp --acl public-read $(pwd)/${OUTPUT_TGZ_FILE} s3://YOURBUCKET/${OUTPUT_TGZ_FILE}"
```
mrbrowning: Could maybe configure `YOUR_BUCKET` via an envvar too?

nickbp: Figured it'd make sense to have the upload be a separate step -- maybe the caller has somewhere else they want to put it. In practice this script would be getting run manually if e.g. the python version needed to be bumped or additional things needed to be added

PR #1968, opened 2017-12-18
INFINITY-159: Elasticsearch 6.0.0

mpereira:
In addition to upgrading the Elastic service base tech to version 6.0.0 this PR makes it possible to disable the StatsD plugin that sends metrics to DC/OS. This is because there were breaking changes in the ES Java API (which the StatsD plugin uses) and I still wasn't able to get it passing tests. I have a wip branch [here](https://github.com/Automattic/elasticsearch-statsd-plugin/compare/master...mpereira:6.0.0.0?expand=1) and I'm seeing pretty much [the same test failures Loren was getting with the RC1](https://github.com/Automattic/elasticsearch-statsd-plugin/issues/23).

If we can't figure out the plugin test failures on time maybe we could release 6.0.0 without StatsD first and then do another release later when we get it working?


nickbp:
FYI merge conflict in `frameworks/elastic/versions.sh`

mpereira:
Closing since we don't want the ability to disable statsd. @loren has a [PR](https://github.com/Automattic/elasticsearch-statsd-plugin/pull/26) that makes it work with the ES6 APIs and will push an SDK PR soon.

PR #1976, opened 2017-12-19
INFINITY-2378/2820 Collect all relevant task logs from test, reduce noise

nickbp:
- Get task ids before and after a test, then collect logs from the tasks that appeared while a failed test was running. Print progress as log collection is performed so that the user knows what's happening.
- Turn down some noise from underlying libraries by setting them to WARNING.
- Emit all test logs to stdout instead of a mixture of stdout+stderr. This gets around teamcity printing the two separately.
- Broaden usage of logging plan state when a test fails.


mbdcos:
LGTM

testing/sdk_plan.py
```diff
@@ -25,6 +27,16 @@ def get_recovery_plan(service_name):
...
 
 
+def list_plans(service_name):
+    def fn():
+        output = sdk_api.get(service_name, '/v1/plans')
+        try:
+            return output.json()
+        except:
+            return False
+    return shakedown.wait_for(fn)
```
NimaVaziri: Maybe use the `retry` library as we're gradually reducing dependency on shakedown?

frameworks/cassandra/tests/conftest.py
```diff
@@ -1,37 +1,21 @@
...
-
 import pytest
+import sdk_plan
 import sdk_repository
 import sdk_security
-import sdk_utils
-import shakedown
 
-log = logging.getLogger(__name__)
+SERVICE_NAME = 'cassandra'
```
NimaVaziri: We usually define `SERVICE_NAME` in [`config.py`](https://github.com/mesosphere/dcos-commons/blob/master/frameworks/cassandra/tests/config.py#L16) for the services so perhaps use/import that for convergence's sake?

nickbp: i tried consuming from there directly, but `import config` failed

NimaVaziri: ah, might need to specify directory i.e. `from tests import config` as in https://github.com/mesosphere/dcos-commons/blob/master/frameworks/cassandra/tests/test_sanity.py#L15

nickbp: aha, yep, forgot about that

PR #1978, opened 2017-12-19
[INFINITY-2848/INFINITY-2871] Fix transient to permanent failure transitions and many flakes

gabrielhartmann:
This PR has three functional changes.   The rest is minor refactoring.

1. We had a regression.  A task which was attempting to recovery from what it thought was a transient failure (`TASK_LOST`) did not transition to permanent failure recovery when an end-user marked the task as permanently failed.  A scheduler restart generated the correct outcome.  To fix this [we no longer count recovery steps which have changed their recovery type as "in progress."](https://github.com/mesosphere/dcos-commons/pull/1978/files#diff-a0f3058f3b77bb8d6d9460af11fbf5feR235)  This has the effect of allowing the DefaultRecoveryPlanManager to generate the appropriate new recovery step.

2. If an end-user attempted to permanently replace a task on a host which had permanently failed we waited until a TaskStatus was returned by Mesos.  This caused us to delay permanent failure recovery substantially.  [Now we consider a task as "needing recovery" if it has been marked as permanently failed.](https://github.com/mesosphere/dcos-commons/pull/1978/files#diff-5bfb70eb3c198d78033152e71bba6e94R94)

3. [We're now more careful about when we should re-use an Executor](https://github.com/mesosphere/dcos-commons/pull/1978/files#diff-c3d66356a3e10f18538075653f41f103R222).  In the past we required that a task have reached the `RUNNING` state.  In fact the correct states were `STAGING, STARTING and RUNNING`.  I noticed this in passing so I made the change.

Note:
I will be adding more integration tests to make sure this doesn't regress again in another PR.  We've written them in the past and somehow this slipped through again.  It's going to take me a little time, but I'd like this fix out for the beta releases today.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ tools/elastic
 tools/helloworld
 tools/template
 /test_*.json
+/logs/
 *.log
 *.pem
 
--- a/conftest.py
+++ b/conftest.py
@@ -6,7 +6,10 @@ E.G. py.test frameworks/<your-frameworks>/tests
 """
 import logging
 import os
+import os.path
+import shutil
 import subprocess
+import sys
 
 import pytest
 import requests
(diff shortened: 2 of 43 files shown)


nickbp:
Superceded by #1994

PR #1981, opened 2017-12-20
Add extensive documentation and examples for goal states

mrbrowning:


mrbrowning:
@crackcode123 your call -- since both upgrade and downgrade will happen seamlessly, I think it's probably sufficient to mention that in the changelog for the release that contains the ONCE/FINISHED work.

PR #1986, opened 2017-12-21
INFINITY-1119/INFINITY-2898 (MINOR DEVTOOL CHANGES) Publish/reuse default CLIs, use local GOPATH

nickbp:
Changes to make builds simpler for projects both inside and outside of `dcos-commons`.

Build structure:
- Update the root-level build.sh to actually work, and to actually build/test all the SDK artifacts.
- `frameworks/*/build.sh` now invoke the root-level build.sh to get SDK artifacts (logic moved out of common `build_framework.sh`, where it didn't make sense for external frameworks). This is effectively the differentiator between frameworks in dcos-commons (use local SDK artifacts, dynamic resource.json) and external frameworks (use SDK artifacts tied to a release, static resource.json). Having it done in each dcos-commons framework's build.sh avoids polluting the common `build_package.sh` with logic that doesn't apply to external frameworks.
- `build_framework.sh` was renamed to `build_package.sh` and updated to no longer do any internal builds. Callers are expected to invoke their builds directly (including the Scheduler), and pass all artifacts into `build_package.sh`. In practice, developers often didn't want the builds produced by `build_framework.sh` and had to manually disable them via `CLI_DIR`/`BOOTSTRAP_DIR`.
- To help migration, the old `build_framework.sh` has been replaced with a script describing the steps to migrate to `build_package.sh`.

Go builds:
- When releasing the SDK, also publish builds of a "default" CLI. All but two services can just use this default CLI, and thereby wouldn't directly need any golang tooling at all in order to build.
- When building Go binaries, create a local `GOPATH` in `$REPO_ROOT/.gopath/` instead of modifying the host `GOPATH`'s contents, thereby leaving the user's own `GOPATH` untouched. This is simplified by the fact that we already require libraries to be vendored.
- As an optimization, `build_go_exe.sh` now supports performing multiple per-platform builds in the same run. In cases where multiple per-platform builds are needed, this greatly speeds up build time, as only one 'canary' build is needed. The change is compatible with existing callers -- to take advantage of this, callers can just specify a space-separated list of platforms to build on.


elezar:
Thanks @nickbp. Does this mean that for non-monorepo frameworks, we can use the CLI in the same way as what we do with the other artefacts?

frameworks/kafka/cli/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
elezar: Would it make sense to link this to the `sdk/cli/build.sh` version of the file? May not be worth the effort though.

frameworks/kafka/cli/build.sh
```diff
@@ -0,0 +1,16 @@
...
+EXE_NAME="dcos-service-cli"
+
+# required env:
+export REPO_ROOT_DIR=$(dirname $(dirname $(dirname $CUR_DIR)))
+export REPO_NAME=$(basename $REPO_ROOT_DIR)
+$REPO_ROOT_DIR/tools/build_go_exe.sh frameworks/kafka/cli/ ${EXE_NAME}-linux linux
+$REPO_ROOT_DIR/tools/build_go_exe.sh frameworks/kafka/cli/ ${EXE_NAME}-darwin darwin
+$REPO_ROOT_DIR/tools/build_go_exe.sh frameworks/kafka/cli/ ${EXE_NAME}.exe windows
+
+echo $(pwd)/${EXE_NAME}-linux $(pwd)/${EXE_NAME}-darwin $(pwd)/${EXE_NAME}.exe
```
elezar: What is the purpose of this final echo? At first I thought it was used to capture the artefact paths, but that seems to be hard-coded dependent on the `CLI_DIR`.

sdk/cli/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
elezar: See comments regarding `frameworks/kafka/cli/build.sh` as well.

tools/build_framework.sh
```diff
@@ -33,7 +33,6 @@ shift
 echo "Building $FRAMEWORK_NAME in $FRAMEWORK_DIR:"
 
 # optional args:
-cli_only=
 custom_artifacts=
 publish_method="no"
 while [ $# -gt 0 ]; do
```
elezar: Out of scope of this PR, but we could consider settling on a single way to handle command-line arguments. We use `getopts` elsewhere.

nickbp: Ended up doing this as part of `build_package.sh`.

frameworks/elastic/build.sh
```diff
@@ -2,16 +2,29 @@
 set -e
 
 FRAMEWORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-BUILD_DIR=$FRAMEWORK_DIR/build/distributions
+REPO_ROOT_DIR=$(dirname $(dirname $FRAMEWORK_DIR))
```
gabrielhartmann: The `build.sh` for each of the frameworks looks very very similar to me.  Is it worth factoring things out or is the copy/paste due to a desire maintain parity with services outside the mono-repo?

Maybe we could maintain parity and DRY things out a little?

nickbp: There's still some mild variation with e.g. per-service utils, kafka's custom CLI, custom doc URLs, etc... Gave an attempt at deduping but it ended up being more than it was worth

PR #1988, opened 2017-12-21
[ELASTIC-161] Increase timeout for elastic install

mbdcos:


mbdcos:
30min shld be enough. Revisit if you see this problem again after http timeout fix in dcos CLI.

PR #1993, opened 2017-12-22
INFINITY-2871 Several improvements to test flakes and logging behavior

nickbp:
- Avoid taking forever to fetch plans if scheduler is gone: Set a short timeout, exit early.
- Task log fetching: Fetch all tasks launched since the last failure in the same test suite. Previous logic only fetched tasks which were launched within the same test. Give better information as tasks are fetched/written.
- Task log writing: Include an index prefix in the log directory, so that test ordering is easy to follow when browsing logs.
- Logs: Include "START/END" headers before/after every test, and various other log cleanup.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -192,7 +192,7 @@ def setup_artifact_path(item: pytest.Item, artifact_name: str):
     return os.path.join(output_dir, artifact_name)
 
 
-def get_task_files_for_id(task_id: str):
+def get_task_files_for_id(task_id: str) -> set:
     try:
         return set(subprocess.check_output(['dcos', 'task', 'ls', task_id, '--all']).decode().split())
     except:
@@ -200,7 +200,7 @@ def get_task_files_for_id(task_id: str):
         return set()
 
 
-def get_task_log_for_id(task_id: str,  task_file: str='stdout', lines: int=1000000):
+def get_task_log_for_id(task_id: str,  task_file: str='stdout', lines: int=1000000) -> str:
     log.info('Fetching {} from {}'.format(task_file, task_id))
     result = subprocess.run(
         ['dcos', 'task', 'log', task_id, '--all', '--lines', str(lines), task_file],
@@ -211,7 +211,7 @@ def get_task_log_for_id(task_id: str,  task_file: str='stdout', lines: int=10000
         errmessage = result.stderr.decode()
         if not errmessage.startswith('No files exist. Exiting.'):
             log.error('Failed to get {} task log for task_id={}: {}'.format(task_file, task_id, errmessage))
-        return None
+        return ''
     return result.stdout.decode()
 
 
--- a/conftest.py
+++ b/conftest.py
@@ -15,6 +15,7 @@ import pytest
 import requests
 import sdk_security
 import sdk_utils
+import teamcity
 
 log_level = os.getenv('TEST_LOG_LEVEL', 'INFO').upper()
 log_levels = ('DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL', 'EXCEPTION')
@@ -130,7 +131,9 @@ def pytest_runtest_makereport(item, call):
 def pytest_runtest_teardown(item):
     '''Hook to run after every test.'''
     # Inject footer at end of test, may be followed by additional teardown.
-    print('''
+    # Don't do this when running in teamcity, where it's redundant.
+    if not teamcity.is_running_under_teamcity():
+        print('''
 ==========
 ======= END: {}::{}
 =========='''.format(sdk_utils.get_test_suite_name(item), item.name))
@@ -139,7 +142,9 @@ def pytest_runtest_teardown(item):
 def pytest_runtest_setup(item):
     '''Hook to run before every test.'''
     # Inject header at start of test, following automatic "path/to/test_file.py::test_name":
-    print('''
+    # Don't do this when running in teamcity, where it's redundant.
+    if not teamcity.is_running_under_teamcity():
+        print('''
 ==========
 ======= START: {}::{}
 =========='''.format(sdk_utils.get_test_suite_name(item), item.name))
--- a/conftest.py
+++ b/conftest.py
@@ -166,7 +166,7 @@ def pytest_runtest_setup(item):
         # 3 Remove any prior logs for the test suite.
         test_log_dir = sdk_utils.get_test_suite_log_directory(item)
         if os.path.exists(test_log_dir):
-            log.info('Deleting existing test suite directory: {}/'.format(test_log_dir))
+            log.info('Deleting existing test suite logs: {}/'.format(test_log_dir))
             shutil.rmtree(test_log_dir)
 
     # Increment the test index (to 1, if this is a new suite), and pass the value to sdk_utils for use internally.
@@ -193,7 +193,11 @@ def setup_artifact_path(item: pytest.Item, artifact_name: str):
 
 
 def get_task_files_for_id(task_id: str):
-    return set(subprocess.check_output(['dcos', 'task', 'ls', task_id, '--all']).decode().split())
+    try:
+        return set(subprocess.check_output(['dcos', 'task', 'ls', task_id, '--all']).decode().split())
+    except:
+        log.exception('Failed to get list of files for task: {}'.format(task_id))
+        return set()
 
 
 def get_task_log_for_id(task_id: str,  task_file: str='stdout', lines: int=1000000):
--- a/conftest.py
+++ b/conftest.py
@@ -39,10 +39,41 @@ for noise_source in [
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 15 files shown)


nickbp:
Closing this in favor of having everything in PR #1978

nickbp:
Nevermind, that was superceded by #1994

conftest.py
```diff
@@ -147,30 +192,55 @@ def setup_artifact_path(item: pytest.Item, artifact_name: str):
     return os.path.join(output_dir, artifact_name)
 
 
-def get_rotating_task_log_lines(task_id: str, known_task_files: set, task_file: str):
+def get_task_files_for_id(task_id: str):
```
elezar: Could we annotate that this returns a `set`?
```python
def get_task_files_for_id(task_id: str) -> set:
```

conftest.py
```diff
@@ -147,30 +192,55 @@ def setup_artifact_path(item: pytest.Item, artifact_name: str):
...
+    result = subprocess.run(
+        ['dcos', 'task', 'log', task_id, '--all', '--lines', str(lines), task_file],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    if result.returncode:
+        errmessage = result.stderr.decode()
+        if not errmessage.startswith('No files exist. Exiting.'):
+            log.error('Failed to get {} task log for task_id={}: {}'.format(task_file, task_id, errmessage))
+        return None
```
elezar: The return type here is inconsistent. Should we not return `""` instead?

testing/sdk_install.py
```diff
@@ -52,30 +77,30 @@ def install(
...
-        expected_running_tasks)
+        timeout_seconds)
 
     # 2. Wait for the scheduler to be idle (as implied by deploy plan completion and suppressed bit)
     # This should be skipped ONLY when it's known that the scheduler will be stuck in an incomplete state.
     if wait_for_deployment:
         # this can take a while, default is 15 minutes. for example with HDFS, we can hit the expected
         # total task count via FINISHED tasks, without actually completing deployment
-        log.info("Waiting for {}/{} to finish deployment plan...".format(
+        log.info('Waiting for package={} service={} to finish deployment plan...'.format(
```
elezar: Since we're editing this line anyway, we could switch to using `%s` in log messages instead of `{}` and `format` to keep linters happy.

testing/sdk_plan.py
```diff
@@ -201,13 +203,22 @@ def get_plans_on_failure(request):
...
-            if not plan:
-                log.error('Unable to fetch {} plan for {}'.format(framework_name, plan_name))
-                continue
-            out_path = os.path.join(
-                sdk_utils.get_test_log_directory(request.node),
-                '{}_plan.txt'.format(plan_name))
-            with open(out_path, 'w') as f:
-                f.write(json.dumps(plan, indent=2))
+        try:
+            log.info('Fetching plans from {}...'.format(framework_name))
```
elezar: Since we don't have an "official" linter, it's not critical, but we could consider using `%s` for log messages instead of `{}`.

testing/sdk_plan.py
```diff
@@ -201,13 +203,22 @@ def get_plans_on_failure(request):
...
+            log.info('Plans for {}: {}'.format(framework_name, plan_names))
+            for plan_name in plan_names:
+                log.info('Fetching {} plan: {}'.format(framework_name, plan_name))
+                plan = get_plan(framework_name, plan_name, 5)
+                if not plan:
+                    log.error('Unable to fetch {} plan for {}'.format(framework_name, plan_name))
+                    continue
+                out_path = os.path.join(
+                    sdk_utils.get_test_log_directory(request.node),
+                    '{}_plan.txt'.format(plan_name))
```
elezar: should the output be `.txt`? Are we not writing a `.json` file? We could also use:
```python
json.dump(plan, f, indent=2)
```
to write directly to the file (unless the contents are important from a logging perspective?

testing/sdk_upgrade.py
```diff
@@ -196,7 +196,7 @@ def _upgrade_or_downgrade(
 
         # this can take a while, default is 15 minutes. for example with HDFS, we can hit the expected
         # total task count via ONCE tasks, without actually completing deployment
-        log.info("Waiting for {}/{} to finish deployment plan...".format(
+        log.info("Waiting for package={} service={} to finish deployment plan...".format(
```
elezar: `%s` instead of `{}`.

testing/sdk_utils.py
```diff
@@ -56,13 +59,40 @@ def dcos_version_less_than(version):
...
+    Should start at 1 for the first test in the suite.'''
+    global test_index
+    test_index = index
+
+
+def get_test_suite_name(pytest_node):
+    '''Returns the test suite name to use for a given test.'''
+    # frameworks/template/tests/test_sanity.py => test_sanity_py
+    # tests/test_sanity.py => test_sanity_py
+    return pytest_node.parent.name.split('/')[-1].replace('.','_')
```
elezar: We could use:
```python
os.path.basename.replace('.', '_')
```
as a "safer" version.

testing/sdk_utils.py
```diff
@@ -56,13 +59,40 @@ def dcos_version_less_than(version):
...
 def get_test_log_directory(pytest_node):
+    '''Returns the directory for the logs of a single test.
+    For the parent test suite directory, see get_test_suite_log_directory().'''
     # full item.listchain() is e.g.:
     # - ['build', 'frameworks/template/tests/test_sanity.py', 'test_install']
     # - ['build', 'tests/test_sanity.py', 'test_install']
     # we want to turn both cases into: 'logs/test_sanity_py/test_install'
-    parent_name = pytest_node.parent.name.split('/')[-1].replace('.','_')
-    return os.path.join('logs', parent_name, pytest_node.name)
+    if test_index >= 0:
```
elezar: Why do we check for 0 if we have: "Should start at 1 for the first test in the suite" where this is referenced above? Also, should we not explicitly indicate that this is a global?

nickbp: `global` not required if the value is only being read

testing/sdk_install.py
nickbp:
```
dcos-commons$ grep -R "{}" testing/ frameworks/*/tests/ | wc -l
499
dcos-commons$ egrep -R "%[sd]" testing/ frameworks/*/tests/ | wc -l
63
```

testing/sdk_plan.py
nickbp: It'd be nice to keep the following logging of the number of bytes. Checked if `json.dump()` returned the number of bytes written, but it didn't.

testing/sdk_utils.py
nickbp: Oh right, good point, done.

testing/sdk_utils.py
nickbp: ...though it would be nice to be explicit here on where `test_index` is coming from.

test.sh
```diff
@@ -208,5 +208,5 @@ docker run --rm \
     -w /build \
     -t \
     -i \
-    mesosphere/dcos-commons:latest \
+    mesosphere/dcos-commons:uninstall_fix \
```
gabrielhartmann: Do we really need to do this?  If we're doing this in a real PR I'd say we can drop this, and on merge to master `mesosphere/dcos-commons:latest` will be appropriate.

Brings up a question.  Does our build use the published image or the one generated by the PR.  I think it should be the latter right?

nickbp: Reverted back to `master`.

No idea.

testing/sdk_utils.py
nickbp: Done. Added `global` statement, checking `> 0`

conftest.py
nickbp: Done.

conftest.py
nickbp: Done.

gabrielhartmann:
@nickbp Your `helloworld` test is going to fail because of Scheduler deadlock. https://jira.mesosphere.com/browse/INFINITY-2875

gabrielhartmann:
@nickbp: I rebased on https://github.com/mesosphere/dcos-commons/pull/1999 `deadlock` branch.

gabrielhartmann:
I don't know if this is a result of changes here or not.  Did you touch Metronome jobs?
https://jira.mesosphere.com/browse/INFINITY-2876

Restarting the Cassandra tests.

nickbp:
I'd guess a new metronome regression. There were recently fixes to the list of jobs in the cassandra tests (specifically a run-once delete job to be run before the test begins) but they shouldn't affect how they're actually _run_. These flakes also only appeared today while the job management fixes were a couple days ago.

testing/sdk_install.py
elezar: Yes, I'm not saying that we're consistent.

PR #2003, opened 2017-12-26
[INFINITY-2881] Move from HTTP to MESOS_HTTP health-checks on schedulers

gabrielhartmann:


gabrielhartmann:
@elezar Thanks.  Good catch.

PR #2005, opened 2017-12-28
[HDFS-491] ZKFC wait for name node resolution

NimaVaziri:


frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -293,7 +293,9 @@ pods:
           {{#SECURITY_KERBEROS_ENABLED}}
           export HADOOP_OPTS="-Djava.security.krb5.conf=$MESOS_SANDBOX/{{TASKCFG_ALL_HDFS_VERSION}}/etc/hadoop/krb5.conf $HADOOP_OPTS" &&
           {{/SECURITY_KERBEROS_ENABLED}}
-          ./bootstrap --resolve-hosts $TASK_NAME.$FRAMEWORK_HOST && ./{{TASKCFG_ALL_HDFS_VERSION}}/bin/hdfs zkfc
+          ./bootstrap --resolve-hosts $TASK_NAME.$FRAMEWORK_HOST &&
```
elezar: `--resolve-hosts` accepts a comma-separated list of hosts, so the name node can be added to the list instead of requiring an additional call to `bootstrap`.

NimaVaziri: Ah yes, it seems so

frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -293,7 +293,8 @@ pods:
           {{#SECURITY_KERBEROS_ENABLED}}
           export HADOOP_OPTS="-Djava.security.krb5.conf=$MESOS_SANDBOX/{{TASKCFG_ALL_HDFS_VERSION}}/etc/hadoop/krb5.conf $HADOOP_OPTS" &&
           {{/SECURITY_KERBEROS_ENABLED}}
-          ./bootstrap --resolve-hosts $TASK_NAME.$FRAMEWORK_HOST && ./{{TASKCFG_ALL_HDFS_VERSION}}/bin/hdfs zkfc
+          ./bootstrap --resolve-hosts $TASK_NAME.$FRAMEWORK_HOST name-$POD_INSTANCE_INDEX-node.$FRAMEWORK_HOST &&
```
benclarkwood: @NimaVaziri, this is missing a comma. This works because --resolve-hosts apparently just tries to resolve the first parameter, and bootstrap doesn't error on extra parameters (cc @nickbp)

PR #2007, opened 2017-12-29
[DCOS-19854] Improve uninstall

gabrielhartmann:
The Uninstall Scheduler had some finicky callback ordering for SchedulerDrivers / Plan construction. 
I could not debug why DCOS-19854 failed from logs.  So I've improved logging of the uninstall plan and removed the callbacks to improve reliability.


nickbp:
needs rebase

PR #2008, opened 2017-12-29
[INFINITY-2895] Switch to using .aws/credentials instead of environment variables

elezar:
This PR removes support for using envvars for AWS and requires that the `~/.aws/credentials` file exists.

@gabrielhartmann @benclarkwood this does not affect any of the tests themselves, but only running locally.


gabrielhartmann:
Can you document an example please?

PR #2010, opened 2018-01-03
[INFINITY-2876] Fix backup/restore metronome job definition

gabrielhartmann:
Metronome unintentionally made the `forcePullImage` field required

### DIFF
--- a/frameworks/cassandra/tests/config.py
+++ b/frameworks/cassandra/tests/config.py
@@ -70,7 +70,7 @@ def _get_test_job(name, commands, node_address, node_port, restart_policy='ON_FA
         'id': 'test.cassandra.' + name,
         'run': {
             'cmd': ' && '.join(commands),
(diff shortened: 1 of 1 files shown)


frameworks/cassandra/tests/config.py
```diff
@@ -70,7 +70,7 @@ def _get_test_job(name, commands, node_address, node_port, restart_policy='ON_FA
         'id': 'test.cassandra.' + name,
         'run': {
             'cmd': ' && '.join(commands),
-            'docker': {'image': 'cassandra:3.0.13'},
+            'docker': {'image': 'cassandra:3.0.13', 'forcePullImage': False},
```
NimaVaziri: Don't unspecified boolean values default to `false` here?

gabrielhartmann: no

PR #2020, opened 2018-01-03
INFINITY-[1532,1533] First pass at releasing testing/+tools/ .whl's

nickbp:
Adds tooling which builds a `.whl` file for each of `testing/` and `tools/`, and updates the release script to upload these `.whl` files to S3.

The provided `tools/pip/build.sh` includes example steps for both executing tools and for importing testing libs. In theory external repos can just point their builds to these `.whl`s and avoid the need to manually copy `testing/` and `tools/`, but for now this is really just setting up a source of truth for use in future tooling.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -99,6 +99,7 @@ fi
 sdk/bootstrap/build.sh # produces sdk/bootstrap/bootstrap.zip
 sdk/cli/build.sh # produces sdk/cli/[dcos-service-cli-linux, dcos-service-cli-darwin, dcos-service-cli.exe]
 ./gradlew :executor:distZip # produces sdk/executor/build/distributions/executor.zip
+tools/pip/build.sh $SDK_VERSION # produces tools/pip/[testing|tools]-[SDK_VERSION]-py3-none-any.whl
 
 # Upload artifacts to S3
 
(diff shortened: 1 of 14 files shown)


tools/pip/setup.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
+
```
benclarkwood: Are these terrifying looking python files just copied boilerplate from somewhere else?

nickbp: This is partially copied from the prior CLI python wrapper (used for 1.7 compat), which was also a `.whl`.

release.sh
```diff
@@ -112,8 +114,17 @@ fi
...
+
 aws s3 cp --acl public-read sdk/bootstrap/bootstrap.zip $S3_DIR/bootstrap.zip 1>&2
+
 aws s3 cp --acl public-read sdk/executor/build/distributions/executor.zip $S3_DIR/executor.zip 1>&2
-aws s3 cp --acl public-read sdk/cli/dcos-service-cli-linux $S3_DIR/dcos-service-cli-linux 1>&2
-aws s3 cp --acl public-read sdk/cli/dcos-service-cli-darwin $S3_DIR/dcos-service-cli-darwin 1>&2
-aws s3 cp --acl public-read sdk/cli/dcos-service-cli.exe $S3_DIR/dcos-service-cli.exe 1>&2
+
+for cli_exe in $(ls sdk/cli/dcos-service-cli*); do
+    aws s3 cp --acl public-read $cli_exe $S3_DIR/$(basename $cli_exe) 1>&2
```
elezar: Does `aws s3 cp` not support copying multiple files? An alternative may be `aws s3 sync` which does.

release.sh
```diff
@@ -112,8 +114,17 @@ fi
...
+done
+
+# note: pip is picky about version formatting, so the outputted filenames may not exactly match $SDK_VERSION due to some mangling.
+# additionally, pip will refuse to handle files without the extra info in the filename, so we leave it in.
+for whl_file in $(ls tools/pip/*.whl); do
```
elezar: Same comment here regarding `aws s3 cp` vs `aws s3 sync`.

sdk/cli/README.md
```diff
@@ -0,0 +1,10 @@
+# Default CLI
+
+This directory contains the default CLI build, used by the majority of services who don't add custom commands to their service CLIs. The default CLI builds are uploaded with every SDK release as of early 2018.
```
elezar: Could we reference a SHA here?

release.sh
nickbp: Done

sdk/cli/README.md
nickbp: Updated the package tooling to support template params like `{{sha256:filename@http://.../SHA256SUMS}}`. Now packages consuming default CLIs will be able to just specify the manifest URL for the SDK release they're using and the sha256sum will be fetched from there automatically.

PR #1996, opened 2017-12-22
Revert "INFINITY-2851 Update upgradesFrom if the package name is chan…

gabrielhartmann:
…ging (#1969)"

This reverts commit 27868ab6d453ad8452fe3536700fa39afea7267e.

PR #2021, opened 2018-01-04
INFINITY-2851 Update upgradesFrom if the package name changes

nickbp:
For example, if `myservice` is being released as `myservice-beta`, the `upgradesFrom`/`upgradesTo` values should default to a preceding `myservice-beta` release.

(reattempt of #1969 which was reverted in #1996 due to a bug)

### DIFF
--- a/tools/release_builder.py
+++ b/tools/release_builder.py
@@ -14,9 +14,9 @@ import re
 import shutil
 import sys
 import tempfile
+import universe
 import urllib.request
 import zipfile
-
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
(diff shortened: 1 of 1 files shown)


tools/release_builder.py
```diff
@@ -480,6 +480,18 @@ def _update_package_json(self, pkgdir):
         package_json['name'] = self._pkg_name
         # Update package's version to reflect the user's input
         package_json['version'] = self._pkg_version
+        # Update package's upgradesFrom/downgradesTo to reflect any package name changes
+        # due to enabling or disabling a beta bit.
+        if self._stub_universe_pkg_name != self._pkg_name:
```
elezar: One comment on this. This effectively means that we are setting an upgrade path that has not been tested.

PR #2023, opened 2018-01-04
Simplify the TaskKiller

gabrielhartmann:


sdk/scheduler/src/test/java/com/mesosphere/sdk/api/PodResourceTest.java
```diff
@@ -405,10 +406,10 @@ public void testRestartPodManyRunning() {
         assertEquals("test-0-c", json.getJSONArray("tasks").get(2));
         assertEquals("test-0-d", json.getJSONArray("tasks").get(3));
 
-        verify(mockTaskKiller).killTask(POD_0_TASK_A.getTaskId(), RecoveryType.TRANSIENT);
-        verify(mockTaskKiller).killTask(POD_0_TASK_B.getTaskId(), RecoveryType.TRANSIENT);
-        verify(mockTaskKiller).killTask(POD_0_TASK_C.getTaskId(), RecoveryType.TRANSIENT);
-        verify(mockTaskKiller).killTask(POD_0_TASK_D.getTaskId(), RecoveryType.TRANSIENT);
+        verify(mockTaskKiller).killTask(POD_0_TASK_A.getTaskId());
```
nickbp: These could be updated to `verify()` that the failure listener was invoked as expected

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -3,32 +3,30 @@
...
- * This interface should be implemented to allow components to request the killing of Mesos Tasks.  This is a normal
- * part of restarting a Task, which is a normal part of updating the Configuration of a Task.  This is also useful for
- * allowing end-users to mitigate problems with Tasks when they manually determine that a Task should be restarted or
- * permanently replaced.
+ * This class is a default implementation of the TaskKiller interface.
  */
-public interface TaskKiller {
+public class TaskKiller {
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private SchedulerDriver driver;
```
nickbp: Driver could be final

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PodResource.java
```diff
@@ -390,4 +388,8 @@ private static JSONObject getPodInstanceStatusJson(
         }
         return Optional.of(stateString);
     }
+
+    public void setTaskKiller(TaskKiller taskKiller) {
```
nickbp: May as well keep the Javadoc from the prior version of this function.

sdk/scheduler/src/main/java/com/mesosphere/sdk/api/PodResource.java
```diff
@@ -299,14 +288,23 @@ private Response restartPod(String podInstanceName, RecoveryType recoveryType) {
             return ResponseUtils.elementNotFoundResponse();
         }
```
nickbp: Is there any problem with invoking the listener even if taskKiller is null? Also probably can't hurt to move the log call above this

gabrielhartmann: There probably isn't a problem as they'll be detected as needing recovery later and killed anyway.  But for consistency / ease of understanding, I think you're right.  We'll do both or neither.

sdk/scheduler/src/test/java/com/mesosphere/sdk/api/PodResourceTest.java
gabrielhartmann: Done

PR #2024, opened 2018-01-04
Task cleaner

gabrielhartmann:
It's possible that Tasks entering non-terminal but "bad" states like `TASK_LOST` or `TASK_UNREACHABLE` could come back.  Unexpected tasks should be killed and their `TaskStatus` messages ignored.
  
The simplification of the TaskKiller is accomplished in a separate PR on which this PR is based: #2023.

Only the final commit entitled `Add TaskCleaner` should be reviewed.


gabrielhartmann:
The simplification of the TaskKiller is accomplished in a separate PR on which this PR is based: https://github.com/mesosphere/dcos-commons/pull/2023

PR #2029, opened 2018-01-04
[INFINITY-2909] Verify self-resolution by default

benclarkwood:
Fix golint
Refactor resolution into components
Leverage refactor to verify task DNS => task IP
Add new flag "--self-resolve" to toggle new behavior of verified self resolution


sdk/bootstrap/main.go
```diff
@@ -80,29 +85,40 @@ func parseArgs() args {
...
 	}
 
 	return args
 }
 
+func parseTaskHost() (string, error) {
+	taskName, taskNameOk := os.LookupEnv("TASK_NAME")
+	frameworkHost, frameworkHostOk := os.LookupEnv("FRAMEWORK_HOST")
+	if !taskNameOk || !frameworkHostOk {
+		return "", fmt.Errorf("Cannot determine TASK_NAME or FRAMEWORK_HOST")
```
nickbp: Could have this return the message currently at L99/L219, and then have L99/L219 log whatever's in `err`

sdk/bootstrap/main.go
```diff
@@ -356,20 +436,20 @@ func getContainerIPAddress() (string, error) {
...
 	if err != nil {
-		log.Fatalf("Failed to SET new LIBPROCESS_IP: ", err)
+		log.Fatalf("Failed to SET new LIBPROCESS_IP: %s", err)
 	}
 
-	if args.getTaskIp {
-		log.Printf("Printing new task IP: %s", pod_ip)
-		fmt.Printf("%s", pod_ip)
+	if args.getTaskIP {
+		log.Printf("Printing new task IP: %s", podIP)
```
nickbp: This seems redundant, remove?

sdk/bootstrap/main.go
```diff
@@ -393,6 +477,6 @@ func main() {
 		installDCOSCertIntoJRE()
 	}
 
-	log.Printf("Local IP --> %s", pod_ip)
+	log.Printf("Local IP --> %s", podIP)
```
nickbp: Remove IMO, looks ugly in practice

sdk/bootstrap/main.go
```diff
@@ -132,47 +197,62 @@ func waitForResolve(resolveHosts []string, resolveTimeout time.Duration) {
...
+
+func verifySelfResolution(podIP string, resolveTimeout time.Duration) {
+	taskHost, err := parseTaskHost()
+	if err != nil {
+		printEnv()
+		log.Printf("Missing required envvars to build task DNS address. " +
+			"Ensure that TASK_NAME and FRAMEWORK_HOST are both set or " +
+			"disable self resolution with --self-resolve=false")
+	}
+	log.Printf("Verifying that %s resolves to %s", taskHost, podIP)
```
nickbp: `Verifying that %s resolves` => `Waiting for %s to resolve`?

nickbp:
Fails kafka integration due to `sdk_hosts.resolve_hosts()` attempting to run bootstrap via `task exec`:

```
2018/01/04 20:44:31 Missing required envvars to build task DNS address. Ensure that TASK_NAME and FRAMEWORK_HOST are both set or disable self resolution with --self-resolve=false
2018/01/04 20:44:31 Verifying that  resolves to 10.0.3.60
2018/01/04 20:44:31 Waiting for '' to resolve...
2018/01/04 20:49:31 Time ran out while resolving ''. Customize timeout with -resolve-timeout, or use -verbose to see attempts.
```

sdk/bootstrap/main.go
```diff
@@ -132,47 +197,62 @@ func waitForResolve(resolveHosts []string, resolveTimeout time.Duration) {
...
+	if !timer.Stop() {
+		<-timer.C
+	}
+}
+
+func verifySelfResolution(podIP string, resolveTimeout time.Duration) {
+	taskHost, err := parseTaskHost()
+	if err != nil {
+		printEnv()
+		log.Printf("Missing required envvars to build task DNS address. " +
```
elezar: Is the intention to continue execution here? Should we exit or break?

sdk/bootstrap/main.go
```diff
@@ -132,47 +197,62 @@ func waitForResolve(resolveHosts []string, resolveTimeout time.Duration) {
...
-				}
+	var timer *time.Timer
+	if resolveTimeout == 0 {
+		timer = nil
+	} else {
+		timer = time.NewTimer(resolveTimeout)
+	}
+
+	for {
+		resolvedIPs := resolveHost(taskHost, timer)
```
elezar: Why is this treated as such as special case? Is this not the same as calling `waitForResolve()` with a list containing the `taskHost` (with the modification that we need add return values from `waitForResolve()`?

sdk/bootstrap/main.go
benclarkwood: The message is slightly different:

@99: "set resolve hosts"

@219: "Time ran out waiting to resolve X to Y"

sdk/bootstrap/main.go
benclarkwood: Whoops. That should be a Fatalf.

sdk/bootstrap/main.go
benclarkwood: I want the custom log lines around resolving to something specific.

benclarkwood:
@nickbp @elezar A re-review when you have time please.

PR #2037, opened 2018-01-05
INFINITY-2917 CI: Include timestamp in fetched log filenames

nickbp:
Puts the timestamp at the start of the file so that files are sorted by time.

We only have access to month/day/hour/min, so here's what it'll look like:
`0104_1709.hello-world.0fe39302-f18b-11e7-a6f9-ae11b3b25138.stdout`
(timestamp: `mmdd_HHMM`)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -7,6 +7,7 @@ E.G. py.test frameworks/<your-frameworks>/tests
 import logging
 import os
 import os.path
+import re
 import shutil
 import subprocess
 import sys
@@ -40,6 +41,20 @@ for noise_source in [
     logging.getLogger(noise_source).setLevel('WARNING')
 
 log = logging.getLogger(__name__)
+
+# Regex pattern which parses the output of "dcos task log ls --long", in order to extract the filename and timestamp.
+# Example inputs:
+#   drwxr-xr-x  6  nobody  nobody      4096  Jul 21 22:07                             jre1.8.0_144
+#   drwxr-xr-x  3  nobody  nobody      4096  Jun 28 12:50                          libmesos-bundle
+#   -rw-r--r--  1  nobody  nobody  32539549  Jan 04 16:31  libmesos-bundle-1.10-1.4-63e0814.tar.gz
+# Example output:
+#   match.group(1): "4096  ", match.group(2): "Jul 21 22:07", match.group(3): "jre1.8.0_144"
+# Notes:
+# - Should also support spaces in filenames.
+# - Doesn't make any assumptions about the contents of the tokens before the timestamp/filename,
+#   just assumes that there are 5 of them.
+#                             TOKENS        MONTH     DAY    HH:MM    FILENAME
+task_ls_pattern = re.compile('^([^ ]+ +){5}([a-zA-z]+ [0-9]+ [0-9:]+) +(.*)$')
 
 # An arbitrary limit on the number of tasks that we fetch logs from following a failed test:
 #     100 (task id limit)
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -192,12 +193,25 @@ def setup_artifact_path(item: pytest.Item, artifact_name: str):
...
-        return set(subprocess.check_output(['dcos', 'task', 'ls', task_id, '--all']).decode().split())
+        # Example output:
+        # drwxr-xr-x  6  nobody  nobody      4096  Jul 21 22:07                             jre1.8.0_144
+        # drwxr-xr-x  3  nobody  nobody      4096  Jun 28 12:50                          libmesos-bundle
+        # -rw-r--r--  1  nobody  nobody  32539549  Jan 04 16:31  libmesos-bundle-1.10-1.4-63e0814.tar.gz
+        ls_lines = subprocess.check_output(['dcos', 'task', 'ls', task_id, '--long', '--all']).decode().split('\n')
+        ret = {}
+        for line in ls_lines:
+            # example elems: ['drwxr-xr-x', '6', 'nobody', 'nobody', '4096', 'Jul 21 22:07', 'jre1.8.0_144']
+            elems = [s.strip() for s in filter(None, s.split('  '))]
```
elezar: Was this meant to be `line.split('  ')`?

conftest.py
```diff
@@ -192,12 +193,25 @@ def setup_artifact_path(item: pytest.Item, artifact_name: str):
...
 
-def get_task_files_for_id(task_id: str) -> set:
+def get_task_files_for_id(task_id: str) -> dict:
     try:
-        return set(subprocess.check_output(['dcos', 'task', 'ls', task_id, '--all']).decode().split())
+        # Example output:
+        # drwxr-xr-x  6  nobody  nobody      4096  Jul 21 22:07                             jre1.8.0_144
+        # drwxr-xr-x  3  nobody  nobody      4096  Jun 28 12:50                          libmesos-bundle
+        # -rw-r--r--  1  nobody  nobody  32539549  Jan 04 16:31  libmesos-bundle-1.10-1.4-63e0814.tar.gz
+        ls_lines = subprocess.check_output(['dcos', 'task', 'ls', task_id, '--long', '--all']).decode().split('\n')
```
elezar: Looking at the `ls --help` output, it may be useful to provide a timestamp format using:
```
   --time-style=STYLE     with -l, show times using style STYLE:
                               full-iso, long-iso, iso, locale, or +FORMAT;
                               FORMAT is interpreted like in 'date'; if FORMAT
                               is FORMAT1<newline>FORMAT2, then FORMAT1 applies
                               to non-recent files and FORMAT2 to recent files;
                               if STYLE is prefixed with 'posix-', STYLE
                               takes effect only outside the POSIX locale
```
Thus: `--time-style=+%m%d_%H%M` would get the timestamp in the desired format.

elezar: (OK, just realized that this isn't calling ls ...)

conftest.py
```diff
@@ -192,12 +193,25 @@ def setup_artifact_path(item: pytest.Item, artifact_name: str):
...
+        ret = {}
+        for line in ls_lines:
+            # example elems: ['drwxr-xr-x', '6', 'nobody', 'nobody', '4096', 'Jul 21 22:07', 'jre1.8.0_144']
+            elems = [s.strip() for s in filter(None, s.split('  '))]
+            # get timestamp: 'Jul 21 22:07' => '0721_2207'
+            timeobj = time.strptime(elems[-2], '%b %d %H:%M')
+            timestamp = '{:02d}{:02d}_{:02d}{:02d}'.format(timeobj.tm_mon, timeobj.tm_mday, timeobj.tm_hour, timeobj.tm_min)
```
elezar: How about:
```python
timestamp = time.strftime('%d%m_%H%M', timeobj)
```
instead?

conftest.py
nickbp: It's specifically `'  '` in order to split the elements correctly. Single space would result in the timestamp being fragmented, or in problems with e.g. filenames containing single spaces (hopefully it's unlikely that filenames will contain two spaces).

conftest.py
nickbp: Aha yes that's much better. Done.

conftest.py
elezar: No, I mean the variable `s` doesn't exist so `s.split('  ')` is invalid?

nickbp: Ohh, whoops. Fixed now. Just in case I manually ran the handler again and it's fixed now.

nickbp: Haha, looks like github is eating double spaces: `foo  bar`

conftest.py
```diff
@@ -192,12 +193,24 @@ def setup_artifact_path(item: pytest.Item, artifact_name: str):
...
-        return set(subprocess.check_output(['dcos', 'task', 'ls', task_id, '--all']).decode().split())
+        # Example output:
+        # drwxr-xr-x  6  nobody  nobody      4096  Jul 21 22:07                             jre1.8.0_144
+        # drwxr-xr-x  3  nobody  nobody      4096  Jun 28 12:50                          libmesos-bundle
+        # -rw-r--r--  1  nobody  nobody  32539549  Jan 04 16:31  libmesos-bundle-1.10-1.4-63e0814.tar.gz
+        ls_lines = subprocess.check_output(['dcos', 'task', 'ls', task_id, '--long', '--all']).decode().split('\n')
+        ret = {}
+        for line in ls_lines:
+            # example elems: ['drwxr-xr-x', '6', 'nobody', 'nobody', '4096', 'Jul 21 22:07', 'jre1.8.0_144']
+            elems = [s.strip() for s in filter(None, line.split('  '))]
```
elezar: Does the filter not mean that the length of `elems` could be variable?

nickbp: In theory this is just to remove some empty entries produced by the `split()`
Actually, it does look like there's some problems with this current handling, note the preceding space on the resulting filename:

```
>>> s = 'drwxr-xr-x  6  nobody  nobody      4096  Jul 21 22:07                             jre1.8.0_144'
>>> s.split('  ')
['drwxr-xr-x', '6', 'nobody', 'nobody', '', '', '4096', 'Jul 21 22:07', '', '', '', '', '', '', '', '', '', '', '', '', '', ' jre1.8.0_144']
>>> [e for e in filter(None, s.split('  '))]
['drwxr-xr-x', '6', 'nobody', 'nobody', '4096', 'Jul 21 22:07', ' jre1.8.0_144']
```

Going to figure out how to do this properly real quick..

nickbp: Ok, replaced the whole thing with a regex. Should be much less brittle now.

PR #2038, opened 2018-01-05
Handle `S3_BUCKET` being an empty string by default now.

mpereira:
This was introduced with
https://github.com/mesosphere/dcos-commons/commit/31224ec8dcddc1e5f4f090acfc38f8c6f7e275e5

`dict.get` does not return the default if the value is an empty string so
we need to check for that.

### DIFF
--- a/tools/publish_aws.py
+++ b/tools/publish_aws.py
@@ -42,7 +42,10 @@ class AWSPublisher(object):
         self._universe_url_prefix = os.environ.get(
             'UNIVERSE_URL_PREFIX',
             'https://universe-converter.mesosphere.com/transform?url=')
-        s3_bucket = os.environ.get('S3_BUCKET') if os.environ.get('S3_BUCKET') else 'infinity-artifacts'
+        s3_bucket = os.environ.get('S3_BUCKET')
+        if not s3_bucket:
+            s3_bucket = 'infinity-artifacts'
+        logger.info('Using artifact bucket: {}'.format(s3_bucket))
         s3_dir_path = os.environ.get('S3_DIR_PATH', 'autodelete7d')
         dir_name = '{}-{}'.format(
             time.strftime("%Y%m%d-%H%M%S"),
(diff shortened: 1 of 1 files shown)


mpereira:
Something like this should appear in the output whenever `publish_aws.py` is called without this change (or without passing an `S3_BUCKET` environment variable). 
```
Invalid bucket name "": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$"
  File "/build/tools/publish_aws.py", line 191, in <module>
    sys.exit(main(sys.argv))
  File "/build/tools/publish_aws.py", line 186, in main
    AWSPublisher(package_name, package_dir_path, artifact_paths).upload()
  File "/build/tools/publish_aws.py", line 135, in upload
    content_type='application/vnd.dcos.universe.repo+json;charset=utf-8')
  File "/build/tools/publish_aws.py", line 102, in _upload_artifact
    raise Exception('Failed to upload {} to S3'.format(filename))
Exception: Failed to upload stub-universe-beta-elastic.json to S3
```

tools/publish_aws.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         self._universe_url_prefix = os.environ.get(
             'UNIVERSE_URL_PREFIX',
             'https://universe-converter.mesosphere.com/transform?url=')
-        s3_bucket = os.environ.get('S3_BUCKET', 'infinity-artifacts')
+        s3_bucket = os.environ.get('S3_BUCKET') if os.environ.get('S3_BUCKET') else 'infinity-artifacts'
```
elezar: I would almost consider doing this in two steps. The line gets a little long otherwise:
```python
s3_bucket = os.environ.get('S3_BUCKET')
if not s3_bucket:
    s3_bucket = 'infinity-artifacts'
log.info('Using artefact bucket: %s', s3_bucket)
```

mpereira: In general I try to avoid multiple assignments to the same variable but this is simple enough: https://github.com/mesosphere/dcos-commons/pull/2038/commits/ac0372f42bbde69c54f0969be151fb7e2633497b

PR #2039, opened 2018-01-05
Fix grep matching and subshell output expansion.

mpereira:
`grep [${aws_profile}]` was matching a bracket regular expression instead of a
string with literal brackets.

Also the expansion of `$( echo "$PROFILES" | grep "\[${aws_profile}\]" )`
needs to be enclosed in double quotes so that the != operator works and
doesn't give us "unary operator expected" errors when the output is empty.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -134,7 +134,7 @@ if [ ! -f "${aws_credentials_path}/credentials" ]; then
 else
     CREDENTIALS_FILE=${aws_credentials_path}/credentials
     PROFILES=$( grep -oE "^\[\S+\]" $CREDENTIALS_FILE )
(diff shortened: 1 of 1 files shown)


mpereira:
I was also unsure about this but it seems that the default matching for grep is "basic regex", which should be the same as the extended regex in GNU grep.
```
       -G, --basic-regexp
              Interpret  PATTERN  as  a  basic  regular  expression  (BRE,  see below).  This is the
              default.
```

nickbp:
Conflict in test.sh, possibly from #2041 just being merged

mpereira:
Resolved: https://github.com/mesosphere/dcos-commons/pull/2039/commits/15b162ca145aa953459e62c3f26cb5bd3c9a3e37

PR #2040, opened 2018-01-05
[INFINITY-2672] Add an SSL SASL kafka test

elezar:
This PR adds a basic SSL-enabled kerberos tests to Kafka. This requires a client with the following:

* A valid kerberos principal and keytab
* A valid TLS certificate

I would like to create a follow-up PR refactoring some of the security tests for Kafka as there are many shared steps (e.g. creation of the `client.properties` files and calling the producer and consumer).

@benclarkwood if you feel it would be better to include those steps in this PR, I could do it that way too.


testing/sdk_tasks.py
```diff
@@ -184,3 +184,39 @@ def task_exec(task_name: str, cmd: str, return_stderr_in_stdout: bool = False) -
...
+EOL\"""".format(output_file=filename, content="\n".join(lines))
+    log.info("Running: %s", output_cmd)
+    rc, stdout, stderr = task_exec(task_name, output_cmd)
+
+    if rc or stderr:
+        log.error("Error creating file %s. rc=%s stdout=%s stderr=%s", filename, rc, stdout, stderr)
+        return False
+
+    linecount_cmd = "wc -l {output_file}".format(output_file=filename)
+    rc, stdout, stderr = task_exec(task_name, linecount_cmd)
```
benclarkwood: Has this been changed under the covers to work in 1.9 -> 1.11+?

benclarkwood: N/A, nothing uses this that runs on 1.9 (but still task_exec should have a shim...)

elezar: Sorry. Was afk for a day. There are modifications to `task_exec` in #2059 that should make local executables in the sandbox work in 1.9, this would not be applicable here since `wc` is a system binary.

PR #2041, opened 2018-01-05
Require complete AWS credentials file path

### LINKED ISSUE #2027: Change test.sh --aws to accept a path to a file instead of directory
The `test.sh` script requires the path to the **directory** containing the credentials file instead of a path to the file itself.  It looks environment vars were removed as part of this PR:

https://github.com/mesosphere/dcos-commons/pull/2008

I have to workaround providing an `.aws/credentials` file right now because we use the Jenkins credentials plugin to provide a "secrets file" for AWS credentials, but it provides a path directly to a file, and I can't specify the name of the file.  I have to copy the secrets file to a local directory in my Jenkins workspace, name it properly, provide a path to that directory, and then I delete the file so it doesn't hang around after the build.  

It would be preferable for the `test.sh --aws` command to accept a direct link to this file, but I have a feeling it would be a breaking change for you.

/cc @benclarkwood @elezar

elezar:
This PR changes the expectation that a directory containing an AWS `credentials` file be provided to `test.sh` to a complete path to the file be provided.

This fixes #2027 reported by @seglo.

I have tested this locally, and `test.sh` still works as expected. @seglo confirms that this also works on Jenkins.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -19,7 +19,7 @@ pytest_m="sanity and not azure"
 pytest_k=""
 azure_args=""
 ssh_path="${HOME}/.ssh/ccm.pem"
-aws_credentials_path="${HOME}/.aws"
+aws_credentials_file="${HOME}/.aws/credentials"
 aws_profile="default"
 DCOS_ENTERPRISE=true
 interactive=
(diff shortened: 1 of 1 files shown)


seglo:
Thanks @elezar !

PR #2046, opened 2018-01-05
HDFS-493 Check for command success in secret store tests

nickbp:
Suspecting that the tests weren't logging when something failed during initialization. Add some gratuitous logging to validate this.


testing/sdk_auth.py
```diff
@@ -150,6 +150,8 @@ def get_download_stream(url: str) -> requests.Response:
         log.error("%s", e)
         raise RuntimeError("Failed to download the keytab file: {}".format(repr(e)))
 
+    log.info("Downloaded %d bytes from %s to %s", os.stat(output_filename).st_size, keytab_url, output_filename)
```
NimaVaziri: We should probably be consistent with string formatting (`"%s"` vs `"{}".format()`) but it's not critical.

nickbp: Yeah I just did it this way on account of the surrounding file doing it

PR #2050, opened 2018-01-05
Fix log fetching: Use svc_cmd, strip filenames

nickbp:


nickbp:
The HDFS flake conveniently shows that log fetching are fixed now. Gonna just merge this now as it's blocking other diagnosis

PR #2051, opened 2018-01-06
Add download of diagnostics bundle to failed test reports

mrbrowning:


benclarkwood:
Merging, as the elastic failure is a known unreliable test and I'd like this capturing to get in.

PR #2053, opened 2018-01-06
INFINITY-2922 Include libcrypto.so for cqlsh which imports Python's hashlib

nickbp:
~TODO: copy openssl-libs.tar.gz build onto prod bucket~


benclarkwood:
@nickbp have you tried this on a soak cluster?

benclarkwood:
Also: https://downloads.mesosphere.com/cassandra/assets/openssl-1.0.2n-libcrypto.tar.gz

nickbp:
Looks like the build also needed `libssl.so`. Gonna put up a new build in nick-dev and try another run with that. Looks like the linking works at least!

PR #2054, opened 2018-01-08
[INFINITY-2927] Add missing subprocess import to conftest.py

elezar:
This adds a missing import for `subprocess` to `conftest`.
 
This requirement was added in  #2051

Causing errors: https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=925773&_focus=1830

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,8 +9,10 @@ import os
 import os.path
 import re
 import shutil
+import subprocess
 import sys
 import time
+
 
 import pytest
 import requests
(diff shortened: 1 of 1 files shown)


elezar:
@nickbp following #2050, I would be happy to use one of the `sdk_cmd` functions here instead. Thoughts?

nickbp:
I think that'd make sense. I found it easy to get the subprocess call wrong, and the debug output provided by sdk_cmd is often useful. But if this is breaking things feel free to just merge and do a separate PR

elezar:
OK. Let me create a JIRA ticket for this.

PR #2059, opened 2018-01-08
[INFINITY-2929] fix kafka topic 1.9

elezar:
This PR tries to add a full path to commands with relative paths starting `./`. Note that this requires a 1.9 cluster to test.


benclarkwood:
Looks like the elastic failure here should be fixed by #2075

PR #2062, opened 2018-01-08
Include missing import

mellenburg:


nickbp:
fixed via #2054 as of just now

PR #2068, opened 2018-01-09
INFINITY-2577: Retry dcos package describe.

mpereira:
`dcos package` commands will output stacktraces in case of Cosmos failures. This
is easily reproducible by killing Cosmos and running dcos package commands.
dcos-cli will get non-JSON error responses from Admin Router which fail to be
parsed.

The bug reported in INFINITY-2577 seems to be recurring in our builds but from
what we can do from our side retrying should be sufficient in cases of temporary
unavailability from Cosmos.

This unreleased dcos-cli change could help us diagnose what exactly the issue is
in the builds: https://github.com/dcos/dcos-cli/commit/1617041db809d3bfe0ee1d0a454eb1e260a5f7d9#diff-27e3858ada053fa20bb44c8ab1a838a0


testing/sdk_upgrade.py
```diff
@@ -286,6 +286,8 @@ def _upgrade_or_downgrade(
         sdk_tasks.check_tasks_updated(service_name, '', task_ids)
 
 
+@retrying.retry(stop_max_attempt_number=5,
+                wait_fixed=5000)
```
nickbp: Maybe try a longer timeout?

PR #2071, opened 2018-01-09
Alter test.sh behavior around -i.

benclarkwood:
Revert to previous behavior of passing -i to docker run, but also introduce a new flag (--headless) to enable the new behavior.

This is a tweak of: #2013.

That unintentionally returned the fun docker behavior of normal, human runs of test.sh bleeding docker containers since the interrupt isn't actually sent to STDIN.


benclarkwood:
cc @seglo

PR #2075, opened 2018-01-10
INFINITY-2930 Resolve elastic flake: data node addition

nickbp:
Adding a 4th data node can sometimes fail due to bad resource packing on the cluster. To get around this, only add one data node in `test_sanity`.
Also cleans up some test logging and upgrade handling.

### DIFF
--- a/testing/sdk_upgrade.py
+++ b/testing/sdk_upgrade.py
@@ -217,7 +217,13 @@ def _get_pkg_version(package_name):
         log.warning('Failed to run "{}":\nSTDOUT:\n{}\nSTDERR:\n{}'.format(cmd, stdout, stderr))
         return None
     try:
-        return json.loads(stdout)['package']['version']
+        describe = json.loads(stdout)
+        # New location (either 1.10+ or 1.11+):
+        version = describe.get('package', {}).get('version', None)
+        if version is None:
+            # Old location (until 1.9 or until 1.10):
+            version = describe['version']
+        return version
     except:
         log.warning('Failed to extract package version from "{}":\nSTDOUT:\n{}\nSTDERR:\n{}'.format(cmd, stdout, stderr))
         log.warning(traceback.format_exc())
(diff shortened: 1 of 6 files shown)


benclarkwood:
I defer to @elezar and @mpereira.

nickbp:
Ping

testing/sdk_upgrade.py
```diff
@@ -201,50 +204,18 @@ def _upgrade_or_downgrade(
...
-    _wait_for_new_default_version(prev_version, default_repo_package_name)
-
-
-def _add_last_repo(repo_name, repo_url, prev_version, default_repo_package_name):
-    assert shakedown.add_package_repo(
-        repo_name,
-        repo_url)
-    # Make sure the new default repo packages are available
-    _wait_for_new_default_version(prev_version, default_repo_package_name)
+        sdk_cmd.run_cli('package describe {}'.format(package_name), print_output=False))
```
elezar: Since we're editing this anyway, could we split this into multiple steps? This may make handling errors a little easier. Also, is the output of `package describe` not already JSON from which the `"version"` field can be read directly without requiring regex?

testing/sdk_upgrade.py
```diff
@@ -201,50 +204,18 @@ def _upgrade_or_downgrade(
...
+    return None if result is None else result.group(1)
 
 
-def _wait_for_new_default_version(prev_version, default_repo_package_name):
-    shakedown.wait_for(lambda: _get_pkg_version(default_repo_package_name) != prev_version, noisy=True)
+def _wait_for_new_default_version(package_name, prev_version):
```
elezar: Is this a sufficient condition to check? Would it not make more sense to just check that the package is being served from the newly added universe URL?

nickbp: Providing useful information like that would be very un-Cosmos-like. We're resorting to the version changing as an indicator that Cosmos actually accepted the repo list change.

Cosmos' repo handling is "asynchronous", i.e. it doesn't produce any useful indication of success nor of failure.

testing/sdk_upgrade.py
nickbp: Restructured this a bit but it's still the same steps. Failure by Cosmos to handle the lookup is treated equivalent to Cosmos returning bad data.

benclarkwood:
Failed template due to cluster acquisition failure. Re-running.

PR #2077, opened 2018-01-10
Update krb5 conf file for Active Directory (HDFS)

benclarkwood:
On discussion with Nima, we are not sure if this is required.

We will need to retest HDFS with AD without the change and see if it can be made to work.


frameworks/hdfs/src/main/dist/krb5.conf
```diff
@@ -6,8 +6,9 @@
 [realms]
   {{SECURITY_KERBEROS_REALM}} = {
     kdc = {{SECURITY_KERBEROS_KDC_HOSTNAME}}:{{SECURITY_KERBEROS_KDC_PORT}}
+    admin_server = {{SECURITY_KERBEROS_KDC_HOSTNAME}}:{{SECURITY_KERBEROS_KDC_PORT}}
```
elezar: Are any of the other settings required here? https://web.mit.edu/kerberos/krb5-1.12/doc/admin/conf_files/krb5_conf.html#realms

frameworks/hdfs/src/main/dist/krb5.conf
```diff
@@ -6,8 +6,9 @@
...
   }
 
 [domain_realm]
-  .{{FRAMEWORK_HOST}} = LOCAL
-  {{FRAMEWORK_HOST}} = LOCAL
+  .{{FRAMEWORK_HOST}} = {{SECURITY_KERBEROS_REALM}}
```
elezar: Is there a ticket to track this for Kafka or ZK too? Do we want to add those as part of this PR?

elezar:
I am closing this as I have a working implementation in #2099 with an integration test.

PR #2079, opened 2018-01-10
HDFS-493 Disable broken HDFS tests

nickbp:
Until fix is found. Current guess is kdc producing bad keytabs.

### DIFF
--- a/frameworks/hdfs/tests/test_auth.py
+++ b/frameworks/hdfs/tests/test_auth.py
@@ -116,6 +116,7 @@ def kerberos(configure_security):
 @pytest.mark.dcos_min_version('1.10')
 @sdk_utils.dcos_ee_only
 @pytest.mark.smoke
+@pytest.mark.sanity
 def test_health_of_kerberized_hdfs():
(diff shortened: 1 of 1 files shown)


frameworks/hdfs/tests/test_auth.py
```diff
@@ -116,6 +116,7 @@ def kerberos(configure_security):
 @pytest.mark.dcos_min_version('1.10')
 @sdk_utils.dcos_ee_only
 @pytest.mark.smoke
+@pytest.mark.sanity
```
nickbp: Noticed this while in here as well. Given that non-helloworld smoke tests arent being run, this was effectively dead code.

benclarkwood:
@gabrielhartmann:

HDFS-493 is blocking. I just assigned it over to myself (from Nima). I will see if I can manually reproduce / diagnose from the existing corpus of data today or tomorrow. If I can't I'll turn the tests back on.

gabrielhartmann:
Ok, thanks @benclarkwood

PR #2081, opened 2018-01-11
INFINITY-2958 Switch bootstrap mustache library

nickbp:


benclarkwood:
Are we able to consume that library directly and get the behavior we want?

benclarkwood:
(I assume yes)

nickbp:
Tried setting the map values to `false` rather than `""` but the library renders that incorrectly. So not quite the correct behavior.
- The fix was very straightforward. Maybe they'll have a quick turnaround?: https://github.com/alexkappa/mustache/pull/3
- In the meantime I'll try giving this one a shot: https://github.com/cbroglie/mustache

Updated the unit test to cover the issue in any case.

nickbp:
OK the cbroglie variant works. This'll do.

sdk/bootstrap/main_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+import "io/ioutil"
+import "log"
+import "os"
+import "testing"
+import "github.com/stretchr/testify/assert"
+
+// Exercise rendering of truthy/falsy values for both regions and variables:
+var templateContent = `dc={{CASSANDRA_LOCATION_DATA_CENTER}}
+{{#PLACEMENT_REFERENCED_ZONE}}
```
elezar: In the tests, `PLACEMENT_REFERENCED_ZONE` and `FOO` get assigned the same value. Could one then not combine the two blocks here:
```
{{#FOO}}
one={{FOO}}
rack={{ZONE}}
{{/FOO}}
{{^FOO}}
two={{FOO}}
rack=rack1
{{/FOO}}
```

sdk/bootstrap/main.go
```diff
@@ -214,15 +212,20 @@ func openTemplate(inPath string, source string, templateMaxBytes int64) []byte {
...
 
 func renderTemplate(origContent string, outPath string, envMap map[string]string, source string) {
-	dirpath, _ := path.Split(outPath)
-	template, err := mustache.ParseStringPartialsRawInDir(origContent, dirpath, nil, false)
-	if err != nil {
-		log.Fatalf("Failed to parse template content from %s at '%s': %s", source, outPath, err)
+	// Env preprocessing: map "false", "False", etc to a false bool, so that it's treated as 'falsy'.
+	varMap := make(map[string]interface{})
+	for k, v := range envMap {
+		if strings.ToLower(v) == "false" {
```
gabrielhartmann: trim whitespace?  I'm of two minds.  Maybe if you put whitespace around your booleans you deserve what you get?

nickbp: Yeah I originally had `TrimSpace()` here and then felt it was sloppy. Who knows, maybe someone will want `false ` to behave non-falsy.

PR #2083, opened 2018-01-11
[INFINITY-2957] Pin bootstrap to 0.31.2

benclarkwood:


benclarkwood:
I think the right solution to this is more complex...

PR #2085, opened 2018-01-11
[KAF-392] Set protocol to 1.0

benclarkwood:


benclarkwood:
I think this will be blocked until we release a 1.0.0 GA

benclarkwood:
Note: According to the Kafka docs, switching the log format version should be a separate rolling restart. Created a ticket to track that.

PR #2089, opened 2018-01-11
INFINITY-2967 Enable cassandra agent shutdown test as the last test to run

nickbp:
This test is important enough to keep in PRs if possible.
It's a destructive test, but if it's the last test to be run then that doesn't matter.

PR #2090, opened 2018-01-12
[INFINITY-2967] Fix shutting down the right host for testing

gabrielhartmann:


nickbp:
Ported these changes into #2089 which was based on a more recent `master`

PR #2091, opened 2018-01-12
[HDFS-493] Retry keytab creation, checking for bad pattern

benclarkwood:
This is one of the stranger things I have encountered. In a nutshell the problem is: the keytab generated by the Kerberos automation contains (extra?) characters/format such that the Java Kerberos libraries cannot properly parse the keytab. The result of this is the error logline

```
Exception in thread "main" org.apache.hadoop.security.KerberosAuthException: Login failure for user: hdfs/journal-0-node.testintegrationhdfs.autoip.dcos.thisdcos.directory@LOCAL from keytab hdfs.keytab javax.security.auth.login.LoginException: Unable to obtain password from user
	at org.apache.hadoop.security.UserGroupInformation.loginUserFromKeytab(UserGroupInformation.java:1133)
	at org.apache.hadoop.security.SecurityUtil.login(SecurityUtil.java:275)
	at org.apache.hadoop.hdfs.qjournal.server.JournalNode.start(JournalNode.java:152)
	at org.apache.hadoop.hdfs.qjournal.server.JournalNode.run(JournalNode.java:134)
	at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:70)
	at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:84)
	at org.apache.hadoop.hdfs.qjournal.server.JournalNode.main(JournalNode.java:307)
```

This user error is because HDFS doesn't fully check the keytab for validity (well, and the Java lib eats an exception), and the keytab is in fact _not_ valid as far as the Java libraries are concerned. Specifically, the java libraries get confused / do a bad job parsing a realm towards the end of the file. You can tell the difference between a "good" and "bad" keytab using strings. A good keytab will look something like:

```
>data-0-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
LOCAL
HTTP
>data-1-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
LOCAL
HTTP
>data-1-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
LOCAL
HTTP
>data-1-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
LOCAL
HTTP
>data-2-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
{code}

A "bad" keytab will look something like:

{code:java}
>data-0-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
ZX@w
LOCAL
HTTP
>data-0-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
ZX@w
LOCAL
HTTP
>data-1-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
ZX@w
LOCAL
HTTP
>data-1-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
ZX@w
LOCAL
HTTP
>data-1-node.testintegrationhdfs.autoip.dcos.thisdcos.directory
ZX@w
LOCAL
HTTP
```

Not the gobbledygook, and especially the ZX. Here's where it gets wild. kinit and pals _are 100% okay with these keytabs. They do not give a hoot. I can kinit against the KDC using one of these keytabs and it works A-ok.

So:
It appears that a "bad" keytab will have ZX<char><char> whereas a good keytab will have ZX<char>, which means `strings <keytab> | grep ^ZX` != 0 is a sufficient check (strings will show only things that are 4 consecutive ascii chars by default)

I've put in a retry around keytab generation that uses just that check. I am hopeful that this is some weird timing thing around the KDC just barely being up, or time as a seed or the capricious whims of an eldritch horror, because:
  1. When I run the same command the automation runs manually, it always works, but this is also well after the first keytab generation
  2. The retry I add in this PR seems to make it work okay.


imaxxs:
This is a nasty issue. Thanks for finding and solving it.

benclarkwood:
Well, neat. First test run, a keytab that passes my check still fails to deploy and doesn't pass my Java checker.

Tomorrow, I will make compile a little java tool to test keytabs and put that into the test pipeline. Yeesh.

testing/sdk_auth.py
```diff
@@ -282,13 +287,37 @@ def create_remote_keytab(self, name: str, principals: list=[]) -> str:
...
+        result = subprocess.run(command,
+               shell=True,
+               stdout=subprocess.PIPE,
+               stderr=subprocess.PIPE)
+
+        # Grep exits 0 if there were matches
+        if result.returncode == 0:
+            log.info("There were some matches when checking the keytab for ^ZX: %s", result.stdout)
+            # Clean up the bad keytab.
+            sdk_tasks.task_exec(self.kdc_host_id)
```
elezar: This command doesn't seem to do anything? (I see the deletion happens in the creation step anyway). Also, wasn't `task_exec` moved to `sdk_cmd`?

benclarkwood: Whoops. I started to type a thing, but then didn't Thanks.

testing/sdk_auth.py
```diff
@@ -282,13 +287,35 @@ def create_remote_keytab(self, name: str, principals: list=[]) -> str:
...
+        command = "strings {} | grep ^ZX".format(output_filename)
+        result = subprocess.run(command,
+               shell=True,
+               stdout=subprocess.PIPE,
+               stderr=subprocess.PIPE)
+
+        # Grep exits 0 if there were matches
+        if result.returncode == 0:
+            log.info("There were some matches when checking the keytab for ^ZX: %s", result.stdout)
+            raise Exception("The keytab is bad :(. We're going to retry generating this keytab. What fun.")
```
nickbp: Could maybe relaunch the kdc via marathon and just re-init from scratch. I assume it doesn't have any persistent volumes...

benclarkwood: It does not, I didn't want to figure out how to unwind things that far, but I want to refactor this majorly anyways.

I'm at least going to incorporate my java checking tool instead of using strings.

gabrielhartmann:
@benclarkwood: Have you filed upstream bug(s)?  Some component is wrong.  Either the Java libraries are failing incorrectly or Kerberos components are accepting keytabs they should not.  My money is of course on the Java libraries being wrong.

Also is the keytab schema versioned?  Are we perhaps mismatching keytab schema and java library?

gabrielhartmann:
@benclarkwood: A quick glance through the [Hadoop 2.6.0 release notes](https://hadoop.apache.org/docs/r2.6.0/hadoop-project-dist/hadoop-common/releasenotes.html) yields many issues around Kerberos (large and small).  We should definitely work around this for our CI, but also at least report the bug upstream and maybe contribute a fix.

elezar:
@benclarkwood I was thinking about this while trying to fall asleep last night. If we add the principals without random passkeys, then it could be possible to get deterministic keytabs -- in which case the check should be more straightforward.

benclarkwood:
@elezar: Keytabs have time in them. They are not deterministic afaict.

@gabrielhartmann: It is at the very least a case of the Java libraries being more "picky" than the kerberos utilities. However, the Java library at fault is part of the JRE itself.

I'm going to merge this, as I _do_ believe it has made it better, and I'd like to see how much better that betterness is.

benclarkwood:
(I will check tomorrow AM, and re-disable the tests if it's still bad. @mpereira, @elezar feel free to disable them if they're blocking a PR for you.

PR #2094, opened 2018-01-12
Retry log downloads during test failure collection

mrbrowning:


conftest.py
```diff
@@ -252,7 +254,11 @@ def get_rotating_task_logs(task_id: str, task_file_timestamps: dict, task_file:
     for filename in rotated_filenames:
         if filename not in task_file_timestamps:
             return  # Reached a log index that doesn't exist, exit early
-        content = get_task_log_for_id(task_id, filename)
+        try:
+            content = get_task_log_for_id(task_id, filename)
+        except ConnectionError:
```
nickbp: Does `retrying` log the error at this point?

Could do e.g. `log.error(traceback.format_exc())` in the `except` (needs an `import traceback` if not already present)

mrbrowning: I removed that log statement on the basis that all it logs is already reported in the `log.error` on 263, but you're right that this would swallow the actual stderr output. I think I'll put that in the exception message and then log it on that same line below?

PR #2096, opened 2018-01-12
[INFINITY-2973] Cleanup Kerberos defaults and tests

benclarkwood:


frameworks/kafka/tests/test_kerberos_auth.py
```diff
@@ -80,6 +80,7 @@ def kafka_server(kerberos):
                         "hostname": kerberos.get_host(),
                         "port": int(kerberos.get_port())
                     },
+                    "realm": sdk_auth.REALM,
```
elezar: Using `kerberos.get_realm()` would be better here as this would allow us to better encapsulate these details.

frameworks/hdfs/universe/config.json
```diff
@@ -82,25 +82,21 @@
                   "properties": {
                     "hostname": {
                       "type": "string",
-                      "description": "The name or address of a host running a KDC for the realm.",
-                      "default": "kdc.marathon.autoip.dcos.thisdcos.directory"
+                      "description": "The name or address of a host running a KDC for the realm."
```
elezar: Should these (and the other properties where the defaults have been removed) now be `required`?

frameworks/kafka/tests/test_zookeeper_auth.py
```diff
@@ -110,8 +110,11 @@ def zookeeper_server(kerberos):
             "security": {
                 "kerberos": {
                     "enabled": True,
-                    "kdc_host_name": kerberos.get_host(),
-                    "kdc_host_port": int(kerberos.get_port()),
+                    "kdc": {
```
elezar: Has this just been passing because of the defaults?

elezar:
The use of `sdk_auth.REALM` is not a blocker for this going in. We can address that change later.

elezar:
OK. I've made changes to the docs as part of #2093.

frameworks/kafka/tests/test_zookeeper_auth.py
benclarkwood: Yes

frameworks/hdfs/universe/config.json
benclarkwood: Short answer: no

Long answer: yes... but only as required if enabled is true. Which the UI does not currently have a good way to display.

PR #2105, opened 2018-01-18
[INFINITY-2783] HDFS: Make kerberos debug a configurable option with default false

benclarkwood:
Tested by hand.

### DIFF
--- a/frameworks/hdfs/src/main/dist/svc.yml
+++ b/frameworks/hdfs/src/main/dist/svc.yml
@@ -57,6 +57,9 @@ pods:
           mv hdfs {{TASKCFG_ALL_HDFS_VERSION}}/bin/ ;
           {{#SECURITY_KERBEROS_ENABLED}}
           export HADOOP_OPTS="-Djava.security.krb5.conf=$MESOS_SANDBOX/{{TASKCFG_ALL_HDFS_VERSION}}/etc/hadoop/krb5.conf $HADOOP_OPTS" &&
+          {{#TASKCFG_ALL_SECURITY_KERBEROS_DEBUG}}
+          export HADOOP_OPTS="-Dsun.security.krb5.debug=true $HADOOP_OPTS" &&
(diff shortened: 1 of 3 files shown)


frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -57,6 +57,9 @@ pods:
           mv hdfs {{TASKCFG_ALL_HDFS_VERSION}}/bin/ ;
           {{#SECURITY_KERBEROS_ENABLED}}
           export HADOOP_OPTS="-Djava.security.krb5.conf=$MESOS_SANDBOX/{{TASKCFG_ALL_HDFS_VERSION}}/etc/hadoop/krb5.conf $HADOOP_OPTS" &&
+          {{#TASKCFG_ALL_SECURITY_KERBEROS_DEBUG}}
```
elezar: I wish there was a better way to plumb this through! This is pretty verbose.

We could consider *always* adding `-Dsun.security.krb5.debug={{TASKCFG_ALL_SECURITY_KERBEROS_DEBUG}}` instead of the conditional mustache?

frameworks/hdfs/universe/marathon.json.mustache
```diff
@@ -96,6 +96,9 @@
     "TASKCFG_ALL_SECURITY_KERBEROS_PRIMARY": "hdfs",
     "TASKCFG_ALL_SECURITY_KERBEROS_PRIMARY_HTTP": "HTTP",
     "TASKCFG_ALL_SECURITY_KERBEROS_REALM": "{{service.security.kerberos.realm}}",
+    {{#service.security.kerberos.debug}}
```
elezar: Note that this conditional section falls away if we decide to *always* supply the flag.

benclarkwood: I think doing it this way is more consistent with how we've done other things (e.g. let presence of the envvar indicate you should do something, rather then using the value of the envvar)

PR #2106, opened 2018-01-18
[INFINITY-2995] Fix transient to custom permanent recovery transition

gabrielhartmann:
1. A task has failed transiently
1. A service provides custom permanent recovery logic
1. A user initiates a permanent replacement of the transiently failed task
1. BOOM!

Here `BOOM!` means that the recovery plan contains phases for both the transient recovery and permanent recovery of the same pod.  The recovery manager doesn't like that and it's also the wrong outcome.

The user initiated permanent replacement should override the previously detected transient failure.  This PR makes that happen.

Note:
The new ServiceTest failed without the fix.  It contained both permanent and transient recovery phases for `hello-0` as expected.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/recovery/DefaultRecoveryPlanManager.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/recovery/DefaultRecoveryPlanManager.java
@@ -122,10 +122,10 @@ public class DefaultRecoveryPlanManager implements PlanManager {
         logger.info("Dirty assets for recovery plan consideration: {}", dirtyAssets);
 
         synchronized (planLock) {
-            Collection<PodInstanceRequirement> podInstanceRequirements = null;
+            Collection<PodInstanceRequirement> newRequirements = null;
 
             try {
-                podInstanceRequirements = getRecoveryRequirements(dirtyAssets);
+                newRequirements = getNewRecoveryRequirements(dirtyAssets);
             } catch (TaskException e) {
                 logger.error("Failed to generate steps.", e);
                 return;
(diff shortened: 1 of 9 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -122,25 +122,24 @@ protected void updatePlan(Collection<PodInstanceRequirement> dirtyAssets) {
...
             } catch (TaskException e) {
                 logger.error("Failed to generate steps.", e);
                 return;
             }
 
             List<PodInstanceRequirement> defaultRequirements = new ArrayList<>();
             List<Phase> phases = new ArrayList<>();
-            for (PodInstanceRequirement requirement : podInstanceRequirements) {
+            for (PodInstanceRequirement requirement : newRequirements) {
                 boolean overriden = false;
```
nickbp: `overridden`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -169,12 +168,31 @@ private Plan createPlan(List<PodInstanceRequirement> defaultRequirements, List<P
                 .collect(Collectors.toList());
     }
 
+    private boolean phaseOverriden(Phase phase, Collection<PodInstanceRequirement> newReqs) {
+        // If at least 1 new requirement conflicts with the phases's steps, the phase has been overridden
```
nickbp: `phase's`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -169,12 +168,31 @@ private Plan createPlan(List<PodInstanceRequirement> defaultRequirements, List<P
                 .collect(Collectors.toList());
     }
 
+    private boolean phaseOverriden(Phase phase, Collection<PodInstanceRequirement> newReqs) {
```
nickbp: could this be `static`?

sdk/testing/src/main/java/com/mesosphere/sdk/testing/ServiceTestRunner.java
```diff
@@ -241,6 +229,15 @@ public ServiceTestRunner setPodEnv(String podType, String... podEnvKeyVals) {
         return setPodEnv(podType, toMap(podEnvKeyVals));
     }
 
+    /**
+     * Allows the specification of custom recoveyr logic just as in
```
nickbp: `recovery`

sdk/testing/src/main/java/com/mesosphere/sdk/testing/Expect.java
```diff
@@ -303,6 +295,35 @@ public String getDescription() {
         };
     }
 
+    /**
+     * Verifies that the indicated recovery phase.step has the expected status.
+     */
+    public static Expect recoveryStepStatus(String phaseName, String stepName, Status expectedStatus) {
```
nickbp: Could have a generic `stepStatus(String planName, String phaseName, ...)`, and then have `recoveryStepStatus(...)` just be a wrapper for that.

Also, could use `DefaultRecoveryPlanManager.DEFAULT_RECOVERY_PLAN_NAME`

PR #2107, opened 2018-01-18
HDFS-495 Add logging to journal node readiness check

nickbp:


frameworks/hdfs/src/main/dist/svc.yml
```diff
@@ -93,17 +93,23 @@ pods:
         readiness-check:
           cmd: >
```
benclarkwood: You could always switch this to `|`, and then do a `set -e` at the top of the command, then you can skip a lot of &&-ing.

nickbp: Done here and throughout the other `cmd`s.

PR #2109, opened 2018-01-18
INFINITY-2990 Add shutdown test for hello-world

nickbp:
Upstreams shared logic from cassandra shutdown test into `sdk_tasks`.

Switches to using `dcos task` as source of truth, instead of extracting from self-reported taskinfos. Felt like a good idea.


frameworks/helloworld/tests/test_zzzrecovery.py
```diff
@@ -30,12 +38,11 @@ def test_kill_hello_node():
 
 
 @pytest.mark.sanity
-@pytest.mark.recovery
 def test_pod_restart():
     hello_ids = sdk_tasks.get_task_ids(config.SERVICE_NAME, 'hello-0')
 
     # get current agent id:
-    jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod info hello-0', json=True)
+    jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod info hello-0', json=True, print_output=False)
```
gabrielhartmann: Why don't we want to print output?  Helpful for debugging no?

frameworks/helloworld/tests/test_zzzrecovery.py
```diff
@@ -296,3 +269,59 @@ def test_config_update_then_zk_killed():
...
+    jsonobj = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod replace world-0', json=True)
+    assert len(jsonobj) == 2
+    assert jsonobj['pod'] == 'world-0'
+    assert len(jsonobj['tasks']) == 1
+    assert jsonobj['tasks'][0] == 'world-0-server'
+
+    sdk_tasks.check_tasks_updated(config.SERVICE_NAME, 'world-0', world_ids)
+    config.check_running()
+
+    # check agent moved (TODO: uncomment if/when agent is guaranteed to change (may randomly move back to old agent))
```
gabrielhartmann: I don't think replace will ever require movement to a new agent.  I'd just drop this code.

PR #2112, opened 2018-01-19
[INFINITY-2981] Fix setting up service accounts and secrets for ZK for Kafka tests

gabrielhartmann:


elezar:
Nice catch @gabrielhartmann! I will create a follow-up ticket to have a look at how our fixtures are used so that these steps are not needed in future -- especially as we implement more inter-service tests. Of course, the visibility makes it more explicit as to what is required when installing services.

PR #2113, opened 2018-01-19
Update airgap linter to accept FRAMEWORK_HOST

benclarkwood:


benclarkwood:
This has a bug in it after I refactored it. Whoops. Fixing.

PR #2114, opened 2018-01-19
INFINITY-3005: Simulate default executor instead of custom executor

nickbp:
Needed some changes to the simulation handling:
- Include additional overhead resources for the default executor to use.
- Use the ExecutorInfo from the parent LaunchGroup, rather than one of the TaskInfos.
- Hello world: Account for the added resources when checking the decommission plan

### DIFF
--- a/frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
+++ b/frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
@@ -19,6 +19,8 @@ import org.junit.After;
 import org.junit.Assert;
 import org.junit.Test;
 import org.mockito.Mockito;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.util.*;
@@ -39,7 +41,7 @@ public class ServiceTest {
      */
     @Test
     public void testDefaultDeployment() throws Exception {
-        runDefaultDeployment();
+        runDefaultDeployment(true);
     }
 
     /**
@@ -203,7 +205,7 @@ public class ServiceTest {
     @Test
     public void testHelloDecommissionNotAllowed() throws Exception {
         // Simulate an initial deployment with default of 2 world nodes (and 1 hello node):
-        ServiceTestResult result = runDefaultDeployment();
+        ServiceTestResult result = runDefaultDeployment(true);
         Assert.assertEquals(
                 new TreeSet<>(Arrays.asList("hello-0-server", "world-0-server", "world-1-server")),
                 result.getPersister().getChildren("/Tasks"));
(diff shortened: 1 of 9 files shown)


nickbp:
ping: green

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
```diff
@@ -270,7 +270,7 @@ public void testWorldDecommission() throws Exception {
 
         // Check initial plan state
         ticks.add(new ExpectDecommissionPlanProgress(Arrays.asList(
-                new StepCount("world-1", 6, 0, 0), new StepCount("world-0", 6, 0, 0))));
+                new StepCount("world-1", 9, 0, 0), new StepCount("world-0", 9, 0, 0))));
```
elezar: In the long-term it would be nice to not have to manually manage these step counts.

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
```diff
@@ -550,18 +550,22 @@ public void expect(ClusterState state, SchedulerDriver mockDriver) throws Assert
...
             Map<String, Status> expectedSteps = new HashMap<>();
             expectedSteps.put(String.format("kill-%s-server", stepCount.phaseName),
                     stepCount.statusOfStepIndex(expectedSteps.size()));
-            for (String resourceId :
-                ResourceUtils.getResourceIds(ResourceUtils.getAllResources(state.getLastLaunchedPod(stepCount.phaseName)))) {
+            LaunchedPod pod = state.getLastLaunchedPod(stepCount.phaseName);
+
+            Collection<String> resourceIds = new ArrayList<>();
+            resourceIds.addAll(ResourceUtils.getResourceIds(ResourceUtils.getAllResources(pod.getTasks())));
+            resourceIds.addAll(ResourceUtils.getResourceIds(pod.getExecutor().getResourcesList()));
```
elezar: Does this mean that if we write a test that forces a custom executor, all we need to do is ensure that `pod.getExecutor()` returns the correct resources?

sdk/testing/src/main/java/com/mesosphere/sdk/testing/Expect.java
```diff
@@ -77,26 +77,35 @@ public void expect(ClusterState state, SchedulerDriver mockDriver) {
...
                     } else if (operation.getType().equals(Protos.Offer.Operation.Type.LAUNCH_GROUP)) {
-                        // New-style launch with default executor
-                        launchedTaskNames.addAll(operation.getLaunch().getTaskInfosList().stream()
+                        // New-style launch with default executor in parent LaunchGroup
+                        launchedExecutor = operation.getLaunchGroup().getExecutor();
+
+                        Collection<Protos.TaskInfo> taskInfos =
+                                operation.getLaunchGroup().getTaskGroup().getTasksList();
+
+                        launchedTaskNames.addAll(taskInfos.stream()
```
elezar: Not critical, but I don't feel as if these two branches don't make the differences clear enough.

sdk/testing/src/main/java/com/mesosphere/sdk/testing/SendOffer.java
```diff
@@ -139,35 +148,55 @@ public String getDescription() {
                 }
             }
         }
+
+        // In addition to the resources required by the tasks, add some resources for the Default Executor overhead.
+        if (!podToReuse.isPresent()) {
+            offerBuilder.addAllResources(DEFAULT_EXECUTOR_RESOURCES);
```
elezar: How would this have to change if we wanted to test *both* default and custom executors?

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
nickbp: Switched to a variable that's assigned depending on whether default or custom executor is being exercised.

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
nickbp: As of the latest revision just now, to use a custom executor you'd call `ServiceTestRunner::setUseCustomExecutor()` before calling `run()`.

sdk/testing/src/main/java/com/mesosphere/sdk/testing/Expect.java
nickbp: Reworded the comments to be a bit clearer

sdk/testing/src/main/java/com/mesosphere/sdk/testing/SendOffer.java
nickbp: If custom executors are being exercised, the scheduler will effectively be getting slightly larger offers than it requires. It would ignore the additional cpu/mem/disk and launch the tasks anyway.

PR #2115, opened 2018-01-20
[INFINITY-2900] Update kafka zk library to enable re-resolution

benclarkwood:
This:
1. Creates a new integration test where ZK is installed on a virtual network
2. Restarts all 3 ZK nodes (to force all to have new IPs)
3. Checks that Kafka is not logging about a loss of connection.
4. Updates to the latest ZK library which re-resolves

I ran the test before updating the library and it does indeed fail.

Note: This points at a stub universe of ZK as a couple virtual network fixes were required over there to enable this test (specifically, making restarts reliable)


frameworks/kafka/tests/test_zookeeper.py
```diff
@@ -0,0 +1,113 @@
...
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+        zookeeper_framework_host = "{}.autoip.dcos.thisdcos.directory:1140".format(ZK_SERVICE_NAME)
+        config.install(
+            config.PACKAGE_NAME,
+            config.SERVICE_NAME,
+            config.DEFAULT_BROKER_COUNT,
+            additional_options={
+                "kafka": {
+                    "kafka_zookeeper_uri": "zookeeper-0-server.{host},zookeeper-0-server.{host},zookeeper-0-server.{host}".format(host=zookeeper_framework_host)
```
elezar: Does the `URI` not include the port?

frameworks/kafka/tests/test_zookeeper.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+@pytest.mark.sanity
+@pytest.mark.zookeeper
+def test_zookeeper_reresolution():
+
+    def restart_zookeeper_node(id: int):
+        sdk_cmd.svc_cli(ZK_PACKAGE, ZK_SERVICE_NAME, "pod restart zookeeper-{}".format(id))
+
+        sdk_plan.wait_for_in_progress_recovery(ZK_SERVICE_NAME)
```
elezar: Should one wait for a deployment as well?

frameworks/kafka/tests/test_zookeeper.py
```diff
@@ -0,0 +1,113 @@
...
+        sdk_plan.wait_for_completed_recovery(ZK_SERVICE_NAME)
+
+    # Restart each zookeeper node, so that each one receives a new IP address
+    # (it's on a virtual network). This will force Kafka to re-resolve ZK nodes.
+    for id in range(0, 3):
+        restart_zookeeper_node(id)
+
+    # Now, verify that Kafka remains happy
+    rc, stdout, stderr = sdk_cmd.run_raw_cli("task log kafka-0-broker --lines 15")
```
elezar: Are 15 lines enough? We could consider running a grep on the broker instead?

Also, should we check all brokers?

frameworks/kafka/tests/test_zookeeper.py
elezar: Scratch that. Just read the code again: :/

frameworks/kafka/tests/test_zookeeper.py
benclarkwood: I'd expect there to be at least _some_ connection blip, and in looking at the logs of a Kafka broker that is totally disconnected, 15 lines will always catch this error (it sits there repeating the same cycle)

benclarkwood: But, a broker that has successfully reconnected will not have this error in the tail of its logs.

frameworks/kafka/tests/test_zookeeper.py
benclarkwood: sdk_install does that, no?

benclarkwood: (I wait for a completed recovery)

frameworks/kafka/tests/test_zookeeper.py
elezar: Yes, but we're only checking a single broker. It could be that it happens to be happy all the time.

frameworks/kafka/tests/test_zookeeper.py
elezar: Sorry, I may have thought that `pod restart` also triggers a `deploy` ...

frameworks/kafka/tests/test_zookeeper.py
```diff
@@ -0,0 +1,113 @@
...
+ZK_SERVICE_NAME = "kafka-zookeeper"
+
+
+# NOTE: This can be removed after we publish a zookeeper that
+# has at least through sha a0d96b28769e4cb871b3e2424f4c6b889f5a06dd
+@pytest.fixture(scope='module', autouse=True)
+def install_zookeeper_stub():
+    zk_url = "https://universe-converter.mesosphere.com/transform?url=https://infinity-artifacts.s3.amazonaws.com/permanent/kafka-zookeeper/assets/sha-a0d96b28769e4cb871b3e2424f4c6b889f5a06dd/stub-universe-kafka-zookeeper.json"
+    try:
+        sdk_cmd.run_cli('package repo add --index=0 {} {}'.format("zk-a0d96b28769e4cb871b3e2424f4c6b889f5a06dd", zk_url))
```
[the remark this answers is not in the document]
benclarkwood: `sdk_repository.add_stub_universe_urls([zk_url, ])` doesn't appear to exist anymore

PR #2116, opened 2018-01-20
INFINITY-3007: Customized release indexing

nickbp:
- Release GAs at every 100 indexes by default, leaving space for hotfixes
- Release Betas sequentially (no index skipping)
- Support manual RELEASE_INDEX value for (GA) hotfixes

### DIFF
--- a/tools/universe/package_publisher.py
+++ b/tools/universe/package_publisher.py
@@ -11,25 +11,74 @@ import shutil
 log = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
+# By default, publish GA releases at 0, 100, 200, 300, ... to allow hotfixes within releases.
+ga_index_multiplier = 100
+# Don't bother with skipping indexes for beta releases, just go sequentially.
+beta_index_multiplier = 1
+
 class UniversePackagePublisher(object):
     '''Creates a PR for a release against the Universe repository at http://github.com/mesosphere/universe.
     '''
 
-    def __init__(self, package_name, package_version, commit_desc, dry_run=False):
+    def __init__(self, package_name, package_version, commit_desc, beta_release, dry_run=False):
         self._pkg_name = package_name
         self._pkg_version = package_version
         self._pr_title = 'Release {} {} (automated commit)\n\n'.format(self._pkg_name, self._pkg_version)
         self._commit_desc = commit_desc
+        self._beta_release = beta_release
         self._dry_run = dry_run
 
         # Optional configuration via envvars:
         self._release_branch = os.environ.get('RELEASE_BRANCH', 'version-3.x')
+        self._release_index = int(os.environ.get('RELEASE_INDEX', -1))
         self._github_user = os.environ.get('GITHUB_USER', 'mesosphere-ci')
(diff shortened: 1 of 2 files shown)


tools/universe/package_publisher.py
```diff
@@ -11,27 +11,76 @@
...
+        if self._release_index >= 0:
+            if os.path.exists(os.path.join(repo_pkg_base, str(self._release_index))):
+                raise Exception('Requested index {} is already occupied in directory {}: {}'.format(
+                    self._release_index, repo_pkg_base, sorted(os.listdir(repo_pkg_base))))
+            this_index = self._release_index
+
+            # Search backwards from this_index to find a prior release to diff against:
+            for num in reversed(range(0, this_index)): # reversed([0, ..., this-1]) => [this-1, ..., 0]
+                if os.path.isdir(os.path.join(repo_pkg_base, str(num))):
+                    last_index = num
```
elezar: Is the diff not applicable in all cases and not only when we specify the release index?

tools/universe/package_publisher.py
```diff
@@ -11,27 +11,76 @@
...
 
+    def _find_release_index(self, repo_pkg_base):
+        '''Returns the correct number/id for this release in the universe tree, and the prior release to diff against.
+
+        Returns a tuple containing two ints: [prior_index (or -1 if N/A), this_index]'''
+
+        # Find the index to be used for this release:
+        last_index = -1
+        if self._release_index >= 0:
+            if os.path.exists(os.path.join(repo_pkg_base, str(self._release_index))):
```
elezar: If we allow for indices to be specified explicitly, then even the automatic case could result in an overlap with an existing release. Should this check not be in the general path?

tools/universe/package_publisher.py
```diff
@@ -11,27 +11,76 @@
...
+            this_index = self._release_index
+
+            # Search backwards from this_index to find a prior release to diff against:
+            for num in reversed(range(0, this_index)): # reversed([0, ..., this-1]) => [this-1, ..., 0]
+                if os.path.isdir(os.path.join(repo_pkg_base, str(num))):
+                    last_index = num
+                    break
+        else:
+            # Iterate over the directory contents and find the largest integer
+            for filename in os.listdir(repo_pkg_base):
```
elezar: Is there a way to restrict this to directories?

tools/universe/package_publisher.py
nickbp: The automatic case is checking what's currently present in the repo to figure out the next number to use as 'latest'. Shouldn't have a conflict assuming things are added sequentially, i.e. unless someone has two parallel PRs out at the same time.

tools/universe/package_publisher.py
nickbp: The diff is performed between last_index and this_index, and both the automatic and manual cases set these two values.

In the automatic case, last_index is found first (largest number in the package), and then this_index is calculated against that.

tools/universe/package_publisher.py
nickbp: Added a directory check, mirroring the one a few lines up in the manual case. _Shouldn't_ be anything else here but can't hurt to be explicit..

PR #2117, opened 2018-01-22
[INFINITY-3003] Fix multiple different constructions of ExecutorInfo

gabrielhartmann:
1. Stop constructing ExecutorInfo twice
1. Make the ExecutorEvaluationStage care solely about the ExecutorId
1. Make the VolumeEvaluationStage take care of setting correct protobufs always
1. Make the ResourceEvaluationStages take care of setting correct protobufs always
1. Construct volume protobufs correctly (w/ DYNAMIC type)

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
@@ -75,18 +75,10 @@ public class ExecutorResourceMapper {
 
     private List<OfferEvaluationStage> getEvaluationStagesInternal() {
         List<ResourceSpec> remainingResourceSpecs = new ArrayList<>();
-
-        // When using a NEW executor add the appropriate evaluation stages.  If the executor ID is NOT empty then the
-        // executor is already running.  Therefore you can't expect cpu, mem or disk resources to be offered as the
-        // executor is already running and consuming those resources.
-        if (useDefaultExecutor && executorInfo.getExecutorId().getValue().isEmpty()) {
+        remainingResourceSpecs.addAll(volumeSpecs);
+        if (useDefaultExecutor) {
             remainingResourceSpecs.addAll(resourceSpecs);
         }
-
-        // Always add volumes for evaluation.  The VolumeEvaluationStage is smart enough to pass even when resources are
-        // not offered because the executor is already running.  It must be added always to make sure that the volume
-        // protobuf is constructed and added appropriately.
-        remainingResourceSpecs.addAll(volumeSpecs);
 
         List<ResourceLabels> matchingResources = new ArrayList<>();
         for (Protos.Resource resource : resources) {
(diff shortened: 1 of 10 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
```diff
@@ -102,6 +106,7 @@ public ExecutorResourceMapper(
                 }
                 matchingResources.add(matchingResource.get());
             } else {
+                LOGGER.warn("Failed to find match for resource: {}", resource);
```
nickbp: Could do a `TextFormat.shortDebugString()` here to make it a one-liner.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -212,7 +193,12 @@ public OfferEvaluator(
             evaluationPipeline.add(new ExecutorEvaluationStage(Optional.empty()));
             evaluationPipeline.addAll(getNewEvaluationPipeline(podInstanceRequirement, allTasks, tlsStageBuilder));
         } else {
-            evaluationPipeline.add(new ExecutorEvaluationStage(getExecutorInfo(thisPodTasks.values())));
+            Optional<Protos.ExecutorInfo> executorInfo = getExecutorInfo(thisPodTasks.values());
+            String executorIdString = executorInfo.get().getExecutorId().getValue();
```
nickbp: Is it actually set to an empty string sometimes? If just checking for presence can do `executorInfo.get().hasExecutorId()`. String fields which are unset will return `""` if you `getX()` against them anyway.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -223,12 +209,18 @@ public OfferEvaluator(
...
-                logger.info("Using existing executor: {}", taskInfo.getExecutor().getExecutorId().getValue());
+                logger.info("Using existing executor: {}", TextFormat.shortDebugString(taskInfo.getExecutor()));
                 return Optional.of(taskInfo.getExecutor());
             }
         }
 
-        return Optional.empty();
+        Protos.ExecutorInfo executorInfo = taskInfos.stream().findFirst().get()
+                .getExecutor().toBuilder()
+                .setExecutorId(Protos.ExecutorID.newBuilder().setValue(""))
```
nickbp: Oh I see this now. Should the field be cleared (`clearExecutorId()`) instead of set to an empty string?

gabrielhartmann: I wondered myself.  Was scared and went with empty string.  I can try it at least.

nickbp: Oh, it might throw at the `build()` call below if the field is `required`. Given that we may be stuck with an empty string.

https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L744

gabrielhartmann:
@nickbp: Adding some comments.  I tried clearing ExecutorID, but [it's a required field](https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L744), and we build the ExecutorInfo and pass it around.  Building it blows up if that require field isn't set.

gabrielhartmann:
Cassandra build is [actually here](https://teamcity.mesosphere.io/viewLog.html?buildId=944999&buildTypeId=DcosIo_DcosCommons_Cassandra_BuildAndTestPr&tab=buildResultsDiv), hopefully TC updates Github status eventually.

PR #2119, opened 2018-01-23
Remove allow_plaintext configuration option.

mpereira:


benclarkwood:
@mpereira there's nothing in the marathon.json that uses this option?

mpereira:
@benclarkwood nope

PR #2120, opened 2018-01-23
[INFINITY-3013] Increase HDFS tls test timeout to align with other tests

gabrielhartmann:


frameworks/hdfs/tests/test_tls.py
```diff
@@ -37,6 +37,7 @@ def hdfs_service_tls(service_account):
     try:
         sdk_install.install(
             config.PACKAGE_NAME,
+            timeout_seconds=30 * 60,
```
elezar: Could we put the timeout at the end as with the other tests?

Note that the first three parameters are not keyword arguments.

gabrielhartmann: I can move it if it makes you happy.  All arguments starting with this one are keyword arguments.

elezar: Not a strict requirement.

PR #2121, opened 2018-01-23
Edit tls test to use try/finally and correct fixtures to run on its own

benclarkwood:


frameworks/kafka/tests/test_tls.py
```diff
@@ -15,49 +15,52 @@
...
-        config.DEFAULT_BROKER_COUNT,
-        additional_options={
-            "service": {
-                "service_account": service_account,
-                "service_account_secret": service_account,
-                "security": {
-                    "transport_encryption": {
-                        "enabled": True
+    try:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
```
elezar: I think the HDFS TLS test is missing this too.

frameworks/kafka/tests/test_tls.py
```diff
@@ -15,49 +15,52 @@
...
+                        }
                     }
                 }
             }
-        }
-    )
+        )
 
-    sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
+        sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
```
elezar: Is this not part of the code in `install`?

frameworks/kafka/tests/test_tls.py
```diff
@@ -15,49 +15,52 @@
...
 
-    # Wait for service health check to pass
-    shakedown.service_healthy(config.SERVICE_NAME)
+        # Wait for service health check to pass
+        shakedown.service_healthy(config.SERVICE_NAME)
```
elezar: Probably out of scope of this PR, but isn't this redundant?

PR #2124, opened 2018-01-24
[INFINITY-2842] Add tls krb5 tests for HDFS

elezar:
This PR adds a basic test for HDFS when both TLS and Kerberos are enabled.

The following basic checks are performed:
* The correct certificate is presented to the client for the `name`, `journal`, and `data` nodes (TLS test)
* Kerberos user authnz behaves as expected. 

I will follow up wit a second PR to remove some duplication for handling TLS. (See #2126)

### DIFF
--- a/frameworks/hdfs/tests/test_ssl_kerberos_auth.py
+++ b/frameworks/hdfs/tests/test_ssl_kerberos_auth.py
@@ -195,7 +195,7 @@ def test_verify_https_ports(hdfs_client, node_type, port):
 def test_user_can_auth_and_write_and_read(hdfs_client, kerberos):
     sdk_auth.kinit(hdfs_client["id"], keytab=config.KEYTAB, principal=kerberos.get_principal("hdfs"))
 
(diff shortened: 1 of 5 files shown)


frameworks/hdfs/tests/test_ssl_kerberos_auth.py
```diff
@@ -0,0 +1,261 @@
...
+        dcos_ca_bundle = auth.fetch_dcos_ca_bundle(client_id)
+
+        yield {**client, **{"dcos_ca_bundle": dcos_ca_bundle}}
+
+    finally:
+        sdk_marathon.destroy_app(client_id)
+
+
+# TODO(elezar) Is there a better way to determine this?
+DEFAULT_JOURNAL_NODE_TLS_PORT = 8481
```
benclarkwood: Parsing the package description is the only way that comes to mind immediately.

elezar: I think we could also download `hdfs-site.xml` and extract the ports that way.

PR #2125, opened 2018-01-24
Warn that cassandra doesn't currently support backup/restore with auth.

mpereira:
Would backup/restore work with [AllowAllAuthenticator](https://docs.datastax.com/en/cassandra/3.0/cassandra/configuration/secureConfigNativeAuth.html) and [AllowAllAuthorizer](https://docs.datastax.com/en/cassandra/3.0/cassandra/configuration/secureConfigInternalAuth.html) (the default auth configuration)?

If yes we should probably mention it too.

### DIFF
--- a/frameworks/cassandra/docs/limitations.md
+++ b/frameworks/cassandra/docs/limitations.md
@@ -50,6 +50,10 @@ To prevent accidental data loss, the service does not support reducing the numbe
 
 To prevent accidental data loss from reallocation, the service does not support changing volume requirements after initial deployment.
 
+## Backup/Restore
+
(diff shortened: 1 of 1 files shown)


mpereira:
I opened a [PR](https://github.com/mesosphere/dcos-cassandra-service/pull/461) for the deprecated service too.

benclarkwood:
It doesn't make any sense to document a limitation for a feature that does not exist (authn/authz).

cc @keithchambers

mpereira:
> Would backup/restore work with AllowAllAuthenticator and AllowAllAuthorizer (the default auth configuration)?
> If yes we should probably mention it too.

@benclarkwood what do you think about this?

PR #2130, opened 2018-01-24
[INFINITY-2540] Restrict resource refinement tests to supported versions

gabrielhartmann:


gabrielhartmann:
@elezar:  Thanks.  Made the recommended change.

nickbp:
The test_taskcfg flake should be fixed in #2131. Given that this PR is fixing unrelated tests, and that it's working going by how the test was correctly disabled in the 1.9 run, I'm going to just merge.

PR #2131, opened 2018-01-24
INFINITY-3026: Reenable shutdown test

nickbp:
Handle the case where multiple tasks are colocated on the host being shut down. Replace em all.

### DIFF
--- a/docs/pages/tutorials/kafka-tutorial.md
+++ b/docs/pages/tutorials/kafka-tutorial.md
@@ -49,6 +49,7 @@ configs:
 ```
 
 9. Save and exit the `svc.yml` file. It should look similar to the following:
+
 ```
 name: {{FRAMEWORK_NAME}}
 scheduler:
@@ -81,6 +82,7 @@ pods:
             template: server.properties.mustache
             dest: kafka_2.11-0.10.0.0/config/server.properties
 ```
+
 The `server.properties.mustache` should live in the current directory (`dcos-commons/examples/kafka/src/main/dist`).
 
 10. Download the `server.properties.mustache` file mentioned above:
--- a/frameworks/cassandra/tests/test_zzzrecovery.py
+++ b/frameworks/cassandra/tests/test_zzzrecovery.py
@@ -69,7 +69,10 @@ def test_shutdown_host():
     candidate_tasks = sdk_tasks.get_tasks_avoiding_scheduler(
         config.SERVICE_NAME, re.compile('^node-[0-9]+-server$'))
     assert len(candidate_tasks) != 0, 'Could not find a node to shut down'
-    # Just pick the first task from the list. In practice, we should never have two node pods on the same machine.
+    # Cassandra nodes should never share a machine
+    assert len(candidate_tasks) == len(set([task.host for task in candidate_tasks])), \
+        'Expected candidate tasks to all be on different hosts: {}'.format(candidate_tasks)
+    # Just pick the first one from the list
     replace_task = candidate_tasks[0]
 
     replace_pod_name = replace_task.name[:-len('-server')]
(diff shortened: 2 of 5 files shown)


gabrielhartmann:
Two failures to provision clusters, re-kicked the runs.

docs/pages/tutorials/kafka-tutorial.md
```diff
@@ -49,6 +49,7 @@ configs:
 ```
 
 9. Save and exit the `svc.yml` file. It should look similar to the following:
+
```
gabrielhartmann: unintentional changes in this PR?

frameworks/cassandra/tests/test_zzzrecovery.py
```diff
@@ -66,9 +66,12 @@ def test_node_replace_replaces_node():
 # @@@@@@@
 @pytest.mark.sanity
 def test_shutdown_host():
-    replace_task = sdk_tasks.get_task_avoiding_scheduler(
+    candidate_tasks = sdk_tasks.get_tasks_avoiding_scheduler(
         config.SERVICE_NAME, re.compile('^node-[0-9]+-server$'))
-    assert replace_task is not None, 'Could not find a node to shut down'
+    assert len(candidate_tasks) != 0, 'Could not find a node to shut down'
```
gabrielhartmann: if we should never have two pods on the same agent, should we:
```
assert len(candidate_tasks) == 1
```
?

nickbp: This is the list of tasks which aren't colocated with the scheduler. Would expect >=2.

frameworks/helloworld/tests/test_zzzrecovery.py
```diff
@@ -288,18 +288,26 @@ def test_pod_replace():
...
+    log.info('Tasks on host {} to be replaced after shutdown: {}'.format(replace_hostname, replace_tasks))
 
     # Instead of partitioning or reconnecting, we shut down the host permanently
-    sdk_cmd.shutdown_agent(replace_task.host)
+    sdk_cmd.shutdown_agent(replace_hostname)
 
-    sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME,
-                    'pod replace {}'.format(replace_pod_name))
+    for task in replace_tasks:
+        pod_name = task.name[:-len('-server')]
```
gabrielhartmann: What does this do?  Don't we want pod names, not task names?

docs/pages/tutorials/kafka-tutorial.md
nickbp: Not entirely unintentional, just sneaking it in to fix a formatting problem: http://mesosphere.github.io/dcos-commons/tutorials/kafka-tutorial.html

frameworks/cassandra/tests/test_zzzrecovery.py
gabrielhartmann: ahhh, my mistake

frameworks/helloworld/tests/test_zzzrecovery.py
nickbp: get pod name for `pod replace` operation: `hello-0-server` => `hello-0`

gabrielhartmann: I see.  Now I can read it.  Thanks.

gabrielhartmann: Should we get a `Set` of pod names before issuing replace calls so we don't call replace on the same pod twice?

nickbp: It shouldn't be a problem here, as there's only 1 task/pod. Could add a sanity check to that effect though.

elezar: Could use rstrip("-server") instead to make the intent clearer. Also safer if the string doesn't contain "-server"

nickbp: doesn't rstrip just strip anything that matches the set?

```
>>> 'hello'.rstrip('lo')
'he'
```

nickbp:
Merging as the failed cassandra tests are unrelated to what's being fixed here.

frameworks/helloworld/tests/test_zzzrecovery.py
elezar: My bad. I have been using it wrong!

PR #2133, opened 2018-01-24
INFINITY-3025: Increase namenode readiness check timeout

nickbp:
A deployment failure in CI appeared to be because the readiness check expired before the name node could finish starting. More diagnosis would depend on INFINITY-2998 (fetching executor logs)


nickbp:
Discussed offline, looks like the current timing was just a bit too tight

PR #2135, opened 2018-01-25
INFINITY-3032: Bump JRE to 1.8-162 (DEV CHANGE: update JAVA_HOME)

nickbp:
Experiments:
- [X] Try using the 'Server JRE', which is significantly smaller than the stock JRE (52.26MB instead of 76.4MB), but which places the JRE itself under e.g. `jdk1.8.0_162/jre/` instead of `jre1.8.0_162/`
- [X] Remove the separate JCE build, as 162 should have unlimited enabled by default


nickbp:
Merging as the only failure is consistently the known issue with azure creds.

PR #2136, opened 2018-01-25
INFINITY-3034 Fix azure backup parameters

nickbp:
Not clear how this could have been passing before, but it's failing now! Remove AZURE_ prefix from some parameters per [documentation](https://mesosphere.github.io/dcos-commons/services/cassandra/disaster-recovery.html) and svc.yml content.


nickbp:
Looks like the creds in CI are now bad

nickbp:
Remaining failures are consistently around bad cred configuration in CI. Merged this as it at least gets us less-broken.

PR #2137, opened 2018-01-25
[INFINITY-3035] toggle kafka security settings

elezar:
This PR adds a test (not marked as sanity) that performs the following steps:

- installs Kafka with the default settings
- updates Kafka to enable kerberos
- updates Kafka to disable kerberos

After every step it runs a client correctly configured to send and receive a message.

This is still a WIP, but could be used to automate the testing of other toggles.

Note: This is based off #2126


frameworks/kafka/tests/test_toggle_kerberos.py
```diff
@@ -0,0 +1,264 @@
...
+    pytest.mark.skipif(sdk_utils.dcos_version_less_than('1.10'), reason="Kerberos tests require DC/OS 1.10"),
+    pytest.mark.skipif(sdk_utils.is_open_dcos(), reason="Kerberos tests required DC/OS EE"),
+]
+
+
+MESSAGES = []
+
+
+@pytest.fixture(scope='module', autouse=True)
+def kerberos(configure_security):
```
gabrielhartmann: A comment similar to the one below for `kafka_server` would be nice.

frameworks/kafka/tests/test_toggle_kerberos.py
```diff
@@ -0,0 +1,264 @@
...
+            additional_options=service_kerberos_options,
+            timeout_seconds=30 * 60)
+
+        yield {**service_kerberos_options, **{"package_name": config.PACKAGE_NAME}}
+    finally:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.fixture(scope='module', autouse=True)
+def kafka_client(kerberos, kafka_server):
```
gabrielhartmann: ditto

frameworks/kafka/tests/test_toggle_tls.py
```diff
@@ -0,0 +1,403 @@
...
+        sdk_cmd.run_cli(
+            "security org groups add_user superusers {name}".format(name=name))
+        yield {"name": name, "secret": secret}
+    finally:
+        sdk_security.delete_service_account(
+            service_account_name=name, service_account_secret=secret)
+
+
+@pytest.fixture(scope='module', autouse=True)
+def kafka_server(service_account):
```
gabrielhartmann: `kafka_server` and `kafka_client` look very similar at first glance to the one used in the other test.

gabrielhartmann:
Metronome has perhaps caused us some headaches for reading/writing/deleting/validating data in C* tests.  However I think we should be careful not to write two ways of doing these kinds of activities.  Here we implement our own "job" execution system by task execing into a Marathon app.  This isn't consistent with how we do it for C* and also seems like us working around our own product.  I'd prefer we do things in Kafka as we do them in C*.

If we want to move off of Metronome later we can do that in a consistent way too.

Anticipating a desire to move from one job execution method to another perhaps we'd like to abstract that capability a bit more.  I don't want to block this PR on that kind of an effort.  I would like to see a consistent approach though to how we read/write/delete/validate data across services though.

elezar:
Thanks for the review @gabrielhartmann. I'll definitely address the lack of comments and some of the duplication. On the latter, I have been trying to follow more of a tick-tock approach for changes like this, where I make localised changes first and then follow up with a refactoring/deduplication PR. In this case, I can definitely start the process with these two tests.

I'm not sure that I agree that what is being done here is equivalent to the jobs in C*. There, we are testing functionality built into the service itself (please correct me if I'm wrong), and here we are testing whether a *client* can connect to the Kafka cluster and perform some basic operations. This then represents more of an end-to-end tests (from a client perspective) than a standard integration test -- which is also why I have not marked these as `sanity` yet.

With that said, I do agree that we should work towards uniformity (along with abstraction) so that we can easily switch job backends/execution frameworks. In this regard, I have been working on getting our client-based tests (currently those found in Kafka and HDFS) to be more uniform. I will have another look at the C* jobs to see how the operations that I am performing would fit into that.

frameworks/kafka/docs/limitations.md
```diff
@@ -17,6 +17,17 @@ The "disk" configuration value is denominated in MB. We recommend you set the co
...
 
+### Toggling Kerberos
+
+Kerberos authentication can be toggled (enabled / disabled), but this triggers a rolling restart of the cluster. Clients configured with the old security settings will lose connectivity during and after this process. It is recommended that backups are made and downtime is scheduled.
+
+### Toggling TLS
+
+Transport encryption using TLS can be toggled (enabled / disabled), but will trigger a rolling restart of the cluster. As each broker restarts, a client may lose connectivity based on its security settings and the value of the `service.security.transport_encryption.allow_plaintext` configuration option. It is recommended that backups are made and downtime is scheduled.
+
+In order to enable TLS, a service account and corresponding secret is required. Since it is not possible to change the service account used by a service, it is recommended that the service is deployed with an explicit service account to allow for TLS to be enabled at a later stage.
```
mpereira: > Since it is not possible to change the service account used by a service

Is this documented somewhere? I just tried it with Elastic and this is what `update status` shows:
```
$ dcos beta-elastic --name=beta-elastic update status
deploy (serial strategy) (ERROR)
├─ master-update (serial strategy) (COMPLETE)
│  ├─ master-0:[node] (COMPLETE)
│  ├─ master-1:[node] (COMPLETE)
│  └─ master-2:[node] (COMPLETE)
├─ data-update (serial strategy) (COMPLETE)
│  ├─ data-0:[node] (COMPLETE)
│  └─ data-1:[node] (COMPLETE)
├─ ingest-update (serial strategy) (COMPLETE)
└─ coordinator-update (serial strategy) (COMPLETE)
   └─ coordinator-0:[node] (COMPLETE)

Errors:
- Field: 'TaskVolumes[taskname:node]'; Transition: '[name: disk, value: type: SCALAR scalar { value: 10000.0 }, role: beta-elastic-role, pre-reserved-role: *, principal: elastic, type: 'ROOT', container-path: 'container-path']' => '[name: disk, value: type: SCALAR scalar { value: 10000.0 }, role: beta-elastic-role, pre-reserved-role: *, principal: elastic-same-key, type: 'ROOT', container-path: 'container-path']'; Message: 'Volumes must be equal.'
- Field: 'TaskVolumes[taskname:node]'; Transition: '[name: disk, value: type: SCALAR scalar { value: 2000.0 }, role: beta-elastic-role, pre-reserved-role: *, principal: elastic, type: 'ROOT', container-path: 'container-path']' => '[name: disk, value: type: SCALAR scalar { value: 2000.0 }, role: beta-elastic-role, pre-reserved-role: *, principal: elastic-same-key, type: 'ROOT', container-path: 'container-path']'; Message: 'Volumes must be equal.'
- Field: 'TaskVolumes[taskname:node]'; Transition: '[name: disk, value: type: SCALAR scalar { value: 2000.0 }, role: beta-elastic-role, pre-reserved-role: *, principal: elastic, type: 'ROOT', container-path: 'container-path']' => '[name: disk, value: type: SCALAR scalar { value: 2000.0 }, role: beta-elastic-role, pre-reserved-role: *, principal: elastic-same-key, type: 'ROOT', container-path: 'container-path']'; Message: 'Volumes must be equal.'
- Field: 'TaskVolumes[taskname:node]'; Transition: '[name: disk, value: type: SCALAR scalar { value: 1000.0 }, role: beta-elastic-role, pre-reserved-role: *, principal: elastic, type: 'ROOT', container-path: 'container-path']' => '[name: disk, value: type: SCALAR scalar { value: 1000.0 }, role: beta-elastic-role, pre-reserved-role: *, principal: elastic-same-key, type: 'ROOT', container-path: 'container-path']'; Message: 'Volumes must be equal.'
```

elezar: I'm not sure whether we have documentation of the limitation (@benclarkwood @gabrielhartmann?).

Have you tried using the `elastic` service account that is automatically generated and just adding a secret to that?

mpereira: Yes the default service account works, what doesn't work is the thing you mentioned: changing it (i.e. `elastic` -> `elastic-same-key`), even though the new one has the same permissions and secret.

I asked mostly because I didn't know of that limitation and couldn't find it in the documentation, and also because the error message shown is kinda cryptic. Maybe I'm unaware of something other than `update status` that shows a better error message...

benclarkwood:
@gabrielhartmann we do all our client testing for security via this methodology. We spin up a Marathon container to act as the "client", and then task exec to it. 

Metronome (or another job system) is not a replacement for what we are doing here, direct client testing.

elezar:
@gabrielhartmann after a discussion with @benclarkwood, I would stick to **not** using metronome here. I do agree that there is duplication that should be removed, and I have created https://jira.mesosphere.com/browse/KAF-406 to follow-up on this.

PR #1910, opened 2017-12-05
[INFINITY-2532] Update Cassandra to 3.0.15

gabrielhartmann:

PR #2139, opened 2018-01-25
[CASSANDRA-669] Update Cassandra to 3.0.15 (#1910)

benclarkwood:


gabrielhartmann:
Would be nice to make these tests look less terrible before approving this.  In particular getting C* to pass.  Maybe we need to spend some time giving the `0.30.X` branch some CI improvement attention.

PR #2146, opened 2018-01-26
[KAF-398] Disable vips when kerberos enabled

benclarkwood:
VIPs will not work with Kerberos. Disable them when Kerberos is enabled.

### DIFF
--- a/frameworks/kafka/docs/limitations.md
+++ b/frameworks/kafka/docs/limitations.md
@@ -1,6 +1,6 @@
 ---
 layout: layout.pug
-navigationTitle: 
+navigationTitle:
 excerpt:
 title: Limitations
 menuWeight: 100
@@ -13,7 +13,9 @@ The "disk" configuration value is denominated in MB. We recommend you set the co
 
 ## Security
 
-The security features introduced in Apache Kafka 0.9 are not supported at this time.
+### Kerberos
+
+When Kerberos is enabled, the broker vip is disabled as it will not function correctly.
 
 ## Out-of-band configuration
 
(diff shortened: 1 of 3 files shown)


frameworks/kafka/tests/test_kerberos_auth.py
```diff
@@ -122,6 +123,13 @@ def kafka_client(kerberos, kafka_server):
         sdk_marathon.destroy_app(client_id)
 
 
+@pytest.mark.dcos_min_version('1.10')
+@sdk_utils.dcos_ee_only
+@pytest.mark.sanity
+def test_no_vip(kafka_client, kafka_server, kerberos):
+    sdk_networks.get_and_test_endpoints(kafka_server["package_name"], kafka_server["service"]["name"], "broker", 2)
```
gabrielhartmann: Looks like this only checks count of endpoints.  Would perhaps be better to check expected information in the endpoints.

frameworks/kafka/docs/limitations.md
```diff
@@ -13,7 +13,9 @@ The "disk" configuration value is denominated in MB. We recommend you set the co
 
 ## Security
 
-The security features introduced in Apache Kafka 0.9 are not supported at this time.
+### Kerberos
+
+When Kerberos is enabled, the broker vip is disabled as it will not function correctly.
```
gabrielhartmann: We may want to be slightly careful with the language here so it's clear which component doesn't like working with VIPs.  Maybe a brief discussion of Kerberos' behavior would be good.

PR #2150, opened 2018-01-26
INFINITY-3037: Document Elastic TLS toggling limitation.

mpereira:


frameworks/elastic/docs/limitations.md
```diff
@@ -15,13 +15,19 @@ The maximum number of deployable nodes is constrained by the DC/OS cluster's res
 
 During deployment and upgrades, the `serial` strategy does not wait for the Elastic service to reach green before proceeding to the next node.
 
+## Toggling TLS requires doing a full-cluster restart
```
elezar: I would keep the heading as just `Toggling TLS` and then start the section with this as a complete sentence.

frameworks/elastic/docs/limitations.md
```diff
@@ -15,13 +15,19 @@ The maximum number of deployable nodes is constrained by the DC/OS cluster's res
...
+
+This is an [Elasticsearch limitation](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-tls.html). Since the service's default update strategy is a rolling upgrade (`serial`) changing its configuration to enable (or disable) transport encryption will result in nodes configured with TLS not being able to communicate with nodes configured with unencrypted networking (and vice-versa). You must do a full-cluster restart (`parallel` update strategy). Make sure to have backups and plan for some downtime.
```
elezar: Is "unencrypted networking" the standard terminology for this in Elasticsearch? In the other services we use "plaintext".

frameworks/elastic/docs/limitations.md
```diff
@@ -15,13 +15,19 @@ The maximum number of deployable nodes is constrained by the DC/OS cluster's res
...
+
+Afterwards, you can set the update strategy back to `serial` for future updates.
```
elezar: Is this a "can", or a "should"?

frameworks/elastic/docs/limitations.md
elezar: The cluster restart sentence could be worded as:
"For this a full-cluster restart is required (enabled using the `parallel` update strategy)"

mpereira: > Is "unencrypted networking" the standard terminology for this in Elasticsearch? In the other services we use "plaintext".

Yes. I tried to use ES terminology for "node" configuration and SDK terminology for "service" configuration. I'll switch to "plaintext" so that we're consistent.

mpereira: > The cluster restart sentence could be worded as:
"For this a full-cluster restart is required (enabled using the parallel update strategy)"

In that sentence, reading through the whole paragraph, doesn't it sound like "this" is referring to "nodes configured with TLS not being able to communicate with nodes configured with unencrypted networking", which could be a bit confusing?

What do you think about "A full-cluster restart is required (using the `parallel` update strategy)"?

frameworks/elastic/docs/limitations.md
mpereira: Hm, good question. It is not absolutely required, but a good idea in most cases. I used the same wording as in the [X-Pack docs](https://github.com/mesosphere/dcos-commons/blob/master/frameworks/elastic/docs/elastic-x-pack.md) fwiw.

Should we change both to "should"?

frameworks/elastic/docs/limitations.md
elezar: Feel free to continue using the ES terminology if you think it's better.

I like the "A full-cluster restart ..." suggestion.

frameworks/elastic/docs/limitations.md
elezar: If "serial" is our recommended mode of operation, then "can" is a bit loose.

I'm OK to leave it as "can" if X-Pack is considered the reference. We could ask @loren or @benclarkwood for their input.

mpereira:
@elezar I kept the "unencrypted networking" ES terminology but I think I got everything else: https://github.com/mesosphere/dcos-commons/pull/2150/commits/3c07d8e9bb5d67d4e12f4be25c61f1238fbca1f5

PR #2151, opened 2018-01-26
INFINITY-2977: Add initial Elastic TLS instructions.

mpereira:
Still pending updating other documentation files to point at this one. Also maybe we don't want to recommend adding service accounts to the `superusers` group for this, and instead add them to a group with the minimum set of permissions necessary.

### DIFF
--- a/frameworks/elastic/docs/tls.md
+++ b/frameworks/elastic/docs/tls.md
@@ -1,0 +1,139 @@
+---
+layout: layout.pug
+navigationTitle:
+excerpt:
+title: Configuring TLS
+menuWeight: 25
+
+---
+
+# Encrypting Network Communication
+
+Elasticsearch requires
+[some work](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-tls.html)
+to get TLS working. The Elastic service takes care of setting everything up when transport encryption is enabled. It will generate and deploy certificates and enable SSL both on the transport and HTTP layers by default.
+
+The following steps describe getting an Elastic service with TLS from scratch. You might already have key pairs, service
+accounts and secrets that you could already use. In that case only run the steps you need.
+
+## Install the Enterprise DC/OS CLI
+
+```
+dcos package install --yes dcos-enterprise-cli
+```
+
+## Create key pair
+
+This generates `elastic.private.pem` and `elastic.public.pem`.
+
+```
+dcos security org service-accounts keypair elastic.private.pem elastic.public.pem
+```
+
+## Create service account
+
+The service account name that is configured for your service (`service.service_account` in the service JSON
+configuration or the `service account` text field in the "Edit Configuration" UI) must match the actual service account
+name, which we're calling `elastic` in this case.
+
+```
+dcos security org service-accounts create -p elastic.public.pem -d "Elastic service account" elastic
+```
+
+## Create service account secret
+
+This stores the private key as a [DC/OS secret](https://docs.mesosphere.com/1.10/security/ent/secrets/) named `elastic-secret`.
+
+```
+dcos security secrets create-sa-secret elastic.private.pem elastic elastic-secret
+```
+
+## Add service account to `superusers` group
+
+```
+dcos security org groups add_user superusers elastic
+```
+
+## Install or update Elastic with TLS enabled
+
+Elasticsearch
+[requires that X-Pack is installed](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-tls.html)
+to enable transport encryption. Make sure `elasticsearch.xpack_enabled` is set to `true` or nodes won't be able to
+start.
+
+### Install
+
+Make sure to replace `$your_service_name` by your actual service name.
+
+```bash
+cat <<EOF > elastic_with_tls.json
+{
+  "service": {
+    "name": "$your_service_name",
+    "service_account": "elastic",
+    "service_account_secret": "elastic-secret",
+    "security": {
+      "transport_encryption": {
+        "enabled": true
+      }
+    }
+  },
+  "elasticsearch": {
+    "xpack_enabled": true
+  }
+}
+EOF
+```
+
+```
+dcos package install --yes --options=elastic_with_tls.json beta-elastic
+```
+
+### Or update a running service with TLS turned off
+
+Updating requires a `parallel` strategy due to a know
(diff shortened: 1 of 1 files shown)


frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+configuration or the `service account` text field in the "Edit Configuration" UI) must match the actual service account
+name, which we're calling `elastic` in this case.
+
+```
+dcos security org service-accounts create -p elastic.public.pem -d "Elastic service account" elastic
+```
+
+## Create service account secret
+
+This stores the private key as a [DC/OS secret](https://docs.mesosphere.com/1.10/security/ent/secrets/) named `elastic-secret`.
```
elezar: We should use a non-versioned URI for the link. Try replacing `1.10` with `latest` or `current`.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+
+```
+dcos security secrets create-sa-secret elastic.private.pem elastic elastic-secret
+```
+
+## Add service account to `superusers` group
```
elezar: I think there are other permissions that we could use here. Perhaps check with @mhrabovcin on this.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+
+---
+
+# Encrypting Network Communication
+
+Elasticsearch requires
+[some work](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-tls.html)
+to get TLS working. The Elastic service takes care of setting everything up when transport encryption is enabled. It will generate and deploy certificates and enable SSL both on the transport and HTTP layers by default.
+
+The following steps describe getting an Elastic service with TLS from scratch. You might already have key pairs, service
```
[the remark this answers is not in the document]
elezar: For elastic it may make sense to mention the X-Pack dependency earlier.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+Elasticsearch
+[requires that X-Pack is installed](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-tls.html)
+to enable transport encryption. Make sure `elasticsearch.xpack_enabled` is set to `true` or nodes won't be able to
+start.
+
+### Install
+
+Make sure to replace `$your_service_name` by your actual service name.
+
+```bash
```
elezar: The norm for our other user docs is to simply have the `json` in a `json` block:
```json
{
  "service": {
    "name": "secure-elastic"
  }
}
```

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+## Install or update Elastic with TLS enabled
+
+Elasticsearch
+[requires that X-Pack is installed](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-tls.html)
+to enable transport encryption. Make sure `elasticsearch.xpack_enabled` is set to `true` or nodes won't be able to
+start.
+
+### Install
+
+Make sure to replace `$your_service_name` by your actual service name.
```
elezar: Rather just specify the string in the example and state that this is the name used. It is not clear that the `$` here represents the start of a bash variable name or some other notation.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+      }
+    }
+  },
+  "elasticsearch": {
+    "xpack_enabled": true
+  }
+}
+EOF
+```
+
```
elezar: I would follow the `json` block with something like:
Here the service name is `"my-elastic"` and the `service_account` and `service_account_secrets` correspond with the values used in the previous steps.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+```
+dcos package install --yes --options=elastic_with_tls.json beta-elastic
+```
+
+### Or update a running service with TLS turned off
+
+Updating requires a `parallel` strategy due to a know
```
elezar: As far as I could tell from my tests, this also *required* that the service account (and possibly secret) had already be setup and used to deploy the service.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+[Elasticsearch limitation](https://github.com/mesosphere/dcos-commons/blob/e4b2ab5d82a6cda31815fc1224d1eca768513aa9/frameworks/elastic/docs/limitations.md#toggling-tls-requires-doing-a-full-cluster-restart).
+
+```bash
```
elezar: Use a `json` block.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+  },
+  "elasticsearch": {
+    "xpack_enabled": true
+  }
+}
+EOF
+```
+
+```
+dcos beta-elastic --name=$your_service_name update start --options=elastic_with_tls.json
```
elezar: `bash` block and precede the command with `$`.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+## Now you can only access your Elasticsearch cluster through HTTPS
+
+```bash
+dcos task exec -it coordinator-0-node curl -k -u elastic:changeme https://coordinator.$your_service_name.l4lb.thisdcos.directory:9200
```
elezar: Precede the `bash` command with a `$`, and don't use the `$your_service_name` variable unless it is explicitly specified.

elezar: This is also a bit misleading as we are not accessing the cluster here. We are just running a command on the node directly. It may be more useful to link to information on how one should connect to a TLS-enabled Elastic cluster.

frameworks/elastic/docs/tls.md
```diff
@@ -0,0 +1,139 @@
+---
+layout: layout.pug
+navigationTitle:
+excerpt:
+title: Configuring TLS
+menuWeight: 25
```
elezar: Is this the correct menu position. @triclambert who do we reach out to w.r.t documentation issues now?

mpereira:
@benclarkwood I'm closing this one since you said you're reuse the same form used by the Kafka security section.

PR #2157, opened 2018-01-29
Fix placement constraint property names.

mpereira:


mpereira:
I also created https://jira.mesosphere.com/browse/INFINITY-3055

mpereira:
Yup!

PR #2162, opened 2018-01-30
[KAF-399] Remove existing ZK jar before moving in the new one

benclarkwood:
It is nice to be able to label the custom ZK lib we inject as something different than what is already there. 

By removing the pre-existing one, we make sure we aren't simply relying on luck vis a vis the classpath.


elezar:
How does this affect Confluent ZK?

PR #2163, opened 2018-01-30
Update Logging

gluehbirnenkopf:
Recommendations.isEmpty should trigger error instead of warning. 
Otherwise Instances, which cant be deployed will be ignored to easy.


gabrielhartmann:
Mesos passes in offers in sets.  It's not totally unexpected that a given set of offers doesn't have the needed resources.  In practice this is actually a very common scenario, but not necessarily a problematic one.  The next set of offers is likely to have what you're looking for, especially if you're expecting reserved resources.  We're pretty much always expecting reserved resources except for the initial deployment of a pod instance.

I tend to think errors are reserved for actual mistakes.  This situation isn't a mistake, but you might want to keep an eye on it.  I think this is the situation for which `warn` was designed.

PR #2165, opened 2018-01-31
Fix zone validation and add validation to Elastic

gabrielhartmann:
I was originally just supposed to be adding zone transition validation to Elastic.  While writing a test validating it worked I found out Zone Validation is broken everywhere.

So this PR:
1. Fixes Zone validation
1. Adds it to Elastic
1. Fixes bad original zone validation tests (they shouldn't reference tasks at all)

I'll add more PRs for more extensive testing of Cassandra and HDFS which already had validation turned on.

I have validated manually that C* and Elastic now act as they're supposed to.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/validation/ZoneValidator.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/validation/ZoneValidator.java
@@ -1,12 +1,13 @@
 package com.mesosphere.sdk.specification.validation;
 
 import com.mesosphere.sdk.config.validate.ConfigValidationError;
-import com.mesosphere.sdk.offer.TaskUtils;
-import com.mesosphere.sdk.offer.taskdata.EnvConstants;
+import com.mesosphere.sdk.offer.evaluate.placement.PlacementRule;
+import com.mesosphere.sdk.offer.evaluate.placement.PlacementUtils;
+import com.mesosphere.sdk.specification.PodSpec;
 import com.mesosphere.sdk.specification.ServiceSpec;
-import com.mesosphere.sdk.specification.TaskSpec;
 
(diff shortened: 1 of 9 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/validation/ZoneValidator.java
```diff
@@ -20,52 +20,44 @@
     public static Collection<ConfigValidationError> validate(
             Optional<ServiceSpec> oldConfig,
             ServiceSpec newConfig,
-            String podType,
-            String taskName) {
+            String podType) {
```
nickbp: Could simplify this for callers with a `String... podTypes` where all passed podTypes are checked. Then e.g. HDFS could call it as `validate(..., NAME_POD_TYPE, JOURNAL_POD_TYPE, DATA_POD_TYPE)`

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/validation/ZoneValidator.java
```diff
@@ -20,52 +20,44 @@
...
-    }
-
-    static List<ConfigValidationError> validateTransition(
-            String oldEnv, String newEnv, String podType, String taskName) {
-        boolean oldZones = Boolean.valueOf(oldEnv);
-        boolean newZones = Boolean.valueOf(newEnv);
+        Optional<PlacementRule> oldRule = oldPod.get().getPlacementRule();
+        Optional<PlacementRule> newRule = newPod.get().getPlacementRule();
 
         ConfigValidationError error = ConfigValidationError.transitionError(
```
benclarkwood: You could construct the error inside the conditional. Not worth redoing CI over, of course.

PR #2171, opened 2018-01-31
[HDFS-493] Switch to java tool for checking validity of keytabs

benclarkwood:
Giving this a shot, rather than checking for a specific string in the `strings` output.


benclarkwood:
This is weird. My tool worked fine for all the kafka keytabs but didn't for any of the HDFS ones. Boo.

elezar:
I'm not sure if this is a keytab validity issue? I have just run the HDFS tests a number of times in a row, and deployment fails every time. Changing to non-foldered service names removes this issue.

elezar:
Could one not use the Java-supplied class directly: https://gist.github.com/incepttechnologies/8243870
```
Alternatively you can also use Klist or Ktab utility that comes with standard java.

C:\>java sun.security.krb5.internal.tools.Klist -k -t krba01.keytab
```

elezar:
I have a minor fix in #2175 that builds on these changes and has a green CI. Should we close this here and review further there?

benclarkwood:
Yes

PR #2172, opened 2018-01-31
[HDFS-433] Add missing HDFS settings

NimaVaziri:
A few settings are currently not exposed, as they must have been missed earlier when most settings became exposed.

This PR exposes these missed settings.

### DIFF
--- a/frameworks/hdfs/universe/config.json
+++ b/frameworks/hdfs/universe/config.json
@@ -185,6 +185,11 @@
           "media": {
             "type": "application/x-zone-constraints+json"
           }
+        },
+        "hadoop_journalnode_opts": {
+          "type": "string",
+          "description": "Options to specify when running the journal node.",
+          "default": ""
         }
       },
       "required": [
@@ -731,6 +736,11 @@
           "description": "Boolean which enables backend datanode-side support for the experimental DistributedFileSystem#getFileVBlockStorageLocations API.",
           "default": false
         },
+        "fsdataset_volume_choosing_policy": {
+          "type": "string",
+          "description": "The class name of the policy for choosing volumes in the list of directories. Defaults to org.apache.hadoop.hdfs.server.datanode.fsdataset.RoundRobinVolumeChoosingPolicy. If you would like to take into account available disk space, set the value to 'org.apache.hadoop.hdfs.server.datanode.fsdataset.AvailableSpaceVolumeChoosingPolicy'.",
+          "default": ""
+        },
         "available-space-volume-choosing-policy_balanced-space-threshold": {
           "type": "string",
           "description": "Only used when the dfs.datanode.fsdataset.volume.choosing.policy is set to org.apache.hadoop.hdfs.server.datanode.fsdataset.AvailableSpaceVolumeChoosingPolicy. This setting controls how much DN volumes are allowed to differ in terms of bytes of free disk space before they are considered imbalanced. If the free space of all the volumes are within this range of each other, the volumes will be considered balanced and block assignments will be done on a pure round robin basis.",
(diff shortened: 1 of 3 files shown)


frameworks/hdfs/universe/config.json
```diff
@@ -778,6 +788,11 @@
           "media": {
             "type": "application/x-zone-constraints+json"
           }
+        },
+        "hadoop_datanode_opts": {
```
elezar: The descriptions for these three options could be improved a bit -- including how they interact with the `hadoop_opts` parameter.

frameworks/hdfs/universe/config.json
```diff
@@ -731,6 +736,11 @@
           "description": "Boolean which enables backend datanode-side support for the experimental DistributedFileSystem#getFileVBlockStorageLocations API.",
           "default": false
         },
+        "fsdataset_volume_choosing_policy": {
```
elezar: Should we not set the default to the default values specified in the description?

NimaVaziri: This setting doesn't have a default according to the official documentation: https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/hdfs-default.xml

frameworks/hdfs/universe/config.json
NimaVaziri: 👍 added a bit more detail to the descriptions

frameworks/hdfs/universe/config.json
elezar: Sorry, I thought a default was mentioned in the description?

NimaVaziri: Yeah, it says the default is `org.apache.hadoop.hdfs.server.datanode.fsdataset.RoundRobinVolumeChoosingPolicy` but then doesn't actually put that in the value field in the chart. Anyhow, let's go with this as the default.

frameworks/hdfs/universe/config.json
```diff
@@ -875,9 +890,14 @@
         },
         "hadoop_opts": {
           "type": "string",
-          "description": "Extra Java runtime options.",
+          "description": "Extra Java runtime options for all Hadoop nodes.",
           "default": ""
         },
+        "hadoop_heapsize": {
+          "type": "string",
```
elezar: Don't we generally use integers for memory values?

NimaVaziri: Ah generally yeah. Earlier versions of Hadoop require you to specify the multiplier type as well as the number eg `1024MB` instead of just `1024`. More recent versions, such as the one here, don't require this anymore so we can change this to a `number` type.

elezar: One last question. Is this kind of change safe across upgrades?

NimaVaziri: In this case, yeah it's safe because Hadoop already implicitly takes this setting into account by setting this value by default to 1000MB. Explicitly setting it doesn't change the functionality, and even upgrading to a package that has this setting with a different number simply changes the heap size available to the Hadoop nodes, so not exactly a breaking change.

elezar: I mean that if one does a `dcos hdfs describe > config.json`, make an edit and then `dcos hdfs update start --options=config.json --version=next-version` where `next-version` is the one that now has an integer in this field and the one that was running had a string.

NimaVaziri: Ah, well `hadoop_heapsize` doesn't currently exist at all in `config.json` so the first version of the package would have this field as an `integer` to begin with.

elezar: (oh right ... it's been a long day :) )

frameworks/hdfs/universe/config.json
```diff
@@ -894,7 +894,7 @@
           "default": ""
         },
         "hadoop_heapsize": {
-          "type": "string",
+          "type": "integer",
```
elezar: The default should change to an integer value too though.

PR #2175, opened 2018-02-01
[HDFS-493] update hdfs kerberos test

elezar:
This PR depends on #2171 and does the following:

* Moves the HDFS test process in line with the other (more stable) HDFS tests (there is still some cleanup to do, but that can be addressed as part of a follow-up PR).
* Instead of retrying keytab creation blindly (mine was failing deterministically), retry only once with a reversed principal list.

Most of the files changed come from #2171, and I was pretty careful to only edit the HDFS-related ones and `sdk_auth`.


elezar:
I'm not :/ must have forgotten to push before rugby.

testing/sdk_auth.py
```diff
@@ -297,24 +296,27 @@ def get_keytab_for_principals(self, principals: list, output_filename: str):
...
-        if result.returncode == 0:
-            log.info("There were some matches when checking the keytab for ^ZX: %s", result.stdout)
+                                shell=True,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+
+        if result.returncode is not 0:
+            log.info(result.stdout)
+            # reverse the principal list before generating again.
+            principals.reverse()
```
elezar: @benclarkwood here is the reversal. I did forget to push it last night.

elezar:
Look at that! Green HDFS Kerberos CI!

PR #2176, opened 2018-02-01
Add environment variable to opt out of log collection

elezar:
This PR adds an `INTEGRATION_TEST__SKIP_LOG_COLLECTION` which disables log collection. This can be quite verbose locally.

We would want to backport this too.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -37,6 +37,10 @@ for noise_source in [
 
 log = logging.getLogger(__name__)
 
+# The following environment variable allows for log collection to be turned off.
+# This is useful, for exampl in testing.
+INTEGRATION_TEST__SKIP_LOG_COLLECTION = os.environ.get('INTEGRATION_TEST__SKIP_LOG_COLLECTION', False)
+
 
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -37,6 +37,10 @@
 
 log = logging.getLogger(__name__)
 
+# The following environment variable allows for log collection to be turned off.
+# This is useful, for exampl in testing.
+INTEGRATION_TEST__SKIP_LOG_COLLECTION = os.environ.get('INTEGRATION_TEST__SKIP_LOG_COLLECTION', False)
```
nickbp: - How about INTEGRATION_TEST_LOG_COLLECTION that defaults to true? Having a boolean value that disables something can get confusing
- Any reason for the double underscore?
- Would probably need to do something like this, as the os.environ value will be a string if set: `bool(os.environ.get('name', 'True'))`

elezar: I didn't want to inadvertently break something in the test, and thus went for the not False approach. I'll make the change though.

The double underscore is from my time with Airflow... Thinking on it now, it's not our norm. Will change.

elezar:
@nickbp I have incorporated your changes.

PR #2177, opened 2018-02-01
INFINITY-3039: Add Cassandra TLS toggling tests.

mpereira:
This PR has sequential tests that will install Cassandra with no TLS, turn it on and then off again, while making sure clients can perform CRUD operations in each step. It also includes a test to make sure that data written in Cassandra with default configs is persisted after toggling TLS.

There is some code duplication that we might want to address after we get the rest of the TLS toggling PRs merged.


benclarkwood:
Looks good to me. Not sure what is up with that test.

mpereira:
@benclarkwood the client was trying to create an encrypted connection to the server when it shouldn't, fixed here: https://github.com/mesosphere/dcos-commons/pull/2177/commits/4a75cffee3efa0051bc3f04dc232c9008edcd33c

nickbp:
Ping for a :+1:

PR #2186, opened 2018-02-03
[KAF-405] Conditionally inject broker count

benclarkwood:
We only need the broker count when SSL Authz is enabled. This wraps the injection such that it will only be injected if:

1. Kerberos is off
2. TE is enabled
3. SSL Auth is enabled
4. Authorization is on


gabrielhartmann:
@benclarkwood: What's the harm in injecting this information unconditionally?

benclarkwood:
@gabrielhartmann the harm is that you get a rolling restart of all the brokers on any scaleup when it is only strictly necessary for TLS authz.

benclarkwood:
@elezar that's the issue, if you inject it to all tasks always, then any time you up the count you roll all the brokers. 

Rolling the brokers is only necessary for SSL authz.

PR #2188, opened 2018-02-03
 INFINITY-3074 Use templates to dedupe service docs

nickbp:
- Upstreams common content from service docs into a shared set of templates.
- Templates support specifying the package name, which should simplify publishing docs for beta vs GA releases. More to be done here in a separate PR
- Some per-service pages had effectively-common content which other services were missing. Now all services include that content.
- Some content is actually service-specific. That content is left as-is in the per-service doc files.


gabrielhartmann:
@nickbp: Sweet.  Not to be too meta, but do we have some documentation about the documentation?  i.e. Something we can reference to show how docs are generated from templates?  It might be in here somewhere already, but the PR's slightly big.

nickbp:
There are some instructions here: https://github.com/mesosphere/dcos-commons/tree/master/docs

gabrielhartmann:
Perfect.  Thanks.

PR #2191, opened 2018-02-04
Make task killing more reliable

gabrielhartmann:
There were two problems to address.

1. Constructing / passing around `TaskKiller`s was messy, error prone and inconvenient (because you had to have a `SchedulerDriver` handy).
1. Killing a task in Mesos isn't "reliable".  They may drop the message without consequence at anytime so your'e on your own. 

So now we enqueue tasks to kill and periodically attempt to kill them.  This in conjunction with the `TaskCleaner` will make it very unlikely any tasks fall through the cracks.  Note that there's no anticipated task kill latency introduced here.  We still issue kill messages as fast ever.  Only in very rare obscure scenarios could some component issue a kill before a driver is available to do the killing.  For example maybe the API server comes up fast and someone very urgently wants to restart a pod.  It may be delayed slightly while we wait to register.

~~I need to add a couple tests to address those very rare cases, but this can mostly be reviewed now.  Getting it through a run of CI will also show that I haven't broken anything major.~~
EDIT:  Tests referenced above have been added.

The last piece of the puzzle will be to trigger implicit reconciliation periodically so Mesos feeds us Tasks we may need to kill, but that's not addressed in this PR.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
@@ -1,31 +1,112 @@
 package com.mesosphere.sdk.scheduler;
 
+import org.apache.mesos.Protos;
 import org.apache.mesos.Protos.TaskID;
 import org.apache.mesos.SchedulerDriver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/**
- * This class is a default implementation of the TaskKiller interface.
- */
-public class TaskKiller {
-    private final Logger logger = LoggerFactory.getLogger(getClass());
-    private final SchedulerDriver driver;
+import java.time.Duration;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
-    public TaskKiller(SchedulerDriver driver) {
-        this.driver = driver;
+/**
+ * This class implements reliable task killing in conjunction with the {@link TaskCleaner}.  Mesos does not provide
+ * reliable task killing.  This class repeatedly attempts to kill a task until Mesos declares it has been killed or that
+ * Mesos doesn't know anything about this task.
+ */
+public final class TaskKiller {
+    private static final Logger LOGGER = LoggerFactory.getLogger(TaskKiller.class);
+
+    private static final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
+    private static final Duration killInterval = Duration.ofSeconds(5);
+    private static final Set<TaskID> tasksToKill = new HashSet<>();
+
+    private static TaskKiller taskKiller = new TaskKiller();
+    private static SchedulerDriver driver;
+    private static final Object lock = new Object();
+
+    private TaskKiller() {
+        executor.scheduleAtFixedRate(
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        // Don't acquire the lock here.  There's no harm killing tasks more than once
+                        // and holding a lock while accessing the driver always causes deadlocks.
+                        for (TaskID taskID : tasksToKill) {
+                            killTaskInternal(taskID);
+                        }
+                    }
+                },
+                killInterval.toMillis(),
+                killInterval.toMillis(),
+                TimeUnit.MILLISECONDS);
     }
 
-    public void killTask(TaskID taskId) {
+    public static void setDriver(SchedulerDriver schedulerDriver) {
+        driver = schedulerDriver;
+    }
+
+    /**
+     * Calling this method will cause the referenced TaskID to be reliably killed.  An attempt to kill the task will be
+     * immediately made.  Continued periodic attempts will be made to kill the task until Mesos indicates that it has
+     * killed the Task or doesn't recongize the task.
+     *
+     * This is still not a guarantee that a task has been killed.  Mesos may not know about a particular TaskID at any
+     * given time.  See the {@link TaskCleaner} for the rest of the task killing system.
+     * @param taskId the TaskID of the Task to be killed.
+     */
+    public static void killTask(TaskID taskId) {
         // In order to update a podinstance its normal to kill all tasks in a pod.
         // Sometimes a task hasn't been launched ever but it has been recorded for
         // resource reservation footprint reasons, and therefore doesn't have a TaskID yet.
         if (taskId.getValue().isEmpty()) {
-            logger.warn("Attempted to kill empty TaskID.");
+            LOGGER.warn("Attempted to kill empty TaskID.");
             return;
         }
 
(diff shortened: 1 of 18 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -1,31 +1,128 @@
...
+        synchronized (lock) {
+            LOGGER.info("Enqueueing kill of task: {}", taskId.getValue());
+            tasksToKill.add(taskId);
+        }
+
+        killTaskInternal(taskId);
+    }
+
+    public static void update(Protos.TaskStatus taskStatus) {
+        // The race here is fine.  If we check for the set containing an element
```
nickbp: Instead of checking for presence and then removing, could just call remove() (within the lock) and then check the boolean return value to see if the removal was successful.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -1,31 +1,128 @@
...
+        // then try to remove it after it's already gone there's no harm.  In return
+        // for this subtlety we avoid acquiring the lock on potentially many TaskStatus
+        // updates for tasks we don't care about killing.
+        if (isDead(taskStatus) && tasksToKill.contains(taskStatus.getTaskId())) {
+            synchronized (lock) {
+                tasksToKill.remove(taskStatus.getTaskId());
+                LOGGER.info("Completed killing: {}", taskStatus.getTaskId());
```
nickbp: Couldn't hurt to also log the other contents of the queue here

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -1,31 +1,128 @@
...
+            killTaskInternal(taskId);
+        }
+    }
+
+    private static void killTaskInternal(TaskID taskId) {
+        if (driver != null) {
+            LOGGER.info("Killing task: {}", taskId.getValue());
+            driver.killTask(taskId);
+        } else {
+            LOGGER.warn("Can't kill '{}' driver not yet set.", taskId);
```
nickbp: - `taskId.getValue()`
- `... '{}', driver ...`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -1,31 +1,128 @@
...
+        }
+    }
+
+    private static boolean isDead(Protos.TaskStatus taskStatus) {
```
nickbp: Upstream into TaskUtils with the other TaskState checks?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -1,31 +1,128 @@
...
+            synchronized (lock) {
+                tasksToKill.remove(taskStatus.getTaskId());
+                LOGGER.info("Completed killing: {}", taskStatus.getTaskId());
+            }
+        }
+    }
+
+    @VisibleForTesting
+    static void killAllTasks() {
+        for (TaskID taskId : tasksToKill) {
```
nickbp: Risk of tasksToKill being modified while we're iterating over the set?

gabrielhartmann: Well, it could be modified, but it's harmless.  Worst case we attempt to kill something that's already killed.  We will get a `TASK_LOST` and then ignore it as it will reference something that's already not in the set of things we're killing.  Unfortunately we can't hold a lock and access the `driver`.  One of the upsides of writing our own Java native driver will be that we can choose how it behaves and potentially avoid this requirement.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
gabrielhartmann: Maybe.  I've got two of these special ones now.  There's another over in the `TaskCleaner`.  They're a bit service specific and probably shouldn't be used by other components.  I could move them up though.  I'll think about it.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
gabrielhartmann: I wanted to avoid acquiring the lock at all if I could.  It's probably not a big deal...

gabrielhartmann: I think you're right.  I'm going to just grab the lock.  It's too finicky.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
gabrielhartmann: I guess we could grab a copy and work off of that.  That'd be a little safer.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -1,31 +1,137 @@
...
+    @VisibleForTesting
+    static void startScheduling() {
+        executor = Executors.newSingleThreadScheduledExecutor();
+        executor.scheduleAtFixedRate(
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        killAllTasks();
+                    }
+                },
```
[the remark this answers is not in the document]
gabrielhartmann: `killAllTasks` gets run a lot an debug logging is already overrun.  Current logging looks sufficient to me.  What errors/exceptions are you expecting?  We generally prefer hard failure.  If the executor goes down I'd actually like the whole scheduler to fail.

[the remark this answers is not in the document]
mpereira: > We generally prefer hard failure. If the executor goes down I'd actually like the whole scheduler to fail.

I might be missing something but it looks like if `killAllTasks` inside the runnable throws an exception the periodic action will (silently) just stop happening, so we might need some additional handling to make the whole scheduler fail.

From java docs:
> [If any execution of the task encounters an exception, subsequent executions are suppressed](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleAtFixedRate(java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit))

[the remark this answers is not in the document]
gabrielhartmann: Fair enough: https://jira.mesosphere.com/browse/INFINITY-3084

PR #2193, opened 2018-02-05
Centralize access to the Mesos SchedulerDriver

gabrielhartmann:
This PR depends on #2191 and so is WIP until that merges.

A desire for a minor improvement has driven the construction of both that PR and this one.  We wanted to start the API server as soon as possible since its startup blocks the acceptance of offers (for good reason).  However the construction of any entity which must communicate with Mesos is tied up with registration and a need to gain access to a Mesos `SchedulerDriver`.

So I made a single global way to get or set a `SchedulerDriver`.  It's in the new `Driver` class.

This allows us to:
1. Construct objects when it makes sense, independent of our access to the `SchedulerDriver`.
1. Simplify method signatures (including constructors) by not having to pass around and store `SchedulerDriver`s in many places.
1. Continue simplification of construction of the "top of the stack" `SchedulerRunner`, `AbstractScheduler`, `DefaultScheduler`, `DefaultPlanScheduler` etc. components.

This PR ultimately does allow the API server to start sooner, but the core change is how we access the `SchedulerDriver`.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/OfferAccepter.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/OfferAccepter.java
@@ -1,6 +1,7 @@
 package com.mesosphere.sdk.offer;
 
 import com.google.protobuf.TextFormat;
+import com.mesosphere.sdk.scheduler.Driver;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.mesos.Protos.Filters;
 import org.apache.mesos.Protos.Offer.Operation;
@@ -35,10 +36,16 @@ public class OfferAccepter {
         return this;
     }
 
-    public List<OfferID> accept(SchedulerDriver driver, List<OfferRecommendation> recommendations) {
+    public List<OfferID> accept(List<OfferRecommendation> recommendations) {
         if (CollectionUtils.isEmpty(recommendations)) {
             LOGGER.warn("No recommendations, nothing to do");
-            return new ArrayList<>();
+            return Collections.emptyList();
+        }
+
+        Optional<SchedulerDriver> driver = Driver.getDriver();
+        if (!driver.isPresent()) {
+            LOGGER.error("No driver present for accepting offers.  This should never happen.");
+            return Collections.emptyList();
         }
 
         List<OfferID> offerIds = getOfferIds(recommendations);
@@ -50,11 +57,11 @@ public class OfferAccepter {
             record(recommendations);
         } catch (Exception ex) {
             LOGGER.error("Failed to record Operations so not launching Task", ex);
-            return new ArrayList<>();
+            return Collections.emptyList();
         }
 
         if (CollectionUtils.isNotEmpty(operations)) {
-            driver.acceptOffers(offerIds, operations, FILTERS);
+            driver.get().acceptOffers(offerIds, operations, FILTERS);
         } else {
             LOGGER.warn("No Operations to perform.");
         }
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/OfferAccepter.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/OfferAccepter.java
@@ -44,8 +44,7 @@ public class OfferAccepter {
 
         Optional<SchedulerDriver> driver = Driver.getDriver();
         if (!driver.isPresent()) {
-            LOGGER.error("No driver present for accepting offers.  This should never happen.");
-            return Collections.emptyList();
+            throw new IllegalStateException("No driver present for accepting offers.  This should never happen.");
         }
 
         List<OfferID> offerIds = getOfferIds(recommendations);
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
@@ -16,8 +16,6 @@ import com.mesosphere.sdk.state.StateStore;
 import org.apache.mesos.Protos;
 import org.apache.mesos.Scheduler;
 import org.apache.mesos.SchedulerDriver;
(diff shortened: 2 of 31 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/OfferAccepter.java
```diff
@@ -35,10 +36,16 @@ public OfferAccepter addRecorder(OperationRecorder recorder) {
...
+    public List<OfferID> accept(List<OfferRecommendation> recommendations) {
         if (CollectionUtils.isEmpty(recommendations)) {
             LOGGER.warn("No recommendations, nothing to do");
-            return new ArrayList<>();
+            return Collections.emptyList();
+        }
+
+        Optional<SchedulerDriver> driver = Driver.getDriver();
+        if (!driver.isPresent()) {
+            LOGGER.error("No driver present for accepting offers.  This should never happen.");
```
elezar: Is logging an `error` sufficient here? We could also print a stack trace, for example -- if we don't throw and exception and exit.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -236,32 +235,20 @@ public void registered(SchedulerDriver driver, Protos.FrameworkID frameworkId, P
                 return;
             }
 
+            Driver.setDriver(driver);
+
             LOGGER.info("Registered framework with frameworkId: {}", frameworkId.getValue());
-            this.driver = driver;
             this.reviveManager = new ReviveManager(driver);
```
elezar: Is there a reason that the `ReviveManager` doesn't use the new singleton driver?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -341,6 +329,7 @@ public void statusUpdate(SchedulerDriver driver, Protos.TaskStatus status) {
             try {
                 processStatusUpdate(status);
                 reconciler.update(status);
+                TaskKiller.update(status);
```
elezar: Is there some way to register components that need to have `update` called here, instead of having them added manually?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/Driver.java
```diff
@@ -0,0 +1,29 @@
...
+public final class Driver {
+    private static SchedulerDriver driver;
+    private static Object lock = new Object();
+
+    private Driver() {
+        // Don't instantiate this class.
+    }
+
+    public static Optional<SchedulerDriver> getDriver() {
+        synchronized (lock) {
```
elezar: Most of the `getDriver` calls have explicit `isPresent()` checks after the calls to log error messages when this is not the case. Would it make sense to just `throw` instead as this seems to indicate a critical bug on our side?

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/PodResourceTest.java
```diff
@@ -407,11 +407,11 @@ public void testRestartPodManyRunning() {
         assertEquals("test-0-c", json.getJSONArray("tasks").get(2));
         assertEquals("test-0-d", json.getJSONArray("tasks").get(3));
 
-        verify(mockTaskKiller).killTask(POD_0_TASK_A.getTaskId());
-        verify(mockTaskKiller).killTask(POD_0_TASK_B.getTaskId());
-        verify(mockTaskKiller).killTask(POD_0_TASK_C.getTaskId());
-        verify(mockTaskKiller).killTask(POD_0_TASK_D.getTaskId());
-        verifyNoMoreInteractions(mockTaskKiller);
+        verify(driver).killTask(POD_0_TASK_A.getTaskId());
```
elezar: So, we are calling the `killTask` on the driver directly instead of using a `TaskKiller` now?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: No, it's an oversight.  I'm guess I missed a couple.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/Driver.java
gabrielhartmann: I wouldn't do it here.  In the places where I've written, "This should never happen." it might make sense to throw.  The reason I chose not to, is that although it shouldn't ever happen it's also possible it will self resolve...  Maybe we should crash / throw though.  `IllegalStateException` might make sense.

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/PodResourceTest.java
gabrielhartmann: No, we're verifying that ultimately the right thing happened.  Ultimately we should really kill the task.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
gabrielhartmann: Good idea, out of scope.  PRs welcome.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/Driver.java
gabrielhartmann: I'm adding throws on the client side.  That is I'm letting consumers decide whether the absence of a driver a particular point is a cause for concern, and how much concern.

elezar: Sounds good. It wasn't clear when it was fatal, and when it could be tolerated which is why I asked.

PR #2198, opened 2018-02-05
INFINITY-3073 Apply parameters to all service docs

nickbp:
- Updates the service docs to have package name/service name broken out as template parameters (and to pass those parameters into the templates). This should simplify releasing docs for both beta and GA packages, with the appropriate package name in both cases.
- Switches from snake_case to camelCase for the template params that were introduced in the previous PR:
    - This is the jekyll convention.
    - This avoids confusing IDEs because _ can mean italics.
- Updates the docs/ README to better reflect the current state of things.


docs/README.md
```diff
@@ -8,15 +8,32 @@ Docs are published automatically by CI to the `gh-pages` branch whenever changes
...
 
-## View docs locally
+## Generate and view docs locally
 
 1. Run `./generate.sh` locally from the `docs/` folder.
-2. Visit the provided `file://` URL in your browser.
+2. Visit the provided `http://` URL in your browser.
+3. When finished, exit `generate.sh` by pressing Ctrl+C. If it's stuck, then run `killall http.py` to get it unstuck.
 
 # Notes
```
benclarkwood: Are the intended uses of all the new variables documented somewhere?

docs/pages/_includes/services/limitations.md
```diff
@@ -4,7 +4,7 @@ Out-of-band configuration modifications are not supported. The service's core re
 - If a task crashes, it will be restarted with the configuration known to the scheduler, not one modified out-of-band.
 - If a configuration update is initiated, all out-of-band modifications will be overwritten during the rolling update.
 
-{{ include.config_blurb }}
+{{ include.configBlurb }}
```
benclarkwood: How are we handling all the service specific limitations?

nickbp: `frameworks/*/docs/limitations.md` has per-service limitations, this template just lists the common bits.

docs/README.md
nickbp: The template variables aren't (yet) universal, some need several while others don't need any. At this point there's somewhat of a convention around packageName/serviceName but that's about it.

PR #2204, opened 2018-02-06
Correct setDriver call to use Driver singleton

elezar:
This fixes a testing bug introduced in #2201 (as a follow-up to #2193) which causes the SDK builds to fail.

I have tested locally with `./gradlew clean build`.

The `TaskKiller` now interacts with the **global** `Driver` and does not maintain its own reference to a `SchedulerDriver`.


elezar:
@gabrielhartmann just to confirm that these changes have not been backported to 0.40?

PR #2205, opened 2018-02-06
[INFINITY-2837] readiness should be checked at startup

elezar:
This is another attempt at #1966.

The readiness check should be used to determine whether a task is at its RUNNING goal state.

If this is not taken into account, a scheduler restart would result in a RUNNING task with a failed readiness check being marked as completed.

Ideally we should use similar logic for task updates and the initial determination of the steps.

Due to the fact that we don't have reliable readiness check information on 1.9 clusters, the default behaviour for a `CustomExecutor` (used as a proxy for 1.9 clusters) is the current behaviour. This should address the failing tests on 1.9 in the previous PR.


frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
```diff
@@ -344,6 +344,150 @@ private void testWorldDecommission(boolean useDefaultExecutor) throws Exception
         runner.run(ticks);
     }
 
+
+    @Test
```
nickbp: It looks like the main differences between these tests are:
- The readiness check exit code
- The resulting status
- Whether to use custom or default executor

Given that, could break the test out into a separate parameterized function and have these three call into that function?

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
```diff
@@ -344,6 +344,150 @@ private void testWorldDecommission(boolean useDefaultExecutor) throws Exception
...
+        ServiceTestRunner runner = new ServiceTestRunner();
+        ServiceTestResult result = runner.run(ticks);
+
+        // Start a new scheduler:
+        ticks.clear();
+
+        ticks.add(Send.register());
+        ticks.add(Expect.reconciledExplicitly(result.getPersister()));
+
+        // Since the readiness check of the task did not pass, we expect it to remain in the PENDING state
```
nickbp: Comment is out of date

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
```diff
@@ -344,6 +344,150 @@ private void testWorldDecommission(boolean useDefaultExecutor) throws Exception
...
+        ServiceTestRunner runner = new ServiceTestRunner().setUseCustomExecutor();
+        ServiceTestResult result = runner.run(ticks);
+
+        // Start a new scheduler:
+        ticks.clear();
+
+        ticks.add(Send.register());
+        ticks.add(Expect.reconciledExplicitly(result.getPersister()));
+
+        // Since the readiness check of the task did not pass, we expect it to remain in the PENDING state
```
nickbp: Comment is out of date

elezar:
Thanks for your comments @nickbp. I have moved these tests to their own test class, and addressed your comments w.r.t the duplication.

PR #2207, opened 2018-02-06
Fix sidecar refinement / MOUNT volume reuse

gabrielhartmann:
### Sidecar Resource Refinement Problem
You need to do 3 things to encounter the bug fixed here.
1. Have a volume at the pod level
1. Have a sidecar task on that pod
1. Consume pre-reserved resources for that pod

`edge-lb` does all three of those things and so can't use `0.40.X` versions of the SDK.  This should fix that.  I'll confirm with them before release of `0.40.2`.

### Fix
This change can be a little bit hard to follow.  There were a couple things that needed to happen.
1. Use the `ResourceBuilder` class to build executor volumes so you get the reservation refinement stuff.
1. Pass around the `sourceRoot` field so we can build `MOUNT` volumes for Executor reuse.

### MOUNT volume problem
We have moved to a consistent construction pattern for protobufs.  That is we don't pull them out of the StateStore and reuse them _ever_.  This had problems for mount volumes in particular, because they're atomic.  The `VolumeSpec` says use `50 GB`, but the MOUNT was `100 GB`.  When reusing an Executor volume we need to indicate `100 GB` not fifty.  That information isn't in the spec.

### Fix
When matching resources and specs, grab the size info and put it in the spec.

### Note
~~The custom executor is only supported on 1.9 and previous.  1.9 and previous clusters do not support reservation refinement.  That's why the custom executor code path remains unmodified.~~
This has been manually tested on 1.9 and 1.10.  It's been tested with DSE as well.

We have to solve the generating clusters with MOUNT volumes problem to get integration tests going again.  This PR isn't going to do that.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
@@ -115,10 +115,18 @@ public class ResourceUtils {
     }
 
     public static Optional<String> getSourceRoot(Resource resource) {
-        if (resource.hasDisk() && resource.getDisk().hasSource()) {
-            return Optional.of(resource.getDisk().getSource().getMount().getRoot());
+        if (!isMountVolume(resource)) {
+            return Optional.empty();
         }
 
-        return Optional.empty();
+        return Optional.of(resource.getDisk().getSource().getMount().getRoot());
+    }
+
+    private static boolean isMountVolume(Resource resource) {
+        return resource.hasDisk()
+                && resource.getDisk().hasSource()
+                && resource.getDisk().getSource().hasType()
+                && resource.getDisk().getSource().getType()
+                .equals(Resource.DiskInfo.Source.Type.MOUNT);
     }
 }
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
@@ -531,7 +531,9 @@ public class TaskUtils {
         // Tasks with a goal state of finished should never leave the purview of their original
         // plan, so they are not the responsibility of recovery.  Recovery only applies to Tasks
         // which reached their goal state of RUNNING and then later failed.
-        if (taskSpec.getGoal().equals(GoalState.ONCE) || taskSpec.getGoal().equals(GoalState.FINISH)) {
+        if (taskSpec.getGoal().equals(GoalState.ONCE)
+                || taskSpec.getGoal().equals(GoalState.FINISH)
+                || taskSpec.getGoal().equals(GoalState.FINISHED)) {
             return false;
         } else {
             return isRecoveryNeeded(taskStatus);
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
@@ -6,7 +6,6 @@ import com.mesosphere.sdk.offer.Constants;
 import com.mesosphere.sdk.specification.PodSpec;
 import com.mesosphere.sdk.specification.ResourceSpec;
 import com.mesosphere.sdk.specification.VolumeSpec;
-import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.mesos.Protos;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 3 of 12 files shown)


nickbp:
Still WIP?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/VolumeEvaluationStage.java
```diff
@@ -58,10 +101,13 @@ public EvaluationOutcome evaluate(MesosResourcePool mesosResourcePool, PodInfoBu
...
 
+            String sourceRoot = null;
+
             Resource volume = PodInfoBuilder.getExistingExecutorVolume(
                     volumeSpec,
-                    resourceId.get(),
-                    persistenceId.get(),
+                    resourceId,
+                    persistenceId,
+                    Optional.ofNullable(sourceRoot),
```
nickbp: Always null?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -115,10 +115,18 @@ public static boolean hasResourceId(Resource resource) {
     }
 
     public static Optional<String> getSourceRoot(Resource resource) {
```
nickbp: Should VolumeEvaluationStage be calling this? Also, can't hurt to check its current callers (there could very well not be any)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -531,7 +531,9 @@ public static boolean needsRecovery(TaskSpec taskSpec, TaskStatus taskStatus) {
         // Tasks with a goal state of finished should never leave the purview of their original
         // plan, so they are not the responsibility of recovery.  Recovery only applies to Tasks
         // which reached their goal state of RUNNING and then later failed.
-        if (taskSpec.getGoal().equals(GoalState.ONCE) || taskSpec.getGoal().equals(GoalState.FINISH)) {
```
nickbp: Could have a switch statement that explicitly lists the goal states. IDEs/linters can then catch if a goal state is created and then isn't added here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
```diff
@@ -171,23 +147,40 @@ public ExecutorResourceMapper(
...
+                resourceLabels.getSourceRoot());
     }
 
     private OfferEvaluationStage newCreateEvaluationStage(ResourceSpec resourceSpec) {
-        return toEvaluationStage(resourceSpec, Optional.empty(), Optional.empty());
+        return toEvaluationStageForNew(resourceSpec);
     }
 
-    private OfferEvaluationStage toEvaluationStage(
+    private OfferEvaluationStage toEvaluationStageForExisting(
```
nickbp: Given that this is only called by newUpdateEvaluationStage(), its content could be folded into there.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
```diff
@@ -171,23 +147,40 @@ public ExecutorResourceMapper(
...
+                    resourceId,
+                    persistenceId,
+                    sourceRoot,
+                    useDefaultExecutor);
         } else {
             return new ResourceEvaluationStage(resourceSpec, resourceId, null);
         }
     }
+
+    private OfferEvaluationStage toEvaluationStageForNew(ResourceSpec resourceSpec) {
```
nickbp: Given that this is only called by newCreateEvaluationStage(), its content could be folded into there.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/VolumeEvaluationStage.java
gabrielhartmann: No, that was part of the WIP version.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
gabrielhartmann: WIP artifact.   Should be called appropriately now.

gabrielhartmann:
@nickbp: Please have another look while I write a ServiceTest for MOUNT volumes.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/VolumeEvaluationStage.java
```diff
@@ -25,24 +25,61 @@
...
         this.persistenceId = persistenceId;
+        this.sourceRoot = sourceRoot;
         this.useDefaultExecutor = useDefaultExecutor;
     }
 
     private boolean createsVolume() {
         return !persistenceId.isPresent();
     }
 
+    private static boolean isMountVolume(MesosResource mesosResource) {
```
nickbp: Duplicate of `ResourceUtils.isMountVolume()`. Might not even be used? Diff doesn't show anyone calling into this.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
```diff
@@ -17,7 +16,7 @@
  * Handles cross-referencing a preexisting {@link Protos.TaskInfo}'s current {@link Protos.Resource}s against a set
  * of expected {@link ResourceSpec}s for that task.
  */
-class TaskResourceMapper {
+public class TaskResourceMapper {
```
nickbp: Is this actually used outside of this package now? Didn't see any external usage in this change. But feel free to ignore if external usage is coming soon.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceLabels.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public ResourceLabels(ResourceSpec resourceSpec, String resourceId) {
+        this(resourceSpec, resourceSpec, resourceId);
+    }
+
+    public ResourceLabels(ResourceSpec original, ResourceSpec updated, String resourceId) {
+        this(original, updated, resourceId, Optional.empty(), Optional.empty());
+    }
+
+    public ResourceLabels(
```
nickbp: IMO could stand to have some javadocs around `original` vs `updated`.

gabrielhartmann: yep.  I'll add some.

PR #2203, opened 2018-02-06
Do not recover tasks with a goal state of FINISHED

gabrielhartmann:
I swear I fixed this once.  Maybe a merge un-fixed it.  Anyway, fixed again.

PR #2208, opened 2018-02-06
[INFINITY-3096] Do not recover tasks with a goal state of FINISHED (#2203)

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -531,7 +531,9 @@ public static boolean needsRecovery(TaskSpec taskSpec, TaskStatus taskStatus) {
         // Tasks with a goal state of finished should never leave the purview of their original
```
mrbrowning: Nit, but I find this comment pretty obtuse. Even knowing how things work, it took me a second to realize that finished tasks "should never leave the purview of their original plan" just means that the plan responsible for finished tasks is going to keep restarting the task until it reaches its goal state, and then since it's done it should never need recovery.

PR #2209, opened 2018-02-07
INFINITY-3081 Cleanup of install docs

nickbp:


docs/pages/_includes/services/install.md
```diff
@@ -27,12 +25,66 @@ All `dcos {{ include.packageName }}` CLI commands have a `--name` argument allow
...
+```
+
+### Virtual networks
+
+{{ include.techName }} supports deployment on virtual networks on DC/OS (including the `dcos` overlay network), allowing each container (task) to have its own IP address and not use port resources on the agent machines. This can be specified by passing the following configuration during installation:
+
+```json
+{
+  "service": {
+    "virtual_network_enabled": true
```
benclarkwood: @mpereira can you make sure this gets updated with the labels parameter as part of your work?

PR #2212, opened 2018-02-07
[INFINITY-2725] fix kdc soak issues

elezar:
This PR does the following:
* Adds support for rerunning `kdc.py` without affecting the current `kdc` app
* Adds support for specifying the secret name (default is still `__dcos_base64__keytab`)
* Adds support for binary secrets (`--binary-secret` command line flag)

I also used this to redeploy a keytab for INFINITY-2997.


elezar:
Thanks for the review @benclarkwood. I have made a further change where I generate a unique remote keytab name so that these are stored and can be retrieved for debugging.

PR #2214, opened 2018-02-07
[INFINITY-3113] Uninstall scheduler not using plan customizer

benclarkwood:
1. Make the uninstall scheduler use the plan customizer
2. Add a unit test to verify the customizer is used
3. Doc the altered behavior of the customizer

### DIFF
--- a/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
+++ b/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
@@ -1,8 +1,8 @@
 package com.mesosphere.sdk.helloworld.scheduler;
 
 import com.mesosphere.sdk.offer.Constants;
-import com.mesosphere.sdk.scheduler.SchedulerConfig;
-import com.mesosphere.sdk.scheduler.plan.*;
+import com.mesosphere.sdk.scheduler.plan.Plan;
+import com.mesosphere.sdk.scheduler.plan.PlanCustomizer;
 
 import java.util.Collections;
 
@@ -12,16 +12,10 @@ import java.util.Collections;
  */
 public class ReversePhasesCustomizer implements PlanCustomizer {
 
-    private final SchedulerConfig schedulerConfig;
-
-    public ReversePhasesCustomizer(SchedulerConfig schedulerConfig) {
-        this.schedulerConfig = schedulerConfig;
-    }
-
     @Override
     public Plan updatePlan(Plan plan) {
         // Definitely don't reverse the uninstall plan!
-        if (!schedulerConfig.isUninstallEnabled() && plan.getName().equals(Constants.DEPLOY_PLAN_NAME)) {
+        if (plan.getName().equals(Constants.DEPLOY_PLAN_NAME)) {
             plan.getChildren().forEach(phase -> Collections.reverse(phase.getChildren()));
         }
         return plan;
--- a/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
+++ b/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
@@ -11,10 +11,17 @@ import java.util.Collections;
  * in reverse order.
  */
 public class ReversePhasesCustomizer implements PlanCustomizer {
+
+    private final SchedulerConfig schedulerConfig;
+
+    public ReversePhasesCustomizer(SchedulerConfig schedulerConfig) {
+        this.schedulerConfig = schedulerConfig;
+    }
+
     @Override
     public Plan updatePlan(Plan plan) {
         // Definitely don't reverse the uninstall plan!
-        if (!SchedulerConfig.fromEnv().isUninstallEnabled() && plan.getName().equals(Constants.DEPLOY_PLAN_NAME)) {
+        if (!schedulerConfig.isUninstallEnabled() && plan.getName().equals(Constants.DEPLOY_PLAN_NAME)) {
             plan.getChildren().forEach(phase -> Collections.reverse(phase.getChildren()));
         }
         return plan;
--- a/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
+++ b/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
@@ -1,8 +1,7 @@
 package com.mesosphere.sdk.helloworld.scheduler;
 
 import com.mesosphere.sdk.offer.Constants;
-import com.mesosphere.sdk.scheduler.plan.Plan;
-import com.mesosphere.sdk.scheduler.plan.PlanCustomizer;
+import com.mesosphere.sdk.scheduler.plan.*;
 
 import java.util.Collections;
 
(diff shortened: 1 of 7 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/PlanCustomizer.java
```diff
@@ -4,5 +4,13 @@
...
  */
 public interface PlanCustomizer {
+
+    /**
+     * Takes a reference to a plan, returning a reference to a modified plan derived from the original.
+     *
+     * Note: ALL plans (including the uninstall deploy plan) are run through the customizer if one is declared.
+     * @param plan
+     * @return
+     */
```
nickbp: Would it make sense to just add an explicit 'Mode' enum? Modes could be DEPLOY or UNINSTALL for now, possibly another someday for e.g. UPDATE. Feels like most customizers will only be applicable to a subset of modes. Also in a module world I'd expect SchedulerConfig to be inaccessible accessible...

Could either update the existing function to have this param, or add a separate default/overrideable updatePlan(Plan, Mode) function which by default only calls the currently-defined updatePlan(Plan) function if not uninstalling (retaining cur semantics for updatePlan(Plan), if it makes sense to do that).

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
```diff
@@ -12,7 +13,8 @@
 public class ReversePhasesCustomizer implements PlanCustomizer {
     @Override
     public Plan updatePlan(Plan plan) {
-        if (plan.getName().equals(Constants.DEPLOY_PLAN_NAME)) {
+        // Definitely don't reverse the uninstall plan!
+        if (!SchedulerConfig.fromEnv().isUninstallEnabled() && plan.getName().equals(Constants.DEPLOY_PLAN_NAME)) {
```
nickbp: SchedulerConfig should be passed in as a constructor param rather than building one locally. Then a mock SchedulerConfig can be passed in for testing, or if someone had a weird custom SchedulerConfig for some other reason then that could be passed in as well.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/PlanCustomizer.java
benclarkwood: Gabriel and I had discussed out of band. I think the three options are:

0. The plan customizer will run updatePlan() against any plan, regardless of whether uninstall is happening or not. They'd then need to always gate the `deploy` plan with the `SchedulerConfig.isUninstallEnabled` flag.
1. Add a second method to the interface (e.g. updateUninstallPlan)
2. Add a `Mode` argument to the method.

I'm okay with any of these options, my only thing is making sure the user is not confused and knows what they need to do.

I hadn't considered how this would work under modularization. We'd definitely want the user to be able to see that uninstall is active, in their module. What you suggest, seems like a reasonable way to accomplish that.

nickbp: Option 1 feels like it'd be the best option. We'd probably want to also have both the functions in the `PlanCustomizer` interface then be `default` functions with noop behavior.

This would specifically make it a bit easier if we wanted to add an additional mode later. With default functions that do nothing, developers wouldn't technically need to do anything in order to be compatible, whereas with an enum they'd need to update their switch statement or similar when bumping the SDK.

benclarkwood: Oh, that is a good point.

I am going with option 1, as we need to practice being good citizens.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/PlanCustomizer.java
```diff
@@ -4,5 +4,22 @@
  * This interface allows an opportunity for plans to be modified before execution.
  */
 public interface PlanCustomizer {
+
+    /**
+     * Takes a reference to a plan, returning a reference to a modified plan derived from the original.
+     * @param plan
+     * @return
+     */
     Plan updatePlan(Plan plan);
```
nickbp: Could have a default no-op here too, if e.g. someone only wanted to customize the uninstall side.

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
```diff
@@ -10,8 +11,10 @@
  * in reverse order.
  */
 public class ReversePhasesCustomizer implements PlanCustomizer {
+
     @Override
     public Plan updatePlan(Plan plan) {
+        // Definitely don't reverse the uninstall plan!
```
nickbp: Comment is sorta redundant but ehh...

Maybe rename the class to `ReverseDeployPhasesCustomizer`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
```diff
@@ -45,16 +45,17 @@ public UninstallScheduler(
...
+        this(serviceSpec, stateStore, configStore, schedulerConfig, planCustomizer, Optional.empty());
     }
 
     protected UninstallScheduler(
             ServiceSpec serviceSpec,
             StateStore stateStore,
             ConfigStore<ServiceSpec> configStore,
             SchedulerConfig schedulerConfig,
-            Optional<SecretsClient> customSecretsClientForTests) {
+            Optional<PlanCustomizer> planCustomizer, Optional<SecretsClient> customSecretsClientForTests) {
```
nickbp: Missing newline, may as well have em all one-per-line

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallSchedulerTest.java
```diff
@@ -206,7 +199,7 @@ public void testAllButDeregisteredPlanCompletes() throws Exception {
                 new StateStore(new MemPersister()),
                 mockConfigStore,
                 SchedulerConfigTestUtils.getTestSchedulerConfig(),
-                Optional.of(mockSecretsClient));
+                Optional.empty(), Optional.of(mockSecretsClient));
```
nickbp: one per line

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallSchedulerTest.java
```diff
@@ -289,7 +300,7 @@ private UninstallScheduler getUninstallScheduler(ServiceSpec serviceSpec) {
                 stateStore,
                 mockConfigStore,
                 SchedulerConfigTestUtils.getTestSchedulerConfig(),
-                Optional.of(mockSecretsClient));
+                Optional.of(mockPlanCustomizer), Optional.of(mockSecretsClient));
```
nickbp: one per line

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/ReversePhasesCustomizer.java
benclarkwood: Oh. Comment is nonsense now. Good catch.

pires:
🎉

PR #2217, opened 2018-02-08
INFINITY-3107 Trim 'beta-' prefix for default service name in CLI

nickbp:
If `--name` (or the `<modname>.service_name` option) isn't specified, the CLI currently defaults to using the package name. However, this doesn't work for `beta-foo` packages, where the default service name is `foo`.

With this change, the CLI default behavior will align with the service name convention used by beta packages. Users interacting with default installs of beta packages will no longer be required to provide the `--name` argument. If they provide a custom name then they will need to specify the `--name` argument, as they would with a GA package.

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -63,11 +63,11 @@ func New() *kingpin.Application {
 		}
 	}
 
-	modName, err := GetModuleName()
+	var err error
+	config.ModuleName, err = GetModuleName()
 	if err != nil {
 		client.PrintMessageAndExit(err.Error())
 	}
-	config.ModuleName = modName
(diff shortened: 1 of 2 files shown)


cli/commands.go
```diff
@@ -63,11 +63,11 @@ func New() *kingpin.Application {
 		}
 	}
 
-	modName, err := GetModuleName()
+	var err error
```
benclarkwood: Little non-idiomatic to split it like this, no?

nickbp: Go absolutely refused to use `:=` against the preexisting `config.PackageName`, and also refused to use `=` without declaring `err` beforehand

PR #2219, opened 2018-02-08
Fix 404 for developer-guide link on readme

drewkerrigan:
The current developer guide link now results in a 404 not found. I've replaced with the correct link.


mpereira:
Thanks @drewkerrigan. I think a recent PR submitted by Nick fixed the 404s, but your changes avoid a redirect.

PR #2221, opened 2018-02-08
[INFINITY-3111] Gracefully handle invalid placement constraints

elezar:
This PR adds better validation support for placement constraints. Invalid constraints are wrapped in a placement rule and a validator is added to check for this.

This means that:
* If the initial deployment has an invalid placement constraint, the scheduler does not crash and one can successfully uninstall it.
* If a config update introduces an invalid constraint, deployment plan has an error as with any other invalid config change.

One question:
* I have only added an overloaded `isValid()` method to the `AndRule` and `InvalidPlacementRule`, but it would make sense to add this to all rules.

Example output of a failed update:
```
deploy (serial strategy) (ERROR)
├─ hello (serial strategy) (COMPLETE)
│  └─ hello-0:[server] (COMPLETE)
└─ world (serial strategy) (COMPLETE)
   ├─ world-0:[server] (COMPLETE)
   └─ world-1:[server] (COMPLETE)

Errors:
- Field: 'PlacementRule'; Value: 'AndRule{rules=[IsLocalRegionRule, InvalidPlacementRule{constraints=[[\"hostname\", "UNIQUE"]], exception=Invalid number of entries in rule. Expected 2 or 3, got 1: [[[\"hostname\"]}]}'; Message: 'The PlacementRule for PodSpec 'hello' had invalid constraints'
```

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/PlacementRule.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/PlacementRule.java
@@ -1,13 +1,13 @@
 package com.mesosphere.sdk.offer.evaluate.placement;
 
-import java.util.Collection;
-
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.mesosphere.sdk.offer.evaluate.EvaluationOutcome;
 import com.mesosphere.sdk.specification.PodInstance;
 import org.apache.mesos.Protos.Offer;
 import org.apache.mesos.Protos.TaskInfo;
-import com.mesosphere.sdk.offer.evaluate.EvaluationOutcome;
-import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+import java.util.Collection;
 
 /**
  * Implements placement constraint logic on an offer, filtering out the Resources which are not
@@ -50,4 +50,12 @@ public interface PlacementRule {
      * com.mesosphere.sdk.specification.TaskSpec)
      */
     boolean equals(Object o);
+
+    /**
+     * All placement rules are valid by default. This can be overridden as
+     * required.
+     */
+    default boolean isValid() {
+        return true;
+    }
 }
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/PlacementRule.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/PlacementRule.java
@@ -51,12 +51,4 @@ public interface PlacementRule {
      */
     boolean equals(Object o);
 
-    /**
-     * All placement rules are valid by default. This can be overridden as
-     * required.
-     */
-    @JsonIgnore
-    default boolean isValid() {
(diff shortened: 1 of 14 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -325,6 +323,7 @@ public boolean equals(ServiceSpec first, ServiceSpec second) {
                 DefaultVolumeSpec.class,
                 ExactMatcher.class,
                 HostnameRule.class,
+                InvalidPlacementRule.class,
```
nickbp: Given that we shouldn't be serializing configs with InvalidPlacementRule in them, is it possible to omit it from the classes available for deserialization? However if that causes other problems then leave it in.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/PlacementRule.java
```diff
@@ -46,8 +46,17 @@
...
      * com.mesosphere.sdk.specification.TaskSpec)
      */
     boolean equals(Object o);
+
+    /**
+     * All placement rules are valid by default. This can be overridden as
+     * required.
+     */
+    @JsonIgnore
+    default boolean isValid() {
```
nickbp: Instead of adding a dedicated function for checking validity, how about just having an `instanceof InvalidPlacementRule` check in the config validator? There's already some precedent for this with `instanceof PassthroughRule` in a few places. Could do something like this in the validator:

```
private static boolean isValid(PlacementRule rule) {
  if (rule instanceof OrRule) {
    return ((OrRule) rule).getRules().stream().allMatch(r -> isValid(r));
  } else if (rule instanceof AndRule) {
    return ((AndRule) rule).getRules().stream().allMatch(r -> isValid(r));
  } else {
    return !(rule instanceof InvalidPlacementRule);
  }
}
```

nickbp: Not that this is necessarily the best solution, but I feel wary about modifying the PlacementRule interface for our internal validation purposes.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
elezar: When not including this, the `ServiceTest` that I added to check this behaviour at a service level did not work - this may just be a side effect of the test or something that I am doing incorrectly.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/placement/PlacementRule.java
elezar: Yes, I wasn't 100% happy with the new method in the interface, but also don't think that the `instanceof` check is ideal (there is only a single "valid" use of it for `PassthroughRule`). I've decided to use your suggestion as this is the only place where we need to then define valid placement rules.

elezar:
@nickbp even though you've already approved, I would appreciate one last look.

PR #2222, opened 2018-02-08
[INFINITY-3133] Fix false-positive HDFS readiness check  

NimaVaziri:
The current readiness check sets some env vars based on a fetch of the HDFS jmx endpoint. These env vars are used for an equality check that determines readiness.

If HDFS is kerberized, these env vars aren't set and the equality passes regardless which provides a false-positive result as to journal nodes' true readiness.


elezar:
This PR is stale as the readiness check was changed in #2354.

PR #2225, opened 2018-02-08
INFINITY-3141 Fix redirects

nickbp:


nickbp:
Going to just merge and see if this fixes the problem with the next master build.

PR #2227, opened 2018-02-08
INFINITY-3103 Break out template parameters into per-service YAMLs

nickbp:
- Cleans up duplicated template values that were being specified across each service's respective docs sections.
- When docs are being built for docs.mesosphere.com, the respective service's YAML file may be edited in-place before rendering to customize the output. For example, `packageName: beta-kafka` => `packageName: confluent-kafka`.

More info on the data file mechanism: https://jekyllrb.com/docs/datafiles/

### DIFF
--- a/docs/pages/_data/services/kafka.yml
+++ b/docs/pages/_data/services/kafka.yml
@@ -1,0 +1,26 @@
+# Common values:
+
+packageName: beta-kafka
+serviceName: kafka
+techName: Apache Kafka
+
+# Values specific to certain templates:
+
+install:
+  minNodeCount: three
+  nodeDescription: with three brokers
+  serviceAccountInstructionsUrl: https://docs.mesosphere.com/services/kafka/kafka-auth/
+
+managing:
+  podType: kafka
+  taskType: broker
+
+supportedVersions:
+  techExampleVersion: 0.11.0.2
+  techLink: "[Apache Kafka](https://kafka.apache.org/downloads)"
+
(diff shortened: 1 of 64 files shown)


frameworks/cassandra/docs/api-reference.md
```diff
@@ -4,12 +4,7 @@ navigationTitle:
 excerpt:
 title: API Reference
 menuWeight: 70
-
-packageName: beta-cassandra
-serviceName: cassandra
 ---
+{% assign data = site.data.services.cassandra %}
```
benclarkwood: It isn't possible to centralize this assignment is it?

nickbp: Not AFAICT, the scopes appear to just be:
- site-wide
- the current page
No separate 'section' scope.

docs/pages/_data/services/kafka.yml
```diff
@@ -0,0 +1,26 @@
...
+  podType: kafka
+  taskType: broker
+
+supportedVersions:
+  techExampleVersion: 0.11.0.2
+  techLink: "[Apache Kafka](https://kafka.apache.org/downloads)"
+
+# Values specific to Kafka's own docs content:
+
+kafka:
```
mrbrowning: What's the value of templating these fields, given that they only apply to this framework?

nickbp: Was imagining that they'd need to be changed on-the-fly as well. It is pretty subjective though.

PR #2236, opened 2018-02-09
Use retrying instead of sleep to wait on goal states in larger clusters

mrbrowning:
The pod pause soak tests fail occasionally when waiting for a task to reach `PAUSED` or `RUNNING`, since the offer cycle is longer on account of the cluster's size. This moves to a smarter way of dealing with that.

### DIFF
--- a/frameworks/helloworld/tests/test_soak.py
+++ b/frameworks/helloworld/tests/test_soak.py
@@ -38,7 +38,7 @@ def get_task_info(pod_info, task_name):
 @retrying.retry(wait_fixed=1000, stop_max_delay=30 * 1000)
 def wait_for_state(state, pod_name, task_names):
     pod_status = sdk_cmd.svc_cli(
-        config.PACKAGE_NAME, config.SERVICE_NAME, 'pod status {} --json'.format(pod_name), json=True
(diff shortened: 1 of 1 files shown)


frameworks/helloworld/tests/test_soak.py
```diff
@@ -34,6 +35,17 @@ def get_task_info(pod_info, task_name):
...
 
+@retrying.retry(wait_fixed=1000, stop_max_delay=30 * 1000)
+def wait_for_state(state, pod_name, task_names):
+    pod_status = sdk_cmd.svc_cli(
+        config.PACKAGE_NAME, config.SERVICE_NAME, 'pod status {} --json'.format(pod_name), json=True
+    )
+
+    for task_name in task_names:
+        task_status = get_task_status(pod_status, '{}-{}'.format(pod_name, task_name))
+        assert task_status['status'] == state
```
elezar: Something like:
```python
task_statuses = map(lambda x: get_task_status(pod_status, '{}-{}'.format(pod_name, x)), task_names
assert all(lambda x: x['status'] == state, task_status) 
```
could make debugging a little easier as one gets a bit more context.

(note that I haven't really checked the exact syntax of the python)

PR #2237, opened 2018-02-09
INFINITY-2974: Backport binary secrets documentation update.

mpereira:


docs/pages/ops-guide/overview.md
```diff
@@ -310,27 +310,39 @@ The path of a secret defines which service IDs can have access to it. You can th
...
+
+```bash
+$ dcos security secrets create --value-file kerb5.keytab.base64-encoded some/path/__dcos_base64__mysecret
 ```
 
-When you reference the `__dcos_base64__mysecret` secret in your service, the content of the secret will be first base64-decoded, and then copied and made available to your service. Refer to the [Developer Guide](../developer-guide/) for more information on how to reference DC/OS secrets as a file in SDK-based services. Refer to a binary secret
-only as a file such that it will be autoatically decoded and made available as a temporary in-memory file mounted within your container (file-based secrets).
+**Note:** The secret name **must** be prefixed with `__dcos_base64__`.
 
+When the `some/path/__dcos_base64__mysecret` secret is [referenced in your service definition](../developer-guide.md#secrets), its base64-decoded contents will be made available as a [temporary file](http://mesos.apache.org/documentation/latest/secrets/#file-based-secrets) in your service task containers. **Note:** Make sure to only refer to binary secrets as files since holding binary content in environment variables is discouraged.
```
elezar: I assume this path is valid in the 0.40 docs too!

mpereira: Yes indeed https://github.com/mesosphere/dcos-commons/blob/sdk-0.40/docs/pages/developer-guide.md#secrets

PR #2238, opened 2018-02-09
INFINITY-3039: Add Cassandra TLS toggling tests. (#2177)

mpereira:


frameworks/cassandra/tests/test_tls.py
```diff
@@ -106,7 +106,7 @@ def test_tls_connection(cassandra_service_tls, dcos_ca_bundle):
 
         sdk_jobs.run_job(config.get_delete_data_job(dcos_ca_bundle=dcos_ca_bundle))
 
-        # Run backup plan, uploading snapshots and schema to the cloudddd
+        # Run restore plan, downloading snapshots and schema from the cloudddd
```
elezar: Not an issue in this PR, but we could correct "cloudddd" in a follow-up.

PR #2239, opened 2018-02-09
INFINITY-3147: Don't swallow exceptions so that uninstall can be retried.

mpereira:
A Kibana uninstall failed due to a package describe timeout that didn't trigger the sdk_install.uninstall retries, causing a subsequent install attempt to fail the build.

```
[2018-02-09 07:03:53,505|tests.test_sanity|INFO]: Uninstall kibana with X-Pack disabled
[2018-02-09 07:03:53,506|sdk_install|INFO]: Uninstalling/janitoring kibana
[2018-02-09 07:04:53,541|sdk_install|INFO]: Got exception when uninstalling package, continuing with janitor anyway: Error while fetching [https://52.26.246.147/package/describe]: HTTP 504: "Gateway Time-out".
```

```
    def install(
            package_name,
            service_name,
            expected_running_tasks,
            additional_options={},
            package_version=None,
            timeout_seconds=TIMEOUT_SECONDS,
            wait_for_deployment=True,
            insert_strict_options=True):
        start = time.time()
    
        # If the package is already installed at this point, fail immediately.
        if sdk_marathon.app_exists(service_name):
>           raise dcos.errors.DCOSException('Service is already installed: {}'.format(service_name))
E           dcos.errors.DCOSException: Service is already installed: kibana
```

This PR makes uninstall re-raise all exceptions. Previously only exceptions with messages containing the `marathon` string were being re-raised.

### DIFF
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -16,6 +16,7 @@ import dcos.packagemanager
 import dcos.subcommand
 import retrying
 import shakedown
+import traceback
 
 import sdk_cmd
 import sdk_marathon
@@ -167,41 +168,38 @@ def _uninstall(
     if sdk_utils.dcos_version_less_than('1.10'):
         log.info('Uninstalling/janitoring {}'.format(service_name))
         try:
-            shakedown.uninstall_package_and_wait(
-                package_name, service_name=service_name)
+            shakedown.uninstall_package_and_wait(package_name, service_name=service_name)
+
+            janitor_start = time.time()
+
+            # leading slash removed, other slashes converted to double underscores:
+            deslashed_service_name = service_name.lstrip('/').replace('/', '__')
+            if role is None:
+                role = deslashed_service_name + '-role'
+            if service_account is None:
+                service_account = service_name + '-principal'
+            if zk is None:
+                zk = 'dcos-service-' + deslashed_service_name
+            janitor_cmd = ('docker run mesosphere/janitor /janitor.py '
+                          '-r {role} -p {service_account} -z {zk} --auth_token={auth}')
+            shakedown.run_command_on_master(
+                janitor_cmd.format(
+                    role=role,
+                    service_account=service_account,
+                    zk=zk,
+                    auth=sdk_cmd.run_cli('config show core.dcos_acs_token', print_output=False).strip()))
+
+            finish = time.time()
+
+            log.info(
+                'Uninstall done after pkg({}) + janitor({}) = total({})'.format(
+                    shakedown.pretty_duration(janitor_start - start),
+                    shakedown.pretty_duration(finish - janitor_start),
+                    shakedown.pretty_duration(finish - start)))
         except (dcos.errors.DCOSException, ValueError) as e:
-            log.info('Got exception when uninstalling package, ' +
-                          'continuing with janitor anyway: {}'.format(e))
-            if 'marathon' in str(e):
-                log.info('Detected a probable marathon flake. Raising so retry will trigger.')
-                raise
-
-        janitor_start = time.time()
-
-        # leading slash removed, other slashes converted to double underscores:
-        deslashed_service_name = service_name.lstrip('/').replace('/', '__')
-        if role is None:
-            role = deslashed_service_name + '-role'
-        if service_account is None:
-            service_account = service_name + '-principal'
-        if zk is None:
-            zk = 'dcos-service-' + deslashed_service_name
-        janitor_cmd = ('docker run mesosphere/janitor /janitor.py '
-                       '-r {role} -p {service_account} -z {zk} --auth_token={auth}')
-        shakedown.run_command_on_master(
-            janitor_cmd.format(
-                role=role,
-                service_account=service_account,
-                zk=zk,
-                auth=sdk_cmd.run_cli('config show core.dcos_acs_token', print_output=False).strip()))
-
-        finish = time.time()
-
-        log.info(
-            'Uninstall done after pkg({}) + janitor({}) = total({})'.format(
-                shakedown.pretty_duration(janitor_start - start),
-                shakedown.pretty_duration(finish - janitor_start),
-                shakedown.pretty_duration(finish - start)))
+            log.info('Got exception when trying to uninstall/janitor {}'.format(service_name))
+            log.info(traceback.format_exc())
+            raise
     else:
         log.info('Uninstalling {}'.format(service_name))
         try:
@@ -228,11 +226,9 @@ def _uninstall(
             shakedown.time_wait(marathon_dropped_service, timeout_seconds=TIMEOUT_SECONDS)
 
         except (dcos.errors.DCOSException, ValueError) as e:
-            log.info(
-                'Got exception when uninstalling package: {}'.format(e))
-            if 'marathon' in str(e):
-                log.info('Detected a probable marathon flake. Raising so retry will trigger.')
-                raise
+            log.info('Got exception when trying to uninstall {}'.format(service_name))
+            log.info(traceback.format_exc())
+            raise
         finally:
             sdk_utils.list_reserved_resources()
 


testing/sdk_install.py
```diff
@@ -228,11 +226,9 @@ def marathon_dropped_service():
...
 
         except (dcos.errors.DCOSException, ValueError) as e:
-            log.info(
-                'Got exception when uninstalling package: {}'.format(e))
-            if 'marathon' in str(e):
-                log.info('Detected a probable marathon flake. Raising so retry will trigger.')
-                raise
+            log.info('Got exception when trying to uninstall {}'.format(service_name))
+            log.info(traceback.format_exc())
+            raise
```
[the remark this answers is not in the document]
mpereira: I assumed it did, behaving like Ruby (without knowing for sure). I just double-checked and it does raise the last exception https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement

testing/sdk_install.py
```diff
@@ -167,41 +168,38 @@ def _uninstall(
...
         log.info('Uninstalling/janitoring {}'.format(service_name))
         try:
-            shakedown.uninstall_package_and_wait(
-                package_name, service_name=service_name)
+            shakedown.uninstall_package_and_wait(package_name, service_name=service_name)
+
+            janitor_start = time.time()
+
+            # leading slash removed, other slashes converted to double underscores:
+            deslashed_service_name = service_name.lstrip('/').replace('/', '__')
```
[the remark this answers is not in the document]
mpereira: I don't see it
```
2 candidates:
./testing/sdk_install.py:182:        deslashed_service_name = service_name.lstrip('/').replace('/', '__')
./testing/sdk_utils.py:52:    return 'dcos-service-{}'.format(service_name.lstrip('/').replace('/', '__'))
```
but it does make sense to extract this to a function.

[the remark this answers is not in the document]
elezar: OK, I thought it was there (didn't check myself). Since this is only two places, I'll leave it up to you whether you want to extract this.

testing/sdk_install.py
```diff
@@ -167,41 +168,38 @@ def _uninstall(
     if sdk_utils.dcos_version_less_than('1.10'):
         log.info('Uninstalling/janitoring {}'.format(service_name))
         try:
-            shakedown.uninstall_package_and_wait(
-                package_name, service_name=service_name)
+            shakedown.uninstall_package_and_wait(package_name, service_name=service_name)
+
+            janitor_start = time.time()
```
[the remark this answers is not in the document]
mpereira: https://github.com/mesosphere/dcos-commons/pull/2239/commits/b5ccd4005f8faa5d2fabe3f5a4bf5d5d8910851c

testing/sdk_install.py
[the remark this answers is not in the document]
mpereira: https://github.com/mesosphere/dcos-commons/pull/2239/commits/bfe3be2df136b1e7afa0ce1af2d078c42afb747e

testing/sdk_install.py
```diff
@@ -133,9 +134,30 @@ def install(
...
 
+def run_janitor(service_name, role, service_account, znode):
+    deslashed_service_name = sdk_utils.get_deslashed_service_name(service_name)
+
+    if role is None:
+        role = deslashed_service_name + '-role'
+    if service_account is None:
+        service_account = service_name + '-principal'
+    if znode is None:
+        znode = 'dcos-service-' + deslashed_service_name
```
elezar: This should just be a call to `sdk_utils.get_zk_path`.

testing/sdk_install.py
```diff
@@ -133,9 +134,30 @@ def install(
...
+
+    cmd = ('docker run mesosphere/janitor /janitor.py '
```
elezar: For complex commands like this (with a lot of `format` specifiers), I generally prefer to use a list and then generate the command using `" ".join(cmd_list)`. For example:
```python
    auth_token = sdk_cmd.run_cli('config show core.dcos_acs_token', print_output=False).strip()

    cmd_list = ["docker", "run", "mesosphere/janitor", "/janitor.py",
                "-r", role,
                "-p", service_account,
                "-z", znode,
                "--auth_token={}".format(auth_token)]
    cmd = " ".join(cmd_list)

    shakedown.run_command_on_master(cmd)
```

testing/sdk_install.py
```diff
@@ -164,77 +186,44 @@ def _uninstall(
...
+    finish = time.time()
+
+    log.info(
+        'Uninstalled {} after pkg({}) + cleanup({}) = total({})'.format(
+            service_name,
+            shakedown.pretty_duration(cleanup_start - start),
+            shakedown.pretty_duration(finish - cleanup_start),
+            shakedown.pretty_duration(finish - start)))
+
+    sdk_utils.list_reserved_resources()
```
elezar: The fact that this is now not in a `finally` block subtly changes how this will behave in the presence of an exception.

testing/sdk_install.py
```diff
@@ -164,77 +186,44 @@ def _uninstall(
...
-
-        except (dcos.errors.DCOSException, ValueError) as e:
-            log.info(
-                'Got exception when uninstalling package: {}'.format(e))
-            if 'marathon' in str(e):
-                log.info('Detected a probable marathon flake. Raising so retry will trigger.')
-                raise
-        finally:
-            sdk_utils.list_reserved_resources()
+          sdk_marathon.wait_for_deployment_and_app_removal(
```
elezar: This indentation here is off by two.

testing/sdk_install.py
```diff
@@ -164,77 +186,44 @@ def _uninstall(
     except KeyError:
         pass # allow tests to 'uninstall' up-front
```
elezar: Could you add an extra space before the `#`?

testing/sdk_install.py
```diff
@@ -164,77 +186,44 @@ def _uninstall(
...
-        except (dcos.errors.DCOSException, ValueError) as e:
-            log.info(
-                'Got exception when uninstalling package: {}'.format(e))
-            if 'marathon' in str(e):
-                log.info('Detected a probable marathon flake. Raising so retry will trigger.')
-                raise
-        finally:
-            sdk_utils.list_reserved_resources()
+          sdk_marathon.wait_for_deployment_and_app_removal(
+              sdk_marathon.get_app_id(service_name), timeout=TIMEOUT_SECONDS)
```
elezar: Since this is only used from "within" `wait_for_deployment_and_app_removal`, we could just pass the service name and call `get_app_id` internally.

mpereira: Makes sense. It probably doesn't matter too much but do you think it makes sense to try to keep non-marathon domain concepts outside the `sdk_marathon` module though? i.e. service names VS application IDs.

testing/sdk_install.py
mpereira: Oof, good catch.

testing/sdk_install.py
elezar: That is true. Let's leave it as is then.

mpereira:
> we really should retry the individual steps as they make sense instead of relying on a global retry. Thoughts?

@elezar I think it's a good idea. One of the "drawbacks" would be having to define retry-able versions of the functions we want retried, which can get a bit verbose (if only decorators were as powerful as lisp macros...).

I noticed the other day that the `retrying` library isn't maintained anymore, and the most widely used fork, [tenacity](http://tenacity.readthedocs.io/en/latest/), supports "standalone usage":
```python
import tenacity
 
r = tenacity.Retrying(
    wait=tenacity.wait_fixed(1),
    retry=tenacity.retry_if_exception_type(IOError))

r.call(do_something)
```
which would allow us to be a bit less ceremonious with retrying functions at a higher granularity.

PR #2243, opened 2018-02-09
Fix reporting of state for overridden tasks that are not pending

mrbrowning:
Previously, if an un-prepared task was paused, the logic for returning that task's state would deem it `PAUSING`, and then `STARTING` when resumed. Since unprepared tasks have never been run and aren't slated to run, their task state in statuses should remain as `<UNKNOWN>`.


nickbp:
just in case: would make sense to backport as well

nickbp:
Unit test failure. Did something sneak in recently?

mrbrowning:
They might be caused by this change -- the unit test failure was down to the scripted scheduler tests, and I could see this changing the expected scheduler responses in some cases. Looking.

PR #2249, opened 2018-02-12
Ensure that all goal states are accounted for when recovering

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -531,13 +531,18 @@ public static boolean needsRecovery(TaskSpec taskSpec, TaskStatus taskStatus) {
         // Tasks with a goal state of finished should never leave the purview of their original
         // plan, so they are not the responsibility of recovery.  Recovery only applies to Tasks
         // which reached their goal state of RUNNING and then later failed.
-        if (taskSpec.getGoal().equals(GoalState.ONCE)
-                || taskSpec.getGoal().equals(GoalState.FINISH)
-                || taskSpec.getGoal().equals(GoalState.FINISHED)) {
-            return false;
-        } else {
-            return isRecoveryNeeded(taskStatus);
+        switch (taskSpec.getGoal()) {
```
elezar: May be overkill, but I thought we added unit tests for this? Or was that indirectly though `StateStoreUtils`?

nickbp: Looks like it's covered `StateStoreUtilsTest`. `StateStoreUtils.fetchTasksNeedingRecovery()` is the sole caller of `TaskUtils.needsRecovery()`. In theory this change should be a noop in any case.

PR #2250, opened 2018-02-12
[INFINITY-3173] Add a mount volume test for hello-world

gabrielhartmann:
This test ensures mount volumes on shared executors and on individual tasks work.  It is currently blocked by DCOS-21261.

### DIFF
--- a/frameworks/helloworld/tests/test_mount_volumes.py
+++ b/frameworks/helloworld/tests/test_mount_volumes.py
@@ -1,0 +1,89 @@
+import logging
+
+import os.path
+import pytest
+
+import sdk_cmd
+import sdk_hosts
+import sdk_install
+import sdk_plan
+import sdk_tasks
+import sdk_utils
+from tests import config
+
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope='module', autouse=True)
+def configure_package(configure_security):
+    try:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+        options = {
+            "service": {
+                "spec_file": "examples/pod-mount-volume.yml"
+            }
+        }
+
+        sdk_install.install(config.PACKAGE_NAME, config.SERVICE_NAME, 2, additional_options=options)
+
+        yield # let the test session execute
+    finally:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.mark.sanity
+def test_kill_node():
+    '''kill the node task, verify that the node task is relaunched against the same executor as before'''
+    verify_shared_executor('hello-0')
+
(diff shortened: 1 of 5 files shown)


nickbp:
Will mount volumes always be added in test clusters, or should this test be gated on whether the test cluster was configured with mount volumes?

May still want the latter even if the former is the case -- e.g. CCM clusters probably won't be getting mount volumes.

benclarkwood:
I would add a "mount_volume" mark to the test. Then a dev can easily disable it when doing manual testing.

gabrielhartmann:
@nickbp @benclarkwood: Thanks guys.  What do you suggest @mellenburg ?  How will we know if a cluster has MOUNT volumes available or not?

mellenburg:
@gabrielhartmann I actually don't have a good method for tests to check if volume mounts are enabled without SSHing into every node or making authenticated mesos requests. I would suggest disabling the test by default and then use some environment variable to enable it when you know the launch config will provide the desired volumes to support the test and then later implement the approach that asks mesos for its resources and skips based on that

gabrielhartmann:
@nickbp @benclarkwood: I spoke with @mellenburg offline.  Every `hello-world` cluster will have MOUNT volumes.  We'll figure out selective enablement later.

frameworks/helloworld/tests/test_mount_volumes.py
```diff
@@ -54,7 +102,7 @@ def test_kill_node():
 
 
 @pytest.mark.sanity
-def test_kill_agent():
+def test_kill_agent(mount_volumes):
```
gabrielhartmann: @mellenburg: Is adding this fixture to both tests going to add mount volumes twice?

frameworks/helloworld/tests/test_mount_volumes.py
```diff
@@ -1,19 +1,67 @@
...
 from tests import config
 
+import dcos_launch
+import dcos_test_utils
+
 log = logging.getLogger(__name__)
 
 
+@pytest.fixture(scope='session')
+def mount_volumes():
```
gabrielhartmann: Why is this added as a fixture in the test code?  Shouldn't it be a script called after cluster provisioning succeeds and run once to provide MOUNT volumes for the cluster?  Then tests can expect a cluster with MOUNT volumes.

frameworks/helloworld/tests/test_mount_volumes.py
mellenburg: No, session scoped dependencies mean only once per run.

frameworks/helloworld/tests/test_mount_volumes.py
mellenburg: Moved the code to `tools` and added a hook to run it in CI instead

PR #2257, opened 2018-02-13
INFINITY-2949 Add instructions for debugging uninstall

nickbp:
Describes using `force-complete` to get unstuck, with `janitor.py` as an escape hatch if there are still issues.

### DIFF
--- a/docs/pages/_includes/services/uninstall.md
+++ b/docs/pages/_includes/services/uninstall.md
@@ -29,7 +29,7 @@ Note that once the uninstall operation has begun, it cannot be cancelled because
 
 #### Debugging an uninstall
 
-In the vast majority of cases, this uninstall process goes off without a hitch. However in certain situations there can be snags along the way. For example, perhaps a machine in the cluster has permanently gone away, and the service being uninstalled had some resources allocated on that machine. This can result in the uninstall being stuck, because that machine's resources never being offered to the uninstalling scheduler. In turn, the uninstalling scheduler will not be able to successfully the resources it has on that machine.
+In the vast majority of cases, this uninstall process goes off without a hitch. However, in certain situations, there can be snags along the way. For example, perhaps a machine in the cluster has permanently gone away, and the service being uninstalled had some resources allocated on that machine. This can result in the uninstall becoming stuck, because Mesos will never offer those resources to the uninstalling scheduler. As such, the uninstalling scheduler will not be able to successfully unreserve the resources it had reserved on that machine.
 
 This situation is indicated by looking at `deploy` plan while the uninstall is proceeding. The `deploy` plan may be viewed using either of the following methods:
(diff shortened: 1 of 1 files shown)


docs/pages/_includes/services/uninstall.md
```diff
@@ -14,6 +14,85 @@ For example, to uninstall the {{ include.data.techName }} instance named `{{ inc
...
+  - As the task resources are offered by Mesos, they are unreserved by the scheduler.
+    - **Warning**: Any data stored in reserved disk resources will be irretrievably lost.
+  - Once all known resources have been unreserved, the scheduler's persistent state in ZooKeeper is deleted.
+- The cluster automatically removes the scheduler task once it advertises the completion of the uninstall process.
+
+Note that once the uninstall operation has begun, it cannot be cancelled because it can leave the service in an uncertain, half-destroyed state.
+
+#### Debugging an uninstall
+
+In the vast majority of cases, this uninstall process goes off without a hitch. However in certain situations there can be snags along the way. For example, perhaps a machine in the cluster has permanently gone away, and the service being uninstalled had some resources allocated on that machine. This can result in the uninstall being stuck, because that machine's resources never being offered to the uninstalling scheduler. In turn, the uninstalling scheduler will not be able to successfully the resources it has on that machine.
```
benclarkwood: "However, in certain situations, there can be snags along the way"

benclarkwood: "This can result in the uninstall becoming stuck because Mesos will never offers those resources to the uninstalling scheduler. As such, the uninstalling scheduler will not be able to successfully unreserve the resources it had reserved on that machine."

docs/pages/_includes/services/uninstall.md
```diff
@@ -14,6 +14,85 @@ For example, to uninstall the {{ include.data.techName }} instance named `{{ inc
...
+│  ├─ unreserve-5dd56b1d-4522-4bbd-88b5-de9fa0f181f2 (COMPLETE)
+│  └─ unreserve-c9915f07-f446-4e14-a6b4-12c8dd2f914b (COMPLETE)
+└─ deregister-service (PENDING)
+   └─ deregister (PENDING)
+```
+
+#### Manual uninstall
+
+If all else fails, one can simply manually perform the uninstall themselves. To do this, perform the following steps:
+- Delete the uninstalling scheduler from Marathon
```
benclarkwood: Do you need to do a service shutdown too?

benclarkwood: I believe otherwise you'll leak an inactive framework.

nickbp: Yes, you would. Added instructions for determining the framework id and unregistering the framework

docs/pages/_includes/services/uninstall.md
nickbp: Done

PR #2259, opened 2018-02-13
Also accept $FRAMEWORK_HOST as a valid URI in airgap_linter (#2256)

benclarkwood:


tools/airgap_linter.py
```diff
@@ -53,6 +53,7 @@ def is_bad_uri(uri, file_name):
         ".mesos:",
         "$MESOS_CONTAINER_IP",
         "{{FRAMEWORK_HOST}}",
+        "$FRAMEWORK_HOST",
```
elezar: Looking at how these are matched to the URIs, that means that the following:
* `$MESOS_CONTAINER_IP_BUT_NOT_REALLY`
* `$FRAMEWORK_HOST_BUT_ALSO_NOT_QUITE`
will also be accepted as valid URIs.

Also, in terms of environment variables: `${FRAMEWORK_HOST}` and `${MESOS_CONTAINER_IP}` are also valid usages of environment variables.

benclarkwood: That is a good point. I will take a pass at writing a better URL evaluator.

I will add both of those alternate versions.

benclarkwood: (in a follow up)

elezar: Ok. That's fine.

PR #2261, opened 2018-02-13
[INFINITY-3174] Add alternate form of accepted envvars in airgap_linter.

benclarkwood:
This adds `${FRAMEWORK_HOST}` and `${MESOS_CONTAINER_IP}` alongside `$FRAMEWORK_HOST` and `$MESOS_CONTAINER_IP` as accepted envvars in the airgap linter.

Note: I'm not going to backport this as no downstreams require this change.


benclarkwood:
I think I was mentally deferring that to when we overhaul tooling.

benclarkwood:
(I don't think the airgap linter has to be perfect)

PR #2262, opened 2018-02-14
Add UPGRADES_FROM envvar to release builder to allow upgrade/downgrade in soak

mrbrowning:
Manually verified that this has the expected effect on the ouput package JSON.


benclarkwood:
@mrbrowning, so if I build a stub universe locally, what is its upgradesFrom/downgradesTo set to?

mrbrowning:
@benclarkwood The current behavior is to set `upgradesFrom` and `downgradesTo` in stub universes to `*`, which I think is appropriate given all of the variables in a development environment.

benclarkwood:
Ah cool, I think that had been broken previously.

PR #2263, opened 2018-02-14
Update HDFS docs to reflect correct minimum node count and DC/OS Vagrant mentions

mrbrowning:


docs/pages/_data/services/hdfs.yml
```diff
@@ -7,7 +7,7 @@ techName: HDFS
 # Values specific to certain templates:
 
 install:
-  minNodeCount: five
+  minNodeCount: three
```
nickbp: From slack, sounds like it's indeed five?

elezar: Does it not depend on the size of the nodes? I'm happy to make it 5 as long as the pre-install notes are then updated accordingly.

mrbrowning:
@elezar it looks like the current pre-install notes are correct: `Default configuration requires 5 agent nodes ...`

PR #2264, opened 2018-02-14
[INFINITY-3035] Add test for toggling Kafka security settings

elezar:
This is a backport of #2137 and #2268.

[INFINITY-3035] Add test and documentation to toggle kafka security settings


elezar:
This should not yet be merged as this will cause the nightly tests to fail on 1.9 (see #2268)

PR #2265, opened 2018-02-14
[INFINITY-3088] Refactor plan customization to guarantee all plans can be customizable

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -262,6 +267,7 @@ public void registered(SchedulerDriver driver, Protos.FrameworkID frameworkId, P
             isInitialized.set(true);
         }
 
+
```
benclarkwood: Extra newline. Not worth fixing now but nuke in the future.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -178,21 +190,22 @@ public AbstractScheduler setOfferQueueSize(int queueSize) {
...
 
     /**
      * Returns a list of API resources to be served by the scheduler to the local cluster.
-     * This may be called before {@link #getPlanCoordinator()} has been called.
      */
     public abstract Collection<Object> getResources();
 
     /**
-     * Returns the {@link PlanCoordinator}.
+     * Provides a callback to indicate that the scheduler has registered with Mesos.
```
benclarkwood: Doc line is in wrong spot

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -178,21 +190,22 @@ public AbstractScheduler setOfferQueueSize(int queueSize) {
...
      */
-    protected abstract PlanCoordinator getPlanCoordinator() throws Exception;
+    protected abstract PlanCoordinator getPlanCoordinator();
+
+    protected abstract void registeredWithMesos();
```
nickbp: Does anything call this? I see declarations, but no calls

gabrielhartmann: Thanks.  Good catch.  I lost that.

pires:
@gabrielhartmann what's the status on this, will it be merged or not? We still need some changes like a few `DefaultScheduler` methods with modified visibility, from private to protected so we can extend them.

frameworks/elastic/tests/test_sanity.py
```diff
@@ -76,6 +76,8 @@ def test_pod_replace_then_immediate_config_update():
 
     # ensure all nodes, especially data-0, get launched with the updated config
     config.check_plugin_installed(plugin_name, service_name=foldered_name)
+    sdk_plan.wait_for_completed_deployment(foldered_name)
```
elezar: Should this not also wait for an in-progress deployment? (This was required for the toggle tests here: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/kafka/tests/test_toggle_security.py#L535)

gabrielhartmann: No.  The Marathon app has been updated.  Assuming the update isn't totally broken it will never accidentally be complete.  In any case it never checked before.  It just fell through.  If it becomes an actual problem we can fix in a different PR.

PR #2273, opened 2018-02-16
[INFINITY-2835] Allow package name to be overridden with an envvar

elezar:
This PR adds support for overriding the package name using an environment variable `PACKAGE_NAME`.

The default behaviour (in terms of getting the package name from the stub universe URL) remains the same.

### DIFF
--- a/tools/release_builder.py
+++ b/tools/release_builder.py
@@ -24,6 +24,51 @@ universe_converter_url_prefix = 'https://universe-converter.mesosphere.com/trans
 
 class UniverseReleaseBuilder(object):
 
+    @staticmethod
+    def get_package_name(stub_universe_url: str) -> str:
+
+        package_name = os.environ.get('PACKAGE_NAME', '')
+
+        log.info("Read PACKAGE_NAME=%s", package_name)
+
+        if not package_name:
+            log.info("Getting package name from stub universe URL")
+
+            name_match = re.match('.+/stub-universe-(.+).(json)$', stub_universe_url)
+            if not name_match:
+                raise Exception('Unable to extract package name from stub universe URL. '
+                                'Expected filename of form "stub-universe-[pkgname].json"')
+
+            package_name = name_match.group(1)
+            log.info("Got package name %s from stub universe URL")
+
+        return package_name
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


tools/release_builder.py
```diff
@@ -38,53 +89,18 @@ def __init__(
         self._force_upload = os.environ.get('FORCE_ARTIFACT_UPLOAD', '').lower() == 'true'
         self._beta_release = beta_release.lower() == 'true'
 
-        name_match = re.match('.+/stub-universe-(.+).(json)$', stub_universe_url)
-        if not name_match:
-            raise Exception('Unable to extract package name from stub universe URL. ' +
-                            'Expected filename of form "stub-universe-[pkgname].json"')
+        self._stub_universe_pkg_name = self.get_package_name(stub_universe_url)
```
elezar: @nickbp this line only seems to match `json` files? I may have missed where the `zip` files where referenced.

elezar:
With regards to
> Now that things have been restructured and support for the zip format has been dropped, would it be straightforward to instead just get the package name value from the json content directly? To be clear I haven't looked at how this retrieval would fit in the current script structure, so there could be some aspect that makes it difficult to do it this way. In any case, I think the package name should just be a name field at the root of the stub universe object.

The use case that prompted this ticket to be required was `confluent-kafka` -- which could be seen as a modified build of the `kafka` package. This meant that we took as input a stub universe with `"name"=="kafka"` and wanted to generated a package with `"name"=="confluent-kafka"` which was not possible if the package name was determined from the stub universe name **or** the JSON contents.

I would be happy to add support for extracting the info from the package itself, but would prefer to make the `PACKAGE_NAME` as input parameter a requirement as this reduces the amount of magic.

benclarkwood:
To me, it makes sense to accept these changes as-is as it is an _override_ of the default behavior (and a useful one at that).

I think a follow-on change is to modify it to introspect the package.json rather than the file name by default, but leave the override in place.

elezar:
Thanks @benclarkwood and @nickbp for your reviews. I have created https://jira.mesosphere.com/browse/INFINITY-3230 to track using the package contents instead of the filename.

PR #2295, opened 2018-02-22
INFINITY-3248: Add note that custom auxiliary plans don't require additional resources.

mpereira:
Prompted by this questions: https://mesosphere.slack.com/archives/C07PM5KHQ/p1519298500000094

This PR also fixes a badly formatted markdown nested list.

### DIFF
--- a/docs/pages/developer-guide.md
+++ b/docs/pages/developer-guide.md
@@ -1777,7 +1777,7 @@ pods:
 
 In this case, the resources are declared separately from the server task in a resource set named `hello-resources`. They are referenced by a `resource-set` element in the task definition. A task continues to be defined as the combination of a process to run and resources to consume. This alternate formulation provides you with increased flexibility: you can now define multiple processes that can consume the same resources.
 
-**Note:** Resource sets are reserved at scheduler startup. [Custom auxiliary plans](#custom-auxiliary-plans) that run tasks that make use resource sets won't require additional resources than what the scheduler already reserved.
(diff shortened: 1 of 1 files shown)


docs/pages/developer-guide.md
```diff
@@ -1775,7 +1775,9 @@ pods:
         resource-set: hello-resources/
 ```
 
-In this case, the resources are declared separately from the server task in a resource set named `hello-resources`. They are referenced by a `resource-set` element in the task definition. A task continues to be defined as the combination of a process to run and resources to consume. This alternate formulation provides you with increased  flexibility:  you can now define multiple processes that can consume the same resources.
+In this case, the resources are declared separately from the server task in a resource set named `hello-resources`. They are referenced by a `resource-set` element in the task definition. A task continues to be defined as the combination of a process to run and resources to consume. This alternate formulation provides you with increased flexibility: you can now define multiple processes that can consume the same resources.
+
+**Note:** Resource sets are reserved at scheduler startup. [Custom auxiliary plans](#custom-auxiliary-plans) that run tasks that make use resource sets won't require additional resources than what the scheduler already reserved.
```
elezar: In terms of style, there seem to be a lot of "that"s in the second sentence.

mpereira: What do you think about
> Custom auxiliary plans that run tasks referencing resource sets won't require [...]

elezar: Sounds good 🎵

PR #2297, opened 2018-02-22
[INFINITY-3255] Cleanup HDFS client docker image

NimaVaziri:
This PR has been somewhat overdue and was triggered by the JIRA to add cURL to the docker image as well as integrate the Dockerfile for said image to the repo.

This just cleans things up, and aggregates all the pieces to the `tools/docker-client` folder.

To note: besides having to perform the `kinit` command, when the client comes up it actually doesn't care itself about whether the HDFS it's talking to is kerberized or not.


frameworks/hdfs/tools/docker-client/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM ubuntu:16.04
+
+RUN apt-get -y update && \
```
benclarkwood: How much size is saved by condensing these three `RUN`s into a single `RUN`?

frameworks/hdfs/tools/docker-client/krb5.conf
```diff
@@ -0,0 +1,121 @@
...
+                }
+        }
+        fcc-mit-ticketflags = true
+
+[realms]
+        LOCAL = {
+                kdc = KDC_ADDRESS
+                admin_server = KDC_ADDRESS
+        }
+        ATHENA.MIT.EDU = {
```
benclarkwood: Can we drop all these extraneous realms?

frameworks/hdfs/tools/docker-client/Dockerfile
benclarkwood: #layers

NimaVaziri: It's not really in consideration of size. It's just best practice to have `update` in the same layer as other `apt-get` runs so as to not cache the `update` layer and reference older packages if at a later the following `apt-get`s are modified in their own layers.

frameworks/hdfs/tools/docker-client/krb5.conf
NimaVaziri: Yeah, hoping these aren't required for our intentions for some non-obvious reasons...

frameworks/hdfs/tools/docker-client/Dockerfile
elezar: Would it not be better to run all `apt` commands once? Finishing off with a cleanup? See https://github.com/mesosphere/dcos-tensorflow/blob/master/images/Dockerfile#L10

frameworks/hdfs/tools/docker-client/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN apt-get update && \
+    apt-get install -y default-jre && \
+    apt-get install -y \
+    curl \
+    dnsutils \
+    less \
+    vim
+
+ENV HADOOP_VERSION hadoop-2.6.0-cdh5.9.1
```
elezar: Let's use this to construct the HDFS download URL so that we don't have to change it in multiple places.

frameworks/hdfs/tools/docker-client/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+# the env var is required for a non-interactive install of krb5-user
+RUN export DEBIAN_FRONTEND=noninteractive && apt-get install -y krb5-user
+COPY krb5.conf /etc/krb5.conf
+RUN chown -R nobody /etc
+
+COPY configure-environment.sh /hadoop-2.6.0-cdh5.9.1/configure-environment.sh
+COPY configure-hdfs.sh /hadoop-2.6.0-cdh5.9.1/configure-hdfs.sh
+
+# working directory will be the HDFS distribution
+RUN chown -R nobody /hadoop-2.6.0-cdh5.9.1
```
elezar: So this isn't going to work on Centos?

frameworks/hdfs/tools/docker-client/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    less \
+    vim
+
+ENV HADOOP_VERSION hadoop-2.6.0-cdh5.9.1
+ENV HDFS_SERVICE_NAME hdfs
+ENV JAVA_HOME /usr/lib/jvm/default-java
+
+# the env var is required for a non-interactive install of krb5-user
+RUN export DEBIAN_FRONTEND=noninteractive && apt-get install -y krb5-user
+COPY krb5.conf /etc/krb5.conf
```
elezar: We could drop this copy and have a "bring your own krb5.conf" approach as is used in (some?) of the integration tests.

frameworks/hdfs/tools/docker-client/configure-environment.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+export PATH=$PATH:/usr/local/sbin:/usr/sbin:/sbin
+
+sed -e "s/KDC_ADDRESS/$KDC_ADDRESS/g" -i /etc/krb5.conf
```
elezar: I would remove the `krb5.conf` file and require that the user supply one?

frameworks/hdfs/tools/docker-client/configure-hdfs.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+usage() {
+    echo "HDFS_SERVICE_NAME=hdfs /$HADOOP_VERSION/configure-hdfs.sh"
+    exit 1
+}
+
+[[ -n "${HDFS_SERVICE_NAME}" ]] || usage
+
+cd /$HADOOP_VERSION
```
elezar: It's out of scope of this PR, but it's on my mind since we are considering the same thing in TensorFlow. Would it not make sense to rather accept the full URL to the where `core-site.xml` and `hdfs-site.xml` could be found.

This is the approach used by Spark (https://docs.mesosphere.com/services/spark/2.1.0-2.2.1-1/hdfs/) and it would be useful to have consistency.

frameworks/hdfs/tools/docker-client/krb5.conf
```diff
@@ -0,0 +1,21 @@
+[libdefaults]
```
elezar: In the tests that I did (including active directory), one can get away with the following minimal `krb5.conf`:
```
[libdefaults]
default_realm = MY.REALM

[realms]
  MY.REALM = {
    kdc = KDC_ADDRESS
  }
```
(see https://github.com/mesosphere/dcos-commons/blob/master/testing/security/kerberos.py#L33)

As already discussed, however, we could consider just dropping it.

frameworks/hdfs/tests/config.py
```diff
@@ -103,7 +103,7 @@ def run_hdfs_command(service_name, command):
     """
     Execute the command using the Docker client
     """
-    full_command = 'docker run -e HDFS_SERVICE_NAME={service_name} {image_name} /bin/bash -c "/configure-hdfs.sh && {cmd}"'.format(
+    full_command = 'docker run -e HDFS_SERVICE_NAME={service_name} {image_name} /bin/bash -c "hadoop-*/configure-hdfs.sh && {cmd}"'.format(
```
elezar: We could have left `configure-hdfs.sh` in the root and then this `*` here would not be required.

NimaVaziri: Yeah, I don't have a strong preference for either way, just thought having everything in a specific, non-root working directory is cleaner.

frameworks/hdfs/tools/docker-client/Dockerfile
NimaVaziri: Upon further thought, will remove this.

So since this is simply a client we use for testing correctness of the service, running the Marathon client app as `root` seems fine to cover both CentOS and CoreOS.

frameworks/hdfs/tools/docker-client/Dockerfile
```diff
@@ -1,37 +1,39 @@
...
+RUN wget https://downloads.mesosphere.com/hdfs/assets/$HADOOP_VERSION-dcos.tar.gz && \
+    tar xf $HADOOP_VERSION-dcos.tar.gz && \
+    rm $HADOOP_VERSION-dcos.tar.gz
+
+
 # the env var is required for a non-interactive install of krb5-user
-RUN export DEBIAN_FRONTEND=noninteractive && apt-get install -y krb5-user
-COPY krb5.conf /etc/krb5.conf
-RUN chown -R nobody /etc
+RUN export DEBIAN_FRONTEND=noninteractive && \
```
elezar: Any reason to not also include `krb5-user` in the `apt-step` above?

frameworks/hdfs/tools/docker-client/Dockerfile
```diff
@@ -1,37 +1,39 @@
...
+    apt-get update && \
+    apt-get install -y krb5-user && \
+    apt-get clean
 
-COPY configure-environment.sh /hadoop-2.6.0-cdh5.9.1/configure-environment.sh
-COPY configure-hdfs.sh /hadoop-2.6.0-cdh5.9.1/configure-hdfs.sh
+COPY krb5.conf /etc/krb5.conf
```
elezar: Would the default `krb5.conf` that comes with `krb5-user` not be sufficient?

frameworks/hdfs/tools/docker-client/Dockerfile
```diff
@@ -1,37 +1,39 @@
...
-# working directory will be the HDFS distribution
-RUN chown -R nobody /hadoop-2.6.0-cdh5.9.1
+COPY configure-environment.sh /$HADOOP_VERSION/configure-environment.sh
+COPY configure-hdfs.sh /$HADOOP_VERSION/configure-hdfs.sh
 
-WORKDIR /hadoop-2.6.0-cdh5.9.1
+WORKDIR /$HADOOP_VERSION
 
 # this script will setup the HDFS configuration regardless of HDFS being kerberized or not
 CMD "./configure-environment.sh"
```
elezar: If we specify the full path here, the image would also work when the user specifies a `-w` flag to the docker command?

frameworks/hdfs/tools/docker-client/krb5.conf
```diff
@@ -1,19 +1,6 @@
...
-                }
-                plain = {
-                        something = something-else
-                }
-        }
-        fcc-mit-ticketflags = true
-
 [realms]
         LOCAL = {
                 kdc = KDC_ADDRESS
```
elezar: Is the `admin_server` section required?

frameworks/hdfs/tools/docker-client/Dockerfile
NimaVaziri: Ah so I tried that the first time and it didn't work for some reason. Tried it again and seems to be fine. Will just include it above.

frameworks/hdfs/tools/docker-client/Dockerfile
NimaVaziri: Makes sense, prefixed `CMD` with hadoop directory

frameworks/hdfs/tools/docker-client/krb5.conf
NimaVaziri: Not necessarily, removing it as well

frameworks/hdfs/tools/docker-client/Dockerfile
NimaVaziri: It has all the correct fields (and some additional ones that aren't needed since we've simplified our krb5 file to include only fields we need).

The default values of the fields are awkward to correct without writing some wrapper scripts specifically for this manipulation (which is more work than needed).

frameworks/hdfs/tools/docker-client/Dockerfile
elezar: You don't need to export it as it is only required for the `apt-get install` command. For example from https://github.com/elezar/tf_workspace/blob/master/Dockerfile.cpu#L4:
```
RUN apt-get update && \
    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
        git \
        tree \
        wget \
        zsh \
    && \
```

frameworks/hdfs/tools/docker-client/configure-environment.sh
elezar: As also mentioned in #2316 we could consider using mustache here instead? Then we have consistency across how we generate these files in production.

frameworks/hdfs/tests/config.py
elezar: Note that #2327 was created to address this script having moved.

frameworks/hdfs/tools/docker-client/configure-environment.sh
NimaVaziri: Yeah we could do that, just that we'd have to add bootstrap to the client bundle as well.

By production you mean our CI? We don't use this client anywhere else, right?

elezar: By production in this case I mean the actual HDFS service. With this kind of replace, we effectively have two kinds of templating.

We would not strictly be limited to adding `bootstrap` to the client image, as we would be able to use any mustache templating library, but bootstrap would give use some other tools like DNS resolution for free.

PR #2300, opened 2018-02-23
[INFINITY-3202] Remove partition tests, save one.

benclarkwood:
This removes all partition tests, as they were mostly just testing Mesos/DCOS, and not well at that.

It does retain one test that _is_ useful, but moves it back into the recovery suite (the test is disabled at present).


frameworks/helloworld/tests/test_zzzrecovery.py
```diff
@@ -329,3 +329,25 @@ def test_shutdown_host():
         log.info('Checking affected task has moved to a new agent:\n'
                  'old={}\nnew={}'.format(replaced_task, new_task))
         assert replaced_task.agent != new_task.agent
+
```
elezar: Note the `@@@@@` warning above ...

PR #2310, opened 2018-02-27
[INFINITY-3080] Use fine-grained permissions for DC/OS 1.11.

elezar:
This PR restructures how service accounts are created in the setup process, and adds fine-grained settings for DC/OS 1.11.

For clarity the effective changes are as follows:
* Use the same service account fixture for *all* TLS tests requiring a service account.
* On 1.11 clusters use fine-grained permissions instead of `superuser` for this service account.
* Use return a dict from the fixture instead of a string so as to allow for latter setting different secrets and service account names.

What has not changes:
* The logic in calling `create_service_account` and `delete_service_account`
* The names of the service account and secrets created (with the exception of one Kafka test where service account name and service account secret are now the same for uniformity).
* The setup of service accounts and secrets in strict mode.

### DIFF
--- a/frameworks/kafka/tests/test_tls.py
+++ b/frameworks/kafka/tests/test_tls.py
@@ -1,12 +1,12 @@
 import pytest
-import shakedown
 
 import sdk_cmd
 import sdk_install
 import sdk_networks
 import sdk_plan
-import sdk_security
 import sdk_utils
+
+from security import transport_encryption
 
 from tests import config
 
@@ -17,19 +17,15 @@ BROKER_TLS_ENDPOINT = 'broker-tls'
 @pytest.fixture(scope='module')
 def service_account(configure_security):
     """
-    Creates service account and yields the name.
+    Creates service account for TLS.
     """
     try:
         name = config.SERVICE_NAME
-        sdk_security.create_service_account(
-            service_account_name=name, service_account_secret=name)
-        # TODO(mh): Fine grained permissions needs to be addressed in DCOS-16475
-        sdk_cmd.run_cli(
-            "security org groups add_user superusers {name}".format(name=name))
-        yield name
+        service_account_info = transport_encryption.setup_service_account(name)
+
+        yield service_account_info
     finally:
-        sdk_security.delete_service_account(
-            service_account_name=name, service_account_secret=name)
+        transport_encryption.cleanup_service_account(service_account_info)
 
 
 @pytest.fixture(scope='module')
@@ -42,8 +38,8 @@ def kafka_service_tls(service_account):
             config.DEFAULT_BROKER_COUNT,
             additional_options={
                 "service": {
-                    "service_account": service_account,
-                    "service_account_secret": service_account,
+                    "service_account": service_account["name"],
+                    "service_account_secret": service_account["secret"],
                     "security": {
                         "transport_encryption": {
                             "enabled": True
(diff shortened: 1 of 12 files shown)


testing/security/transport_encryption.py
```diff
@@ -6,11 +6,63 @@
...
 
 
+def setup_service_account(service_name: str,
+                          service_account_secret: str=None) -> dict:
+    """
+    Setup the service account for TLS
+    """
+
+    if sdk_utils.is_open_dcos():
+        log.warn("The setup of a service account requires DC/OS EE")
```
benclarkwood: I'd throw an exception here and crash the test.

testing/security/transport_encryption.py
```diff
@@ -6,11 +6,63 @@
...
+        return {}
+
+    name = service_name
+    secret = name if service_account_secret is None else service_account_secret
+    sdk_security.create_service_account(service_account_name=name,
```
benclarkwood: Does this simply ensure the service account exists? It will already exist in the strict case, yes?

frameworks/kafka/tests/test_tls.py
```diff
@@ -17,19 +17,15 @@
 @pytest.fixture(scope='module')
 def service_account(configure_security):
     """
-    Creates service account and yields the name.
+    Creates service account for TLS.
```
benclarkwood: Creates or _updates_?

testing/security/transport_encryption.py
```diff
@@ -6,11 +6,63 @@
...
+                        name, acl["rid"], acl["action"]
+                        ]
+
+            output = sdk_cmd.run_cli(" ".join(cmd_list))
+            log.info("output=%s", output)
+
+    return {"name": name, "secret": secret}
+
+
+def cleanup_service_account(service_account_info: dict):
```
benclarkwood: Won't this delete the service account in the middle of the strict suite?

elezar: This is only called from within the `service_account` fixture for TLS tests (I didn't want to do a more large-scale refactor here). The only difference between the new behaviour and the existing behaviour is that we are not calling `delete_service_account` directly from the fixture.

What may be happening is that the `configure_security` fixture is using a different service account (`'service-acct'`) and such we have not picked it up until now. I will investigate further tomorrow and create a JIRA if applicable.

testing/security/transport_encryption.py
elezar: The behaviour is to first delete the service account if it exists:
```
log.info('Remove any existing service account and/or secret')
    delete_service_account(service_account_name, service_account_secret)
```
I think this is also related to the fact that in the strict case, we're using a different service account (`'service-acct'`) for the non-TLS tests. Ad discussed with the deletion, I will dig a bit further tomorrow.

frameworks/kafka/tests/test_tls.py
elezar: So, since `create_service_account` deletes an existing account -- this may beed to be updated.

What about `"Sets up the service account for the secure tests"`?

testing/security/transport_encryption.py
benclarkwood: Since we were using `superuser` before, it just "worked" in strict.

We'll need to make sure that whatever SA we use has the right permissions to run in strict.

frameworks/kafka/tests/test_tls.py
benclarkwood: Ah right. No the original wording is fine...

elezar:
@benclarkwood thanks for the comments. You raised some good points, and my suggestion would be to address the minor comments (doc-strings and throwing on failure) here and then log the work on improving the way we setup security (I think there may even be a JIRA for that) to address in a later PR. I have (or will shortly) update the effective changes of this PR to make it clearer as to what the behaviour should be.

Alternatively, we could rip apart the security now and ensure consistency across the board. (although this is fun, I would rather avoid it as I still have to move these changes to ZK and DSE).

testing/security/transport_encryption.py
elezar: Yes, just thought of that as well ... will check tomorrow.

benclarkwood:
Right, so we should definitely try it in strict mode before we check in any changes :)

elezar:
@benclarkwood any objections to merging this now?

All the nightlies looked sane as well: https://teamcity.mesosphere.io/project.html?projectId=DcosIo_DcosCommons_Nightly&branch_DcosIo_DcosCommons_Nightly=INFINITY-3080-use-fine-grained-permissions-for-tls

benclarkwood:
Nope!

PR #2313, opened 2018-02-28
[INFINITY-3280] Regression in advertised listeners behavior

benclarkwood:
Only advertise IPs when security is not enabled.

This is consistent with the current behavior in 2.0.X.

### DIFF
--- a/frameworks/kafka/setup-helper/setup_helper.go
+++ b/frameworks/kafka/setup-helper/setup_helper.go
@@ -175,12 +175,11 @@ func setListeners() error {
 	// NOTE: To be consistent with the legacy behavior of the 2.0.X Kafka series,
 	// when there is no security enabled, we must honor the kafka.kafka_advertise_host_ip
 	// configuration parameter
-	if !kerberosEnabled && !tlsEncryptionEnabled && !getBooleanEnvvar(advertiseHostIPEnvvar) {
-		// Write an empty file.
-		err = writeToWorkingDirectory(advertisedListenersProperty, "")
-	} else {
+	if kerberosEnabled || tlsEncryptionEnabled || getBooleanEnvvar(advertiseHostIPEnvvar) {
 		err = writeToWorkingDirectory(advertisedListenersProperty,
 			"advertised.listeners="+strings.Join(advertisedListeners, ","))
+	} else {
+		err = writeToWorkingDirectory(advertisedListenersProperty, "")
 	}
 
 	return err
--- a/frameworks/kafka/setup-helper/setup_helper.go
+++ b/frameworks/kafka/setup-helper/setup_helper.go
@@ -26,6 +26,7 @@ const (
 	authorizationEnvvar   = "SECURITY_AUTHORIZATION_ENABLED"
 	superUsersEnvvar      = "SECURITY_AUTHORIZATION_SUPER_USERS"
 	brokerCountEnvvar     = "BROKER_COUNT"
+	advertiseHostIPEnvvar = "KAFKA_ADVERTISE_HOST"
 
 	listenersProperty           = "listeners"
 	advertisedListenersProperty = "advertised.listeners"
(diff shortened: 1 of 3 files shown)


frameworks/kafka/setup-helper/setup_helper.go
```diff
@@ -162,14 +163,26 @@ func setListeners() error {
...
 
 	err := writeToWorkingDirectory(listenersProperty,
 		"listeners="+strings.Join(listeners, ","))
-	err = writeToWorkingDirectory(advertisedListenersProperty,
-		"advertised.listeners="+strings.Join(advertisedListeners, ","))
+
+	// NOTE: To be consistent with the legacy behavior of the 2.0.X Kafka series,
+	// when there is no security enabled, we must honor the kafka.kafka_advertise_host_ip
+	// configuration parameter
+	if !kerberosEnabled && !tlsEncryptionEnabled && !getBooleanEnvvar(advertiseHostIPEnvvar) {
```
elezar: How about `a || b || c` and swap the execution blocks?

(i.e. `(!a && !b && !c) == !(a || b || c)` - my boolean algebra may be a bit rusty though)

frameworks/kafka/setup-helper/setup_helper.go
```diff
@@ -162,14 +163,26 @@ func setListeners() error {
...
+		// Write an empty file.
+		err = writeToWorkingDirectory(advertisedListenersProperty, "")
```
elezar: Isn't `advertisedListeners` empty in this case anyway? So we just need to ensure that we don't write `advertised.listeners=` if this is the case?

frameworks/kafka/setup-helper/setup_helper_test.go
```diff
@@ -41,22 +41,32 @@ func TestCalculateSettingsListenersError(t *testing.T) {
...
+		kerberosEnvvarValue:         "false",
+		tlsEncryptionEnvvarValue:    "false",
+		tlsAllowPlainEnvvarValue:    "false",
+		expectedListeners:           "listeners=PLAINTEXT://127.0.0.1:1000",
+		expectedAdvertisedListeners: "advertised.listeners=PLAINTEXT://127.0.0.1:1000",
 	},
 	{ // None of the booleans set.
 		expectedListeners:           "listeners=PLAINTEXT://127.0.0.1:1000",
-		expectedAdvertisedListeners: "advertised.listeners=PLAINTEXT://a-task.a-framework:1000",
+		expectedAdvertisedListeners: "",
```
elezar: OK. this answers my question regarding the `advertisedListeners`.

frameworks/kafka/universe/config.json
```diff
@@ -306,7 +306,7 @@
           "default": true
         },
         "kafka_advertise_host_ip": {
-          "description": "Automatically configure advertised.host.name with the tasks' IP address",
+          "description": "DEPRECATED: This configuration parameter has been deprecated and will be removed in a future release. When set to true, and no security settings are enabled, kafka's advertised.listeners property is set to PLAINTEXT://<broker-ip>:<broker-port>. When the setting is false and no security is enabled, then advertised.listeners is NOT set. When security is enabled, the advertised.listeners property will be set to broker hostnames.",
```
elezar: I think in a recent HDFS PR we used `[DEPRECATED]` as the marker for deprecation.

frameworks/kafka/setup-helper/setup_helper.go
```diff
@@ -26,6 +26,7 @@ const (
 	authorizationEnvvar   = "SECURITY_AUTHORIZATION_ENABLED"
 	superUsersEnvvar      = "SECURITY_AUTHORIZATION_SUPER_USERS"
 	brokerCountEnvvar     = "BROKER_COUNT"
+	advertiseHostIPEnvvar = "KAFKA_ADVERTISE_HOST"
```
elezar: Does this meant hat we no longer use the value set through config.json?

benclarkwood: No, I'm obeying the envvar that the config.json value sets.

frameworks/kafka/universe/config.json
benclarkwood: I don't see any `[DEPRECATED]` markers in hdfs config.json?

frameworks/kafka/setup-helper/setup_helper.go
benclarkwood: I'm just ensuring that the file is there so that `cat` can't possibly fail in the svc.yml.

frameworks/kafka/setup-helper/setup_helper.go
benclarkwood: si

frameworks/kafka/setup-helper/setup_helper.go
benclarkwood: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/kafka/universe/marathon.json.mustache#L116

frameworks/kafka/universe/config.json
NimaVaziri: The fields intended for deprecation were removed from master and marked as deprecated on `sdk-0.40`: https://github.com/mesosphere/dcos-commons/blob/sdk-0.40/frameworks/hdfs/universe/config.json#L1358

benclarkwood: Ah, thanks!

I'm going to leave the deprecated fields in master as this field will probably stick around awhile

benclarkwood:
@elezar @NimaVaziri Should be g2g

frameworks/kafka/setup-helper/setup_helper.go
elezar: (the question was more as to whether we could set the condition on the length of the advertised listeners instead of checking the multi-term boolean expression)

PR #2316, opened 2018-02-28
[INFINITY-3256] MIT KDC swap in place of Heimdal

NimaVaziri:
This PR replaces the Heimdal implementation of the KDC with the MIT implementation.

- [X] Replace KDC server implementation
- [ ] Replace client side implementation


tools/kdc/Dockerfile
```diff
@@ -1,11 +1,40 @@
...
+    apt-get clean
+
+
+# The following Kerberos distribution is derived from
+# https://kerberos.org/dist/krb5/1.15/krb5-1.15.2.tar.gz
+RUN wget https://s3-us-west-2.amazonaws.com/nima-dev/krb5-1.16.tar.gz
+RUN tar xvf krb5-1.16.tar.gz  && rm krb5-1.16.tar.gz
+RUN cd krb5-1.16/src && ./configure && make && make install
+
+RUN mkdir /kdc && \
```
NimaVaziri: Still configuring this, need to enable functionality in strict.

tools/kdc/krb5.conf
```diff
@@ -0,0 +1,17 @@
+[logging]
+ default = FILE:/var/log/krb5libs.log
+ kdc = FILE:/var/log/krb5kdc.log
+ admin_server = FILE:/var/log/kadmind.log
+
+[libdefaults]
+ default_realm = LOCAL
```
elezar: Should this not refer to `REALM` as well?

tools/kdc/setup_kdc.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# setup the database used by the KDC
+/setup_kdc_db.sh
+
+sed -i "s/PORT_KDC/$PORT_KDC/g" /kdc/kdc.conf
```
elezar: How about using mustache for the templating instead of bash?

tools/kdc/setup_kdc.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
elezar: Should we add a `set -e` here so that this fails fast?

tools/kdc/setup_kdc.sh
```diff
@@ -0,0 +1,9 @@
...
+# setup the database used by the KDC
+/setup_kdc_db.sh
+
+sed -i "s/PORT_KDC/$PORT_KDC/g" /kdc/kdc.conf
+sed -i "s/PORT_KDC/$PORT_KDC/g; s/REALM/$REALM/g" /kdc/krb5.conf
+krb5kdc
```
elezar: Does this start in background mode?

tools/kdc/setup_kdc_db.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/expect
+
+spawn /usr/local/sbin/kdb5_util create -s -r $env(REALM)
```
elezar: According to https://linux.die.net/man/8/kdb5_util one could use the following option:
```
-P password
specifies the master database password. This option is not recommended.
```

tools/kdc/Dockerfile
```diff
@@ -1,11 +1,40 @@
...
+      curl \
+      expect \
+      less \
+      vim \
+      wget && \
+    apt-get clean
+
+
+# The following Kerberos distribution is derived from
+# https://kerberos.org/dist/krb5/1.15/krb5-1.15.2.tar.gz
```
elezar: What is the reason for the version change in your tar.gz file?

NimaVaziri:
The switch doesn't add any benefits as we previously thought it would.

PR #2317, opened 2018-02-28
INFINITY-3211 Fix flakes in TaskKiller unit tests

nickbp:
Support disabling the background executor in tests, so that it doesn't kill tasks when we don't expect it.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
@@ -24,36 +24,43 @@ import java.util.stream.Collectors;
 public final class TaskKiller {
     private static final Logger LOGGER = LoggerFactory.getLogger(TaskKiller.class);
 
-    private static final Duration killInterval = Duration.ofSeconds(5);
-    private static final Set<TaskID> tasksToKill = new HashSet<>();
-    private static final Object lock = new Object();
+    private static final Duration KILL_INTERVAL = Duration.ofSeconds(5);
+    private static final Set<TaskID> TASKS_TO_KILL = new HashSet<>();
+    private static final Object LOCK = new Object();
 
+    /**
+     * After the first task kill, this executor will be created and will periodically reissue kill invocations for any
+     * tasks which haven't produced a dead or unknown status. We do this because the Mesos kill command is best-effort.
+     */
     private static ScheduledExecutorService executor;
-    private static TaskKiller taskKiller = new TaskKiller();
+
+    /**
+     * Whether the above executor should be running. Only disabled for tests.
+     */
+    private static boolean executorEnabled = true;
 
     private TaskKiller() {
-        startScheduling();
+        // Do not instantiate
     }
 
+    /**
+     * Resets the {@link TaskKiller}'s internal state for tests.
+     *
+     * @param executorEnabled whether the background kill executor should be enabled, should only be disabled in tests
+     */
     @VisibleForTesting
-    static void shutdownScheduling() throws InterruptedException {
-        executor.shutdownNow();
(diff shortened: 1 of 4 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -24,36 +24,43 @@
...
+     * Resets the {@link TaskKiller}'s internal state for tests.
+     *
+     * @param executorEnabled whether the background kill executor should be enabled, should only be disabled in tests
+     */
     @VisibleForTesting
-    static void shutdownScheduling() throws InterruptedException {
-        executor.shutdownNow();
-        executor.awaitTermination(killInterval.toMillis(), TimeUnit.MILLISECONDS);
-    }
+    public static void reset(boolean executorEnabled) throws InterruptedException {
```
benclarkwood: I assume we have to test the TaskKiller outside of its own package?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -74,25 +81,44 @@ public static void killTask(TaskID taskId) {
...
-            tasksToKill.add(taskId);
+        synchronized (LOCK) {
+            TASKS_TO_KILL.add(taskId);
+            LOGGER.info("Enqueued kill of task: {}, {} tasks to kill: {}",
+                    taskId.getValue(),
+                    TASKS_TO_KILL.size(),
+                    TASKS_TO_KILL.stream().map(t -> t.getValue()).collect(Collectors.toList()));
+
+            // Initialize the executor if enabled and not already running.
+            if (executor == null && executorEnabled) {
```
benclarkwood: Is there any circumstance where there are tasks that need to be killed, but `killTask` won't be called?

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/PodResourceTest.java
```diff
@@ -116,6 +119,16 @@
 
     private PodResource resource;
 
+    @BeforeClass
+    public static void beforeAll() throws InterruptedException {
```
benclarkwood: Something something, Singletons are the root of all evil

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
```diff
@@ -74,25 +81,44 @@ public static void killTask(TaskID taskId) {
             return;
         }
```
benclarkwood: Having a log statement before lock acquisition would be good IMO. It isn't _actually_ possible to achieve a deadlock here, but I'm always leery of having locks that are "silently" acquired.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
nickbp: Some tests will invoke task kill (PodResource, service tests) and have expectations on how often driver.kill() is issued, so they'd want to disable the background thread as well.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/TaskKiller.java
nickbp: Not that I'm aware of. The only shortcut would be to invoke the driver directly.

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/PodResourceTest.java
gabrielhartmann: False.

PR #2319, opened 2018-02-28
[INFINITY-3283] Allow restart and force-complete of all plan elements

gabrielhartmann:


sdk/scheduler/src/main/java/com/mesosphere/sdk/http/PlansResource.java
```diff
@@ -421,4 +419,8 @@ private static JSONObject getCommandResult(String command) {
                 "message",
                 String.format("Received cmd: %s", command)));
     }
+
+    static JSONObject getPlanCommandResult(String cmd, String plan, String phase, String step) {
```
nickbp: This could be `private` but don't block on this

gabrielhartmann: oh yeah, thanks.  Artifact of earlier version.

PR #2321, opened 2018-02-28
INFINITY-3211 Fix flakes in TaskKiller unit tests (#2317)

nickbp:
BACKPORT of #2317

Support disabling the background executor in tests, so that it doesn't kill tasks when we don't expect it.


elezar:
I merged this PR with failing HDFS tests -- as these are a known issue and are being addressed in #2327.

PR #2323, opened 2018-03-01
[KAF-409] Add all the missing properties to the plumbing

benclarkwood:
There were a number of properties missing.

I converted the table of documented configs [here]() and used that to update the description of existing configs, and add the ones that are missing. I *did* flip the default on two parameters (enable delete topic, and enable unsafe election). The former because the default changed awhile ago, and the latter because we were not conforming to the documented default.

Here is the list of _new_ config properties:
transaction.state.log.segment.bytes
transaction.remove.expired.transaction.cleanup.interval.ms
log.retention.minutes
queued.max.request.bytes
transaction.max.timeout.ms
log.cleaner.min.compaction.lag.ms
group.initial.rebalance.delay.ms
transaction.state.log.num.partitions
replica.fetch.response.max.bytes
log.roll.ms
log.roll.jitter.ms
log.flush.start.offset.checkpoint.interval.ms
delete.records.purgatory.purge.interval.requests
transaction.abort.timed.out.transaction.cleanup.interval.ms
transaction.state.log.load.buffer.size
transactional.id.expiration.ms
transaction.state.log.replication.factor
log.flush.interval.ms
transaction.state.log.min.isr
replication.quota.window.num
log.retention.ms
replication.quota.window.size.seconds

I interleaved them into logical places in the configuration. They were added as a block to the marathon.json. I apologize that this will be maddening to review.


frameworks/kafka/universe/config.json
```diff
@@ -324,15 +324,21 @@
...
+          "description": "Specify the final compression type for a given topic. This configuration accepts the standard compression codecs ('gzip', 'snappy', 'lz4'). It additionally accepts 'uncompressed' which is equivalent to no compression; and 'producer' which means retain the original compression codec set by the producer.",
           "type": "string",
           "default": "producer"
         },
         "delete_topic_enable": {
           "title": "delete.topic.enable",
           "description": "Enables delete topic. Delete topic through the admin tool will have no effect if this config is turned off",
           "type": "boolean",
-          "default": false
+          "default": true
```
elezar: If i recall correctly, this change is also made in the Confluent mangle?

frameworks/kafka/universe/config.json
```diff
@@ -352,6 +358,11 @@
           "type": "string",
           "default": "9223372036854775807"
         },
+        "log_flush_interval_ms": {
+        "title": "log.flush.interval.ms",
```
elezar: Indentation.

frameworks/kafka/universe/config.json
```diff
@@ -364,18 +375,44 @@
           "type": "string",
           "default": "9223372036854775807"
         },
+        "log_flush_start_offset_checkpoint_interval_ms": {
+          "title": "log.flush.start.offset.checkpoint.interval.ms",
+          "description": "The frequency with which we update the persistent record of log start offset",
+          "type": "integer",
+          "default": 60000
```
elezar: For this and other settings where a range is specified in the https://kafka.apache.org/documentation/ documentation. Should we add a `minimum` specifier to the JSON Schema? (Or at least create a ticket to do this?)

frameworks/kafka/universe/config.json
```diff
@@ -498,25 +535,31 @@
...
+        "queued_max_request_bytes": {
+          "title": "queued.max.request.bytes",
+          "description": "The number of queued bytes allowed before no more requests are read",
+          "type": "integer",
+          "default": -1
+        },
         "quota_consumer_default": {
           "title": "quota.consumer.default",
-          "description": "Any consumer distinguished by clientId/consumer group will get throttled if it fetches more bytes than this value per-second",
+          "description": "DEPRECATED: Used only when dynamic default quotas are not configured for or in Zookeeper. Any consumer distinguished by clientId/consumer group will get throttled if it fetches more bytes than this value per-second",
```
elezar: `[DEPRECATED]`?

elezar: (I was wondering if there was some way to mark this in the schema directly.

frameworks/kafka/universe/config.json
```diff
@@ -698,13 +765,19 @@
           "title": "log.cleaner.io.max.bytes.per.second",
           "description": "The log cleaner will be throttled so that the sum of its read and write i/o will be less than this value on average",
           "type": "number",
-          "default": 1.7976931348623157e+308
```
elezar: The docs I saw still referenced the old value.

frameworks/kafka/universe/config.json
```diff
@@ -698,13 +765,19 @@
...
+          "default": 1.80E+308
         },
         "log_cleaner_min_cleanable_ratio": {
           "title": "log.cleaner.min.cleanable.ratio",
           "description": "The minimum ratio of dirty log to total log for a log to eligible for cleaning",
-          "type": "number",
-          "default": 0.5
+          "type": "double",
+          "default": "0.5"
```
elezar: Should this not just be `"number"` with a value of `0.5`? (not a string?)

frameworks/kafka/universe/config.json
```diff
@@ -744,7 +817,7 @@
         },
         "max_connections_per_ip": {
           "title": "max.connections.per.ip",
-          "description": "mum number of connections we allow from each ip address",
+          "description": "The maximum number of connections we allow from each ip address",
```
elezar: This should have a `minimum` of `1`.

frameworks/kafka/universe/marathon.json.mustache
```diff
@@ -201,7 +201,41 @@
     "TASKCFG_ALL_KAFKA_REPLICA_HIGH_WATERMARK_CHECKPOINT_INTERVAL_MS": "{{kafka.replica_high_watermark_checkpoint_interval_ms}}",
     "TASKCFG_ALL_KAFKA_OFFSETS_TOPIC_SEGMENT_BYTES": "{{kafka.offsets_topic_segment_bytes}}",
     "TASKCFG_ALL_KAFKA_LOG_CLEANER_IO_MAX_BYTES_PER_SECOND": "{{kafka.log_cleaner_io_max_bytes_per_second}}",
-    "TASKCFG_ALL_KAFKA_COMPRESSION_TYPE": "{{kafka.compression_type}}"
+    "TASKCFG_ALL_KAFKA_COMPRESSION_TYPE": "{{kafka.compression_type}}",
+
+    {{#kafka.log_flush_interval_ms}}
```
elezar: I was wondering whether `0` was a Falsey value in the package manager mustache library. Do you know?

frameworks/kafka/universe/config.json
benclarkwood: Whoops! Missed that one, thanks.

frameworks/kafka/universe/config.json
benclarkwood: That's from the Kafka docs. They deprecated them.

I'll probably just remove the deprecated word.

frameworks/kafka/universe/config.json
benclarkwood: That we set it to true?

frameworks/kafka/universe/marathon.json.mustache
benclarkwood: @nickbp do you know off the top of your head?

nickbp: Not certain but I think it isn't? Could get around this by having the value be a string in config.json with a default value of "". Or if it really needs to default to 0, could add some translation in kafka's Main.java. Can't hurt to check what mustache actually does, though.

frameworks/kafka/universe/config.json
benclarkwood: Ah hah! That would explain my failed CI. My script to build all these didn't account for "double"

frameworks/kafka/universe/config.json
benclarkwood: Heh, this is the result of me parsing it to a value.

The docs have the old value.

frameworks/kafka/universe/config.json
benclarkwood: We do not appear to.

frameworks/kafka/universe/config.json
nickbp: FWIW it may make sense to also include the script in this PR. I imagine it'll be useful for resyncing as options are added/changed in Kafka over time.

benclarkwood:
@elezar I have:
1. Added minimums for "everything" (I may have missed one or two)
2. Fixed the other things you noticed.

frameworks/kafka/universe/config.json
elezar: No, I like the deprecated. If this is automated, then let's leave it as is.

frameworks/kafka/universe/config.json
elezar: Good old floating-point arithmetic! I still remember the days where that was what put food on the table.

frameworks/kafka/universe/config.json
elezar: OK, I may have been misremembering.

frameworks/kafka/universe/marathon.json.mustache
elezar: The reason for asking is that 0 is a valid value for some of these, and if the conditional block would evaluate to false in this case, then it would use the Kafka default instead.

benclarkwood: I will give it a quick test.

benclarkwood: Ok, it works in Marathon ok.

benclarkwood: And, it also works in bootstrap's mustache.
```
log.roll.ms=1000
log.roll.hours=168
log.roll.jitter.ms=0
log.roll.jitter.hours=0
```

PR #2324, opened 2018-03-01
[KAF-410] Document topic command security limitation

benclarkwood:


benclarkwood:
I don't know that it's worth making them work.

I think the best path is to simply make them report that they are disabled when security is enabled.

benclarkwood:
But yes, there's a ticket out to correct this.

PR #2326, opened 2018-03-01
[INFINITY-2887][BACKPORT] Update the way tests are run

elezar:
This is a backport of #2283 and contains tooling changes.

The default package version when the `{{dcos-sdk-version}}` template is used has been updated to `"0.40.0-SNAPSHOT"`.


elezar:
The HDFS test failures will be addressed by #2327

PR #2329, opened 2018-03-01
Tooling/fix new standalone framework

elezar:
This PR adds a tooling feature based of the work for #2283.

The purpose of this is to replace the `new-standalone-framework.sh` script (which was not not updated with new SDK features (e.g. pre-built bootstrap and executors).

I spent time on this now because new-starters may need a more self-contained way of starting a framework from scratch.

This can be seen as an iteration on the tooling topic for the next release.


docs/pages/developer-guide.md
```diff
@@ -62,16 +62,32 @@ They store the desired configuration of a service and all relevant information r
...
-1. Create your framework.
+The `mesosphere/dcos-commons:latest` Docker image is able to set up a development environment for creating a new standalone framework.
 
-   ```bash
-   $ ./new-framework.sh frameworks/myframework
-   $ cd frameworks/myframework
-   ```
+In order to set up this environment run the following command:
+```bash
+$ TARGET_FOLDER=/home/elezar/frameworks
```
mpereira: `s/TARGET_FOLDER/FRAMEWORK_ROOT/`?

tools/ci/init
```diff
@@ -101,21 +104,224 @@ def update_sdk(output_path: str, target_version: str):
...
+TEMPLATE_DOCUMENTATION_PATH="http://YOURNAMEHERE.COM/DOCS" \\
+TEMPLATE_ISSUES_PATH="http://YOURNAMEHERE.COM/SUPPORT" \\
+    $FRAMEWORK_DIR/tools/build_package.sh \\
+        ${FRAMEWORK_NAME} \\
+        $FRAMEWORK_DIR \\
+        -a "$BUILD_DIR/${FRAMEWORK_NAME}-scheduler.zip" \\
+        $@
+"""
+    write_file(os.path.join(output_path, "build.sh"),
+               contents.replace("{{framework_name}}", framework_name))
```
mpereira: I think we need a more unique replace target, this replaces `./bootstrap -template=false` with `./bootstrap -$framework_name=false` in marathon.json.mustache. Do you mind if push a fix for this @elezar?

elezar: This only replaces the literal string `"{{framework_name}}"` in the local variable `contents` with `framework_name`.

tools/ci/init
```diff
@@ -101,21 +104,224 @@ def update_sdk(output_path: str, target_version: str):
...
+             "src/main/java/com/mesosphere/sdk/template/"]
+
+    for f in folders:
+        shutil.move(os.path.join(output_path, f),
+                    os.path.join(output_path, f[:-len("/template/")], framework_name))
+
+
+def update_files(output_path: str, framework_name: str):
+
+    template_pattern = "([^-]?)template([^_]?)"
```
elezar: It could be that the `?` in the regex is not playing along with what we want to do. (i.e. replacing `-template=` with the framework name.

PR #2331, opened 2018-03-01
INFINITY-3240 Support taskids with multiple entries

nickbp:
Update current code to support a format that may be used in the future:
- Current format: `<name>__<uuid>`
- Potential future format: `<otherthings>__<name>__<uuid>`

### DIFF
--- a/sdk/common/src/main/java/com/mesosphere/sdk/offer/CommonIdUtils.java
+++ b/sdk/common/src/main/java/com/mesosphere/sdk/offer/CommonIdUtils.java
@@ -3,9 +3,6 @@ package com.mesosphere.sdk.offer;
 import java.util.UUID;
 
 import org.apache.mesos.Protos;
-import org.apache.mesos.Protos.ExecutorID;
-import org.apache.mesos.Protos.SlaveID;
-import org.apache.mesos.Protos.TaskID;
 
 /**
  * Various utility methods for manipulating data in {@link Protos.TaskInfo}s.
@@ -13,7 +10,7 @@ import org.apache.mesos.Protos.TaskID;
 public class CommonIdUtils {
 
     /** Used in task and executor IDs to separate the task/executor name from a UUID. */
-    public static final String NAME_ID_DELIM = "__";
+    private static final String NAME_ID_DELIM = "__";
 
     private CommonIdUtils() {
         // do not instantiate
@@ -26,7 +23,7 @@ public class CommonIdUtils {
      *
      * @see #toTaskId(String)
      */
-    public static String toTaskName(TaskID taskId) throws TaskException {
+    public static String toTaskName(Protos.TaskID taskId) throws TaskException {
         return extractNameFromId(taskId.getValue());
     }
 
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorTest.java
```diff
@@ -60,7 +60,7 @@ public void testReserveLaunchScalar() throws Exception {
         Assert.assertEquals(getResourceId(reserveResource), getResourceId(launchResource));
         String executorId = launchOperation.getLaunchGroup().getExecutor().getExecutorId().getValue();
 
-        String prefix = TestConstants.POD_TYPE + CommonIdUtils.NAME_ID_DELIM;
+        String prefix = TestConstants.POD_TYPE + "__";
```
benclarkwood: Any reason to not keep the delimiter as public?

sdk/common/src/test/java/com/mesosphere/sdk/offer/CommonIdUtilsTest.java
```diff
@@ -11,33 +10,116 @@
...
+    }
 
-        Assert.assertNotNull(executorID);
-        final String value = executorID.getValue();
-        Assert.assertTrue(StringUtils.isNotBlank(value));
-        Assert.assertTrue(value.contains("__"));
-        Assert.assertNotNull(UUID.fromString(value.split("__")[1]));
+    @Test
+    public void testMultiValidToEmptyExecutorName() throws Exception {
+        Protos.ExecutorID validExecutorId = getExecutorId(TEST_OTHER_NAME + "____id");
```
benclarkwood: Should this case throw an exception instead?

sdk/common/src/main/java/com/mesosphere/sdk/offer/CommonIdUtils.java
```diff
@@ -3,17 +3,14 @@
 import java.util.UUID;
```
benclarkwood: Is it possible for us to wind up in the scenario of:
1. User upgrades to new version
2. Task gets relaunched with new style task id
3. User downgrades to old version
4. We then somehow try to regenerate the task name for the relaunched task and are trying to generate one with out the "other stuff" prefix.

gabrielhartmann: That is a good question.

gabrielhartmann: @nickbp: Can we write some test(s) that prove this out one way or another?

nickbp: We only retrieve the task name in a couple places. One is to figure out the goal state from a TaskStatus and the other is to validate during initial launch that the task id isn't malformed (i.e. we don't actually use the name value).

In practice the task id is generated from the task spec from scratch at launch time. It just gets regenerated each time since the UUID is supposed to change.

That said I could still do a ServiceTest or something if it's desirable

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorTest.java
mrbrowning: This is kind of an issue with the Utils idiom, IMO, since it doesn't actually encapsulate the data it operates on. That won't change here for a while, though, so I vote for public too.

sdk/common/src/main/java/com/mesosphere/sdk/offer/CommonIdUtils.java
nickbp: OK, yeah this really shouldn't be an issue now that I think about it more. Task names are generated from the service specs. Where TaskInfos are concerned, task name is a separate field from task id. The task id -> task name extraction only really comes up with TaskStatuses, where there is no task name field. Hence the above case where we're figuring out the goal state from a task status (e.g. to figure out if 'FINISHED' is the desired state, after a task has finished).

PR #2337, opened 2018-03-03
[TOOLS] Fixed uninitialized variable in release_builder.py::apply_beta_version

susanxhuynh:
In release_builder.py, the variable `stripped_version` is not initialized if `is_beta` is false. (Compare with the method above it, apply_beta_prefix())


nickbp:
Going to just merge as this isn't exercised by CI

PR #2338, opened 2018-03-03
Move placement constraints to broker section (DCOS_KAFKA-26)

deric:
As discussed in [DCOS_KAFKA-26](https://jira.mesosphere.com/browse/DCOS_KAFKA-26) having `placement_constraint` for brokers in "service" section could be confusing for new users, when the next tab is called "brokers".


elezar:
@deric moving this to the `brokers` section is a breaking change for existing users and we would have to add a better transitional path.

deric:
@elezar How would you imagine a smooth transitional path? The PR is targeted for `beta-kafka` where users shouldn't expect stable API. AFAIK there are not upgrade scripts in DC/OS packages where we could check for such configuration incompatibilities.

The only thing I can think of is checking for defined variable in case someone is trying to deploy new package with old config:
```
{{#brokers.placement_constraint}}
"PLACEMENT_CONSTRAINTS": "{{{brokers.placement_constraint}}}",
{{/brokers.placement_constraint}}
# for backward compatibility, will be removed in release x.x.x
{{#service.placement_constraint}}
"PLACEMENT_CONSTRAINTS": "{{{service.placement_constraint}}}",
{{/service.placement_constraint}}
```

elezar:
Since the `beta-kafka` package generally becomes a GA version at some point, there should be a valid upgrade path from the existing GA packages where this setting is under the `service` section: https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/K/kafka/44/config.json#L33

The check you propose [above](https://github.com/mesosphere/dcos-commons/pull/2338#issuecomment-370382019) should work, but we would not want *both* sections specified (as this will be invalid JSON).

```
{{#service.placement_constraint}}
"PLACEMENT_CONSTRAINTS": "{{{service.placement_constraint}}}",
{{/service.placement_constraint}}
{{^service.placement_constraint}}
{{#brokers.placement_constraint}}
"PLACEMENT_CONSTRAINTS": "{{{brokers.placement_constraint}}}",
{{/brokers.placement_constraint}}
{{/service.placement_constraint}}
```

deric:
@elezar Sure, I've updated the PR accordingly.

elezar:
Hi @deric. Sorry for lagging behind on this PR. We have recently moved Kafka out of this repo. Please consider recreating this PR at https://github.com/mesosphere/dcos-kafka-service

PR #2345, opened 2018-03-06
[INFINITY-3294] Add retry script and retry gradle command.

mpereira:
This script could probably be used in other parts as well, for now I only used it on the gradle command that might fail ocasionally due its usage of public repositories.


build.sh
```diff
@@ -60,7 +60,8 @@ fi
 # Build Java bits: Executor
 if [ "$EXECUTOR_DIR" != $DISABLED_VALUE ]; then
     echo "Building executor in $EXECUTOR_DIR"
-    ./gradlew distZip -p $EXECUTOR_DIR
+    # Retrying gradle command to handle transient failures with downloading jars from public repositories.
+    $REPO_ROOT_DIR/tools/retry -- ./gradlew distZip -p $EXECUTOR_DIR
```
elezar: This will retry all failures. Could we split the gradle command to first download the prerequisites (possibly with `--refresh-dependencies  Refresh the state of dependencies.`?) and then run the build afterwards?

tools/retry
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env bash
```
nickbp: Maybe just do a quick version of this in python

mpereira: What do you think of using https://github.com/strizhechenko/shell-retry?

build.sh
mpereira: https://github.com/mesosphere/dcos-commons/pull/2345/commits/4581445c82ecc62bf2a3909f0d476ce1775e3cab

mpereira:
Closing.

PR #2348, opened 2018-03-07
[ELASTIC-175] Make X-Pack install successfully.

mpereira:
wip


mpereira:
The current test failure should be gone after https://github.com/mesosphere/dcos-commons/pull/2361 gets merged.

PR #2349, opened 2018-03-07
[ELASTIC-172] Add test that verifies that Kibana works against an ES cluster with TLS.

mpereira:


frameworks/elastic/tests/test_tls.py
```diff
@@ -91,3 +115,10 @@ def test_crud_over_tls(elastic_service_tls):
 
     assert document
     assert document['_source']['name'] == 'Loren'
+
+
+@pytest.mark.tls
+@pytest.mark.sanity
+@pytest.mark.dcos_min_version('1.10')
```
elezar: Since all tests are `dcos_min_version('1.10')` we could add this to the module-level `pytestmark`.

frameworks/elastic/tests/test_tls.py
```diff
@@ -60,6 +61,29 @@ def elastic_service_tls(service_account):
...
+                "elasticsearch_tls": True,
+                "elasticsearch_url": elasticsearch_url
+            }
+        },
+        timeout_seconds=config.DEFAULT_KIBANA_TIMEOUT,
+        wait_for_deployment=False)
+
+    yield
+
+    sdk_install.uninstall(config.KIBANA_PACKAGE_NAME, config.KIBANA_PACKAGE_NAME)
```
elezar: We may want a `try - finally` block here to ensure that the uninstall is executed even if there is a failure.

elezar: (and above for elastic!)

mpereira: https://github.com/mesosphere/dcos-commons/pull/2349/commits/e07b1a750211a6faa1a2e50579a4adb51cebad48

frameworks/elastic/tests/test_tls.py
mpereira: https://github.com/mesosphere/dcos-commons/pull/2349/commits/a6b8f0c1b61f37bae4e7dc071a0aff30a6cacbc6

frameworks/elastic/tests/test_tls.py
```diff
@@ -2,15 +2,18 @@
...
 from security import transport_encryption
 
 from tests import config
 
-pytestmark = pytest.mark.skipif(sdk_utils.is_open_dcos(),
-                                reason='Feature only supported in DC/OS EE')
+pytestmark = [pytest.mark.skipif(sdk_utils.is_open_dcos(),
+                                 reason="Feature only supported in DC/OS EE"),
+              pytest.mark.skipif(sdk_utils.dcos_version_less_than("1.10"),
+                                 reason="TLS tests require DC/OS 1.10 or higher")]
```
elezar: I've seen a lot of other messages using `1.10+`. Not sure which you prefer, but it's not critical. (only change it if applicable and if there are other changes that need the tests to run again)

frameworks/elastic/tests/test_tls.py
```diff
@@ -30,46 +33,69 @@ def service_account(configure_security):
...
+            service_name=config.KIBANA_PACKAGE_NAME,
+            expected_running_tasks=0,
+            additional_options={
+                "kibana": {
+                    "xpack_enabled": True,
+                    "elasticsearch_tls": True,
+                    "elasticsearch_url": elasticsearch_url
+                }
+            },
+            timeout_seconds=config.DEFAULT_KIBANA_TIMEOUT,
```
elezar: Could we change this to be `KIBANA_DEFAULT_TIMEOUT` as this gives better grouping with the other `KIBANA_*` globals.

frameworks/elastic/tests/test_tls.py
```diff
@@ -30,46 +33,69 @@ def service_account(configure_security):
...
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
 
-    sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+@pytest.fixture(scope='module')
+def kibana_application_tls(elastic_service_tls):
+    try:
+        elasticsearch_url = "https://" + sdk_hosts.vip_host(config.SERVICE_NAME, "coordinator", 9200)
+
+        sdk_install.install(
```
elezar: This is going to break on strict due to a misconfigured service account and secret, or is that not an issue for non-SDK applictaionts

frameworks/elastic/tests/test_tls.py
```diff
@@ -91,3 +117,9 @@ def test_crud_over_tls(elastic_service_tls):
 
     assert document
     assert document['_source']['name'] == 'Loren'
+
+
+@pytest.mark.tls
+@pytest.mark.sanity
+def test_kibana_tls(kibana_application_tls):
+    config.check_kibana_adminrouter_integration("service/{}/login".format(config.KIBANA_PACKAGE_NAME))
```
elezar: This should be `KIBANA_SERVICE_NAME` right?

frameworks/elastic/tests/test_tls.py
```diff
@@ -30,46 +33,69 @@ def service_account(configure_security):
...
-    )
+        )
 
-    sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
+        sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
 
-    # Wait for service health check to pass
-    shakedown.service_healthy(config.SERVICE_NAME)
+        # Wait for service health check to pass
+        shakedown.service_healthy(config.SERVICE_NAME)
```
elezar: Is this valid? As far as I know, the scheduler returns healthy long before the deployment has completed.

frameworks/elastic/tests/test_tls.py
```diff
@@ -30,46 +33,69 @@ def service_account(configure_security):
...
-        expected_running_tasks=config.DEFAULT_TASK_COUNT,
-        additional_options={
-            "service": {
-                "service_account_secret": service_account["name"],
-                "service_account": service_account["secret"],
-                "security": {
-                    "transport_encryption": {
-                        "enabled": True
+    try:
+        sdk_install.install(
```
elezar: We could add an `uninstall` call here to ensure that there is no elastic installed before trying to install it.

frameworks/elastic/tests/test_tls.py
```diff
@@ -30,46 +33,69 @@ def service_account(configure_security):
...
+            config.PACKAGE_NAME,
+            service_name=config.SERVICE_NAME,
+            expected_running_tasks=config.DEFAULT_TASK_COUNT,
+            additional_options={
+                "service": {
+                    "service_account_secret": service_account["name"],
```
elezar: These two fields are the swapped.

frameworks/elastic/tests/test_tls.py
```diff
@@ -30,46 +33,69 @@ def service_account(configure_security):
...
+                    "service_account": service_account["secret"],
+                    "security": {
+                        "transport_encryption": {
+                            "enabled": True
+                        }
```
elezar: The formatting of the dict looks off. (the `}` closing the `security` section.

frameworks/elastic/tests/test_tls.py
mpereira: I think this would only be needed/desired for services that define health checks. Since Elastic doesn't and  `sdk_install.install` by default waits for the deployment to finish we can remove it.

frameworks/elastic/tests/test_tls.py
mpereira: Good catch. I'll fix them here too: https://github.com/mesosphere/dcos-commons/blob/a6b8f0c1b61f37bae4e7dc071a0aff30a6cacbc6/frameworks/hdfs/tests/test_tls.py#L45-L46

frameworks/elastic/tests/test_tls.py
mpereira: You sure? If you mean the above `}` it's closing `transport_encryption`.

frameworks/elastic/tests/test_tls.py
mpereira: Yes, maybe just drop the prefix entirely.

frameworks/elastic/tests/test_tls.py
mpereira: Yes!

frameworks/elastic/tests/test_tls.py
mpereira: I can change it, but I probably copied from [here](https://github.com/mesosphere/dcos-commons/blob/a6b8f0c1b61f37bae4e7dc071a0aff30a6cacbc6/frameworks/hdfs/tests/test_kerberos_auth.py#L25-L26) fwiw.

frameworks/elastic/tests/test_tls.py
mpereira: I *think* it's fine for Kibana since it's non-SDK and doesn't need to generate artifacts, secrets, etc., right? Its config schema doesn't even define service account/secret.

frameworks/elastic/tests/test_tls.py
mpereira: https://github.com/mesosphere/dcos-commons/pull/2349/commits/ea0aa2e401f8b275f78577ed3dca99af999f1e0a

frameworks/elastic/tests/test_tls.py
mpereira: https://github.com/mesosphere/dcos-commons/pull/2349/commits/554730b7ecfadde27a7cd43be93408238688d641

frameworks/elastic/tests/test_tls.py
mpereira: https://github.com/mesosphere/dcos-commons/pull/2349/commits/b50c4a2c33c936c959fa364b7b92800442a4ff96

frameworks/elastic/tests/test_tls.py
mpereira: https://github.com/mesosphere/dcos-commons/pull/2349/commits/02ad4ea8eb1318f79cff4dd6015499ea0895739a

frameworks/elastic/tests/test_tls.py
mpereira: https://github.com/mesosphere/dcos-commons/pull/2349/commits/cc962044e532c009daab5e8afe3c9dbe9b16cb4f

frameworks/elastic/tests/test_tls.py
```diff
@@ -30,46 +33,65 @@ def service_account(configure_security):
...
+                    "elasticsearch_url": elasticsearch_url
+                }
+            },
+            timeout_seconds=config.KIBANA_DEFAULT_TIMEOUT,
+            wait_for_deployment=False)
 
-    sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+        yield
+    finally:
+        sdk_install.uninstall(config.KIBANA_PACKAGE_NAME, config.KIBANA_PACKAGE_NAME)
```
elezar: The second parameter should be `config.KIBANA_SERVICE_NAME`. This is only an issue for beta versions.

mpereira: https://github.com/mesosphere/dcos-commons/pull/2349/commits/37239966cd1f5b3b469ff804ba43708e2f5d18aa

PR #2351, opened 2018-03-07
Fix failure escalation detection

gabrielhartmann:
This is hacky.  I know it's hacky.  It works and we're short on time and it's a small change.   We need to do a redesign of recovery.  This approach is at the end of its rope.


gabrielhartmann:
@nickbp: We had reasonable unit test coverage.  Things went bad in the real world.  @NimaVaziri is adding a missing integration test.

PR #2352, opened 2018-03-07
Journal node replace integration tests

NimaVaziri:
This is based on the commit made in #2351 to include the fix and allow passing of added tests in CI.

### DIFF
--- a/frameworks/hdfs/tests/test_sanity.py
+++ b/frameworks/hdfs/tests/test_sanity.py
@@ -204,23 +204,6 @@ def test_kill_all_datanodes():
 
 @pytest.mark.sanity
 @pytest.mark.recovery
-def test_permanently_replace_namenodes():
-    replace_node(0, "name")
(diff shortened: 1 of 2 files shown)


frameworks/hdfs/tests/test_sanity.py
```diff
@@ -205,9 +205,17 @@ def test_kill_all_datanodes():
...
+    replace_node(1, "name")
+    replace_node(0, "name")
+
+
+@pytest.mark.sanity
+@pytest.mark.recovery
+def test_permanently_replace_journalnodes():
+    replace_node(0, "journal")
+    replace_node(1, "journal")
+    replace_node(2, "journal")
```
gabrielhartmann: Let's replace `0` again as in the `name-node` test.

gabrielhartmann: This is going to fail pretty frequently until we get the readiness check under control.

PR #2364, opened 2018-03-09
Add region config change validator and builder method for ensuring single-region constraint

mrbrowning:
Adding tests for this functionality now, but I wanted to make sure we were all on the same page about how it works. The downside I can see is that, if you have placed your scheduler in a single region explicitly and then change that region in your config, `RegionValidator` won't catch that and so you'll only find out once offers start getting rejected. This is a hazard of `ZoneValidator` too, though, and since computing equality over things like a regex-based placement rule isn't feasible I think that's a limitation we'll have to accept.

### DIFF
--- a/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
+++ b/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
@@ -49,6 +49,15 @@ public class Main {
                         schedulerConfig,
                         yamlSpecFile.getParentFile());
                 break;
+            case MULTI_REGION:
+                yamlSpecFile = new File(args[0]);
+                serviceSpec = DefaultServiceSpec
+                        .newGenerator(yamlSpecFile, SchedulerConfig.fromEnv())
+                        .build();
+                builder = DefaultScheduler.newBuilder(serviceSpec, SchedulerConfig.fromEnv())
+                        .withSingleRegionConstraint();
+                runner = SchedulerRunner.fromSchedulerBuilder(builder);
+                break;
             case CUSTOM_PLAN:
                 yamlSpecFile = new File(args[0]);
                 serviceSpec = DefaultServiceSpec
@@ -78,7 +87,8 @@ public class Main {
         YAML,
         Java,
         CUSTOM_PLAN,
-        CUSTOM_DECOMMISSION
+        CUSTOM_DECOMMISSION,
+        MULTI_REGION
     }
 
     private static final String SCENARIO_KEY = "SCENARIO";
@@ -86,6 +96,8 @@ public class Main {
     private static final String JAVA_FLAG = "JAVA";
     private static final String CUSTOM_PLAN_FLAG = "CUSTOM_PLAN";
     private static final String CUSTOM_DECOMISSION_FLAG = "CUSTOM_DECOMMISSION";
+    private static final String MULTI_REGION_FLAG = "MULTI_REGION";
+
 
     private static Scenario getScenario() {
         String flag = System.getenv().get(SCENARIO_KEY);
@@ -97,6 +109,8 @@ public class Main {
                 return Scenario.CUSTOM_PLAN;
             case CUSTOM_DECOMISSION_FLAG:
                 return Scenario.CUSTOM_DECOMMISSION;
+            case MULTI_REGION_FLAG:
+                return Scenario.MULTI_REGION;
             case YAML_FLAG:
             default:
                 return Scenario.YAML;
--- a/frameworks/helloworld/tests/test_region_awareness.py
+++ b/frameworks/helloworld/tests/test_region_awareness.py
@@ -18,7 +18,11 @@ POD_NAMES = ['hello-0', 'world-0', 'world-1']
 @pytest.mark.region_awareness
 @sdk_utils.dcos_ee_only
 def test_nodes_deploy_to_local_region_by_default():
-    sdk_install.install(config.PACKAGE_NAME, config.SERVICE_NAME, 3)
+    sdk_install.install(
+        config.PACKAGE_NAME,
+        config.SERVICE_NAME,
+        3,
+        additional_options={"service": {"scenario": "MULTI_REGION"}})
 
     sdk_plan.wait_for_completed_deploy(config.SERVICE_NAME)
 
@@ -42,7 +46,7 @@ def test_nodes_can_deploy_to_remote_region():
         config.PACKAGE_NAME,
         config.SERVICE_NAME,
         3,
-        additional_options={"service": {"region": "Europe"}})
+        additional_options={"service": {"scenario": "MULTI_REGION", "region": "Europe"}})
 
     sdk_plan.wait_for_completed_deploy(config.SERVICE_NAME)
 
@@ -63,7 +67,11 @@ def test_nodes_can_deploy_to_remote_region():
 @pytest.mark.region_awareness
 @sdk_utils.dcos_ee_only
 def test_region_config_update_does_not_succeed():
-    sdk_install.install(config.PACKAGE_NAME, config.SERVICE_NAME, 3)
+    sdk_install.install(
+        config.PACKAGE_NAME,
+        config.SERVICE_NAME,
+        3,
+        additional_options={"service": {"scenario": "MULTI_REGION"}})
 
     sdk_plan.wait_for_completed_deploy(config.SERVICE_NAME)
     change_region_config('Europe')
--- a/frameworks/helloworld/tests/test_region_awareness.py
+++ b/frameworks/helloworld/tests/test_region_awareness.py
@@ -1,0 +1,88 @@
+import json
+import logging
+
+import pytest
+import sdk_cmd
+import sdk_install
+import sdk_marathon
+import sdk_plan
+import sdk_utils
+from tests import config
+
+log = logging.getLogger(__name__)
+
+POD_NAMES = ['hello-0', 'world-0', 'world-1']
+
+
+@pytest.mark.dcos_min_version('1.11')
+@pytest.mark.region_awareness
+@sdk_utils.dcos_ee_only
+def test_nodes_deploy_to_local_region_by_default():
+    sdk_install.install(config.PACKAGE_NAME, config.SERVICE_NAME, 3)
+
+    sdk_plan.wait_for_completed_deploy(config.SERVICE_NAME)
+
+    for pod_name in POD_NAMES:
+        info = sdk_cmd.service_request(
+            'GET', config.SERVICE_NAME, '/v1/pod/{}/info'.format(pod_name)
+        ).json()[0]['info']
+
+        assert (
+            [l['value'] for l in info['labels']['labels'] if l['key'] == 'offer_region'][0] == 'USA'
+        )
+    
+    sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.mark.dcos_min_version('1.11')
+@pytest.mark.region_awareness
+@sdk_utils.dcos_ee_only
+def test_nodes_can_deploy_to_remote_region():
+    sdk_install.install(
+        config.PACKAGE_NAME,
+        config.SERVICE_NAME,
+        3,
+        additional_options={"service": "region": "Europe"})
+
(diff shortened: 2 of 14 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerBuilder.java
```diff
@@ -215,6 +217,36 @@ public SchedulerBuilder setPlanCustomizer(PlanCustomizer planCustomizer) {
         return this;
     }
 
+    public SchedulerBuilder withSingleRegionConstraint() {
```
nickbp: API looks good. Subjective: Maybe defer the actual serviceSpec manipulation until `build()`? This could do something like set a boolean to true, which then gets checked for in `build()`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -221,6 +226,10 @@ public String getDcosSpace() {
         return envStore.getOptional(MARATHON_APP_ID_ENV, "/");
     }
 
+    public String getSchedulerRegion() {
```
nickbp: Could return an `Optional<String>` here

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/RegionCannotChange.java
```diff
@@ -0,0 +1,36 @@
...
+public class RegionCannotChange implements ConfigValidator<ServiceSpec> {
+
+    @Override
+    public Collection<ConfigValidationError> validate(Optional<ServiceSpec> oldConfig, ServiceSpec newConfig) {
+        if (!oldConfig.isPresent()) {
+            return Collections.emptyList();
+        }
+
+        List<ConfigValidationError> errors = new ArrayList<>();
+        if (oldConfig.get().getRegion() != null && !oldConfig.get().getRegion().equals(newConfig.getRegion())) {
```
gabrielhartmann: Someone's being pretty malicious if they packaged a `null` in an `Optional`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerBuilder.java
```diff
@@ -215,6 +217,36 @@ public SchedulerBuilder setPlanCustomizer(PlanCustomizer planCustomizer) {
...
+                 .collect(Collectors.toList());
+
+         serviceSpec = DefaultServiceSpec.newBuilder(serviceSpec).pods(updatedPodSpecs).build();
+
+         return this;
+    }
+
+    @VisibleForTesting
+    static PlacementRule getRegionRule(String schedulerRegion) {
+        if (schedulerRegion == null) {
```
gabrielhartmann: Probably want `StringUtils.isBlank()`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -221,6 +226,10 @@ public String getDcosSpace() {
...
+        return envStore.getOptional(SERVICE_REGION_ENV, null);
```
gabrielhartmann: This follows the pattern in the rest of this file, but I wonder why we're not using `Optional` here and elsewhere.  Generally speaking we don't pass around `null` a lot in the SDK.

sdk/scheduler/src/test/java/com/mesosphere/sdk/config/validate/RegionCannotChangeTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+import static org.mockito.Mockito.when;
+
+public class RegionCannotChangeTest {
```
gabrielhartmann: I think we're missing the "had a region, but it was removed" test.

gabrielhartmann:
@mrbrowning: I don't understand your comment about rejected offers in the description.  If validation fails, the proposed configuration is rejected and the last good configuration is maintained as the target configuration.  Errors are reported on the deployment plan and in logs, but the new configuration is never accepted as a target.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
mrbrowning: That's sort of the confusing aspect of the `getOptional` pattern -- since it doesn't return an `Optional`, one has to indicate absence of value with null. I'd be happy to create a ticket to refactor that elsewhere.

mrbrowning:
@gabrielhartmann it's an artifact of the fact that I was trying to avoid adding region even as a field to the `ServiceSpec`, but that made config time validation impossible (since you can't really traverse nested placement rules to test for equality). It's no longer true, obviously.

sdk/scheduler/src/test/java/com/mesosphere/sdk/config/validate/RegionCannotChangeTest.java
mrbrowning: Yep, will add.

gabrielhartmann:
Ah.  Got it.  Thanks.

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -0,0 +1,98 @@
...
+log = logging.getLogger(__name__)
+
+POD_NAMES = ['hello-0', 'world-0', 'world-1']
+
+
+@pytest.mark.dcos_min_version('1.11')
+@pytest.mark.region_awareness
+@sdk_utils.dcos_ee_only
+def test_nodes_deploy_to_local_region_by_default():
+    sdk_install.install(
```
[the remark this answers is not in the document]
elezar: https://jira.mesosphere.com/browse/INFINITY-3390

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -0,0 +1,98 @@
...
+
+    sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
+    change_region_config('Europe')
+    plan = sdk_plan.get_deployment_plan(config.SERVICE_NAME)
+    assert plan.get('errors', [])
+
+    sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, 'pod replace hello-0')
+    sdk_plan.wait_for_completed_recovery(config.SERVICE_NAME)
+
+    info = sdk_cmd.service_request(
```
elezar: Getting the pod information and asserting that the value is as desired could be perfect for a utility function.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/validation/RegionValidator.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * This class validates that referencing Regions in placement constraints can only support the following transitions.
+ *
+ * 1. null  --> false
```
elezar: It is not clear how the `null` to `false` transition is validated. (and by extension that `null` to `true` is *not* allowed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerBuilder.java
```diff
@@ -215,6 +217,40 @@ public SchedulerBuilder setPlanCustomizer(PlanCustomizer planCustomizer) {
...
+         return this;
+    }
+
+    @VisibleForTesting
+    static PlacementRule getRegionRule(Optional<String> schedulerRegion) {
+        if (!schedulerRegion.isPresent()) {
+            return new IsLocalRegionRule();
+        }
+
+        return RegionRuleFactory.getInstance().require(ExactMatcher.create(schedulerRegion.get()));
```
elezar: Would it not make sense to have the `RegionRuleFactory` contain the logic for constructing the `IsLocalRegionRule` too? That would leave us with a *single* point for defining region -> region rule Mappings.

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -49,6 +49,15 @@ public static void main(String[] args) throws Exception {
                         schedulerConfig,
                         yamlSpecFile.getParentFile());
                 break;
+            case MULTI_REGION:
+                yamlSpecFile = new File(args[0]);
+                serviceSpec = DefaultServiceSpec
+                        .newGenerator(yamlSpecFile, SchedulerConfig.fromEnv())
+                        .build();
+                builder = DefaultScheduler.newBuilder(serviceSpec, SchedulerConfig.fromEnv())
+                        .withSingleRegionConstraint();
```
elezar: It seems a bit strange to have the `MULTI_REGION` case call a method with the name of `withSingleRegionConstraint()`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerBuilder.java
mrbrowning: Potentially -- those classes already existed as is, but I intend to do a refactor pass where I could adjust that.

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
mrbrowning: `MULTI_REGION` describes the testing scenario -- whether in a multi-region cluster or not, we _always_ mandate that the scheduler run its tasks in a single region, so the point here is that `withSingleRegionConstraint` really only makes sense to assert when in a multi-region environment in the first place.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/RegionCannotChange.java
```diff
@@ -0,0 +1,43 @@
...
+                newConfig.getRegion().isPresent();
+        boolean regionWasUnset = oldConfig.isPresent() && oldConfig.get().getRegion().isPresent() &&
+                !newConfig.getRegion().isPresent();
+        boolean regionWasChanged = oldConfig.isPresent() && oldConfig.get().getRegion().isPresent() &&
+                newConfig.getRegion().isPresent() &&
+                !newConfig.getRegion().get().equals(oldConfig.get().getRegion().get());
+        if (regionWasAdded || regionWasUnset || regionWasChanged) {
+            errors.add(ConfigValidationError.transitionError(
+                    "region",
+                    oldConfig.get().getUser(),
```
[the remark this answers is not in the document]
mrbrowning: Yes.

frameworks/helloworld/tests/test_region_awareness.py
mrbrowning: I think DRY is less important in test code -- better in my opinion to be able to read the test at a glance to see what's going on. Also since the JSON schema for pod info is both a broad and deep tree, I've always found it non-obvious which utility function is actually going to get me what I need.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/RegionCannotChange.java
```diff
@@ -0,0 +1,43 @@
...
+public class RegionCannotChange implements ConfigValidator<ServiceSpec> {
+
+    @Override
+    public Collection<ConfigValidationError> validate(Optional<ServiceSpec> oldConfig, ServiceSpec newConfig) {
+        if (!oldConfig.isPresent()) {
+            return Collections.emptyList();
+        }
+
+        List<ConfigValidationError> errors = new ArrayList<>();
+        boolean regionWasAdded = oldConfig.isPresent() && !oldConfig.get().getRegion().isPresent() &&
```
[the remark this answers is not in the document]
mrbrowning: Yes, this was an artifact of not previously returning `Optional<String>`, will clean up.

PR #2367, opened 2018-03-10
[INFINITY-3133] Added missing Kerberos auth for readiness check

NimaVaziri:
Missed this the first time around - the kerberos tests don't do a restart/replace of the journal nodes so the readiness check doesn't actually succeed after the first time it passes.


benclarkwood:
Sounds like there should be a restart integration test in the Kerberos suite then, no?

NimaVaziri:
Yup, will add

NimaVaziri:
@benclarkwood added restart integration test

PR #2361, opened 2018-03-09
[INFINITY-3321] Kibana CI install failure

mpereira:
The build script for the Elastic service builds 2 packages: Elastic and Kibana. A stub Universe is created for each, and for PR builds their URLs are appended to a file (`$UNIVERSE_URL_PATH`), one URL per line.

https://github.com/mesosphere/dcos-commons/blob/3814626ec94197386a0fe2ced623b988e5d4507b/tools/publish_aws.py#L93-L98
https://github.com/mesosphere/dcos-commons/blob/3814626ec94197386a0fe2ced623b988e5d4507b/frameworks/elastic/build.sh#L35

The TC test runner script (TEST_SCRIPT in the TC build parameters) runs `export STUB_UNIVERSE_URL=$(cat $UNIVERSE_URL_PATH)`, making `$STUB_UNIVERSE_URL` a newline-separated string.

During PR build test runs the code that adds the stub Universe repos splits that string by comma (instead of by newline), ending up sending a 2 line command to sdk_cmd, which interprets each line as a separate command.

https://github.com/mesosphere/dcos-commons/blob/3814626ec94197386a0fe2ced623b988e5d4507b/testing/sdk_repository.py#L22-L24

The Elastic repo is added successfully (first line), but the Kibana repo (second line) is interpreted as a shell command, which fails.

```
[2018-03-09 12:45:08,919|sdk_cmd|INFO]: STDERR:
/bin/sh: 2: https://universe-converter.mesosphere.com/transform?url=https://infinity-artifacts-ci.s3.amazonaws.com/autodelete7d/kibana/20180309-123133-S99ev8xdsYyxzKbU/stub-universe-kibana.json: not found
```

This means Kibana tests have been running against the Universe version for some time now, which explains some weirdness I had been seeing when working on the TLS tickets.

This PR:
1. Makes PR test builds fail when there's an error adding the PR's branch build as a stub Universe (what's the point of testing a PR against what's on the official Universe, right?)
2. Fixes the issue with Kibana, making the PR branch Kibana build repo be added successfully. I decided to make the python code that parses `$UNIVERSE_URL_PATH` more robust instead of just making the build shell script append to the same line (or change the TC script), but we could do either one of those instead too.

PR #2369, opened 2018-03-10
[INFINITY-3321] Kibana CI install failure (#2361)

gabrielhartmann:
* Make STUB_UNIVERSE_URL parsing more robust.

* Improve logging.

* Fail when repo can't be added.

* Python3.

* Add return statement

### DIFF
--- a/testing/sdk_repository.py
+++ b/testing/sdk_repository.py
@@ -9,11 +9,31 @@ import logging
 import os
 import random
 import string
+from itertools import chain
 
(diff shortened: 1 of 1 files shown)


benclarkwood:
A single Elastic test (kibana + xpack) is expected to fail.

#2370 fixes it, but relies on the cross package functionality added here.

elezar:
Merging as only the expected test failed.

PR #2375, opened 2018-03-13
Filter out resources not intended for this framework

gabrielhartmann:
In order to provide a guarantee of safety regarding offer handling we should not even consider resources which are not intended for us.  [The crux of the logic is here](https://github.com/mesosphere/dcos-commons/compare/resource_filter?expand=1#diff-07dc1ba43f35caa9802d00f9ddf66dc0R154).

If no dynamic reservations have been applied then we can feel free to consume the resource.  If reservations have been applied we should be certain it we applied them.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
@@ -132,6 +132,24 @@ public class ResourceUtils {
         return hasResourceId && matchingRoles;
     }
 
+    public static Collection<Resource.ReservationInfo> getReservations(Resource resource) {
+        Set<Resource.ReservationInfo> reservations =
+                new HashSet<>(resource.getReservationsList().stream().collect(Collectors.toSet()));
+        if (resource.hasReservation()) {
+            reservations.add(resource.getReservation());
+        }
+
+        return reservations;
+    }
+
+    public static Collection<Resource.ReservationInfo> getDynamicReservations(Resource resource) {
+        return getReservations(resource).stream()
+                .filter(reservationInfo -> reservationInfo.hasType())
+                .filter(reservationInfo -> reservationInfo.getType()
+                        .equals(Protos.Resource.ReservationInfo.Type.DYNAMIC))
+                .collect(Collectors.toSet());
+    }
+
     private static Set<String> getRoles(FrameworkInfo frameworkInfo) {
         Set<String> roles = frameworkInfo.getRolesList().stream().collect(Collectors.toSet());
         if (frameworkInfo.hasRole()) {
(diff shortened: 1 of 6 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -132,6 +132,29 @@ public static boolean isOwnedByThisFramework(Resource resource, FrameworkInfo fr
...
+    public static Collection<Resource.ReservationInfo> getDynamicReservations(Resource resource) {
+        return getReservations(resource).stream()
+                .filter(reservationInfo -> reservationInfo.hasType())
+                .filter(reservationInfo -> reservationInfo.getType()
+                        .equals(Protos.Resource.ReservationInfo.Type.DYNAMIC))
+                .collect(Collectors.toSet());
+    }
+
+    public static boolean isProcessable(Protos.Resource resource, Protos.FrameworkInfo frameworkInfo) {
+        boolean hasDynamicReservations = ResourceUtils.getDynamicReservations(resource).size() > 0;
```
benclarkwood: If you wanted to optimize out pulling the dynamic reservations, you could check isOwnedBy first, return if true, then check dynamic reservations.

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/OfferUtilsTest.java
```diff
@@ -82,6 +85,22 @@ public void testDeclineOffers() {
...
     }
 
+    @Test
+    public void testUnreservedResourcesIsProcessable() {
+        Protos.Resource resource = ResourceTestUtils.getUnreservedCpus(1.0);
+        Assert.assertTrue(ResourceUtils.isProcessable(resource, null));
+    }
+
+    @Test
+    public void testReservedByThisFrameworkIsProcessable() {
```
benclarkwood: Test for alien resources aren't processable?

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/OfferUtilsTest.java
```diff
@@ -1,8 +1,10 @@
 package com.mesosphere.sdk.offer;
```
benclarkwood: I may just not be seeing where you test the alien case of isProcessable or clean in OfferUtils?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
gabrielhartmann: They're of a similar time complexity.

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/OfferUtilsTest.java
gabrielhartmann: https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/ResourceUtilsTest.java#L18-L120

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/OfferUtilsTest.java
gabrielhartmann: https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/ResourceUtilsTest.java#L18-L120

PR #2380, opened 2018-03-14
Install botocore first

mellenburg:
if not, python-analytics or another library will
install 2.7.0 and botocore will block usage


test_requirements.txt
```diff
@@ -1,3 +1,4 @@
+botocore
 py==1.5.1
```
elezar: Is this a known botocore issue, or is it one of our other packages?

mellenburg: I don't think its even an "issue". They said there python-dateutil requirement is <2.7.0 and then apparently python-dateutil had a promotion to 2.7 so now that's an issue if anything else is installed before it decides to take latest

elezar:
I'm going to go ahead an merge, since the `dcos-launch` command in the docker container succeeded.

PR #2381, opened 2018-03-14
[ELASTIC-177] Fix Kibana test failure on 1.9.

mpereira:


frameworks/elastic/tests/config.py
```diff
@@ -112,7 +112,7 @@ def wait_for_expected_nodes_to_exist(service_name=SERVICE_NAME, task_count=DEFAU
     stop_max_delay=DEFAULT_ELASTIC_TIMEOUT*1000,
     retry_on_result=lambda res: not res)
 def check_kibana_plugin_installed(plugin_name, service_name=SERVICE_NAME):
-    cmd = "bash -c '$MESOS_SANDBOX/kibana-$ELASTIC_VERSION-linux-x86_64/bin/kibana-plugin list'"
+    cmd = "KIBANA_DIRECTORY=$(ls -d $MESOS_SANDBOX/kibana-*-linux-x86_64); $KIBANA_DIRECTORY/bin/kibana-plugin list"
```
elezar: The `;` is not strictly required, right?

testing/sdk_cmd.py
```diff
@@ -271,12 +271,19 @@ def task_exec(task_name: str, cmd: str, return_stderr_in_stdout: bool = False) -
...
+        # so we inject these manually.
+        environment_variables = [
+            "LIBPROCESS_IP=0.0.0.0",
+            "MESOS_SANDBOX={}".format(sandbox_path)
+        ]
+
+        if cmd.startswith("./"):
+            cmd = os.path.join(sandbox_path, cmd)
+
+        full_cmd = "bash -c '{}; {}'".format(' '.join(environment_variables), cmd)
```
elezar: Is there a semantic difference when including the `;`?

mpereira: Yes. Prepending env var assignments to a command doesn't reflect on the actual command. Bash seems to expand `$A` before the assignment happens. Breaking it into two commands makes it work. 💩 

```
$ dcos task exec kibana bash -c 'A=1 echo $A'

$ dcos task exec kibana bash -c 'A=1; echo $A'
1
```

frameworks/elastic/tests/config.py
mpereira: It is in this case, same issue as in the comment below.

testing/sdk_cmd.py
mpereira: > Have you run this PR against a 1.9 cluster?
I have.

mpereira:
@elezar: Have you run this PR against a 1.9 cluster?

I have.

PR #2385, opened 2018-03-14
Switch to LoggingUtils wrapper for creating Loggers

nickbp:
Adds a wrapper around `LoggingFactory`. To start with, this will result in logs omitting the `com.full.package.path` when displaying the class name, making logs much easier to read. `LoggingFactory`'s `getLogger(Class<?>)` method is hardcoded to the full class name, with package, whereas `getLogger(String)` allows us to customize.
If we want to revisit the format later, we can just do it in `LoggingUtils`.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
@@ -29,7 +29,7 @@ import java.util.stream.Collectors;
  */
 public class CuratorPersister implements Persister {
 
-    private static final Logger logger = LoggingUtils.getLogger(CuratorPersister.class);
+    private static final Logger LOGGER = LoggingUtils.getLogger(CuratorPersister.class);
 
     /**
      * Number of times to attempt an atomic write transaction in storeMany().
@@ -200,7 +200,7 @@ public class CuratorPersister implements Persister {
              * The best we can do is to wipe everything under the root node. A proposed way to "fix" things
              * lives at https://jira.mesosphere.com/browse/INFINITY-1470.
              */
-            logger.debug("Deleting children of root {}", path);
+            LOGGER.debug("Deleting children of root {}", path);
             try {
                 CuratorTransactionFinal transaction = client.inTransaction().check().forPath(serviceRootPath).and();
                 Set<String> pendingDeletePaths = new HashSet<>();
@@ -222,7 +222,7 @@ public class CuratorPersister implements Persister {
             set(unprefixedPath, null);
         } else {
             // Normal case: Delete node itself and any/all children.
-            logger.debug("Deleting {} (and any children)", path);
+            LOGGER.debug("Deleting {} (and any children)", path);
             try {
                 client.delete().deletingChildrenIfNeeded().forPath(path);
             } catch (KeeperException.NoNodeException e) {
@@ -237,7 +237,7 @@ public class CuratorPersister implements Persister {
     @Override
     public void set(String unprefixedPath, byte[] bytes) throws PersisterException {
         final String path = withFrameworkPrefix(unprefixedPath);
-        logger.debug("Setting {} => {}", path, getInfo(bytes));
+        LOGGER.debug("Setting {} => {}", path, getInfo(bytes));
         try {
             try {
                 client.create().creatingParentsIfNeeded().forPath(path, bytes);
@@ -260,7 +260,7 @@ public class CuratorPersister implements Persister {
         for (Map.Entry<String, byte[]> entry : unprefixedPathBytesMap.entrySet()) {
             pathBytesMap.put(withFrameworkPrefix(entry.getKey()), entry.getValue());
         }
-        logger.debug("Updating {} entries: {}", pathBytesMap.size(), pathBytesMap.keySet());
+        LOGGER.debug("Updating {} entries: {}", pathBytesMap.size(), pathBytesMap.keySet());
         runTransactionWithRetries(new SetTransactionFactory(pathBytesMap));
     }
 
@@ -272,7 +272,7 @@ public class CuratorPersister implements Persister {
         Collection<String> paths = unprefixedPaths.stream()
                 .map(unprefixedPath -> withFrameworkPrefix(unprefixedPath))
                 .collect(Collectors.toList());
-        logger.debug("Deleting {} entries: {}", paths.size(), paths);
+        LOGGER.debug("Deleting {} entries: {}", paths.size(), paths);
         runTransactionWithRetries(new ClearTransactionFactory(paths));
     }
 
@@ -289,7 +289,7 @@ public class CuratorPersister implements Persister {
                     } catch (Exception e) {
                         // Transaction failed! Bad connection? Existence check rendered invalid?
                         // Swallow exception and try again
-                        logger.error(String.format("Failed to complete transaction attempt %d/%d: %s",
+                        LOGGER.error(String.format("Failed to complete transaction attempt %d/%d: %s",
                                 i + 1, ATOMIC_WRITE_ATTEMPTS, transaction), e);
                     }
                 } else {
@@ -307,7 +307,7 @@ public class CuratorPersister implements Persister {
         if (unprefixedPaths.isEmpty()) {
             return Collections.emptyMap();
         }
-        logger.debug("Getting {} entries: {}", unprefixedPaths.size(), unprefixedPaths);
+        LOGGER.debug("Getting {} entries: {}", unprefixedPaths.size(), unprefixedPaths);
 
         Map<String, byte[]> result = new TreeMap<>();
         // Unlike with writes, there is not an atomic read operation. Therefore we wing it with a series of plain reads.
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
@@ -1,6 +1,7 @@
 package com.mesosphere.sdk.curator;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.mesosphere.sdk.offer.LoggingUtils;
 import com.mesosphere.sdk.specification.ServiceSpec;
 import com.mesosphere.sdk.storage.Persister;
 import com.mesosphere.sdk.storage.PersisterException;
@@ -16,7 +17,6 @@ import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.ZooDefs;
 import org.apache.zookeeper.data.ACL;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.nio.charset.StandardCharsets;
 import java.util.*;
@@ -29,7 +29,7 @@ import java.util.stream.Collectors;
  */
 public class CuratorPersister implements Persister {
 
-    private static final Logger logger = LoggerFactory.getLogger(CuratorPersister.class);
+    private static final Logger logger = LoggingUtils.getLogger(CuratorPersister.class);
 
     /**
      * Number of times to attempt an atomic write transaction in storeMany().
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointsResource.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointsResource.java
@@ -14,6 +14,7 @@ import javax.ws.rs.core.Response;
 import com.google.protobuf.TextFormat;
 import com.mesosphere.sdk.http.types.EndpointProducer;
 import com.mesosphere.sdk.offer.Constants;
+import com.mesosphere.sdk.offer.LoggingUtils;
 import com.mesosphere.sdk.offer.TaskException;
 import com.mesosphere.sdk.offer.taskdata.AuxLabelAccess;
 import com.mesosphere.sdk.offer.taskdata.TaskLabelReader;
@@ -29,14 +30,13 @@ import org.apache.mesos.Protos.TaskStatus;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * A read-only API for accessing information about how to connect to the service.
  */
 @Path("/v1/endpoints")
 public class EndpointsResource {
-    private static final Logger LOGGER = LoggerFactory.getLogger(EndpointsResource.class);
+    private static final Logger LOGGER = LoggingUtils.getLogger(EndpointsResource.class);
 
     private static final String RESPONSE_KEY_DNS = "dns";
     private static final String RESPONSE_KEY_ADDRESS = "address";
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
@@ -9,8 +9,6 @@ import org.slf4j.LoggerFactory;
  */
 public class LoggingUtils {
 
-    private static boolean loggingNamesEnabled = true;
-
     private LoggingUtils() {
         // do not instantiate
     }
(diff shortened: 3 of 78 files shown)


benclarkwood:
@nickbp can you provide an example log line or two?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Creates a logger which is tagged with the provided class.
+     *
+     * @param clazz the class using this logger
+     */
+    public static Logger getLogger(Class<?> clazz) {
+        // Note: This currently results in a name like "LoggingUtils" in logs.
+        // Consider including an abbreviated package prefix, e.g. "c.m.s.o.LoggingUtils"?
+        return LoggerFactory.getLogger(clazz.getSimpleName());
```
benclarkwood: Fun JVM trivia, getSimpleName() is uncached string manipulation. Ask me how I know?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Creates a logger which is tagged with the provided class.
+     *
+     * @param clazz the class using this logger
+     */
+    public static Logger getLogger(Class<?> clazz) {
+        // Note: This currently results in a name like "LoggingUtils" in logs.
+        // Consider including an abbreviated package prefix, e.g. "c.m.s.o.LoggingUtils"?
```
benclarkwood: I do think it is worth having an abbreviated prefix. We might have collisions otherwise.

Something like this (I can't pass up an opportunity to do lambdas...)

```
    public static Logger getLogger(Class<?> clazz) {
        // Note: This currently results in a name like "LoggingUtils" in logs.
        // Consider including an abbreviated package prefix, e.g. "c.m.s.o.LoggingUtils"?
        return getLogger(getAbbreviatedClassName(clazz));
    }

    /**
     * Creates a logger which is tagged with the provided class and the provided custom label.
     *
     * @param clazz the class using this logger
     * @param name  an additional context label detailing e.g. the name of the service being managed
     */
    public static Logger getLogger(Class<?> clazz, String name) {
        if (StringUtils.isEmpty(name) || !loggingNamesEnabled) {
            return getLogger(clazz);
        } else {
            return getLogger(String.format("(%s) %s", name, getAbbreviatedClassName(clazz)));
        }
    }

    private static Logger getLogger(String name) {
        return LoggerFactory.getLogger(name);
    }

    private static String getAbbreviatedClassName(Class<?> clazz) {
        String fullName = clazz.getName();
        List<String> nameParts = Arrays.asList(fullName.split("."));
        String prefix = nameParts.subList(0, nameParts.size())
                .stream()
                    .map(namePart -> namePart.substring(0, 1).toLowerCase())
                    .collect(Collectors.joining(","));

        return String.format("%s.%s", prefix, nameParts.get(nameParts.size() - 1));
    }

```

benclarkwood: It's definitely possible there exists something for this already...

benclarkwood: https://stackoverflow.com/questions/7891910/log4j-abbreviate-shorten-package-names

Hm, we can probably just configure our logger to do this.

benclarkwood:
```
%c{1.}	org.apache.commons.Foo	results in o.a.c.Foo
```

nickbp: Are we using apache logging on the backend?

nickbp: Also, I'd be somewhat surprised that would work, as `LoggerFactory` itself hardcodes the label to `class.getName()`: https://www.slf4j.org/api/src-html/org/slf4j/LoggerFactory.html#line.379

nickbp: Oh hey, it works!

nickbp: So, with package vs without. Personally I'm still a little partial to without:

```
INFO  2018-03-14 17:26:06,728 [Test worker] c.m.s.s.p.DefaultPlanCoordinator:getCandidates(73): Updated dirtied assets: []
INFO  2018-03-14 17:26:06,728 [Test worker] c.m.s.s.p.DefaultPlanCoordinator:getCandidates(79): Got total candidates: [unreserve-e3e6d735-4dde-4b23-80ff-e73b1cd0de2f]
INFO  2018-03-14 17:26:06,728 [Test worker] c.m.s.s.AbstractScheduler:processQueuedOffers(429): InProgress Steps: []
INFO  2018-03-14 17:26:06,728 [Test worker] c.m.s.s.ReviveManager:revive(72): Candidates, old: [kill-world-1-server [N/A]], current: [unreserve-e3e6d735-4dde-4b23-80ff-e73b1cd0de2f [N/A]], new:[unreserve-e3e6d735-4dde-4b23-80ff-e73b1cd0de2f [N/A]]
WARN  2018-03-14 17:26:06,728 [Test worker] c.m.s.s.ReviveManager:revive(90): Revive attempt has been throttled.
INFO  2018-03-14 17:26:06,728 [Test worker] c.m.s.s.AbstractScheduler:processQueuedOffers(437): Processing 1 offer against 1 step:
INFO  2018-03-14 17:26:06,729 [Test worker] c.m.s.s.AbstractScheduler:processQueuedOffers(441):   1: AN OFFER HERE
INFO  2018-03-14 17:26:06,729 [Test worker] c.m.s.s.p.DefaultPlanScheduler:resourceOffers(71): Processing resource offers for step: unreserve-e3e6d735-4dde-4b23-80ff-e73b1cd0de2f
INFO  2018-03-14 17:26:06,729 [Test worker] c.m.s.s.u.ResourceCleanupStep:start(31): Setting state to Prepared for resource e3e6d735-4dde-4b23-80ff-e73b1cd0de2f
```

```
INFO  2018-03-14 17:26:06,728 [Test worker] DefaultPlanCoordinator:getCandidates(73): Updated dirtied assets: []
INFO  2018-03-14 17:26:06,728 [Test worker] DefaultPlanCoordinator:getCandidates(79): Got total candidates: [unreserve-e3e6d735-4dde-4b23-80ff-e73b1cd0de2f]
INFO  2018-03-14 17:26:06,728 [Test worker] AbstractScheduler:processQueuedOffers(429): InProgress Steps: []
INFO  2018-03-14 17:26:06,728 [Test worker] ReviveManager:revive(72): Candidates, old: [kill-world-1-server [N/A]], current: [unreserve-e3e6d735-4dde-4b23-80ff-e73b1cd0de2f [N/A]], new:[unreserve-e3e6d735-4dde-4b23-80ff-e73b1cd0de2f [N/A]]
WARN  2018-03-14 17:26:06,728 [Test worker] ReviveManager:revive(90): Revive attempt has been throttled.
INFO  2018-03-14 17:26:06,728 [Test worker] AbstractScheduler:processQueuedOffers(437): Processing 1 offer against 1 step:
INFO  2018-03-14 17:26:06,729 [Test worker] AbstractScheduler:processQueuedOffers(441):   1: AN OFFER HERE
INFO  2018-03-14 17:26:06,729 [Test worker] DefaultPlanScheduler:resourceOffers(71): Processing resource offers for step: unreserve-e3e6d735-4dde-4b23-80ff-e73b1cd0de2f
INFO  2018-03-14 17:26:06,729 [Test worker] ResourceCleanupStep:start(31): Setting state to Prepared for resource e3e6d735-4dde-4b23-80ff-e73b1cd0de2f
```

Here are all duplicate .java files in dcos-commons:

```
dcos-commons$ find . -iname *.java -exec basename {} \; | sort | uniq -c | grep -v "      1 "
      2 LaunchedTask.java
      8 Main.java
      6 ServiceTest.java
```
For `LaunchedTask`, there's one in `executor` (custom executor) and one in `testing` (service tests).

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -29,7 +29,7 @@
  */
 public class CuratorPersister implements Persister {
 
-    private static final Logger logger = LoggerFactory.getLogger(CuratorPersister.class);
+    private static final Logger logger = LoggingUtils.getLogger(CuratorPersister.class);
```
elezar: A bit of a consistency nit. Should we use `LOGGER` or `logger` everywhere?

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointsResource.java
```diff
@@ -29,14 +30,13 @@
...
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * A read-only API for accessing information about how to connect to the service.
  */
 @Path("/v1/endpoints")
 public class EndpointsResource {
-    private static final Logger LOGGER = LoggerFactory.getLogger(EndpointsResource.class);
+    private static final Logger LOGGER = LoggingUtils.getLogger(EndpointsResource.class);
```
elezar: Another inconsistency / question. What is the difference between using `.class` and `getClass()`?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Creates a logger which is tagged with the provided class and the provided custom label.
+     *
+     * @param clazz the class using this logger
+     * @param name  an additional context label detailing e.g. the name of the service being managed
+     */
+    public static Logger getLogger(Class<?> clazz, String name) {
+        if (StringUtils.isEmpty(name) || !loggingNamesEnabled) {
```
elezar: Nit: `isBlank` would handle the case where `name` is just whitespace.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Disables logger names in output.
+     *
+     * @see #getLogger(Class, String)
+     */
+    public static void disableNames() {
+        loggingNamesEnabled = false;
+    }
+}
```
elezar: Newline.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
```diff
@@ -0,0 +1,51 @@
...
+            return LoggerFactory.getLogger(String.format("(%s) %s", name, clazz.getSimpleName()));
+        }
+    }
+
+    /**
+     * Disables logger names in output.
+     *
+     * @see #getLogger(Class, String)
+     */
+    public static void disableNames() {
```
elezar: Does it really make sense to have this helper? I didn't see anywhere where we call `disableNames()` (or call `getLogger()` twice in the same class -- which is the only time that changing this setting would have effect.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
```diff
@@ -32,7 +31,6 @@ public ExecutorResourceMapper(
             Collection<ResourceSpec> resourceSpecs,
             Protos.ExecutorInfo executorInfo,
             boolean useDefaultExecutor) {
-        this.executorInfo = executorInfo;
         this.volumeSpecs = podSpec.getVolumes();
```
elezar: Is the member unused? Would it make sense to create a follow-up ticket to change the constructor argument list? I assume that it makes sense since `ExecutorResourceMapper` doesn't implement or extend anything.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointsResource.java
benclarkwood: In some classes, we use a private static final logger and use `.class`.

In other classes, we use a per instance logger and use `.getClass()`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
benclarkwood: Yeah, looking at the two side by side the abbreviated is kinda terrible looking.

I'm okay with just using `simpleName()` for now.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointsResource.java
nickbp: Another difference, in the case of an abstract class, is that `getClass()` will result in logging as the final concrete class while `.class` will result in logging as the abstract class.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
nickbp: The `executorInfo`'s resources are extracted a couple lines down from here. It's just the `executorInfo` member itself that was unused.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
nickbp: In theory any `public static final` should be `LOGGER` while just `public final` should be `logger`. It looks like there are a few cases of `public static final logger`, so I'll update those. Didn't see any cases of the opposite.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
nickbp: Going with getSimpleName(). Now that we have the utility used everywhere, we can always revisit later

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/LoggingUtils.java
nickbp: Removed this for now. In theory it'd be a global switch to disable names, but I'm not so sure it'd actually work properly anyway given that many Logger instances could have been statically created before it was called.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
elezar: sorry missed the differences. Must have compared two inconsistent ones.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointsResource.java
elezar: Once again I may have been comparing a subset. Thanks for the clarification.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
elezar: Me and my lazyness in expanding github diffs.

nickbp:
Yes that's fixing a linter warning

PR #2393, opened 2018-03-16
Enable region support in Cassandra

mrbrowning:
This change is based off the SDK region awareness change which I made a small change to and am waiting for tests to pass again. Once that's merged, I'll rebase. The relevant changes here are only in the Cassandra framework.

### DIFF
--- a/frameworks/cassandra/universe/config.json
+++ b/frameworks/cassandra/universe/config.json
@@ -84,7 +84,7 @@
           "default": "serial"
         },
         "region": {
-          "description": "The name of the region that nodes in the Cassandra cluster should run in.",
+          "description": "The Cassandra cluster should run in this region.  When no region is specified the cluster is constrained to the local region.",
           "type": "string",
           "default": ""
         },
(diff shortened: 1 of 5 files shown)


frameworks/cassandra/universe/config.json
```diff
@@ -83,6 +83,11 @@
           ],
           "default": "serial"
         },
+        "region": {
+          "description": "The name of the region that nodes in the Cassandra cluster should run in.",
```
gabrielhartmann: Ending a sentence with a preposition

gabrielhartmann:
Ideally we'll have some integration tests with this.  Still working on cluster generation with regions?

frameworks/cassandra/universe/config.json
gabrielhartmann:
```
The Cassandra cluster should run in this region.
```

frameworks/cassandra/universe/config.json
```diff
@@ -83,6 +83,11 @@
           ],
           "default": "serial"
         },
+        "region": {
```
elezar: From a customer UX perspective, would it make sense to have the "region" field and the "rack" and "data_center" fields adjacent to each other, as they both have to do with placement of the service. This is of course something that can be changed after the fact.

PR #2396, opened 2018-03-19
Ensure service does not remain installed after multi-region tests

mrbrowning:


frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -14,18 +14,30 @@
 POD_NAMES = ['hello-0', 'world-0', 'world-1']
 
 
+def with_service_installation(additional_options):
```
elezar: This could just be a `pytest.fixture`. The default scope is per test. https://stackoverflow.com/questions/22627659/run-code-before-and-after-each-test-in-py-test

(These can also be parameterized).

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -14,18 +14,38 @@
 POD_NAMES = ['hello-0', 'world-0', 'world-1']
 
 
-@pytest.mark.dcos_min_version('1.11')
-@pytest.mark.region_awareness
-@sdk_utils.dcos_ee_only
-def test_nodes_deploy_to_local_region_by_default():
+@pytest.fixture
+def local_service():
     sdk_install.install(
```
elezar: We do need the `try - finally` block here so that uninstall is run even if the install fails.

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -14,18 +14,45 @@
...
+def local_service():
+    try:
+        sdk_install.install(
+            config.PACKAGE_NAME,
+            config.SERVICE_NAME,
+            3,
+            additional_options={
+                "service": {"scenario": "MULTI_REGION", "allow_region_awareness": "true"}
+            })
+        sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
```
elezar: `sdk_install.install` already waits for completed deployment. Is there a specific reason to call this again?

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -14,18 +14,45 @@
...
+
+        yield
+    finally:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.fixture
+def remote_service():
```
elezar: The `try - finally` block is needed for this installation too.

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -21,7 +21,9 @@ def local_service():
             config.PACKAGE_NAME,
             config.SERVICE_NAME,
             3,
-            additional_options={"service": {"scenario": "MULTI_REGION"}})
+            additional_options={
+                "service": {"scenario": "MULTI_REGION", "allow_region_awareness": "true"}
```
elezar: Some notes:
* Formatting the python dict with one entry per line improves readability and also makes future diffs easier to process.
* The `allow_region_awareness` field is a `boolean` value, and should be `True` and not `"true"`.
* Is it *required* to specify the `allow_region_awareness` field? The default value is `true` and as such it may be better to leave it out here.

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -35,7 +37,11 @@ def remote_service():
         config.PACKAGE_NAME,
         config.SERVICE_NAME,
         3,
-        additional_options={"service": {"scenario": "MULTI_REGION", "region": "Europe"}})
+        additional_options={
+            "service": {
+                "scenario": "MULTI_REGION", "allow_region_awareness": "true", "region": "Europe"
```
elezar: See the comments above for the other `additional_options` dict.

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -60,19 +79,11 @@ def test_nodes_can_deploy_to_remote_region():
...
 @sdk_utils.dcos_ee_only
-def test_region_config_update_does_not_succeed():
-    sdk_install.install(
-        config.PACKAGE_NAME,
-        config.SERVICE_NAME,
-        3,
-        additional_options={"service": {"scenario": "MULTI_REGION"}})
-
+def test_region_config_update_does_not_succeed(local_service):
     sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
```
elezar: There should be no need to wait for completed deployment here.

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -14,18 +14,45 @@
...
+
+    sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.mark.dcos_min_version('1.11')
+@pytest.mark.region_awareness
+@sdk_utils.dcos_ee_only
+def test_nodes_deploy_to_local_region_by_default(local_service):
     for pod_name in POD_NAMES:
         info = sdk_cmd.service_request(
```
elezar: I still think that a function:
```python
def get_pod_region(service_name: str, pod_name: str) -> bool:
    r = sdk_cmd.service_request('GET', service_name, '/v1/pod/{}/info'.format(pod_name))

    info = r.json()[0]['info']

    return [l['value'] for l in info['labels']['labels'] if l['key'] == 'offer_region'][0]
```
would give us some utility. For example:
```python
    for pod_name in POD_NAMES:
        pod_region = get_pod_region(config.SERVICE_NAME, pod_name)
        expected_region = "Europe"
        assert pod_region == expected_region, "Region mismatch for pod {}. " \
                                              "Expected: {} Found: {}".format(pod_name, expected_region, pod_region)
```

Some notes:
* Should the "USA" and "Europe" regions be global constants (`LOCAL_REGION` and `REMOTE_REGION`)?
* Where does the default of "USA" come from? Is this part of the cluster definition?
* What behaviour do we expect when the pod info has:
    * No region label? (Currently we're going to get an index out of bounds exception
    * Multiple region labels?

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -89,8 +100,6 @@ def test_region_config_update_does_not_succeed():
         [l['value'] for l in info['labels']['labels'] if l['key'] == 'offer_region'][0] == 'USA'
     )
 
-    sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
-
 
 def change_region_config(region_name):
     service_config = sdk_marathon.get_config(config.SERVICE_NAME)
```
elezar: Should we not wait for a triggered and completed deployment in this case?

frameworks/helloworld/tests/test_region_awareness.py
```diff
@@ -60,19 +79,11 @@ def test_nodes_can_deploy_to_remote_region():
...
-    sdk_install.install(
-        config.PACKAGE_NAME,
-        config.SERVICE_NAME,
-        3,
-        additional_options={"service": {"scenario": "MULTI_REGION"}})
-
+def test_region_config_update_does_not_succeed(local_service):
     sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
     change_region_config('Europe')
     plan = sdk_plan.get_deployment_plan(config.SERVICE_NAME)
```
elezar: If we expect the region change to fail, would we not expect the config validation errors to show up in the error list?

mrbrowning: That's what `assert plan.get('errors', [])` does -- if the list is empty, the value is falsy, and the assert will fail.

frameworks/helloworld/tests/test_region_awareness.py
mrbrowning: I don't think so, why?

frameworks/helloworld/tests/test_region_awareness.py
elezar: Sorry, my bad. Would it make sense to check for an explicit error message instead of relying on any error?

frameworks/helloworld/tests/test_region_awareness.py
elezar: Sorry, I might have been referring to the wrong line here.

elezar:
@mrbrowning I merged the changes so that we can unblock CI.

PR #2398, opened 2018-03-19
Set region awareness envvar to true in Cassandra when region is specified

mrbrowning:


frameworks/cassandra/universe/marathon.json.mustache
```diff
@@ -74,6 +74,7 @@
     "VIRTUAL_NETWORK_PLUGIN_LABELS": "{{service.virtual_network_plugin_labels}}",
     {{/service.virtual_network_enabled}}
     {{#service.region}}
+    "ALLOW_REGION_AWARENESS": "true",
```
gabrielhartmann: Why have this envvar at all if it's hard-coded to true?  Couldn't it just be added to Cassandra's Scheduler config programmatically?

mrbrowning:
@gabrielhartmann just to save a step, since the one implies the other, and otherwise the user would have to specify some `config.allow_region_awareness` in addition to specifying a remote region if they wanted to run in one.

frameworks/cassandra/universe/marathon.json.mustache
mrbrowning: Oh, I see what you mean. I suppose it could.

mrbrowning: Well, `SchedulerConfig` would have to be modified to support overrides.

PR #2399, opened 2018-03-20
Clean up logic in region matching and report region mismatch correctly

mrbrowning:
This fixes error reporting when a region config update fails to validate. Unfortunately, localizing creation of `IsLocalRegionRule` to `RegionRuleFactory` is not compatible with the current interface to `RuleFactory`, and so I'll defer that to a later cleanup.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/RegionCannotChange.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/RegionCannotChange.java
@@ -21,17 +21,10 @@ public class RegionCannotChange implements ConfigValidator<ServiceSpec> {
         }
 
         List<ConfigValidationError> errors = new ArrayList<>();
-        boolean regionWasAdded = oldConfig.isPresent() && !oldConfig.get().getRegion().isPresent() &&
-                newConfig.getRegion().isPresent();
-        boolean regionWasUnset = oldConfig.isPresent() && oldConfig.get().getRegion().isPresent() &&
-                !newConfig.getRegion().isPresent();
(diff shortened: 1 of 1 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/RegionCannotChange.java
```diff
@@ -21,17 +21,10 @@
...
-        boolean regionWasChanged = oldConfig.isPresent() && oldConfig.get().getRegion().isPresent() &&
-                newConfig.getRegion().isPresent() &&
-                !newConfig.getRegion().get().equals(oldConfig.get().getRegion().get());
-        if (regionWasAdded || regionWasUnset || regionWasChanged) {
+        if (!oldConfig.get().getRegion().equals(newConfig.getRegion())) {
             errors.add(ConfigValidationError.transitionError(
                     "region",
-                    oldConfig.get().getUser(),
+                    oldConfig.get().getRegion().toString(),
                     newConfig.getUser(),
```
elezar: The `newConfig` region should also be used.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/RegionCannotChange.java
```diff
@@ -21,17 +21,10 @@
...
                     "Region for old service must remain the same across deployments."
```
elezar: What does "old" service mean in this context?

mrbrowning: It means the `ServiceSpec` generated from the last config target, if any. That's the name that's used in the interface definition.

PR #2404, opened 2018-03-20
[INFINITY-3363] Add retry on re-resolution test (#2395)

benclarkwood:
* Add retry on re-resolution test

* Improve ZK reresolution check

* Check for failure followed by success

* Only retry three times


elezar:
Looks like the check is not 100% yet...

elezar:
I am closing this as we should backport #2407 if that succeeds.

PR #2405, opened 2018-03-21
[INFINITY-3396] Add service TLD top level override

benclarkwood:
This PR accomplishes a single goal: Make it possible for a SDK user to inject a custom TLD for their service. 

This supports the use case of exposing the service outside the cluster with custom DNS (a lot of custom DNS setup is required on the operator's part).

Changes:
- Route SchedulerConfig to the right places
- Add an accessor to SchedulerConfig for a custom service tld
- Make EndpointUtils use the custom TLD 
- Add an integration test in HWorld for the behavior

WIP:
I accidentally left in some misc Kafka changes. Need to remove those.


frameworks/helloworld/universe/config.json
```diff
@@ -112,6 +113,11 @@
           "description": "Allow region awareness",
           "type": "boolean",
           "default": true
+        },
+        "custom_service_tld": {
```
elezar: Would `custom_top_level_domain` not be better here? Duplicating `service` may be unnecessary?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/security/CertificateNamesGenerator.java
```diff
@@ -41,15 +42,19 @@
...
 
-    public CertificateNamesGenerator(String serviceName, TaskSpec taskSpec, PodInstance podInstance) {
+    public CertificateNamesGenerator(String serviceName,
+                                     TaskSpec taskSpec,
+                                     PodInstance podInstance,
+                                     SchedulerConfig schedulerConfig) {
         this.serviceName = serviceName;
         this.taskInstanceName = TaskSpec.getInstanceName(podInstance, taskSpec);
         // Task can specify its own service discovery name
         if (taskSpec.getDiscovery().isPresent() && taskSpec.getDiscovery().get().getPrefix().isPresent()) {
```
elezar: This seems to indicate that we already have some form of discovery in the `taskSpec`. Could we not use that for the custom TLD too?

testing/sdk_hosts.py
```diff
@@ -14,7 +14,7 @@
 
 
 SYSTEM_HOST_SUFFIX = 'mesos'
-AUTOIP_HOST_SUFFIX = 'autoip.dcos.thisdcos.directory'
+AUTOIP_HOST_SUFFIX = 'na2c.dcos.fakejpmc.net'
```
elezar: we may not want to have `jpmc` here. Also, this breaks all tests which are not using custom TLDs.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
```diff
@@ -84,7 +86,7 @@ private static String renderTemplate(File configFile, String serviceName) throws
         // Simulate the envvars that would be passed to a task. We want to render the config as though it was being done
         // in a task. Manually copy over a couple envvars which is included in tasks by default.
         Map<String, String> env = new HashMap<>(new TaskEnvRouter().getConfig("ALL"));
-        env.put(EnvConstants.FRAMEWORK_HOST_TASKENV, EndpointUtils.toAutoIpDomain(serviceName));
+        env.put(EnvConstants.FRAMEWORK_HOST_TASKENV, EndpointUtils.toAutoIpDomain(serviceName, schedulerConfig));
```
elezar: Out of scope of this PR: It seems strange to have these environment variables set here as well as in https://github.com/mesosphere/dcos-commons/pull/2405/files#diff-01c250e232e8ece8351364fa59a4cfd5R398

frameworks/helloworld/src/main/dist/custom_tld.yml
```diff
@@ -0,0 +1,27 @@
+name: {{FRAMEWORK_NAME}}
+scheduler:
+  principal: {{SERVICE_PRINCIPAL}}
+  user: {{SERVICE_USER}}
+pods:
+  hello:
+    count: 1
+    uris:
+      - {{BOOTSTRAP_URI}}
```
elezar: Future note: Should we think about adding this URI automatically?

frameworks/helloworld/tests/test_custom_service_tld.py
```diff
@@ -0,0 +1,51 @@
...
+    # all TLD usage with a different TLD.
+
+    # A service yaml is used which will make sure task DNS resolves and then sleep so advertised endpoints
+    # can be checked.
+    custom_tld = "autoip.dcos.{}.dcos.directory".format(crypto_id.strip())
+    sdk_install.install(
+        config.PACKAGE_NAME,
+        config.SERVICE_NAME,
+        1,
+        additional_options={"service": {
```
elezar: Could we split the lines in the dict to make future diffs easier to process.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
```diff
@@ -42,24 +46,31 @@ public static String toEndpoint(String hostname, int port) {
     /**
      * Returns the correct DNS domain for tasks within the service.
      */
-    public static String toAutoIpDomain(String serviceName) {
+    public static String toAutoIpDomain(String serviceName, SchedulerConfig schedulerConfig) {
+        Optional<String> customTLD = schedulerConfig.getCustomServiceTLD();
```
elezar: Would it make sense to move this logic into the `SchedulerConfig` instance? i.e that we have a `SchedulerConfig.getTLD()` method that returns the default or the custom one.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -375,7 +378,7 @@ public void setExecutorVolume(VolumeSpec volumeSpec) {
      */
     @VisibleForTesting
     public static Map<String, String> getTaskEnvironment(
-            String serviceName, PodInstance podInstance, TaskSpec taskSpec) {
+            String serviceName, PodInstance podInstance, TaskSpec taskSpec, SchedulerConfig schedulerConfig) {
```
elezar: In terms of the hierarchy in the `svc.yml` file, we have:
* service (`name`)
* scheduler
* pods
* tasks

Would it make sense to move the `schedulerConfig` parameter to after `serviceName` to better reflect this?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/security/CertificateNamesGenerator.java
```diff
@@ -41,15 +42,19 @@
     // ub-common-name INTEGER ::= 64
     private static final int CN_MAX_LENGTH = 64;
 
-    public CertificateNamesGenerator(String serviceName, TaskSpec taskSpec, PodInstance podInstance) {
+    public CertificateNamesGenerator(String serviceName,
```
elezar: The same comment here as for `getTaskEnvironment ` w.r.t the ordering of the parameters.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -351,6 +356,13 @@ public boolean isregionAwarenessEnabled() {
         return Boolean.valueOf(envStore.getOptional(ALLOW_REGION_AWARENESS_ENV, "false"));
     }
 
+    /**
+     * Returns an optional of the custom Service TLD.
+     */
+    public Optional<String> getCustomServiceTLD() {
```
elezar: As already mentioned, would a `getTLD` method that returns a `String` make sense instead?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
benclarkwood: SchedulerConfig is the class for accessing the environment of the scheduler. It isn't a part of the yaml.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
benclarkwood: It would split the two kinds of TLDs (non-vip and vip).

VIPs will never work outside of a cluster so it doesn't make sense to support a custom TLD for them.

frameworks/helloworld/universe/config.json
benclarkwood: I was matching it to the name of the envvar used

frameworks/helloworld/tests/test_custom_service_tld.py
benclarkwood: Yes

frameworks/helloworld/src/main/dist/custom_tld.yml
benclarkwood: Yes

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
benclarkwood: It's because they aren't currently injected into the scheduler, and this requires rendering a template.

benclarkwood: (Arguably we could always inject them into the scheduler)

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
benclarkwood: I don't want to split half of the functionality of EndpointUtils into SchedulerConfig.

And so here it needs to be an Optional as the _absence_ of the custom TLD is meaningful.

benclarkwood:
I used the SchedulerConfig as I believe at present a sneaky Envvar is the correct interface for this.

If we decide to make it more "official" we'd probably move it into the yaml. Maybe.

frameworks/helloworld/universe/config.json
elezar: We could have the envvar be `SERVICE_CUSTOM_TLD` in the same way that we use `SECURITY_KERBEROS_ENABLED` to also (partially) encode the json heirarchy.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
elezar: I don't quite see how this has anything to do with `VIPs`. The only use of `Constants.DNS_TLD`  in this class is here. This is also the only place where we call the `getCustomServiceTLD` method.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
elezar: Ok. The name was a bit confusing. I suppose `SchedulerMarathonEnv` would be more accurate. (not that it should change now).

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/security/CertificateNamesGenerator.java
elezar: Also, while we're making the change, could we use:
```java
        String taskName = taskSpec.getDiscovery().flatMap(d -> d.getPrefix())
                .map(prefix -> String.format("%s-%d", prefix, podInstance.getIndex()))
                .orElse(this.taskInstanceName);
        this.autoIpHostname = EndpointUtils.toAutoIpHostname(serviceName, taskName, schedulerConfig);
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
elezar: OK. It was just not clear that the absence was meaningful, as the only time the absence is used to make a decision is to revert to the default.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
benclarkwood: Ah, I wasn't getting what ya'll were putting down.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
```diff
@@ -42,24 +46,31 @@ public static String toEndpoint(String hostname, int port) {
...
     /**
      * Returns the correct DNS hostname:port endpoint for the provided task and port running within the service.
      */
-    public static String toAutoIpEndpoint(String serviceName, String taskName, int port) {
-        return toEndpoint(toAutoIpHostname(serviceName, taskName), port);
+    public static String toAutoIpEndpoint(String serviceName,
+                                          String taskName,
+                                          int port,
+                                          SchedulerConfig schedulerConfig) {
+        return toEndpoint(toAutoIpHostname(serviceName, taskName, schedulerConfig), port);
```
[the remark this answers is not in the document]
benclarkwood: I will defer that to some future work. But I agree that it makes sense to have.

frameworks/helloworld/src/main/dist/custom_tld.yml
elezar: True. Not all pods need it. I was letting the lines between pods and the scheduler (where we now always use it) blur a little.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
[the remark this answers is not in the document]
elezar: @nickbp could you create a jira ticket referencing this PR and the discussion on data-agility today?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/security/CertificateNamesGenerator.java
benclarkwood: That would require first classing this in the YAML.

I am hesitant to do so. Leaving it as a sneaky envvar keeps it secreter.

benclarkwood:
@nickbp @elezar Review feedback incorporated

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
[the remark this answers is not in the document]
nickbp: Filed.

frameworks/helloworld/universe/config.json
elezar: FWIW, I still think `service.custom_tld` is better than `service.custom_service_tld`, but I'm not adamant about it.

PR #2409, opened 2018-03-22
INFINITY-3409 Persistent storage changes: Namespacing, FrameworkStore, Schema check

nickbp:
- Support namespacing in `StateStore`/`ConfigStore`.
- Move FrameworkId storage from `StateStore` into a separate `FrameworkStore` for which namespacing does not apply.
- Remove erasing storage from `StateStore` and do it at the `Persister` level instead. Now that `StateStore` can be namespaced, erasing storage there would be confusing.
- Move schema version check upstream into the `SchedulerRunner`, instead of doing it in `StateStore`/`ConfigStore` (once each). This allows upstream to determine the expected schema version
- Clean up storage initialization in SchedulerBuilder: Just allow upstream to provide the `Persister`, and build `*Store`s based on that `Persister`.


sdk/scheduler/src/main/java/com/mesosphere/sdk/state/FrameworkStore.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param persister The persister which holds the data
+     */
+    public FrameworkStore(Persister persister) {
+        this.persister = persister;
+    }
+
+    /**
+     * Returns the underlying {@link Persister}.
+     */
+    public Persister getPersister() {
```
benclarkwood: When would something need to access the persister?

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/SchemaVersionStore.java
```diff
@@ -65,36 +52,44 @@
      * compatible. Note that this may auto-populate the underlying schema version if the value isn't
      * currently present.
      *
-     * @return the current schema version, which may be auto-populated with the current version
+     * @param expectedVersion the expected schema version to be stored if no version is currently set or to be checked
```
benclarkwood: Java doc summary is stale

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreUtils.java
```diff
@@ -215,14 +216,17 @@ static void repairTaskIDs(StateStore stateStore) {
...
+     * uninstalling.
      */
     public static boolean isUninstalling(StateStore stateStore) throws StateStoreException {
         return fetchBooleanProperty(stateStore, UNINSTALLING_PROPERTY_KEY);
     }
 
     /**
-     * Sets an 'uninstall' property in the provided {@link StateStore} to {@code true}.
+     * Sets an 'uninstall' property in the provided {@link StateStore} to {@code true}. If the {@link StateStore} was
+     * created against a namespaced service, then this returns whether that service is uninstalling.
```
benclarkwood: I don't see a return? Copy paste error I think...

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/FrameworkStore.java
nickbp: I think this (will) come up when we're clearing the persister data during uninstall, or something like that.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStoreUtils.java
nickbp: Fixed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/SchemaVersionStore.java
nickbp: Updated javadocs across the class.

PR #2412, opened 2018-03-22
INFINITY-3407/3408 Add service name to task id and reservation labels

nickbp:
Updates offer evaluation to embed service name in a couple places:
- Add service name to the start of the task id (TODO see how it behaves with folders in integration tests...)
- Add service name to reservation labels in reserved resources (adjacent to `resource_id`)

Other changes:
- Remove unused transient label: It wasn't even being written because we forgot a `setLabels()` call
- Include service name in some loggers, more will be added later

PR #2419, opened 2018-03-26
[INFINITY-2971] improve kafka authz test coverage

elezar:
This PR extends the AuthZ tests for Kafka and includes:

* Adds the same authz tests that exist for SSL auth to Kerberos auth.
* Adds the same authz tests that exist for SSL auth to Kerberos auth with a custom Kerberized ZooKeeper
* Add the same authz tests that exist for SSL auth to SSL-Kerberos auth

It also: 
* applies some minor flake8 corrections to the kafka tests.
* Correctly replaces the `wait_for_brokers` function with `sdk_cmd.resolve_hosts` (see #2412)

### DIFF
--- a/frameworks/kafka/src/main/dist/server.properties.mustache
+++ b/frameworks/kafka/src/main/dist/server.properties.mustache
@@ -92,9 +92,11 @@ socket.request.max.bytes={{KAFKA_SOCKET_REQUEST_MAX_BYTES}}
 authorizer.class.name=kafka.security.auth.SimpleAclAuthorizer
 super.users={{SETUP_HELPER_SUPER_USERS}}
 allow.everyone.if.no.acl.found={{SECURITY_AUTHORIZATION_ALLOW_EVERYONE_IF_NO_ACL_FOUND}}
+{{^SECURITY_KERBEROS_ENABLED}}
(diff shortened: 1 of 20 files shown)


frameworks/kafka/src/main/dist/server.properties.mustache
```diff
@@ -92,9 +92,11 @@ socket.request.max.bytes={{KAFKA_SOCKET_REQUEST_MAX_BYTES}}
 authorizer.class.name=kafka.security.auth.SimpleAclAuthorizer
 super.users={{SETUP_HELPER_SUPER_USERS}}
 allow.everyone.if.no.acl.found={{SECURITY_AUTHORIZATION_ALLOW_EVERYONE_IF_NO_ACL_FOUND}}
+{{^SECURITY_KERBEROS_ENABLED}}
```
elezar: @benclarkwood this is for https://jira.mesosphere.com/browse/INFINITY-3435. Would you prefer that I disable the affected test in this PR and then fix this in a separate one?

elezar: I went ahead and merged this. More overhead in breaking it out.

PR #2420, opened 2018-03-27
Add MESOS_ZK_URI scheduler config variable

brugidou:
This allows to use a different zookeeper URI for mesos masters.


brugidou:
Hey, I don't have access to the teamCity project to see why the checks failed. Can you help?

elezar:
The checks failed because we don't allow external users to trigger the builds by default. I will have a look at the changes and start the builds.

Would you be able to create a JIRA ticket explaining the use-case for the additional environment variable. There may be a different way to achieve what you are doing.

For example if you are developing your own framework, the zookeeper connection can be specified in the `svc.yml` file.

elezar:
The checks failed because we don't allow external users to trigger the builds by default. I will have a look at the changes and start the builds.

Would you be able to create a JIRA ticket explaining the use-case for the additional environment variable. There may be a different way to achieve what you are doing.

For example if you are developing your own framework, the zookeeper connection can be specified in the `svc.yml` file.

brugidou:
Hi @elezar, I was following the [guidelines](https://github.com/mesosphere/dcos-commons/blob/master/CONTRIBUTING.md) and opened an issue (#2421).

I would be happy to also open a JIRA ticket, but perhaps you could update the guidelines to point to the correct project in jira? I assume it's https://jira.mesosphere.com but I have no idea which project I should use, it seems that internally you use INFINITY project but I don't have access to view this project.

Also, I am aware that the `svc.yml` file allow for zookeeper configuration but this configuration is also used at the framework's StateStore. The goal of this PR is to let users use a different zookeeper store for their framework than the mesos one. Since most of the scheduler configuration is done via environment variables I tried to replicate this.

Let me know what I can do to clarify, I would be happy to implement this differently if my approach is not aligned with what you would like.

benclarkwood:
@brugidou I don't understand what you gain from this over using the ServiceSpec field.

> Also, I am aware that the svc.yml file allow for zookeeper configuration but this configuration is also used as the framework's StateStore. The goal of this PR is to let users use a different zookeeper store for their framework than the mesos one.

It's up to a service author if they want to expose the ability to override the framework store via the svc.yml. We typically expose things as an envvar when they are not ready to be first classed in the ServiceSpec.

brugidou:
I'm sorry if I was not very clear. I want my service to use:

- one zookeeper (usually zk://leader.mesos:2181/mesos for typical DC/OS) for the MesosScheduler to connect to Mesos masters
- another different zookeeper to store the service state

PR #2425, opened 2018-03-27
Bump to 0.41.1-snapshot

benclarkwood:


benclarkwood:
@nickbp, we shipped 0.41, so the next snapshot to do is 0.41.1.

PR #2431, opened 2018-03-28
INFINITY-3406 Break out Framework vs Scheduler config handling

nickbp:
- `FrameworkConfig`: Handles options relating to framework init. With a single service, it's derived from the service spec, whereas with multiple services it's derived from envvars (with reasonable defaults in most cases)
- `SchedulerConfig`: `EnvStore` is broken out to be shared with `FrameworkConfig`.

Includes other cascading changes relating to this:
- Started testing `ServiceRunner`, needed to update `CuratorLocker` (invoked by `ServiceRunner`) to support that.
- Created initial version of `FrameworkRunner` which currently just handles creating the `FrameworkInfo` object (this will do more later).
- Consolidates existing region `PlacementRule` initialization within `SchedulerBuilder`.

Also has some nitpicky reformatting/ordering fixes, such as giving `ServiceSpec`/`DefaultServiceSpec` a consistent field ordering, and addressing protobufs as `Proto.Foo` in several places.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerBuilder.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerBuilder.java
@@ -9,6 +9,8 @@ import com.mesosphere.sdk.config.validate.DefaultConfigValidators;
 import com.mesosphere.sdk.config.validate.PodSpecsCannotUseUnsupportedFeatures;
 import com.mesosphere.sdk.curator.CuratorPersister;
 import com.mesosphere.sdk.dcos.Capabilities;
+import com.mesosphere.sdk.framework.FrameworkConfig;
+import com.mesosphere.sdk.framework.FrameworkRunner;
 import com.mesosphere.sdk.http.endpoints.ArtifactResource;
 import com.mesosphere.sdk.http.queries.ArtifactQueries;
 import com.mesosphere.sdk.http.types.EndpointProducer;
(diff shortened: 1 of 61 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerBuilder.java
```diff
@@ -246,6 +221,56 @@ public SchedulerBuilder setTemplateUrlFactory(ArtifactQueries.TemplateUrlFactory
      * @throws IllegalArgumentException if validating the provided configuration failed
      */
     public AbstractScheduler build() {
+        // If region awareness is enabled (via java bit or via env) and the cluster supports it, update the ServiceSpec
+        // to include region constraints.
+        final ServiceSpec serviceSpec;
+        if (Capabilities.getInstance().supportsDomains()) {
```
nickbp: cc @mrbrowning for sanity check on region PlacementRule init being performed here (also wrote related tests in `SchedulerBuilderTest.java`

mrbrowning:
FWIW, a rebase here will be necessary -- previously `Capabilities` independently reinstated a scheduler config from the environment, bypassing the config modification that individual schedulers such as Cassandra do to set `ALLOW_REGION_AWARENESS`. I think that will need to be taken account of in the `EnvStore` class now.

benclarkwood:
Shouldn't a service that is going to allow region awareness just set ALLOW_REGION_AWARENESS in the Marathon app?

nickbp:
The prior behavior was to do two things: enable a bit in scheduler builder, AND inject an envvar into the map that was passed to scheduler config. Now either the envvar (for operators, exercised in hello world) OR the bit (for devs, exercised in cass/kaf/elastic/hdfs) is sufficient to provide intent.

PR #2442, opened 2018-03-30
INFINITY-3410 De-monolith AbstractScheduler, move things into fwk pkg, filter resource roles up-front

nickbp:
- Moves per-framework classes into `framework` package. Future PRs will make changes to them at this new location.
- Breaks out offer queueing/processing from `AbstractScheduler` into separate `OfferProcessor` class
- Breaks out Mesos `Scheduler` implementation from `AbstractScheduler` into separate `FrameworkScheduler` class
- Moves filtering of unknown resources upstream. Now they're filtered up-front in `FrameworkScheduler`


sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
```diff
@@ -0,0 +1,269 @@
...
+ */
+public class FrameworkScheduler implements Scheduler {
+
+    private static final Logger LOGGER = LoggingUtils.getLogger(FrameworkScheduler.class);
+
+    /**
+     * Mesos may call registered() multiple times in the lifespan of a Scheduler process, specifically when there's
+     * master re-election. Avoid performing initialization multiple times, which would cause queues to be stuck.
+     */
+    private final AtomicBoolean isRegisterStarted = new AtomicBoolean(false);
```
gabrielhartmann: This variable name isn't clear to me.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
```diff
@@ -0,0 +1,269 @@
...
+            LOGGER.info("Already registered, calling reregistered()");
+            reregistered(driver, masterInfo);
+            return;
+        }
+
+        Driver.setDriver(driver);
+        abstractScheduler.registeredWithMesos();
+
+        LOGGER.info("Registered framework with frameworkId: {}", frameworkId.getValue());
+        this.taskCleaner = new TaskCleaner(stateStore, multithreaded);
```
gabrielhartmann: Does this need to be initialized post registration?  StateStore is available on construction.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
```diff
@@ -0,0 +1,269 @@
...
+            IsLocalRegionRule.setLocalDomain(masterInfo.getDomain());
+        }
+    }
+
+    @Override
+    public void resourceOffers(SchedulerDriver driver, List<Protos.Offer> offers) {
+        Metrics.incrementReceivedOffers(offers.size());
+
+        if (!readyToAcceptOffers.get()) {
+            LOGGER.info("Declining {} offer{}: Waiting for API Server to start.",
```
gabrielhartmann: This log message seems off.  I think `readyToAcceptOffers` might be false for more reasons than just the API server not being up.  If not we should change the variable name.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
```diff
@@ -0,0 +1,269 @@
...
+        // Filter any bad resources from the offers before they even enter processing.
+        offerProcessor.enqueue(offers.stream()
+                .map(offer -> filterBadResources(offer))
+                .collect(Collectors.toList()));
+    }
+
+    /**
+     * Before we forward the offers to the processor queue, lets filter out resources that don't belong to us.
+     * Resources can look like one of the following:
+     * 1. Dynamic against our-role or refined-role/our-role (belongs to us)
```
gabrielhartmann: `refined-role/our-role` --> `pre-reserved-role/our-role`

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
```diff
@@ -0,0 +1,269 @@
...
+     * 2. Static against refined-role (we can reserve against it)
```
gabrielhartmann: `refined` --> `pre-reserved`

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /**
+     * Dequeues and processes any elements which are present on the offer queue, potentially blocking for offers to
+     * appear.
+     */
+    private void processQueuedOffers() {
+        LOGGER.info("Waiting for queued offers...");
+        List<Protos.Offer> offers = offerQueue.takeAll();
+        try {
```
gabrielhartmann: Not clear to me why this try block exists.  Doesn't mean it shouldn't.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
nickbp: `multithreaded` may be set false in tests

gabrielhartmann: Slightly unfortunate.  Can threading be enabled/disabled on construction or do tests require doing it post construction?  Is it a convenience thing?

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
nickbp: It's mainly for the following `finally`, to ensure we increment the offers processed in the metrics, even if an exception is thrown

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
nickbp: Renamed to `registerCalled`

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
nickbp: It's mainly for convenience -- tests may disable threading at the AbstractScheduler level and then it cascades into here. For what it's worth this will end up going away when FrameworkScheduler is finished being built out, this part is in an interim state with this PR.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
nickbp: In practice this is only for the API server. Renamed `readyToAcceptOffers` to `apiServerStarted`, and `setReadyToAcceptOffers()` to `setApiServerStarted()`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
nickbp: Done

PR #2444, opened 2018-03-30
Allow to override EndpointUtils

brugidou:
This is very useful for users that have custom service discovery
mechanisms.

I know that this pull request may not be accepted since it is a bit intrusive, I just wanted to submit it just in case so we can discuss.


nickbp:
SDK 0.52.0 may have what you want, see this PR: https://github.com/mesosphere/dcos-commons/pull/2592

The values used by EndpointUtils are now effectively configurable via envvars in the scheduler:
- autoip domain
- vip domain
- name of marathon instance

https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java#L176

elezar:
Hi @brugidou. As @nickbp please check out the changes in #2592 

If these are not sufficient, then please create a JIRA ticket to discuss this further. I am closing this PR for the time being.

PR #2445, opened 2018-03-30
INFINITY-3411/3469 New reconcilers and other framework-level refactoring

nickbp:
- Break up implicit (per-framework) and explicit (per-service) reconciliation into separate classes. Perform implicit periodically on a background thread, with timing knobs controlled by envvars, just in case.
- Fill in the rest of `FrameworkRunner`, which handles running the Mesos framework thread, and running a stub scheduler API in the case of completed uninstall.
- Move per-service offer handling from (per-framework) `OfferProcessor` into (per-service) `AbstractScheduler`
- Fix `TokenBucket` flakes?


sdk/scheduler/src/test/java/com/mesosphere/sdk/framework/ApiServerTest.java
```diff
@@ -19,36 +30,122 @@
...
+    public static class TestResourcePod {
+
+        @Path("/{name}/info")
+        @GET
+        public Response getPlan(@PathParam("name") String podName) {
+            return Response.ok().entity("Service Pod: " + podName).build();
+        }
+    }
+
+    @Path("/v1/jobs")
```
gabrielhartmann: If these `jobs` endpoints should land on this side of the SDK vs. Queue divide (I think they do), let's call them `services` instead of `jobs` to align with what the SDK thinks instead of what Queue thinks.  A queue may consider a service to be a job, but that's its business.

sdk/scheduler/src/test/java/com/mesosphere/sdk/framework/ImplicitReconcilerTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Test
+    public void testSingleThread() {
+        reconciler.disableThreading().start();
+        verify(mockSchedulerDriver).reconcileTasks(Collections.emptyList());
+        try {
+            reconciler.start();
+            Assert.fail("Expected exception from double start");
+        } catch (IllegalStateException e) {
+            // expected
```
gabrielhartmann: There's an argument you can pass to the `@Test` annotation for this.  It's what's used in other tests in the SDK for this pattern.

gabrielhartmann: e.g.
```
 @Test(expected = IllegalArgumentException.class)
```

sdk/scheduler/src/test/java/com/mesosphere/sdk/framework/ApiServerTest.java
nickbp: Yes they're under e.g. `/v1/service/<name>/plans/deploy` in the 'real' version to be upstreamed soon. This test was originally written against an early pass. One catch is this means the full URL would look like `http://cluster.com/service/queue/v1/service/job/plans/deploy`

sdk/scheduler/src/test/java/com/mesosphere/sdk/framework/ImplicitReconcilerTest.java
nickbp: Switched to `expected` (will be in the following PR). Also broke this out into two test cases: One for the double start (expect error) and another for the single threaded behavior (don't expect error)

sdk/scheduler/src/test/java/com/mesosphere/sdk/framework/ImplicitReconcilerTest.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    @Test
+    public void testScheduledThread() throws InterruptedException {
+        when(mockSchedulerConfig.getImplicitReconcileDelayMs()).thenReturn(0L);
+        when(mockSchedulerConfig.getImplicitReconcilePeriodMs()).thenReturn(1L);
+        reconciler.start();
+        Thread.sleep(1000);
```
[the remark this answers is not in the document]
nickbp: Switched to using a timeout (will be in the following PR)

PR #2447, opened 2018-04-02
Bump libmesos bundle version

takirala:
Addresses [INFINITY-3455](https://jira.mesosphere.com/browse/INFINITY-3455). 

`libmesos-bundle-1.9.0-rc2-1.2.0-rc2-1.tar.gz` is being used `queries/testdata` of cli as well. Comment if I am incorrect in assuming `testdata` need not be updated.


benclarkwood:
@takirala go ahead and backport this to sdk-0.40

PR #2449, opened 2018-04-02
INFINITY-3445/3469 Upstream offer evaluation and implement MesosEventClient

nickbp:
This one's the largest of the set, as it involves a lot of interconnected pieces. Tried to just do a subset of this but it didn't work out. Still, not _too_ bad at ~2k lines...

- Introduces a `MesosEventClient` interface which is implemented by `AbstractScheduler` (therefore `DefaultScheduler` and `UninstallScheduler`), and which will also be implemented by an upcoming multi-service router (wrapping multiple `AbstractScheduler`s).
- Consolidates multiple ways of cleaning up tasks into a single flow. This cleanup handling is now a first-class part of the offer evaluation cycle driven by `OfferProcessor`. Uninstalling a service vs decommissioning a task vs wiping unknown resources are now all handled through the same flow.
- For uninstall, cleaned resource tallying is done by just removing them from `TaskInfo`s, rather than marking them with tombstones in those `TaskInfo`s. This mirrors the behavior when decommissioning tasks, and reduces the risk of leaving behind corrupted information in a world where a service uninstall doesn't necessarily mean that the entire framework is being torn down.
- Offer evaluation is now largely driven by `OfferProcessor`, instead of being entirely within `AbstractScheduler`. The division here is effectively per-Framework handling in `OfferProcessor`, vs per-Service handling in the `AbstractScheduler`(s). Now, the `AbstractScheduler` (as a `MesosEventClient`) returns the operations it would liked to be performed, and `OfferProcessor` hands them to an `OfferAcceptor` itself (or declines them itself). Previously this was all done within `AbstractScheduler`, whereas now that is more concerned with just the evaluation step proper.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallRecorder.java
```diff
@@ -1,108 +1,148 @@
...
-                .anyMatch(taskInfoResource -> resourcesMatch(taskInfoResource, resource));
+        // Notify the resource steps in the uninstall plan or decommission plan about these resource ids.
+        resourceSteps.forEach(step -> step.updateResourceStatus(allResourceIds));
     }
 
-    private static boolean resourcesMatch(Protos.Resource resource1, Protos.Resource resource2) {
-        return ResourceUtils.getResourceId(resource1).equals(ResourceUtils.getResourceId(resource2));
-    }
+    /**
+     * Returns an updated copy of any {@code taskInfos} which have omit any resources matching the provided
```
benclarkwood: Spelling

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallRecorder.java
```diff
@@ -1,108 +1,148 @@
...
+            if (resourceId.isPresent() && resourceIdsToRemove.contains(resourceId.get())) {
+                // Matching resource found. Omit it from the filtered list.
+                anyUpdates = true;
             } else {
-                updatedResources.add(resource);
+                filteredResources.add(resource);
             }
         }
-        return updatedResources;
+        return anyUpdates ? filteredResources : null;
```
benclarkwood: Returning an empty list might be cleaner, no?

PR #2453, opened 2018-04-03
[INFINITY-3441] Remove beta notice from preinstall notes

benclarkwood:


benclarkwood:
@elezar Adding it in the future sounds great. We could do it automatically when you indicate something is a beta.

PR #2456, opened 2018-04-03
DCOS_ELASTIC-21 Adding support for Kibana Plugins and therefore also …

devpie:
…increasing the grace period.

### DIFF
--- a/frameworks/elastic/universe-kibana/config.json
+++ b/frameworks/elastic/universe-kibana/config.json
@@ -31,6 +31,11 @@
           "type": "boolean",
           "default": false
         },
+        "plugins": {
+          "description": "An optional comma-separated list of URLs to ZIP files containing Kibana plugins.",
+          "type": "string",
+          "default": ""
+        },
         "elasticsearch_url": {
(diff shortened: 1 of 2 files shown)


devpie:
Hi, would somebody tell me why all the checks fail?

elezar:
For security reasons, we don't by default allow external contributions to trigger our CI.

jan-ruv:
Ah ok, thanks. That makes sense :smile:

keithchambers:
Hi Jan. Thank you for the PR. This is something that aligns with our roadmap.

Let me check with Engineering to see if they can code review and if anything is missing. Get back with you shortly ...

vishnu2kmohan:
@mpereira Does this make sense for us to include this (if this is not already the case) for our Elastic 6.6 release?
cc @imaxxs

mpereira:
@vishnu2kmohan it does. I added https://jira.mesosphere.com/browse/DCOS-39949 to the current sprint. It's already under the Elastic 6.6.0 epic.

takirala:
Elastic framework has been moved to https://github.com/mesosphere/dcos-elastic-service

PR #2457, opened 2018-04-03
Updating min dcos release version from 1.9 to 1.10

kvish:
small change to test CI, part of changes to remove custom executor and move to default executor in dcos versions greater than 1.9


gabrielhartmann:
@kvish: LGTM, do you have plans to do this for all the services outside the mono-repo as well?

benclarkwood:
To close the loop: We don't need to change anything outside of dcos-commons yet.

PR #2459, opened 2018-04-04
[INFINITY-3446] Reject Service Names that violate DNS length requirements.

benclarkwood:
did:
ERROR new deployments on a too long DNS name
Do nothing to old deployments (but do warn)

I can't decide if we can actually do this or not. It's probably possible for a user to wind up in a situation where they can't re-install their service with a too long name.

Maybe we don't even work at all with too long names? I'll find out...

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ServiceNameCannotBreakDNS.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ServiceNameCannotBreakDNS.java
@@ -1,0 +1,47 @@
+package com.mesosphere.sdk.config.validate;
+
+import com.mesosphere.sdk.http.EndpointUtils;
+import com.mesosphere.sdk.offer.LoggingUtils;
+import com.mesosphere.sdk.specification.ServiceSpec;
+import org.slf4j.Logger;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+
+public class ServiceNameCannotBreakDNS implements ConfigValidator<ServiceSpec> {
+
+    public static final Logger LOGGER = LoggingUtils.getLogger(ServiceNameCannotBreakDNS.class);
+
+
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ServiceNameCannotBreakDNS.java
```diff
@@ -0,0 +1,47 @@
...
+import com.mesosphere.sdk.specification.ServiceSpec;
+import org.slf4j.Logger;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+
+public class ServiceNameCannotBreakDNS implements ConfigValidator<ServiceSpec> {
+
+    public static final Logger LOGGER = LoggingUtils.getLogger(ServiceNameCannotBreakDNS.class);
```
nickbp: could be `private`

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ServiceNameCannotBreakDNS.java
```diff
@@ -0,0 +1,47 @@
...
+            // There is. We can at most WARN that the name is going to cause truncated DNS.
+            if (!nameWithinSubDomainLimit(newConfig.getName())) {
+                LOGGER.warn("The service name (without slashes) exceeds the maximum size allowed in a DNS subdomain.");
+            }
+
+            return Collections.emptyList();
+        } else {
+            // This is a new deployment, we can fail it if the service name is too large.
+            if (!nameWithinSubDomainLimit(newConfig.getName())) {
+                return Collections.singletonList(ConfigValidationError.valueError(
```
nickbp: If you wanted to be fancy here, the `(without slashes)` in the error could be included only when the service name actually has slashes. But this is such a corner case already...

PR #2460, opened 2018-04-04
INFINITY-3472/3475 Finish first draft of multi-service support

nickbp:
This PR mainly focuses on adding a first pass at multi-service support, but also catches up on a couple loose ends that didn't fit in prior PRs, as well as some requested changes in the prior #2449 . Bad news: Big PR. Good news: Much of the PR (>30%) is just boilerplate for the additional HTTP endpoints. Better news: Nearly all of the PR is new code, so hopefully not much risk of creating regressions (just bugs in the new functionality)

- Adds a new set of `Multi*Resource`s which allow querying underlying services in a multi-service scheduler. All existing single-service queries are supported via these new endpoints, with the exception of `HealthResource` (only meant for servicing _overall_ Scheduler health in health checks) and `OfferOutcomeResource` (could add later? or not?).
  - For example, the deploy plan for a given sub-service can be queried using `MultiPlansResource` at `/v1/services/<name>/plans/deploy`. Multi-service schedulers would use these endpoints instead of the default ones like `PlansResource`.
- Adds support for specifying service-level `GoalState`s at the Java API level (only). In a multi-service scheduler, a service with a `GoalState` of `FINISH` can be automatically uninstalled once its `deploy` and `recovery` plans are `COMPLETE`. The default `GoalState` is `RUNNING`, meaning a service which runs continuously until told to stop.
- Simplifies multi-service construction in `SchedulerBuilder`: Reduced two calls to one, and removed the need to manually provide a `namespace` value (should always be the service name). An added `frameworkName` param is needed in order to correctly generate artifact URLs in `MultiArtifactResource`, but this param should be straightforward for multi-service developers to provide.
- Adds an initial implementation of multi-service scheduler management for use by developers (MEATY PART):
  - `MultiServiceManager`: In-memory management of running services. Adding/removing services by developers is done here.
  - `MultiServiceEventClient`: Implementation of `MesosEventClient` which routes Mesos events (`Offer`s, `TaskStatus`es, etc) to individual services which are stored within the `MultiServiceManager`. For example, this routes a received `TaskStatus` to the service which owns that task.
  - `MultiServiceRunner`: Runs the main Framework thread, passing events from Mesos to a provided `MultiServiceEventClient`. This mirrors today’s `ServiceRunner` used by mono-schedulers.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -220,6 +235,12 @@ private static void killUnneededTasks(StateStore stateStore, Set<String> taskToD
 
     @Override
     protected OfferResponse processOffers(Collection<Protos.Offer> offers, Collection<Step> steps) {
+        if (serviceSpec.getGoal() == GoalState.FINISH
```
gabrielhartmann: This is just slightly hacky.  We call `processOffers` regardless of whether we have any offers to process so it'll definitely work.  It's our core polling mechanism I suppose.  Ideally I think we'd check for eligibility for uninstall independently of our offer processing cycle.  I think changing this as part of this PR would only make it bigger, but it would be good to file a blocking JIRA to revisit our design around service lifecycle vs. offer processing.

nickbp: There's currently a TODO tucked away in MesosEventClient (merged previously), proposing a separate getStatus() (or similar) call for misc checks like this: https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/MesosEventClient.java#L76

Filed INFINITY-3493 to track it as well.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -0,0 +1,402 @@
...
+        boolean anyClientsNotReady = false;
+
+        List<OfferRecommendation> recommendations = new ArrayList<>();
+        List<Protos.Offer> remainingOffers = new ArrayList<>();
+        remainingOffers.addAll(offers);
+
+        Collection<String> finishedServices = new ArrayList<>();
+        Collection<String> uninstalledServices = new ArrayList<>();
+
+        Collection<AbstractScheduler> services = multiServiceManager.lockAndGetServices();
```
gabrielhartmann: Does holding this lock imply that add/remove/update operations on services are impossible now?  Do we want to lock, get copies, and then release the lock right away?  Do the add/remove/update operations just block?  It might be ok for now but this might be another place we want to file a JIRA to follow up and make sure we allow responsive API calls to clients.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -0,0 +1,402 @@
...
+                    remainingOffers = updatedRemainingOffers;
+                }
+                recommendations.addAll(response.recommendations);
+                LOGGER.info("  {} offer result: {}[{} rec{}], {} offer{} remaining",
+                        serviceName,
+                        response.result,
+                        response.recommendations.size(), response.recommendations.size() == 1 ? "" : "s",
+                        remainingOffers.size(), remainingOffers.size() == 1 ? "" : "s");
+
+                switch (response.result) {
```
gabrielhartmann: This is another case where combine service lifecycle and offer processing logic.  We should note this as a place where we might want to pursue de-coupling.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -0,0 +1,402 @@
...
+        // Map reserved resources (and their parent offers) to the service that they're assigned to.
+        // We can then query those services with the subset of offers that belong to them.
+        for (Protos.Offer offer : unusedOffers) {
+            for (Protos.Resource resource : offer.getResourcesList()) {
+                Optional<String> serviceName = ResourceUtils.getNamespace(resource);
+                if (serviceName.isPresent()) {
+                    // Found service name: Store resource against serviceName+offerId to be evaluated below.
+                    getEntry(offersByService, serviceName.get(), offer).add(resource);
+                } else if (ResourceUtils.getReservation(resource).isPresent()) {
+                    // (CASE 1) Malformed: Reserved resource which is missing a service name.
```
gabrielhartmann: This is safe because this code only ever runs in the multi-service case?

nickbp: Yes, `MultiServiceEventClient` is only relevant when running in multi-service mode. It's effectively slotting in where an `AbstractScheduler` which go in a mono-service situation. (Both implement `MesosEventClient`)

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -0,0 +1,402 @@
...
+                        response.result,
+                        response.offerResources.stream()
+                                .collect(Collectors.summingInt(or -> or.getResources().size())),
+                        response.offerResources.size(),
+                        response.offerResources.size() == 1 ? "" : "s");
+                switch (response.result) {
+                case FAILED:
+                    // We should be able to safely skip this service and proceed to the next one. For this round,
+                    // the service just won't have anything added to unexpectedResources. We play it safe by telling
+                    // upstream to do a short decline for the unprocessed resources.
```
gabrielhartmann: This statement confuses me.  You can only accept/decline offers as a whole.  A short decline of a subset of resources is not a thing as far as I know.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -0,0 +1,402 @@
...
+            LOGGER.info("Received status for unknown task {}: {}",
+                    status.getTaskId().getValue(), TextFormat.shortDebugString(status));
+            return StatusResponse.unknownTask();
+        }
+        LOGGER.info("Received status for task {}: {}", status.getTaskId().getValue(), status.getState());
+        return service.get().status(status);
+    }
+
+    /**
+     * Returns a set of queues-specific endpoints to be served by the scheduler. This effectively overrides the
```
gabrielhartmann: We should not include references to the word `queues`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -0,0 +1,402 @@
...
+                new MultiEndpointsResource(frameworkName, multiServiceManager, schedulerConfig),
+                new MultiPlansResource(multiServiceManager),
+                new MultiPodResource(multiServiceManager),
+                new MultiStateResource(multiServiceManager, new StringPropertyDeserializer())));
+        endpoints.addAll(customEndpoints);
+        return endpoints;
+    }
+
+    /**
+     * Finds the requested entry in the provided map, initializing the entry if needed.
```
gabrielhartmann: This comment could be better.  I don't really know what an `entry` is here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -0,0 +1,402 @@
...
+            serviceOffers = new HashMap<>();
+            map.put(serviceName, serviceOffers);
+        }
+        return getEntry(serviceOffers, offer);
+    }
+
+    /**
+     * Finds the requested entry in the provided map, initializing the entry if needed.
+     */
+    private static OfferResources getEntry(Map<Protos.OfferID, OfferResources> map, Protos.Offer offer) {
```
gabrielhartmann: ditto

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
nickbp: MultiServiceManager supports separate read (shared) vs read/write (exclusive) locking. Exclusive locks are only used when actually adding/removing a service from the set. All other access is only grabbing a shared lock.

This call is only getting a shared lock, so other shared-lock operations (like most API calls) can occur against any/all services at the same time as this is occurring.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
```diff
@@ -0,0 +1,282 @@
...
+                return Optional.empty();
+            }
+            return Optional.ofNullable(services.get(originalName));
+        } finally {
+            rlock.unlock();
+        }
+    }
+
+    /**
+     * Adds or replaces a running service according to the name in its ServiceSpec. Note: If the service was marked for
```
gabrielhartmann: What's the difference between add/replace?  Does replace == update?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
```diff
@@ -0,0 +1,282 @@
...
+     * @throws IllegalArgumentException if the service name collides with an existing service after reducing slashes to
+     *                                  periods
+     */
+    public MultiServiceManager putService(AbstractScheduler service) {
+        String originalName = service.getServiceSpec().getName();
+        String sanitizedName = CommonIdUtils.toSanitizedServiceName(originalName);
+        rwlock.lock();
+        try {
+            // NOTE: If the service is uninstalling, it should already be passed to us as an UninstallScheduler.
+            // See SchedulerBuilder.
```
gabrielhartmann: I guess this occurs through some factory type thing that detects whether uninstall is still active for the service?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
nickbp: Add = start from scratch (new 'install')
Replace = start from previous state ('reconfigure')

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
```diff
@@ -0,0 +1,282 @@
...
+
+            // Update the sanitized=>original mapping, and check for a colliding sanitized name,
+            // e.g. "/path/to/service" vs "/path/to.service".
+            // This differs from an exact match, which we treat as a reconfiguration/replacement of the prior service.
```
gabrielhartmann: `replace` has a special meaning in the SDK and is destructive.  I think you mean `update` here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
nickbp: Yes, `SchedulerBuilder` checks for a few cases (between service removal vs global uninstall) here: https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerBuilder.java#L281-L314

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
```diff
@@ -0,0 +1,282 @@
...
+     *   a. Upstream builds a new foo using SchedulerBuilder, which internally finds the uninstall bit and returns a
+     *      new UninstallScheduler
+     *   b. putService(foo) is called with the UninstallScheduler
+     *   c. The UninstallScheduler resumes cleanup from where it left off...
+     * 5. Sometime after the UninstallScheduler finishes cleanup, it returns FINISHED in response to offers.
+     * 6. We remove the service and invoke uninstallCallback.uninstalled(), telling upstream that it's gone. If upstream
+     *    invokes putService(foo) again at this point, the service will be relaunched from scratch because the uninstall
+     *    bit in ZK will have been cleared.
+     */
+    public void uninstallServices(Collection<String> finishedServiceNames) {
```
gabrielhartmann: Because uninstall is going to be happening so frequently now, we should probably put a timeout on it.  If we're waiting for resource `abc-123` for more than 10 minutes or something, consider it gone and let uninstall proceed.  We don't want to have to handle many or very old uninstalls which are stuck.  Garbage collection will happen eventually anyway if resources from unexpected services ever show up again.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
```diff
@@ -0,0 +1,282 @@
...
+            rwlock.unlock();
+        }
+    }
+
+    /**
+     * Removes the specified services after they have completed uninstall. Any unknown service names are ignored.
+     *
+     * @return the number of services which are still present after this removal
+     */
+    public int removeServices(Collection<String> uninstalledServiceNames) {
```
gabrielhartmann: Having this return value while perhaps convenient seems weird.  I'd just have `getServicesCount` method somewhere and keep it separate.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
```diff
@@ -0,0 +1,282 @@
...
+
+            return services.size();
+        } finally {
+            rwlock.unlock();
+        }
+    }
+
+    /**
+     * Gets a shared/read lock on the underlying data store, and returns all available services.
+     * Upstream MUST call {@link #unlockServices()} after finishing.
```
gabrielhartmann: I've noted elsewhere that I'm a little concerned about how long this lock can be held.  We should see if we can do something about that.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * Unlocks a previous lock which was obtained via {@link #lockAndGetServices()}.
+     */
+    public void unlockServices() {
+        rlock.unlock();
+    }
+
+    /**
+     * Notifies underlying services that a registration or re-registration has occurred. After this has been invoked,
```
gabrielhartmann: Why do underlying services care about registration events?  Not clear to me that they should.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -84,6 +84,10 @@ public DefaultServiceSpec(
         this.role = role;
         this.principal = principal;
         this.user = getUser(user, pods);
+        this.goalState = goalState == null ? GoalState.RUNNING : goalState;
+        if (goalState == GoalState.FINISHED) {
+            throw new IllegalArgumentException("Service goal state is deprecated FINISHED. Did you mean FINISH?");
```
gabrielhartmann: Doesn't a goal state of `ONCE` really make more sense at the service level when it's representing a job?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
gabrielhartmann: Can we just ignore malformed resources?  I'm nervous about eating other frameworks resources again.

nickbp: From convo elsewhere:
- if service name label is present and maps to known service: map to service
- if service name label is present but maps to unknown service: assume it's from a valid service which was since removed
- if service name label isn't present: uh-oh, better leave this one alone

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
nickbp: (INFINITY-3493)

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
nickbp: Cleaned this up in a follow-up PR.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
nickbp: Changed this to 'Adds or updates'

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
nickbp: Switched to `update`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
nickbp: INFINITY-3496

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
nickbp: Fixed in upcoming PR.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
nickbp: it's mainly telling them that they can start interacting with the `SchedulerDriver`. looks like DefaultScheduler uses this signal to kill any remaining decommission-pending tasks in its StateStore

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
nickbp: Not sure, I just went by the comments in `GoalState`:

```
    /**
     * Running and then exiting successfully. Should not be restarted upon a successful exit.
     */
    FINISH,

    /**
     * Running only once over the lifetime of a service. Should not be run again for the duration of this install.
     */
    ONCE
```

PR #2462, opened 2018-04-05
INFINITY-3491/3495 Fix invalid reservation handling, add hello-world examples

nickbp:
Fixes:
- When in multi-service mode, we would previously clean reservations that lacked a service name label. Instead, just complain and leave them be.
- CuratorLocker had again started getting invoked _after_ other access had already happened. Fix the problem once and for all by moving locking immediately before building the CuratorPersister object itself.
- Other fixes from #2460

New:
- Updates hello-world to include a reference implementation of running a 'dynamic' multi-service scheduler, as well as a 'static' multi-service scheduler, in addition to the current single-service options. Dynamic mode is run when no yaml files are provided, while static is run when multiple (comma or space-separated) yaml files are provided. Next is to add integration tests against this implementation (JIRA filed)

### DIFF
--- a/frameworks/helloworld/universe/config.json
+++ b/frameworks/helloworld/universe/config.json
@@ -68,48 +68,13 @@
           "default": true
         },
         "scenario": {
-          "description": "A value to test various SDK customization scenarios",
+          "description": "Comma-separated list of SDK customization scenarios to perform on the service(s)",
           "type": "string",
-          "enum": [
-            "YAML",
-            "JAVA",
-            "CUSTOM_PLAN",
-            "CUSTOM_DECOMMISSION",
-            "MULTI_REGION"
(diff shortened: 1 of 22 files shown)


frameworks/helloworld/universe/config.json
```diff
@@ -68,48 +68,13 @@
...
-            "JAVA",
-            "CUSTOM_PLAN",
-            "CUSTOM_DECOMMISSION",
-            "MULTI_REGION"
-          ],
           "default": "YAML"
         },
         "yaml": {
-          "description": "An example YAML file to deploy",
+          "description": "Comma-separated list of zero or more example YAML files to deploy",
```
gabrielhartmann: add another field `yamls` for this.

PR #2465, opened 2018-04-09
Add package registry option in stub universe tooling

takirala:
Addresses [INFINITY-3486](https://jira.mesosphere.com/browse/INFINITY-3486)

#### CI Changes
- Add `PACKAGE_REGISTRY_ENABLED` in docker run environment and set it to `true`.
- Add `PACKAGE_REGISTRY_STUB_URL` to parameters. Pass it to docker run.

#### TODO
~~Check partner repo changes~~ None applicable as of now. We can make some changes in future.
~~Create ticket to clean up stub universe for package registry~~ [INFINITY-3510](https://jira.mesosphere.com/browse/INFINITY-3510)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -5,12 +5,13 @@ Note: pytest must be invoked with this file in the working directory
 E.G. py.test frameworks/<your-frameworks>/tests
 """
 import logging
-import os
 import os.path
 import sys
 
 import pytest
 import sdk_diag
+import sdk_repository
+import sdk_package_registry
 import sdk_utils
 import teamcity
 
@@ -38,10 +39,18 @@ for noise_source in [
 log = logging.getLogger(__name__)
 
 # The following environment variable allows for log collection to be turned off.
-# This is useful, for exampl in testing.
+# This is useful, for example in testing.
 INTEGRATION_TEST_LOG_COLLECTION = str(
     os.environ.get('INTEGRATION_TEST_LOG_COLLECTION', "True")
 ).lower() in ["true", "1"]
+
+
+@pytest.fixture(scope='session', autouse=True)
+def configure_universe(tmpdir_factory):
+    if os.environ.get('PACKAGE_REGISTRY_ENABLED', '') == 'true':
+        yield from sdk_package_registry.package_registry_session(tmpdir_factory)
+    else:
+        yield from sdk_repository.universe_session()
 
 
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
--- a/conftest.py
+++ b/conftest.py
@@ -40,14 +40,14 @@ log = logging.getLogger(__name__)
 
 # The following environment variable allows for log collection to be turned off.
 # This is useful, for example in testing.
-INTEGRATION_TEST_LOG_COLLECTION = str(
-    os.environ.get('INTEGRATION_TEST_LOG_COLLECTION', "True")
-).lower() in ["true", "1"]
+INTEGRATION_TEST_LOG_COLLECTION = sdk_utils.is_env_var_set(
+    'INTEGRATION_TEST_LOG_COLLECTION', default=str(True)
+)
 
 
 @pytest.fixture(scope='session', autouse=True)
 def configure_universe(tmpdir_factory):
-    if os.environ.get('PACKAGE_REGISTRY_ENABLED', '') == 'true':
+    if sdk_utils.is_env_var_set('PACKAGE_REGISTRY_ENABLED', default=''):
         yield from sdk_package_registry.package_registry_session(tmpdir_factory)
     else:
         yield from sdk_repository.universe_session()
--- a/frameworks/cassandra/tests/conftest.py
+++ b/frameworks/cassandra/tests/conftest.py
@@ -1,12 +1,6 @@
 import pytest
-import sdk_repository
 import sdk_security
 from tests import config
-
-
-@pytest.fixture(scope='session')
-def configure_universe():
-    yield from sdk_repository.universe_session()
 
 
 @pytest.fixture(scope='session')
--- a/frameworks/hdfs/tests/conftest.py
+++ b/frameworks/hdfs/tests/conftest.py
@@ -1,12 +1,6 @@
 import pytest
-import sdk_repository
 import sdk_security
 from tests import config
-
-
-@pytest.fixture(scope='session')
-def configure_universe():
-    yield from sdk_repository.universe_session()
 
 
 @pytest.fixture(scope='session')
--- a/test.sh
+++ b/test.sh
@@ -131,7 +131,7 @@ case $key in
     package_registry="true"
     ;;
     --dcos-files-path)
-    dcos_files_path="$2"
+    dcos_files_path="$(abs_path "$2")"
     shift
     ;;
     --gradle-cache)
@@ -283,8 +283,8 @@ docker run --rm \
     -e STUB_UNIVERSE_URL="$STUB_UNIVERSE_URL" \
     -e PACKAGE_REGISTRY_ENABLED="${package_registry}" \
     -e PACKAGE_REGISTRY_STUB_URL="${PACKAGE_REGISTRY_STUB_URL}" \
-    -e DCOS_FILES_PATH="$(abs_path "${dcos_files_path}")" \
-    -v "$(abs_path "${dcos_files_path}")":"$(abs_path "${dcos_files_path}")" \
+    -e DCOS_FILES_PATH="${dcos_files_path}" \
+    -v "${dcos_files_path}":"${dcos_files_path}" \
     -v $(pwd):$WORK_DIR \
     -v $ssh_path:/ssh/key \
     -w $WORK_DIR \
--- a/test.sh
+++ b/test.sh
@@ -10,6 +10,15 @@
 # Exit immediately on errors
 set -e
 
+function abs_path() { (
+    if [[ -d $1 ]]; then
+        cd $(dirname $1)
+        echo $(pwd)/$(basename $1)
+    else
+        echo ""
+    fi
+)}
+
 REPO_ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 if [ -d $REPO_ROOT_DIR/frameworks ]; then
     FRAMEWORK_LIST=$(ls $REPO_ROOT_DIR/frameworks | sort)
@@ -29,10 +38,11 @@ aws_profile="default"
 enterprise="true"
 interactive="false"
 headless="false"
+package_registry="false"
 
 function usage()
 {
-    echo "Usage: $0 [-m MARKEXPR] [-k EXPRESSION] [-p PATH] [-s] [-i|--interactive] [--headless] [--aws|-a PATH] [--aws-profile PROFILE] [all|<framework-name>]"
+    echo "Usage: $0 [-m MARKEXPR] [-k EXPRESSION] [-p PATH] [-s] [-i|--interactive] [--headless] [--package-registry] [--dcos-files-path DCOS_FILES_PATH] [--aws|-a PATH] [--aws-profile PROFILE] [all|<framework-name>]"
     echo "-m passed to pytest directly [default -m \"${pytest_m}\"]"
     echo "-k passed to pytest directly [default NONE]"
     echo "   Additional pytest arguments can be passed in the PYTEST_ARGS"
@@ -42,6 +52,8 @@ function usage()
     echo "-s run in strict mode (sets \$SECURITY=\"strict\")"
     echo "--interactive start a docker container in interactive mode"
     echo "--headless leave STDIN available (mutually exclusive with --interactive)"
+    echo "--package-registry Use package registry to install packages"
+    echo "--dcos-files-path DCOS_FILES_PATH sets the path to look for .dcos files. If empty, use stub universe urls to build .dcos file(s)."
     echo "--gradle-cache PATH sets the gradle cache to the specified path [default ${gradle_cache}]."
     echo "               Setting PATH to \"\" will disable the cache."
     echo "--aws-profile PROFILE the AWS profile to use [default ${aws_profile}]"
@@ -115,6 +127,13 @@ case $key in
     --headless)
     headless="true"
     ;;
+    --package-registry)
+    package_registry="true"
+    ;;
+    --dcos-files-path)
+    dcos_files_path="$2"
+    shift
+    ;;
     --gradle-cache)
     gradle_cache="$2"
     shift
@@ -173,6 +192,7 @@ fi
 
 echo "interactive=$interactive"
 echo "headless=$headless"
+echo "package-registry=${package_registry}"
 echo "security=$security"
 echo "enterprise=$enterprise"
 
@@ -241,6 +261,12 @@ if [ -n "$pytest_m" ]; then
     PYTEST_ARGS="$PYTEST_ARGS-m \"$pytest_m\""
 fi
 
+if [ x"$package_registry" == x"true" ]; then
+    if [ -z "$PACKAGE_REGISTRY_STUB_URL" ]; then
+        echo "PACKAGE_REGISTRY_STUB_URL not found in environment. Exiting..."
+        exit 1
+    fi
+fi
 
 docker run --rm \
     -v ${aws_credentials_file}:/root/.aws/credentials:ro \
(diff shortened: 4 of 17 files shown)


testing/sdk_utils.py
```diff
@@ -109,6 +112,20 @@ def is_strict_mode():
     return os.environ.get('SECURITY', '') == 'strict'
 
 
+def random_string(length=8):
+    return ''.join(
```
elezar: Why not just use `str(uuid.uuid4())`?

testing/sdk_cmd.py
```diff
@@ -154,6 +147,27 @@ def run_raw_cli(cmd, print_output=True):
...
+
+    eg. `cmd`= "package install <package-name>" results in:
+    $ dcos package install <package-name>
+    """
+    dcos_cmd = "dcos {}".format(cmd)
+    log.info("(CLI) {}".format(dcos_cmd))
+    return run_command(dcos_cmd, print_output)
+
+
+def dcos_version() -> str:
```
elezar: The other DC/OS version related functions are in `sdk_utils`. Is the only reason we have it here the circular dependency that would result?

If we do decide to keep it here, we could at least add a `lrucache` decorator as we wouldn't expect this to change at all.

testing/sdk_cmd.py
```diff
@@ -154,6 +147,27 @@ def run_raw_cli(cmd, print_output=True):
     return result.returncode, stdout, stderr
 
 
+
+def run_raw_cli(cmd, print_output=True):
+    """Runs the command with `dcos` as the prefix to the shell command
+    and returns the resulting output (stdout separated from stderr by a newline).
```
elezar: This docstring is incorrect as this returns an `rc, stdout, stderr` tuple.

testing/sdk_repository.py
```diff
@@ -96,18 +100,194 @@ def remove_universe_repos(stub_urls):
...
 
 
-def universe_session():
+def add_package_registry_stub() -> dict:
+    stub_url = os.environ['PACKAGE_REGISTRY_STUB_URL']
+    with urllib.request.urlopen(stub_url) as url:
+        repo = json.loads(url.read().decode())
+        min_supported = repo["packages"][0]["minDcosReleaseVersion"]
+
+    version = sdk_cmd.dcos_version()
```
elezar: Is there any reason why we could we not just call `sdk_utils.dcos_version_at_least(min_supported)`?

testing/sdk_repository.py
```diff
@@ -96,18 +100,194 @@ def remove_universe_repos(stub_urls):
...
+            'registry': {
+                'service-account-secret-path': service_secret_path
+            }
+        },
+        wait_for_deployment=False,
+        insert_strict_options=False
+    )
+
+    pkg_reg_repo = add_stub_universe_urls(
+        ['https://registry.marathon.l4lb.thisdcos.directory/repo']
```
elezar: This is dependent on `PACKAGE_REGISTRY_SERVICE_NAME`.

testing/sdk_repository.py
```diff
@@ -96,18 +100,194 @@ def remove_universe_repos(stub_urls):
...
+    )
+    # If `describe` endpoint is working, registry is writable by AR.
+    package_name = 'hello'
+    package_version = 'world'
+    expected_msg = 'Version [{}] of package [{}] ' \
+                   'not found'.format(package_version, package_name)
+    loop_until_cli_condition(
```
elezar: This is a bit different to how we currently handle "wait-for" or "expect" events in the rest of the codebase. See for example the `wait_for_*` functions in `sdk_plans.py`.

testing/sdk_repository.py
```diff
@@ -96,18 +100,194 @@ def remove_universe_repos(stub_urls):
...
+        'registry describe --package-name={} --package-version={}'.format(
+            package_name, package_version
+        ),
+        lambda code, out, err: code == 1 and expected_msg in err
+    )
+    return pkg_reg_repo
+
+
+def add_dcos_files_to_registry(
+        temp_dir
```
elezar: Would it make sense to add type annotations to the parameters as well as the function return types now?

testing/sdk_repository.py
```diff
@@ -96,18 +100,194 @@ def remove_universe_repos(stub_urls):
...
+            '--build-definition-file={}'.format(str(package_json_file)),
+            '--output-directory={}'.format(dcos_files_path),
+            '--json'
+        ]))
+        assert rc == 0
+    assert os.path.isfile(target)
+    return target
+
+
+def grant_perms_for_registry_account(service_uid: str) -> None:
```
elezar: It may be useful to have this in the `sdk_security` module instead?

frameworks/cassandra/tests/conftest.py
```diff
@@ -3,10 +3,14 @@
 import sdk_security
 from tests import config
 
+from sdk_utils import random_string
```
elezar: We generally prefer: `import sdk_utils` and then use `sdk_utils.random_string()` explicitly.

frameworks/cassandra/tests/conftest.py
```diff
@@ -3,10 +3,14 @@
...
+
 
 @pytest.fixture(scope='session')
-def configure_universe():
-    yield from sdk_repository.universe_session()
+def configure_universe(tmpdir_factory):
+    yield from sdk_repository.universe_session(
+        tmpdir_factory.mktemp(basename=random_string())
```
elezar: Is it needed to inject that here?

testing/sdk_repository.py
```diff
@@ -96,18 +100,194 @@ def remove_universe_repos(stub_urls):
...
-        yield from sdk_repository.universe_session()
+    def configure_universe(temp_dir):
+        yield from sdk_repository.universe_session(temp_dir)
     """
     stub_urls = {}
+    pkg_reg = {}
+    package_registry_enabled = os.environ['PACKAGE_REGISTRY_ENABLED'] == 'true'
     try:
-        stub_urls = add_universe_repos()
+        if package_registry_enabled:
```
elezar: We could move the package registry specifics to their own function and also use the `try-yield-finally` semantics there. That would keep this function much simpler. for example:
```
if package_registry_enabled:
    yield sdk_package_registry.setup_package_registry()
else:
     try:
         stub_urls = add_stub_universe_urls(get_universe_repos())
         yield
      finally:
         remove_universe_repos(stub_urls)
```

frameworks/hdfs/tests/conftest.py
```diff
@@ -3,10 +3,14 @@
 import sdk_security
 from tests import config
 
+from sdk_utils import random_string
+
 
 @pytest.fixture(scope='session')
-def configure_universe():
-    yield from sdk_repository.universe_session()
+def configure_universe(tmpdir_factory):
```
elezar: Since the `configure_universe` fixture is not dependent on any specific framework code, we could consider moving it into the root `conftest.py` as a fixture and then also adding the `autouse=True` parameter.

testing/sdk_repository.py
```diff
@@ -96,18 +100,194 @@ def remove_universe_repos(stub_urls):
...
+def build_dcos_files_from_stubs(
+        stub_universe_urls: list,
+        dcos_files_path: str,
+        temp_dir
+) -> list:
+    if not len(stub_universe_urls):
+        return stub_universe_urls
+    package_file_paths = []
+    for repo_url in stub_universe_urls:
+        headers = {
```
elezar: Is this not also supposed to handle the authentication with the DC/OS service? We have some other "request" functions including `service_request` and `cluster_request` in `sdk_cmd` which may be doing similar things. If would be good to understand what is being done differently here.

It may also make sense to group these types of functions more logically, although that is probably out of scope for this PR.

testing/sdk_repository.py
```diff
@@ -96,18 +100,194 @@ def remove_universe_repos(stub_urls):
     log.info('Finished removing universe repos')
 
 
-def universe_session():
+def add_package_registry_stub() -> dict:
+    stub_url = os.environ['PACKAGE_REGISTRY_STUB_URL']
+    with urllib.request.urlopen(stub_url) as url:
```
elezar: This seems like a very specific check, and I would argue that this is applicable to all universe repos.

Also, is it guaranteed that a stub universe URL only contains a single package?

takirala: Yes, this was intentional. In long term, we will get rid of this method and install package registry from the cluster. I added a TODO for the same.

testing/sdk_repository.py
takirala: No, the urls are supposed to be publicly accessible (at least as of today). An example url would be `https://universe-converter.mesosphere.com/transform?url=https://infinity-artifacts.s3.amazonaws.com/autodelete7d/template/20180410-181638-EhC0WXM14amJN6c7/stub-universe-template.json`

testing/sdk_repository.py
takirala: I tried to do that initially. But since the set of operations are specific to registry and all the `sdk_security` stuff use the `http` api, it didn't feel right to put this in there. We can always move this to `sdk_security` if/when we need to reuse a similar functionality.

testing/sdk_repository.py
takirala: I have looked at those functions. I still feel this is reasonable enough for my use case as I am not interested in the return value and the lambda is generic so the `@retry` can be used across functions. I renamed the function to match the convention.

testing/sdk_cmd.py
takirala: Fixed the stale doc string

testing/sdk_cmd.py
takirala: I did not see the dcos version utils in `sdk_utils.py`. Fixed this now.

testing/sdk_utils.py
takirala: This is easier if debugging for service uid / secret paths. Also wherever needed, we can just use `uuid.uuid()` directly instead of calling `random_string()`.

conftest.py
```diff
@@ -38,12 +39,20 @@
...
 # The following environment variable allows for log collection to be turned off.
-# This is useful, for exampl in testing.
+# This is useful, for example in testing.
 INTEGRATION_TEST_LOG_COLLECTION = str(
     os.environ.get('INTEGRATION_TEST_LOG_COLLECTION', "True")
 ).lower() in ["true", "1"]
 
 
+@pytest.fixture(scope='session', autouse=True)
+def configure_universe(tmpdir_factory):
```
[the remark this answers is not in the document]
takirala: @benclarkwood FYI, Evan suggested that https://github.com/mesosphere/dcos-commons/pull/2465#discussion_r180445827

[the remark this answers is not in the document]
benclarkwood: Cool 👍

benclarkwood:
@takirala I'm cool with this merging if you're pretty confident in it.

benclarkwood:
@elezar Got any other feedback on this?

elezar:
Sorry, been swamped with the TF stuff. Will have a look after lunch.

testing/sdk_package_registry.py
```diff
@@ -0,0 +1,243 @@
...
+        rc, _, _ = sdk_cmd.run_raw_cli(' '.join([
+            'registry',
+            'build',
+            '--build-definition-file={}'.format(str(package_json_file)),
+            '--output-directory={}'.format(dcos_files_path),
+            '--json'
+        ]))
+        assert rc == 0
+    assert os.path.isfile(target), 'No valid .dcos file is built'
+    return target, name, version
```
elezar: Would it perhaps make more sense to return a dict so that the semantics of the tuple are clearer where it is used?

takirala: This seems good enough for now. fwiw, this tuple definition is internal to `sdk_package_registry` and we can change to dict if the complexity of tuple increases beyond 3 values.

testing/sdk_package_registry.py
```diff
@@ -0,0 +1,243 @@
...
+    if os.path.isfile(target):
+        log.info('Skipping build, using cached file : {}'.format(target))
+    else:
+        del package['releaseVersion']
+        del package['selected']
+        package_json_file = tmpdir_factory\
+            .mktemp(sdk_utils.random_string())\
+            .join(sdk_utils.random_string())
+        package_json_file.write(json.dumps(package))
+        rc, _, _ = sdk_cmd.run_raw_cli(' '.join([
```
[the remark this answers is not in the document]
takirala: Yes, it returns a json object containing the name of the target. but i already have an `assert(os.path.isfile(target))` that takes care of both outcomes of if statement so i am ignoring the stdout.

testing/sdk_package_registry.py
```diff
@@ -0,0 +1,243 @@
...
+    target = os.path.join(dcos_files_path, '{}-{}.dcos'.format(name, version))
+    if os.path.isfile(target):
+        log.info('Skipping build, using cached file : {}'.format(target))
+    else:
+        del package['releaseVersion']
+        del package['selected']
+        package_json_file = tmpdir_factory\
+            .mktemp(sdk_utils.random_string())\
+            .join(sdk_utils.random_string())
+        package_json_file.write(json.dumps(package))
```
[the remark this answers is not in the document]
takirala: - `dumps` seemed more cleaner way to do it as i need to get the abs path of the file as a `str` to pass to the registry command and i don't want more nesting.
- Passing named temporary file may not be what we want because we have a case where if we already have a `.dcos` file there is no work to be done and we don't need any temporary files. so passing `tmpdir_factory` makes sense (as this lazily creates files/directories as needed).

tools/ci/test_runner.sh
```diff
@@ -6,6 +6,9 @@ set -e -x
 export DCOS_ENTERPRISE
 export PYTHONUNBUFFERED=1
 export SECURITY
+export PACKAGE_REGISTRY_ENABLED
```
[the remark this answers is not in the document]
takirala: yes, and that is why i am setting the "default" value only in test.sh and not assuming any such defaults beyond the scope of test.sh or test_runner.sh

testing/sdk_utils.py
```diff
@@ -109,6 +117,20 @@ def is_strict_mode():
...
+        random.choice(
+            string.ascii_lowercase +
+            string.digits
+        ) for _ in range(length)
+    )
+
+
+def is_repo_supported(dcos_ver: str, min_dcos_version: str) -> bool:
+    return LooseVersion(dcos_ver.rstrip("-dev")) > \
```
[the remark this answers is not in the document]
takirala: wow, i did NOT know that. otoh, i don't need that method anymore as i am using `sdk_utils.dcos_version_at_least`. i should've cleaned this up.

test.sh
```diff
@@ -250,16 +276,20 @@ docker run --rm \
...
+    ${azure_args} \
     -e SECURITY="$security" \
     -e PYTEST_ARGS="$PYTEST_ARGS" \
-    $FRAMEWORK_ARGS \
+    ${FRAMEWORK_ARGS} \
     -e STUB_UNIVERSE_URL="$STUB_UNIVERSE_URL" \
+    -e PACKAGE_REGISTRY_ENABLED="${package_registry}" \
+    -e PACKAGE_REGISTRY_STUB_URL="${PACKAGE_REGISTRY_STUB_URL}" \
+    -e DCOS_FILES_PATH="$(abs_path "${dcos_files_path}")" \
+    -v "$(abs_path "${dcos_files_path}")":"$(abs_path "${dcos_files_path}")" \
```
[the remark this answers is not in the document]
takirala: Yes, and from what I tested locally, its almost like a "noop" on `-v` when we pass `-v "":""`.

test.sh
```diff
@@ -250,16 +276,20 @@ docker run --rm \
...
-    $azure_args \
+    ${azure_args} \
     -e SECURITY="$security" \
     -e PYTEST_ARGS="$PYTEST_ARGS" \
-    $FRAMEWORK_ARGS \
+    ${FRAMEWORK_ARGS} \
     -e STUB_UNIVERSE_URL="$STUB_UNIVERSE_URL" \
+    -e PACKAGE_REGISTRY_ENABLED="${package_registry}" \
+    -e PACKAGE_REGISTRY_STUB_URL="${PACKAGE_REGISTRY_STUB_URL}" \
+    -e DCOS_FILES_PATH="$(abs_path "${dcos_files_path}")" \
```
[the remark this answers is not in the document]
takirala: That was intentional for couple of reasons: 
- In `sdk_package_registry.py` the business logic varies based upon whether `DCOS_FILES_PATH` is set to a valid path or not.
- There is no harm in doing what I am doing, on the upside, if i pass something like `/Users/myusername/cachefile` as `DCOS_FILES_PATH`, it would be nice to see the same thing in logs as well. And I see no gain by introducing a new string called "/dcos_files" in `test.sh`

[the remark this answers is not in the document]
takirala: yes the subshell from `$()` takes care of that.

conftest.py
```diff
@@ -38,12 +39,20 @@
...
+    if os.environ.get('PACKAGE_REGISTRY_ENABLED', '') == 'true':
+        yield from sdk_package_registry.package_registry_session(tmpdir_factory)
```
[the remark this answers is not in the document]
takirala: The directories/files are created only when we call tmpdir_factory.mk* methods, so i guess there is no harm in passing around the fixture.

testing/sdk_package_registry.py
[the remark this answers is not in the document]
elezar: Thanks for the clarification.

takirala:
Thanks for the review @elezar 
The test is not flaky, I was passing incorrect arguments to `shakedown.get_service_ips`. I fixed it, tested it locally and pushed the changes. Will merge once the test passes.

elezar:
>The test is not flaky, I was passing incorrect arguments to shakedown.get_service_ips. I fixed it, tested it locally and pushed the changes. Will merge once the test passes.

Thanks! I didn't have much time to look into it.

PR #2468, opened 2018-04-10
INFINITY-3500 Upstream default service state manager

nickbp:
Removes the previous requirement that the developer implement their own tracking of which services are running in a multi-service scheduler. Instead, the developer only needs to implement a callback method which consistently constructs (or reconstructs) the service from its name and any context data. The "context data" (max 100KB) allows the developer to store any configuration/context needed to reconstruct the service by their own callback.

Switches the example `hello-world` implementation to use this new mechanism.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
```diff
@@ -187,8 +189,8 @@ public void uninstallService(String serviceName) {
     public void uninstallServices(Collection<String> finishedServiceNames) {
```
benclarkwood: Could be protected, no?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/ServiceStore.java
```diff
@@ -0,0 +1,187 @@
...
+import com.mesosphere.sdk.scheduler.AbstractScheduler;
+import com.mesosphere.sdk.scheduler.SchedulerUtils;
+import com.mesosphere.sdk.storage.Persister;
+import com.mesosphere.sdk.storage.PersisterException;
+import com.mesosphere.sdk.storage.PersisterUtils;
+import com.mesosphere.sdk.storage.StorageError.Reason;
+
+/**
+ * Default implementation of persistent storage which keeps track of the services that have been added to a dynamic
+ * multi-scheduler. For each service entry, the developer is provided with
```
benclarkwood: Finish sentence?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceManager.java
nickbp: No particular reason to, theres a public `uninstallService` which is just wrapping this call. Its conceivable that there would be some utility in marking multiple services for uninstall atomically.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/ServiceStore.java
nickbp: Whoops, will merge for now and fix in a following PR

PR #2469, opened 2018-04-10
Update airgap_linter.py

benclarkwood:


hectorj2f:
@benclarkwood I tested it, and it still complains when using `dcos config set core.dcos_url "https://master.mesos/"` 🤔. Could we add `leader.mesos` and `master.mesos` instead ?

benclarkwood:
@hectorj2f I realized what the issue was. Can you try this version?

PR #2472, opened 2018-04-11
Drop custom executor

kvish:


kvish:
Fixed in the place mentioned by @nickbp

PR #2474, opened 2018-04-12
 INFINITY-3493/3496 Separate getStatus() call, timeout for service removal

nickbp:
Changes to the internals of multi-service status handling. There should be no effect on single-service schedulers.

- Break out the piggybacked `offer()` status checks into a separate dedicated `status()` call. New status codes are added for `DEPLOYING` and `UNINSTALLING`, which are then used to implement the following internal features:
- When a service is being removed from a multi-service scheduler, wait up to 10 minutes (default, configurable via `SchedulerConfig`/env) for the teardown of that service to complete. If the timeout is exceeded, the uninstalling service is forcibly removed, and if the resources associated with that service are offered again later, they will be automatically cleaned up. This timeout is ONLY for removing services from a running scheduler, NOT for uninstall of the scheduler itself, whether in single-mode or multi-mode. For uninstalling the scheduler, there continues to be no timeout (let the operator figure it out). The reason for this timeout in the multi-service removal case is because we're concerned about a running service gradually (and silently) accumulating garbage about long-lost services/resources as agents come and go in the cluster.


sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
```diff
@@ -217,13 +226,42 @@ private void processQueuedOffers() {
         }
     }
 
-    private void evaluateOffers(List<Protos.Offer> offers) {
-        if (isDeregistered.get()) {
-            // Just in case there are unflushed offers in the queue when we're deregistered
-            LOGGER.info("Dropping {} queued offers: Framework is deregistered", offers.size());
-            return;
+    private boolean checkStatuses() {
+        StatusResponse statusResponse = mesosEventClient.status();
```
benclarkwood: So, in the MSScheduler case, this is the global status across all services, right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
```diff
@@ -29,6 +29,7 @@
 import com.mesosphere.sdk.scheduler.Metrics;
 import com.mesosphere.sdk.scheduler.OfferResources;
 import com.mesosphere.sdk.scheduler.MesosEventClient.OfferResponse;
+import com.mesosphere.sdk.scheduler.MesosEventClient.StatusResponse;
```
benclarkwood: Does `ClientStatusResponse` make sense?

It just feels weird to have 1 differentiated StatusResponse and then the undifferentiated StatusResponse that's for a specific purpose.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
nickbp: Yes, this is returned by `MultiServiceEventClient`, which in practice would only return `uninstalled` or `running`, with `uninstalled` being "all services have been wound down, now the scheduler can be taken down"

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
nickbp: SGTM, will rename now while fixing conflicts

benclarkwood: 👍

nickbp:
Actually, I'll break out the deploy exclusivity into a separate PR to follow this one, as it's turning into a pretty large change on its own to do properly. It had started off being a relatively small addition.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
```diff
@@ -227,6 +230,41 @@ private void processQueuedOffers() {
         }
     }
 
+    private boolean checkStatuses() {
```
gabrielhartmann: The method name is a bit confusing to me.  Maybe we could change it to something starting with the word `is`.  Maybe something like `isActive`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -102,6 +119,124 @@ public void unregistered() {
         deregisterStep.setComplete();
     }
 
+    @Override
+    public ClientStatusResponse getClientStatus() {
+        // If the entire scheduler is uninstalling and there are no clients left to uninstall, then tell upstream that
```
benclarkwood: Technically it's "no clients (services)", right?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -102,6 +119,124 @@ public void unregistered() {
...
+        boolean noClients = false;
+
+        Collection<String> finishedServices = new HashSet<>();
+        Collection<AbstractScheduler> uninstalledServices = new HashSet<>();
+
+        Collection<AbstractScheduler> services = multiServiceManager.sharedLockAndGetServices();
+        LOGGER.info("Checking status of {} service{}:", services.size(), services.size() == 1 ? "" : "s");
+        try {
+            if (services.isEmpty()) {
+                // If we don't have any clients, then WE aren't ready.
```
benclarkwood: `s/clients/services`?

I get that AbstractScheduler == MesosEventClient but I find it a little confusing.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -102,6 +119,124 @@ public void unregistered() {
         deregisterStep.setComplete();
     }
 
+    @Override
+    public ClientStatusResponse getClientStatus() {
+        // If the entire scheduler is uninstalling and there are no clients left to uninstall, then tell upstream that
+        // we're uninstalled.
+        boolean noClients = false;
```
benclarkwood: I'd say `noServices`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -115,32 +250,21 @@ public void unregistered() {
      */
     @Override
     public OfferResponse offers(Collection<Protos.Offer> offers) {
-        // If uninstalling, then finish uninstall. Otherwise decline short.
-        boolean noClients = false;
-        // Decline short.
+        // Decline short if any client isn't ready.
         boolean anyClientsNotReady = false;
```
benclarkwood: I like the explicitness of anyServicesNotReady.

But they are technically MesosEventClients so /shrug.

PR #2475, opened 2018-04-12
DCOS_CASSANDRA-60 Backup to S3 compatible object storage server

RHabereder:
Added plans backup-minio and restore-minio (Request #9012 Cassandra Backup to an internal S3 Storage)

### DIFF
--- a/frameworks/cassandra/src/main/dist/svc.yml
+++ b/frameworks/cassandra/src/main/dist/svc.yml
@@ -135,16 +135,8 @@ pods:
         cmd: >
                 export PATH=$MESOS_SANDBOX/python-dist/bin:$PATH ;
                 export LD_LIBRARY_PATH=$MESOS_SANDBOX/openssl-dist:$LD_LIBRARY_PATH ;
-                aws s3 cp container-path/snapshot/ s3://${S3_BUCKET_NAME}/${SNAPSHOT_NAME}/node-${POD_INSTANCE_INDEX}/ --recursive
-        resource-set: sidecar-resources
-      upload-minio:
-        goal: ONCE
-        cmd: >
-                export PATH=$MESOS_SANDBOX/python-dist/bin:$PATH ;
-                export LD_LIBRARY_PATH=$MESOS_SANDBOX/openssl-dist:$LD_LIBRARY_PATH ;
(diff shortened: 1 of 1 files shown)


keithchambers:
Thank you for this PR Raphael.

I would like to see us support Minio and other S3 compatible storage. Is there a way you can see to rework the existing S3 functionality to make this work for both AWS and non-AWS S3?

RHabereder:
Hey Keith. Would you mind to elaborate?

As far as I understand the current functionality, cassandra supports AWS S3 via the "backup-s3" plan. 
With this change it would support all other S3-compatible storages. 

Or do you mean to "merge" the S3-Plans into one intelligent plan that would switch between AWS S3/non-AWS S3?

keithchambers:
I am more thinking the later. i.e., the user wants to backup and restore to storage that supports the S3 API - they don't care if it is AWS S3 or Minio S3.

RHabereder:
I agree, that would reduce the duplicated code a lot too. 
Though I'm not quite sure on how to implement change without it having an impact on AWS S3-Users. 
Maybe I can whip up a prototype this weekend.

Do you think we could merge this change as is, so users could benefit of the already implemented non-aws benefit faster?
I would develop the merge under a separate improvement-story in JIRA, if that is alright with you.

keithchambers:
Thank you for going the extra mile see if there is some way we can support AWS and non-AWS with a single backup S3 plan.

No, we cannot merge this as-is. We support this as part of our Enterprise product. UX is important has changing APIs around on our customers is something we try to minimize.

RHabereder:
This is probably as pretty as it can get. Unfortunately we can't get around an additional parameter (--endpoint-url) if we want to target Non-AWS S3 Storages. 
According to https://github.com/aws/aws-cli/issues/1270 it has to be a parameter as of now. 

So minio/ceph/etc. Users have to pass S3_ENDPOINT_URL to the plan.
I tested backup and restore from AWS and minio, the result was as expected.

What are your thoughts on the current solution?

RHabereder:
@keithchambers Just pinging for a sitrep. Is there anything I can do for this PR, or is the current solution okay to you?

keithchambers:
I will drop a JIRA in the team backlog to review this PR. Sprint run for 2 weeks so expect a reply in that timeframe.

PR #2476, opened 2018-04-12
Using relative hostpath for isolation

kvish:
Using relative hostpath for isolation to support DCOS version < 1.11

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
@@ -570,17 +570,11 @@ public class PodInfoBuilder {
 
         if (isTaskContainer && podSpec.getIsolateTmp()) {
             // Isolate the tmp directory of tasks
+            //switch to SANDBOX SELF after dc/os 1.13
             containerInfo.addVolumes(Protos.Volume.newBuilder()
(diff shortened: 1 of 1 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -570,17 +570,11 @@ private static String getConfigTemplateDownloadPath(ConfigFileSpec config) {
...
-                .setMode(Protos.Volume.Mode.RW)
-                .setSource(Protos.Volume.Source.newBuilder()
-                        .setType(Protos.Volume.Source.Type.SANDBOX_PATH)
-                        .setSandboxPath(Protos.Volume.Source.SandboxPath.newBuilder()
-                                .setPath("tmp")
-                                .setType(Protos.Volume.Source.SandboxPath.Type.SELF)
-                                .build())
-                        .build())
-            );
+                    .setHostPath("tmp")
```
benclarkwood: Indent is funky

benclarkwood: But that isn't the end of the world.

kvish: ill fix it up

PR #2478, opened 2018-04-12
[INFINITY - 2985] backport isolation to sdk 0.40

kvish:


kvish:
backport requires a change in the conditional structure due to useDefaultExecutor flag which no longer exists in master

PR #2480, opened 2018-04-14
DCOS-29245 Reservation discipline: limit simultaneous reservations

nickbp:
Adds support for reservation disciplines, which limits the number of services that may be reserving resources at the same time. This reduces the risk of a deadlock when multiple services are simulatenously consuming available resources. For example, imagine two services which each wanted 51% of the cluster's CPU. By default, this setting is disabled (no limit on services reserving in parallel), but operators may enable it in accordance with any quota configuration they may have.

At the moment, services return `RESERVING` for their entire deploy plan, so this isn't very efficient when  enabled, but again it's disabled by default for now. A separate PR will update the per-service deployment logic into footprint+launch stages, where `RESERVING` would only be advertised during the relatively fast footprint stage, and the launch of tasks into that footprint would instead be advertised as `RUNNING`.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/OfferDiscipline.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Interface used by {@link MultiServiceEventClient} to enforce the services which should or shouldn't be receiving
+ * offers.
+ */
+public interface OfferDiscipline {
+
+    /**
+     * Default implementation for creating an instance, based on the {@link SchedulerConfig}.
+     */
+    public static OfferDiscipline create(SchedulerConfig schedulerConfig, Persister persister) {
```
benclarkwood: I'd rather this logic not reside in the interface from which all disciplines will ultimately inherit.

It looks like we call this only in the Main.java for Hello World. Can you move it there?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/ReservationDiscipline.java
```diff
@@ -0,0 +1,127 @@
...
+        this.reservingMax = reserveLimit;
+        this.selectionStore = selectionStore;
+        this.selectedReservingServices = Optional.empty();
+    }
+
+    /**
+     * Prunes any entries from the persister for services which no longer exist.
+     */
+    @Override
+    public void updateServices(Collection<AbstractScheduler> services) throws PersisterException {
```
benclarkwood: What happens if the MultiService is running with `reservingMax == 10`, 10 services are persisted as reserving, the MultiService is restarted and `reservingMax == 5`?

I think we need to truncate the stored list in that case, no?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/OfferDiscipline.java
```diff
@@ -0,0 +1,44 @@
...
+    public void updateServices(Collection<AbstractScheduler> services) throws Exception;
+
+    /**
+     * Returns whether the specified service should be provided with offers, based on its status response.
+     *
+     * @param statusResponse the status response returned by the service
+     * @param service the active service
+     * @return whether to provide the service with offers
+     */
+    public boolean offersEnabled(ClientStatusResponse statusResponse, AbstractScheduler service);
```
benclarkwood: Is it possible to drop the need for updateServices?

We could track for the user what services they've said should be receiving offers. We could supply it to them in offersEnabled.

e.g. `public boolean offersEnabled(ClientStatusResponse statusResponse, AbstractScheduler candidate, ImmutableSet<AbstractScheduler> receiving)`

Otherwise the user has to use the two methods together perfectly (and keep track of... something). Smells of the annoying tracking of things mesos doesn't do.

nickbp:
(remaining failure in hello-world was due to the known mesos crash bug)

PR #2481, opened 2018-04-16
[INFINITY-2985] Isolation support for elastic search

kvish:
Consume isolation in elastic search pods


frameworks/elastic/universe/config.json
```diff
@@ -124,6 +124,11 @@
           "type": "number",
           "default": 1.0
         },
+        "isolation": {
+          "description": "boolean flag to control isolation",
```
elezar: Would "Enable node isolation" not be a better description?

benclarkwood:
(Also, please include the ticket # in the PR title, e.g. `[PROJECT-####] Title`

frameworks/elastic/universe/config.json
kvish: we arent templating this so this will be removed anyway

PR #2483, opened 2018-04-17
[INFINITY-3520] Make Kafka ciphers configurable and have a secure default.

mpereira:
This PR adds plumbing for a new configuration for TLS cipher suites: `service.security.transport_encryption.ciphers`. That value is plumbed directly to Kafka's `server.properties` `ssl.cipher.suites` configuration.

~However, Kafka doesn't seem to be accepting the current default configuration value. It is a comma-separated list of cipher suite names prefixed with `TLS_`. I added the prefix after seeing it on a couple of places. I've tried it without the prefix too, to no avail. If `service.security.transport_encryption.ciphers` is left blank Kafka accepts connections with the 21 ciphers we saw before. The [documentation](http://kafka.apache.org/documentation/#brokerconfigs) isn't very specific and doesn't provide example values.~

~I've tried:~
~- With `TLS_` prefix, separated by dashes. e.g.: `TLS_AES128-GCM-SHA256,...`~
~- Without `TLS` prefix, separated by dashes. e.g.: `AES128-GCM-SHA256,...`~
~- With `TLS_` prefix, separated by underscores. e.g.: `TLS_AES128_GCM_SHA256,...`~
~- Without `TLS` prefix, separated by underscores. e.g.: `AES128_GCM_SHA256,...`~

~I'll look into this tomorrow morning unless someone feels like taking a swing at it. The added test is currently failing due to the above.~

~Examples of configuring `ssl.cipher.suites` in the wild:~
~- https://github.com/Shopify/sarama/issues/643#issuecomment-216505536~
~- https://phabricator.wikimedia.org/T182993#3855423~


mpereira:
I found out the issue. OpenSSL and the JVM use different names for cipher suites:
- https://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#ciphersuites
- https://testssl.sh/openssl-rfc.mapping.html

mpereira:
Surprisingly enough, Java is following the standard in this case!

PR #2484, opened 2018-04-17
Parameterize version of package from build.sh scripts

takirala:
Now, `build.sh` can take a `version` parameter to populate the package version. Something like : 
`.../build.sh .dcos 1.2.3`. This however forces `elastic` to have same version for `elastic` and `kibana` (which is acceptable for now).

### DIFF
--- a/tools/publish_dcos_file.py
+++ b/tools/publish_dcos_file.py
@@ -28,14 +28,15 @@ logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 _REGISTRY_URL_TEMPLATE = "https://downloads.mesosphere.com/package-registry/" \
                          "binaries/cli/{os}/x86-64/latest/dcos-registry-{os}"
 
+
 class DCOSFilePublisher(object):
 
     def __init__(
             self,
             package_name,
+            package_version,
             input_dir_path,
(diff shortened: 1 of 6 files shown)


tools/publish_dcos_file.py
```diff
@@ -50,32 +51,33 @@ def __init__(
...
-        # Sample s3_directory:
-        # infinity-artifacts/autodelete7d/kafka/20160815-134747-S6vxd0gRQBw43NNy
-        s3_directory_url = os.environ.get(
-            'S3_URL',
-            's3://{}/{}/{}/{}'.format(
-                s3_bucket,
-                s3_dir_path,
-                package_name,
-                dir_name
+        s3_dir_path = os.environ.get('S3_DIR_PATH')
```
benclarkwood: Hm. So this would allow us to drop the second job entire then, right? We could just have the "build permanent" option in the first job.

takirala: Yep, I created a `PUBLISH_DCOS` drop down that has three options:

- DISABLED
- AUTODELETE7D
- PERMANENT

It is `DISABLED` by default and yes, no changes to second job.

PR #2488, opened 2018-04-18
Bump the JRE to 1.8 u171

benclarkwood:


vishnu2kmohan:
+1 for 8u172

benclarkwood:
The guidance from Oracle is to consume the CPU unless you know you need something in the PSU.

http://www.oracle.com/technetwork/java/javase/8u172-relnotes-4308893.html. There also isn't much in the PSU.

benclarkwood:
But I'm a people pleaser, so the people shall receive that which they've asked for.

mpereira:
HW `test_resource_refinement.test_marathon_volume_collission` seems to be failing consistently with this branch.

benclarkwood:
@mpereira it's definitely increasingly flaky.

cc @kvish

PR #2489, opened 2018-04-19
[INFINITY-2985] backport isolation changes

kvish:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -379,9 +380,8 @@ public void setExecutorVolume(VolumeSpec volumeSpec) {
         // This includes networks, rlimits, secret volumes...
         executorInfoBuilder.setContainer(getContainerInfo(podSpec, true, false));
 
-        if (podSpec.getIsolateTmp() && !useDefaultExecutor) {
-            executorInfoBuilder.setContainer(executorInfoBuilder.getContainerBuilder().addVolumes(
-                    Protos.Volume.newBuilder()
+        if (!useDefaultExecutor) {
```
crackcode123: for default executor case what will happen?

benclarkwood: That logic is elsewhere. 

In the 1.9 case we set the tmp isolation directly on the custom executor (only place it’s supported) and for the default executor we set it on the tasks themselves.

crackcode123: ok- in real world, so this code will be actually executed only for 1.9. Thanks for the update.

PR #2491, opened 2018-04-25
INFINITY-3568: Clean up dependencies, add shadowjar output

nickbp:
- Removes a bunch of unused dependencies and drops the separate versions (cross-referencing gets annoying)
- Folds the `common` package into `scheduler`, since `executor` is gone now and there's nothing to be in `common` with.
- Adds a separate `shadowjar` jar output to the scheduler build. This produces a single monolithic JAR of the scheduler and all its dependencies. It also renames the package path of some dependencies from `foo` to `sdkshadow.foo`. This solves some library conflicts encountered downstream when linking the SDK into existing code.

How the shadow jar looks. Note regular jar in `scheduler` and shadow jar in `scheduler-shadow`. Projects that want to use the shadow jar would just need to link against `scheduler-shadow`:
```
$ ./gradlew publishToMavenLocal
$ ls -l ~/.m2/repository/mesosphere/scheduler*/*/
/home/nick/.m2/repository/mesosphere/scheduler/0.50.0-SNAPSHOT/:
-rw-rw-r-- 1 nick nick 853570 May  1 12:07 scheduler-0.50.0-SNAPSHOT.jar
-rw-rw-r-- 1 nick nick 482658 May  1 12:07 scheduler-0.50.0-SNAPSHOT-sources.jar

/home/nick/.m2/repository/mesosphere/scheduler-shadow/0.50.0-SNAPSHOT/:
-rw-rw-r-- 1 nick nick 32095885 May  1 12:08 scheduler-shadow-0.50.0-SNAPSHOT.jar
-rw-rw-r-- 1 nick nick   482658 May  1 12:07 scheduler-shadow-0.50.0-SNAPSHOT-sources.jar
```


sdk/scheduler/build.gradle
```diff
@@ -1,149 +1,173 @@
...
+shadowJar {
+    classifier = ''
+
+    mergeServiceFiles()
+
+    // Rename the following packages so that they do not conflict with downstream builds.
+    // For now, this list specifically consists of libraries that collide within the Jenkins Mesos
+    // plugin, but other cases may be added as needed.
+
+    // io.dropwizard.metrics:metrics-core
```
gabrielhartmann: Want to remove all the commented packages?

nickbp: I left these here to list the packages that were being affected by the rename

sdk/common/build.gradle
```diff
@@ -1,108 +0,0 @@
...
-            }
-        }
-    }
-}
-
-dependencies {
-    compile "com.google.protobuf:protobuf-java:${protobufVer}"
-    compile "commons-collections:commons-collections:${commonsCollectionsVer}"
-    compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
-    compile group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.8.1'
```
jimmidyson: These dependencies have now been lost by anything using this since the removal of common module. I'm not sure if these should be reinstated to scheduler module or left up to the module consumer to add dependency? I don't mind submitting a PR if you let me know which way to go.

PR #2493, opened 2018-04-28
Add get{Optional,Required}Double methods to EnvStore

jimmidyson:
Useful for parsing CPU requirements from env store, consistent with strings, ints, longs


jimmidyson:
@elezar They're as well covered as the other `get{Optional,Required}*` methods - as in, they're not covered at all.

PR #2495, opened 2018-05-03
Fix up plan customization for multiservice schedulers and enable further plan customization based on existing tasks

jimmidyson:
As discussed on Slack - feedback really appreciated!

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
@@ -78,7 +78,7 @@ public class UninstallScheduler extends AbstractScheduler {
             Optional<String> namespace,
             Optional<SecretsClient> customSecretsClientForTests,
             TimeFetcher timeFetcher) {
-        super(serviceSpec, stateStore, planCustomizer, namespace);
+        super(serviceSpec, stateStore, null, planCustomizer, namespace);
         this.logger = LoggingUtils.getLogger(getClass(), namespace);
         this.configStore = configStore;
 
@@ -114,6 +114,8 @@ public class UninstallScheduler extends AbstractScheduler {
                     ? Optional.empty()
                     : Optional.of(timeFetcher.getCurrentTimeMillis() + (uninstallTimeoutSecs * 1000));
         }
+
+        customizePlans();
     }
 
     @Override
--- a/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/DecomissionCustomizer.java
+++ b/frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/DecomissionCustomizer.java
@@ -33,7 +33,11 @@ public class DecomissionCustomizer implements PlanCustomizer {
                             Collections.emptyList()));
         }
 
-        return new DefaultPlan(plan.getName(), updatedPhases, new ParallelStrategy<>());
+        return new DefaultPlan(
+                plan.getName(),
+                updatedPhases,
+                new ParallelStrategy<>()
+        );
     }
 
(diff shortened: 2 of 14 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Plan.java
```diff
@@ -44,4 +47,8 @@ default boolean isRecoveryPlan() {
     default boolean isDecommissionPlan() {
         return getName().equals(DECOMMISSION_PLAN_NAME);
     }
+
+    Optional<String> getNamespace();
+
+    Collection<Protos.TaskInfo> getExistingTasks();
```
nickbp: What calls this? I didn't see anything using it in the PR but could've missed something.

Could the user of this information just query the StateStore directly? The tasks passed into the Plan at construction could fall out of date.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Plan.java
```diff
@@ -44,4 +47,8 @@ default boolean isRecoveryPlan() {
     default boolean isDecommissionPlan() {
         return getName().equals(DECOMMISSION_PLAN_NAME);
     }
+
+    Optional<String> getNamespace();
```
nickbp: Instead of having the plans advertise the service they belong to, how about passing different customizer objects (each bond to a certain namespace) to each of the services when they're being built?

In other words, the logic here is something like this:
```
customization = new ExampleCustomization();
// common customization object, can't tell which service is invoking it:
service1 = new Service(namespace1, customization)
service2 = new Service(namespace2, customization)
```

when it could be:
```
customization = new ExampleCustomization();
// separate per-service subobjects, each bound to specific namespaces:
service1 = new Service(namespace1, customization.getForNamespace(namespace1))
service2 = new Service(namespace2, customization.getForNamespace(namespace2))
```

In the latter structure, the per-service subobjects would already know the context that they're operating in

nickbp: This might do it?: https://github.com/mesosphere/dcos-kubernetes/pull/540

jimmidyson: OK that sounds like a plan (get it?).

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Plan.java
jimmidyson: Much better idea.

jimmidyson:
@nickbp @benclarkwood Any chance of review/merge?

jimmidyson:
Friendly tap :)

Now we're using this for node pools in Kubernetes package and I can't get that merged until this is at least merged...

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallScheduler.java
```diff
@@ -78,7 +78,7 @@ protected UninstallScheduler(
             Optional<String> namespace,
             Optional<SecretsClient> customSecretsClientForTests,
             TimeFetcher timeFetcher) {
-        super(serviceSpec, stateStore, planCustomizer, namespace);
+        super(serviceSpec, stateStore, null, planCustomizer, namespace);
```
nickbp: Can this pass up a constructed PlanCoordinator?

benclarkwood:
Fix is being tracked here: https://issues.apache.org/jira/browse/MESOS-8945

If we can get Kafka green, I'm okay with this going in with bad HW tests (with this known cause)

benclarkwood:
Ah. The kafka test is failing globally as well. Merging...

PR #2498, opened 2018-05-10
[DCOS-29232] sdk_security: more the package functions more configurable

hectorj2f:
It'd be good to make more configurable the sdk_security package. In our case, we need to add additional permissions to the default ones, but we'd like to define the service account or even the linux user.

related to: https://jira.mesosphere.com/browse/DCOS-29232

### DIFF
--- a/frameworks/template/tests/conftest.py
+++ b/frameworks/template/tests/conftest.py
@@ -1,8 +1,12 @@
 import pytest
 import sdk_security
 from tests import config
+from typing import List
 
 
 @pytest.fixture(scope='session')
 def configure_security(configure_universe):
-    yield from sdk_security.security_session(config.SERVICE_NAME)
+    service_account_name = 'service-acct'
+    # Add the permissions you want to grant/revoke during the installation
+    permissions = []
+    yield from sdk_security.security_session(config.SERVICE_NAME, permissions, config.DEFAULT_LINUX_USER, service_account_name)
--- a/testing/sdk_security.py
+++ b/testing/sdk_security.py
@@ -70,7 +70,7 @@ def _revoke(user: str, acl: str, description: str, action: str="create") -> None
     log.info("Want to delete {user}+{acl}".format(user=user, acl=acl))
 
 
-def get_permissions(service_account_name: str, role: str, linux_user: str) -> List[dict]:
+def get_default_permissions(service_account_name: str, role: str, linux_user: str) -> List[dict]:
     return [
         # registration permissions
         {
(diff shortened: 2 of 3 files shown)


frameworks/template/tests/conftest.py
```diff
@@ -1,8 +1,15 @@
...
 
+# Add the permissions you want to grant/revoke during the installation
+def additional_permissions(linux_user, service_account_name) -> List[dict]:
+    return []
 
 @pytest.fixture(scope='session')
 def configure_security(configure_universe):
-    yield from sdk_security.security_session(config.SERVICE_NAME)
+    linux_user = 'root'
+    service_account_name = 'service-acct'
```
mpereira: Maybe it would be more consistent if these values were moved to config.py as constants, e.g. `config.DEFAULT_LINUX_USER` or something like that? Also, does it make sense for `additional_permissions` to be a parameter of `configure_security` instead of a function that is called by it?

hectorj2f:
@mpereira Do you know how I could create this entries using an API (such as `{dcosurl}/acs/api/v1/acls/{acl}/users/{user}/{action}`) endpoint for these permissions ?
```
    dcos:secrets:default:/blabla/* full
    dcos:secrets:list:default:/blabla read
```
I got errors because of the existence of `/` and `*` characters, which makes sense. It forms a weird url.

frameworks/template/tests/conftest.py
elezar: Note that the `linux_user` used to be `nobody` and not `root`. Does this not change behaviour then?

testing/sdk_security.py
```diff
@@ -129,17 +129,19 @@ def get_permissions(service_account_name: str, role: str, linux_user: str) -> Li
         }]
 
 
-def grant_permissions(linux_user: str, role_name: str, service_account_name: str) -> None:
+def grant_permissions(linux_user: str, role_name: str, service_account_name: str, additional_permissions: List[dict]) -> None:
     log.info("Granting permissions to {account}".format(account=service_account_name))
-    permissions = get_permissions(service_account_name, role_name, linux_user)
+    default_permissions = get_permissions(service_account_name, role_name, linux_user)
```
elezar: What if one needs to remove some of the default permissions? Would we not *only* have permissions as an input parameter with `default_permissions` used if these are `None`?

testing/sdk_security.py
```diff
@@ -129,17 +129,19 @@ def get_permissions(service_account_name: str, role: str, linux_user: str) -> Li
...
+    permissions = default_permissions + additional_permissions
     for permission in permissions:
         _grant(**permission)
     log.info("Permission setup completed for {account}".format(account=service_account_name))
 
 
-def revoke_permissions(linux_user: str, role_name: str, service_account_name: str) -> None:
+def revoke_permissions(linux_user: str, role_name: str, service_account_name: str, additional_permissions: List[dict]) -> None:
```
elezar: Should we not keep track of the permissions that were granted and instead pass them in to `revoke_permissions` instead of constructing the list of permissions again?

testing/sdk_security.py
```diff
@@ -215,64 +219,79 @@ def setup_security(framework_name: str,
 
     log.info("Setting up strict-mode security")
```
elezar: Out of scope for this PR, but we could split this into a separate function as we are now only setting the permissions.

testing/sdk_security.py
```diff
@@ -215,64 +219,79 @@ def setup_security(framework_name: str,
...
     @pytest.fixture(scope='session')
     def configure_security(configure_universe):
-        yield from sdk_security.security_session(framework_name)
+        yield from sdk_security.security_session(framework_name, additional_permissions, 'nobody', 'service-acct')
     """
     try:
         is_strict = sdk_utils.is_strict_mode()
         if is_strict:
-            setup_security(framework_name)
+            setup_security(framework_name, additional_permissions, linux_user, service_account, service_account_secret)
```
elezar: As mentioned above, we could return the list of permissions actually granted here and then explicitly remove them in the cleanup step below.

frameworks/template/tests/conftest.py
hectorj2f: I'll change the linux user to `nobody`

testing/sdk_security.py
hectorj2f: I will change the code to pass the permissions or use the defaulted ones, in case the passed list is empty

testing/sdk_security.py
hectorj2f: I agree, let's pass them as a parameter

benclarkwood:
@elezar when you have time

hectorj2f:
@benclarkwood Do you want me to do something here ? What is the status of this PR ?

PR #2500, opened 2018-05-15
DCOS-37527 Upstreamed fixes to test tooling

nickbp:
Various fixes made while catching up spark-build, getting it to where its tooling is largely stock. These changes should also allow us to remove the test runner script in teamcity in favor of just running test.sh directly (spark-build CI is now just running test.sh).

- test.sh:
  - Allow runs from CI: If the user doesn't provide a URL, attempt to create one automatically (was supported already, just removed a gate in test.sh)
  - Improve --help output
  - Clean up docker argument handling, log resulting docker command
  - Handle credentials in env and from disk
- sdk_tasks: Support task names containing spaces (as produced by spark)


test.sh
```diff
@@ -268,28 +266,70 @@ if [ x"$package_registry" == x"true" ]; then
...
+
+while read line; do
+    # Prefix match, then strip prefix in envfile:
+    if [[ "${line:0:8}" = "TEST_SH_" ]]; then
+        echo ${line#TEST_SH_} >> $envfile
+    fi
+done < <(env)
+
+CMD="docker run --rm \
+${docker_interactive_arg} \
```
elezar: `-t` and `-i` are very often specified together. Would it make sense to include them here together?

test.sh
```diff
@@ -268,28 +266,70 @@ if [ x"$package_registry" == x"true" ]; then
...
+--env-file $envfile \
+-v $credsfile:/root/.aws/credentials:ro \
+${ssh_key_volume_arg} \
```
elezar: Could we have a single `volume_args` containing all of these?

testing/sdk_auth.py
```diff
@@ -191,16 +191,14 @@ def _install_marathon_app(app_definition):
             return success
 
         if sdk_marathon.app_exists(self.app_definition["id"]):
-            log.info("Found installed KDC app")
-            return _get_kdc_task(self.app_definition["id"])
+            log.info("Found installed KDC app, destroying it first")
+            sdk_marathon.destroy(self.app_definition["id"])
```
elezar: Why was this change required? We need persistence for the KDC app.

It's a bit roundabout, but basically the `kdc.py` tooling that we use in soak (and for testing) imports `sdk_auth`. And we want the **same** KDC app to be used for multiple runs of the tool.

testing/sdk_tasks.py
```diff
@@ -63,28 +63,34 @@ class Task(object):
...
-            cli_task_tokens[4],
-            cli_task_tokens[5])
+            ' '.join(cli_task_tokens[0:len(cli_task_tokens)-(min_count-1)]),
+            cli_task_tokens[-7],
+            cli_task_tokens[-6],
+            cli_task_tokens[-5],
+            cli_task_tokens[-4],
+            cli_task_tokens[-3],
+            cli_task_tokens[-2],
+            cli_task_tokens[-1])
```
elezar: Will this not break on non-1.11 clusters where there are not zones / regions specified?

testing/sdk_auth.py
nickbp: This was mainly a problem when cleanup() isn't correctly called, e.g. if a test run is aborted. The following run would fail due to the stale kdc.

Would it make sense to start fresh each time for CI tests, but to reuse prior instances for soak tests?

test.sh
```diff
@@ -268,28 +266,70 @@ if [ x"$package_registry" == x"true" ]; then
...
+fi
+
+cat >> $envfile <<EOF
+AWS_PROFILE=$aws_profile
+CLUSTER_URL=$CLUSTER_URL
+DCOS_ENTERPRISE=$enterprise
+DCOS_FILES_PATH=$dcos_files_path
+DCOS_LOGIN_PASSWORD=$DCOS_LOGIN_PASSWORD
+DCOS_LOGIN_USERNAME=$DCOS_LOGIN_USERNAME
+FRAMEWORK=$framework
```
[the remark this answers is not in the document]
nickbp: Done. test.sh now returns an error if:
- Multiple frameworks are available and the user didn't specify one
- The user specified multiple frameworks to test (only allow one at a time)
- The user specified "all" (error message telling user to pick a framework)

test.sh
nickbp: Moved `-t` up to here.

test.sh
nickbp: Done. Volumes are now just added to a single `volume_args` variable.

testing/sdk_tasks.py
nickbp: I *think* the CLI outputs "---" if region/zone are unavailable: https://github.com/dcos/dcos-cli/blob/master/python/lib/dcoscli/dcoscli/tables.py#L45

But agreed it'd make sense to just parse JSON here at this point. I'll try doing that in a following revision to this PR

testing/sdk_auth.py
elezar: Let's make the default behaviour the cleanup behaviour and I can add the current functionality when using `kdc.py`. IIRC we pin to a SHA for soak, so we could do it as a follow-up (feel free to create a ticket if you think it's not part of this PR).

nickbp: To be clear, should the change being made here be retained, or should it be reverted to the prior/current behavior? (I'm fine with either)

nickbp:
Two remaining issues, working on them now:
- Missing ciphers in `testing/security/cipher_suites.py` following ubuntu 16.04->18.04 bump
- The task list parsing fixes in this PR broke `helloworld/tests/test_placement.py:get_task_host(task_name)`

nickbp:
Gonna merge this now, as this contains fixes for many other test flakes and the only remaining failure is due to the mesos regression

PR #2501, opened 2018-05-17
Update to bionic beaver!

benclarkwood:


benclarkwood:
I think a race condition with the docker image generation.

Re-rolling

colin-msphere:
This is failing when installing the `python-software-properties` package through `apt-get`.

benclarkwood:
@colin-msphere thanks!

PR #2504, opened 2018-05-22
Added force-pull flag for the task container images.

joerg84:


joerg84:
I suppose we should add a test...

elezar:
> I suppose we should add a test...

Yes, that would be great as well as documentation for the addition to the spec.

kaiwalyajoshi:
@joerg84 @vishnu2kmohan 
Hey Joerg, we've received a few requests to add this feature, could we resurrect this PR on the lastest master? I'd be more than happy to guide this so that we can land it into master.

PR #2507, opened 2018-05-23
DCOS-34487 Use autoip instead of VIP for fetching config artifacts

nickbp:
With this PR, config fetching is switched from a VIP at port 80 to an autoip endpoint directly on the dynamic port.

Usage of VIPs was originally introduced before autoip was an option. VIPs have the issue of always running on the same port 80: Customers may have implemented firewall rules on agents which effectively block this access.


sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
```diff
@@ -55,20 +60,32 @@ public static String toAutoIpDomain(String serviceName, SchedulerConfig schedule
      */
     public static String toAutoIpHostname(String serviceName, String taskName, SchedulerConfig schedulerConfig) {
         // Unlike with VIPs and mesos-dns hostnames, dots are converted to dashes with autoip hostnames. See DCOS-16086.
```
benclarkwood: It would be nice for the comment to have an example of the translation

benclarkwood: Also, I thought it was just /test/integration/serivce => testintegrationservice?

benclarkwood: I am also confused by the reference to dots

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/taskdata/EnvConstants.java
```diff
@@ -25,9 +25,10 @@ private EnvConstants() {
      *  under this domain. For example, <VIP_NAME>.<FRAMEWORK_VIP_HOST> => vip hostname. */
     public static final String FRAMEWORK_VIP_HOST_TASKENV = "FRAMEWORK_VIP_HOST";
 
-    /** Provides the hostname of the Scheduler's own API. Pods within the service may access scheduler-hosted APIs under
-     *  this hostname. For example, <SCHEDULER_API_HOSTNAME>/v1/state/files/notavirus.exe.txt */
+    /** Provides the hostname/port of the Scheduler's own API. Pods within the service may access scheduler-hosted APIs
+     *  under this hostname. For example, &lt;hostname>:&lt;port>/v1/state/files/notavirus.exe.txt */
```
benclarkwood: I don't know what happened to these angle brackets

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
benclarkwood: Oh, this is a task name thing?

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/endpoints/MultiArtifactResourceTest.java
```diff
@@ -1,28 +1,42 @@
...
-                MultiArtifactResource.getUrlFactory("fwk-name", "job-name").get(uuid, "some-pod", "some-task", "some-config"));
+                MultiArtifactResource.getUrlFactory("fwk-name", "job-name", mockSchedulerConfig).get(uuid, "some-pod", "some-task", "some-config"));
     }
 
     @Test
     public void testGetQueuesTemplateUrlSlashed() {
         UUID uuid = UUID.randomUUID();
         assertEquals(
-                "http://api.pathtofwk-name.marathon.l4lb.thisdcos.directory/v1/service/path.to.job-name/artifacts/template/"
+                "http://pathtofwk-name.marathon.some.tld:1234/v1/service/path.to.job-name/artifacts/template/"
```
benclarkwood: This expected is wrong should be fwk-name-to-path?

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
nickbp: Turns out autoip has different string mangling depending on whether it's the task name or the framework name

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/taskdata/EnvConstants.java
nickbp: this fixes rendering in javadoc html

nickbp:
Only remaining failures are due to the known mesos bug

PR #2510, opened 2018-05-25
Make sure MESOS_CONTAINER_IP is set before running helper

benclarkwood:
setup-helper relies on MESOS_CONTAINER_IP in the no security case which.... is not set yet. Whoops!

### DIFF
--- a/frameworks/kafka/src/main/dist/svc.yml
+++ b/frameworks/kafka/src/main/dist/svc.yml
@@ -87,7 +87,8 @@ pods:
 
           # setup-helper determines the correct listeners and security.inter.broker.protocol.
           # it relies on the task IP being stored in MESOS_CONTAINER_IP
-          ./bootstrap -resolve=false -template=false
+          export MESOS_CONTAINER_IP=$( ./bootstrap --get-task-ip )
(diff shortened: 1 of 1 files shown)


frameworks/kafka/src/main/dist/svc.yml
```diff
@@ -86,6 +86,8 @@ pods:
       {{/TASKCFG_ALL_SECURITY_KERBEROS_ENABLED}}
 
           # setup-helper determines the correct listeners and security.inter.broker.protocol.
+          # it relies on the task IP being stored in MESOS_CONTAINER_IP
+          ./bootstrap -resolve=false -template=false
```
elezar: As a medium-term change we may want to make it more explicit as to what `bootstrap` is doing.

elezar:
@benclarkwood if I recall correctly from one of our other discussions with a community member, this is not sufficient to address the some issues (#2512)?

frameworks/kafka/src/main/dist/svc.yml
benclarkwood: Yes. I always forget that Go don't set parent envvars. Drat.

PR #2514, opened 2018-05-31
Add sdk_install.get_package_json

colin-msphere:
Add `get_package_json` to `sdk_install` that returns the Marathon app definition for a package
as a dict object. This can then be used in a later `POST` to Marathon directly.

Example:

```
options = {"service": {"name": service_name}}
pkg_json = sdk_install.get_package_json('jenkins', None, options)
client = shakedown.marathon.create_client()
client.add_app(pkg_json)
```

### DIFF
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -32,6 +32,26 @@ Used by post-test diagnostics to retrieve stuff from currently running services.
 _installed_service_names = set([])
 
 
+def get_package_json(package: str,
+                     version: str=None,
+                     options: dict=None) -> dict:
+    """Get the Marathon app JSON for a given package with the desired
+    options.
+
+    Args:
+        package: Package name
+        version: Package version
+        options: Package installation options
+
+    Returns: Marathon app def
+
+    """
+    pkg_version = dcos.packagemanager.CosmosPackageVersion(
+            package, version, dcos.cosmos.get_cosmos_url())
+    opts = options if options else dict()
+    return pkg_version.marathon_json(opts)
+
+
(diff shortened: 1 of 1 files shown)


mpereira:
Thanks @colin-msphere.

The name `get_package_json()` might confuse someone thinking this actually returns the contents of one of our `package.json` files (e.g. [Cassandra's package.json](https://github.com/mesosphere/dcos-commons/blob/master/frameworks/cassandra/universe/package.json#L15)).

Do you think `get_marathon_json()` might be a better name?

elezar:
In many (if not most) other cases, we interact with a cluster using the CLI. This was prompted by issues in upstream dependencies. If I recall correctly, one can also render the marathon json using the CLI?

Also 👍 to @mpereira's comment regarding the naming of the function.

testing/sdk_install.py
```diff
@@ -32,6 +32,26 @@
 _installed_service_names = set([])
 
 
+def get_package_json(package: str,
```
benclarkwood: As @mpereira suggested, let's call this `get_marathon_json`.

testing/sdk_install.py
```diff
@@ -32,6 +32,26 @@
...
+
+    Args:
+        package: Package name
+        version: Package version
+        options: Package installation options
+
+    Returns: Marathon app def
+
+    """
+    pkg_version = dcos.packagemanager.CosmosPackageVersion(
```
benclarkwood: As Evan said, if at all possible we should hit the CLI rather than the underlying library as the CLI team plans to drop the library.

PR #2521, opened 2018-06-06
Add ipc.maximum.data.length config option for hdfs

swoehrl-mw:
We encountered a problem with HDFS where hdfs had too many files and block reports to the namenodes failed because the reports were too big (see https://community.hortonworks.com/questions/101841/issue-requested-data-length-146629817-is-longer-th.html for more info). To fix this the option ipc.maximum.data.length needs to be set to a higher value (default is 64MB).
This PR adds the setting to the core-site.xml and as a optional config option to the framework.

We run this change in our production cluster with a value of 256MB.

### DIFF
--- a/frameworks/hdfs/src/main/dist/core-site.xml
+++ b/frameworks/hdfs/src/main/dist/core-site.xml
@@ -41,6 +41,10 @@
         <value>{{IPC_CLIENT_CONNECT_MAX_RETRIES}}</value>
     </property>
     <property>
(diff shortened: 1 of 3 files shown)


vishnu2kmohan:
@swoehrl-mw Thank you for your patience, we've adopted your changes for the impending update to HDFS (2.6.0-cdh5.16.1): https://github.com/mesosphere/universe/pull/2107

PR #2522, opened 2018-06-06
DCOS-37981 Just retrieve individual packages, instead of whole repo

nickbp:
- Use new endpoint to just download info for the targeted package, instead of all packages
- Just use urllib instead of `requests` -> `curl` fallback

### DIFF
--- a/tools/universe/package_manager.py
+++ b/tools/universe/package_manager.py
@@ -9,13 +9,9 @@ import subprocess
 import json
 import os
 import tempfile
+import urllib.request
 
 from . import package
-try:
-    import requests
-    _HAS_REQUESTS = True
-except ImportError:
-    _HAS_REQUESTS = False
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


tools/universe/package_manager.py
```diff
@@ -9,129 +9,64 @@
...
 
 LOGGER = logging.getLogger(__name__)
 
 
 class PackageManager:
     """A simple package manager for retrieving universe packages"""
-    def __init__(self, universe_url="https://universe.mesosphere.com/repo",
-                 dcos_version="1.10",
+    def __init__(self,
+                 universe_package_prefix="https://universe.mesosphere.com/package/",
```
elezar: When I run: 
```
curl -L --header "User-Agent: dcos/1.11" --header "Accept: application/vnd.dcos.universe.repo+json;charset=utf-8;version=v4" https://universe.mesosphere.com/package/
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>865E5C49FA467D26</RequestId><HostId>zel8PgJwmSQXkuFzhuoFpqaLX2F5dQ7ugbeB7snM7BtbMia45hbnZHZOd2t+j5xc0eHrYsy8fMw=</HostId></Error>%```

(whereas `repo` still returns data).

Does one need to be authenticated for this?

Comment (elezar) [reply tools/universe/package_manager.py]: OK. One also needs the package name.

File: tools/universe/package_manager.py [hunk 2]
```diff
@@ -9,129 +9,64 @@
...
-            self._get_packages = _get_packages_with_curl
 
     def get_package_versions(self, package_name):
         """Get all versions for a specified package"""
+        if self._dry_run:
+            return DryRunPackages()
+
+        if package_name not in self.__package_cache:
+            LOGGER.info("Retrieving information for package: %s", package_name)
+            url = self._universe_package_prefix + package_name
```
elezar: We could use `urllib.parse.urljoin(self._universe_package_prefix, package_name)` instead?

nickbp: Good point, done.

PR #2524, opened 2018-06-11
make S3 dir name configurable via env var in same fashion as s3 bucket and path

jimmidyson:
We use this in Kubernetes builds to publish our nightly builds to predictable paths. We have a local update for it, but adding this to dcos-commons would simplify our upgrade process and lessen the chances of us breaking our nightly pipeline on upgrade.

### DIFF
--- a/tools/publish_aws.py
+++ b/tools/publish_aws.py
@@ -50,33 +50,8 @@ class AWSPublisher(object):
                 raise Exception(err)
             self._artifact_paths.append(artifact_path)
 
-        s3_bucket = os.environ.get('S3_BUCKET')
-        if not s3_bucket:
-            s3_bucket = 'infinity-artifacts'
-        logger.info('Using artifact bucket: {}'.format(s3_bucket))
-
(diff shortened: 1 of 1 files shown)


tools/publish_aws.py
```diff
@@ -56,9 +56,11 @@ def __init__(
         logger.info('Using artifact bucket: {}'.format(s3_bucket))
 
         s3_dir_path = os.environ.get('S3_DIR_PATH', 'autodelete7d')
-        dir_name = '{}-{}'.format(
-            time.strftime("%Y%m%d-%H%M%S"),
-            ''.join([random.SystemRandom().choice(string.ascii_letters + string.digits) for i in range(16)]))
+        s3_dir_name = os.environ.get('S3_DIR_NAME')
```
elezar: Not that it's critical, but we could break out the handling of these three variables to a function that returns a tuple `(bucket, dir_path, dir_name)`? Or even *just* `(s3_directory_url, http_directory_url)`.

jimmidyson: I've given it a go... looks OK to me, let me know what you think.

PR #2527, opened 2018-06-13
Add strict checking to svc yml tests

elezar:
While reviewing mesosphere/prometheus#12 I noted that there were issues due to undefined envvars in `svc.yml` were not triggering any errors in the unit tests.

This PR adds a strict mode in the `RawServiceSpec.Builder` that is enabled for the tests.

It may be good to backport the change too so that it ends up in an earlier SDK package.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawServiceSpec.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawServiceSpec.java
@@ -65,8 +65,8 @@ public class RawServiceSpec {
         /**
          * Set whether the rendering should be strict.
          */
-        public Builder setStrictRendering(boolean isStrict) {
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/RawServiceSpec.java
```diff
@@ -61,6 +62,14 @@ public Builder setEnv(Map<String, String> env) {
             return this;
         }
 
+        /**
+         * Set whether the rendering should be strict.
+         */
+        public Builder setStrictRendering(boolean isStrict) {
```
nickbp: How about an `enableStrictRendering()` that sets the value to `true`?

elezar: Done.

PR #2528, opened 2018-06-14
[DCOS-38168] Port testing changes from spark-build.

mpereira:
1. Fix sdk_marathon.install_app infinite waiting bug.
The shakedown.wait_for_task call keeps blocking even after the app task starts running.

2. Allow installation of Marathon apps with slashes in the name.

### DIFF
--- a/testing/sdk_marathon.py
+++ b/testing/sdk_marathon.py
@@ -88,6 +88,22 @@ def get_config(app_name, timeout=TIMEOUT_SECONDS):
     return config
 
 
+def is_app_running(app: dict) -> bool:
+    return app['tasksStaged'] == 0 and app['tasksUnhealthy'] == 0 and app['tasksRunning'] > 0
+
+
+def wait_for_deployment_and_app_running(app_name: str, timeout: int):
+    shakedown.deployment_wait(timeout, app_name)
+
+    def app_running():
+        cmd = 'marathon app show {}'.format(app_name)
+        log.info('Running %s', cmd)
+        app = sdk_cmd.get_json_output(cmd)
+        return is_app_running(app)
+
+    shakedown.time_wait(app_running, timeout_seconds=timeout)
+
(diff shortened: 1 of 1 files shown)


testing/sdk_marathon.py
```diff
@@ -88,6 +88,22 @@ def wait_for_response():
...
+def wait_for_deployment_and_app_running(app_name: str, timeout: int):
+    shakedown.deployment_wait(timeout, app_name)
+
+    def app_running():
+        cmd = 'marathon app show {}'.format(app_name)
+        log.info('Running %s', cmd)
+        app = sdk_cmd.get_json_output(cmd)
+        return is_app_running(app)
+
+    shakedown.time_wait(app_running, timeout_seconds=timeout)
```
elezar: I would love to use `retrying` instead, but since we already rely on `shakedown` here, I'm happy to leave this as is.

testing/sdk_marathon.py
```diff
@@ -88,6 +88,22 @@ def wait_for_response():
     return config
 
 
+def is_app_running(app: dict) -> bool:
+    return app['tasksStaged'] == 0 and app['tasksUnhealthy'] == 0 and app['tasksRunning'] > 0
```
elezar: If these keys don't exist in the `app` dict, we're going to get some verbose stack traces, but that's probably OK.

mpereira: https://github.com/mesosphere/dcos-commons/pull/2528/commits/53475bc91153c6abbdb00fc38277b9b8d99654ba

testing/sdk_marathon.py
mpereira: https://github.com/mesosphere/dcos-commons/pull/2528/commits/18f68c6cb4bf3a4708effdadf5c1ad8c0ed010d6

PR #2537, opened 2018-06-19
DCOS-38151 Suppress offers when service(s) are idle

nickbp:
When the MesosEventClient returns an `IDLE` status (indicating that _all_ plans are `COMPLETE`), suppress offers from Mesos.

- `ReviveManager` (handling offer revive/suppress calls themselves) is moved from the per-service level to the global/framework level, with its own internal `TokenBucket` to handle rate limiting. Meanwhile the work-tracking part stays at the per-Service level in the form of `WorkSetTracker`. Services now just track their pending work locally, tell the upstream framework when they have new work pending.
- `ClientStatusResponse` statuses are now organized as follows: `WORKING/FOOTPRINT`, `WORKING/LAUNCH`, `IDLE/START_UNINSTALL`, `IDLE/REMOVE_CLIENT`, and `IDLE/NONE`. The `WORKING` states all indicate that the service should be receiving offers, while the `IDLE` states all indicate that the service doesn't need offers.
- Within the `ReviveManager` the framework is considered still suppressed until we receive a new offer, indicating that the framework is no longer suppressed. This reduces the risk of e.g. a revive call getting lost in transit. Any retries are subject to rate limiting via `TokenBucket`.
- As a backstop measure (and to support performance testing), the suppress call can be disabled by manually setting a `DISABLE_SUPPRESS` envvar in the scheduler.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -176,83 +164,74 @@ public OfferResponse offers(Collection<Protos.Offer> offers) {
...
-                            offerResponse.recommendations.size(),
-                            offerResponse.recommendations.size() == 1 ? "" : "s",
-                            remainingOffers.size(),
-                            remainingOffers.size() == 1 ? "" : "s");
-                }
-                if (!readyForOffers) {
-                    anyServicesNotReady = true;
-                }
+            if (allServicesIdle) {
+                // ALL services are idle, so we can tell upstream to suppress.
```
benclarkwood: I would just drop this comment. It's talking about an upstream effect outside the scope of this class.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -176,83 +164,74 @@ public OfferResponse offers(Collection<Protos.Offer> offers) {
...
+                clientStatusToReturn = ClientStatusResponse.idle();
+            } else if (anyServicesFootprint) {
+                // One or more of the services is getting footprint, so tell upstream that we're getting footprint.
+                clientStatusToReturn = ClientStatusResponse.footprint(anyServicesHaveNewWork);
+            } else {
+                // Otherwise, one or more services isn't idle, and none of them is getting footprint.
```
benclarkwood: s/is/are

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -282,6 +261,58 @@ public OfferResponse offers(Collection<Protos.Offer> offers) {
...
+        // Decline short if any service isn't ready.
+        boolean anyServicesNotReady = false;
+        List<OfferRecommendation> recommendations = new ArrayList<>();
+
+        List<Protos.Offer> remainingOffers = new ArrayList<>();
+        remainingOffers.addAll(offers);
+        for (String serviceName : serviceNamesToGiveOffers) {
+            Optional<AbstractScheduler> service = multiServiceManager.getService(serviceName);
+            if (!service.isPresent()) {
+                // Service was removed since getClientStatus() calculated this list?
```
benclarkwood: Can we log this discrepancy?

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
```diff
@@ -207,12 +231,18 @@ private void processQueuedOffers(Duration queueWait) {
...
+                } else {
+                    // Offers not needed, at least not right now. Decline long.
+                    declineLong(offers);
                 }
             } finally {
                 context.stop();
             }
+
+            // After the status check, see if a revive is now needed:
+            reviveManager.reviveIfRequested();
```
benclarkwood: My only gripe with using the timeout of `OfferQueue.takeAll(...)` to turn the crank in the suppressed case is that we will be bounded by the offer wait time (5 seconds) with respect to recovery of failed tasks.

I can't immediately think of a better way to do things. Ideally, the `RecoveryManager(s)` would have a fast-lane to triggering a revive in the suppressed case.

benclarkwood: Yeah, it's almost certainly safer to keep the logic the same for all cases, but that 5 seconds is annoying.

Really, we just need them to change suppress to not include reservations.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -176,83 +164,74 @@ public OfferResponse offers(Collection<Protos.Offer> offers) {
...
                 ClientStatusResponse statusResponse = service.getClientStatus();
-                if (statusResponse.result != ClientStatusResponse.Result.RUNNING) {
-                    // Only log status when it's special/unusual
-                    LOGGER.info("{} status: {}", serviceName, statusResponse.result);
+                if (!statusResponse.equals(ClientStatusResponse.idle())) {
+                    // Only log status when it's active
+                    LOGGER.info("{} status: {}", serviceName, statusResponse);
                 }
 
                 // Note: We ALWAYS invoke the offer discipline regardless of status response. This allows the offer
```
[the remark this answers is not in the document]
nickbp: Renamed to `updateServiceStatus` and changed related javadocs/comments.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/OfferProcessor.java
nickbp: I mean, to be clear, there should only be a delay here if the revive is throttled. Otherwise the revive is sent out right as the current offer cycle finishes processing.

benclarkwood: @greggomann @gkleiman this is where having suppress _not_ suppress reservations would be <3

PR #2539, opened 2018-06-20
DCOS-38238: Temporarily disable tests that are failing due to Mesos Fetcher regression.

mpereira:
https://issues.apache.org/jira/browse/MESOS-9008

### DIFF
--- a/frameworks/elastic/tests/test_sanity.py
+++ b/frameworks/elastic/tests/test_sanity.py
@@ -167,7 +167,8 @@ def test_custom_yaml_base64():
 
 @pytest.mark.sanity
 @pytest.mark.timeout(60 * 60)
-@pytest.mark.skip(reason='MESOS-9008: Mesos Fetcher fails to extract Kibana archive')
+@pytest.mark.skipif(sdk_utils.is_master_channel(),
+                    reason='MESOS-9008: Mesos Fetcher fails to extract Kibana archive')
 def test_xpack_toggle_with_kibana(default_populated_index):
     log.info("\n***** Verify X-Pack disabled by default in elasticsearch")
     config.verify_commercial_api_status(False, service_name=foldered_name)
--- a/frameworks/elastic/tests/test_sanity.py
+++ b/frameworks/elastic/tests/test_sanity.py
@@ -167,8 +167,7 @@ def test_custom_yaml_base64():
 
 @pytest.mark.sanity
 @pytest.mark.timeout(60 * 60)
-@pytest.mark.skipif(sdk_utils.is_master_channel(),
-                    reason='MESOS-9008: Mesos Fetcher fails to extract Kibana archive')
+@pytest.mark.skip(reason='MESOS-9008: Mesos Fetcher fails to extract Kibana archive')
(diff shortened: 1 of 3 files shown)


frameworks/elastic/tests/test_sanity.py
```diff
@@ -167,6 +167,7 @@ def test_custom_yaml_base64():
 
 @pytest.mark.sanity
 @pytest.mark.timeout(60 * 60)
+@pytest.mark.skip(reason='MESOS-9008: Mesos Fetcher fails to extract Kibana archive')
```
elezar: Could we restrict this to the master builds?

elezar:
We also need to ensure that we track enabling the tests again.

frameworks/elastic/tests/test_sanity.py
mpereira: Good idea! https://github.com/mesosphere/dcos-commons/pull/2539/commits/9e8fc02d270418b1b495c26bf02d4cb392da8555

mpereira:
> We also need to ensure that we track enabling the tests again.

I created a ticket to track that: https://jira.mesosphere.com/browse/DCOS-38370

testing/sdk_utils.py
```diff
@@ -120,6 +121,18 @@ def is_strict_mode():
     return os.environ.get('SECURITY', '') == 'strict'
 
 
+def is_channel(channel):
+    '''Determine if the tests are being run on a DC/OS cluster with a particular channel.'''
+    active_channel = os.environ.get('DCOS_LAUNCH_CONFIG_CHANNEL', '')
+    return channel and active_channel and re.match(channel, active_channel)
```
elezar: Is `re.match` not overkill here? Don't we just want `channel == active_channel`?

testing/sdk_utils.py
```diff
@@ -120,6 +121,18 @@ def is_strict_mode():
     return os.environ.get('SECURITY', '') == 'strict'
 
 
+def is_channel(channel):
```
elezar: This is good, but changes here would cause all frameworks to be rebuilt. I think we could use `dcos_version > 1.11` as a substitute.

frameworks/elastic/tests/test_sanity.py
```diff
@@ -167,6 +167,8 @@ def test_custom_yaml_base64():
 
 @pytest.mark.sanity
 @pytest.mark.timeout(60 * 60)
+@pytest.mark.skipif(sdk_utils.is_master_channel(),
```
elezar: How about `sdk_utils.dcos_version_at_least('1.12')` to prevent changes to `sdk_utils` and rebuilds of all frameworks?

testing/sdk_utils.py
mpereira: `DCOS_LAUNCH_CONFIG_CHANNEL` gets appended the cluster security mode, e.g. `testing/master/strict`.

testing/sdk_utils.py
mpereira: I agree, that's better.

PR #2540, opened 2018-06-20
DCOS-37673: Upstream buildchain modifications from edgelb

vespian:
## High-level description

This PR upstreams some the build chain modifications we had in edgelb's SDK integration[1]:
* permit specifying custom Go linker flags in `build_go_exe.sh`. We were using it to add version information to `edgelb` CLI via `git describe --tags`
* force rebuilding of all the native executable deps when deciding if the rebuild of all the executables is necessary.
* airgap linter should not try to lint binary files

I do not have much information on the technical reasoning behind the second bullet point apart from the edgelb's original git commit message [2]:

```
Switched cli builds to `go build -a`. There is some weird caching that
breaks compiling without the `-a` flag
```

[1] https://github.com/mesosphere/dcos-edge-lb/tree/master/framework
[2] https://github.com/mesosphere/dcos-edge-lb/commit/94635e25fea71c9f4f138b50f68710a9660099d2

## Corresponding DC/OS tickets 

* https://jira.mesosphere.com/browse/DCOS-37673 `edgelb: SDK buildscripts cleanup`

### DIFF
--- a/tools/airgap_linter.py
+++ b/tools/airgap_linter.py
@@ -13,9 +13,18 @@ import sys
 import os
 
 
+def readlines_if_text_file(filename):
+    try:
+        with open(filename, "r", encoding="utf8",) as fh:
+            return fh.readlines()
+    except UnicodeDecodeError as e:
+        msg = "Skipping extracting uris from file `{}`, looks like binary one: {}"
+        print(msg.format(filename, e))
+        return []
+
+
 def extract_uris(file_name):
-    with open(file_name, "r") as file:
-        lines = file.readlines()
+    lines = readlines_if_text_file(file_name)
 
     matcher = re.compile(".*https?:\/\/([^\?\s]*)", re.IGNORECASE)
     matches = []
@@ -72,6 +81,7 @@ def is_bad_uri(uri, file_name):
 
     return True
 
+
 def get_files_to_check_for_uris(framework_directory):
     # There's a set of files that will always be present.
     files = [os.path.join(framework_directory, "universe", "config.json"),
@@ -100,9 +110,8 @@ def validate_all_uris(framework_directory):
 def validate_images(framework_directory):
     files = get_files_to_check_for_uris(framework_directory)
 
-    for file in files:
-        with open(file, "r") as file:
-            lines = file.readlines()
+    for file_name in files:
+        lines = readlines_if_text_file(file_name)
 
         bad_image = False
         for line in lines:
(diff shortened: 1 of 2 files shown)


tools/airgap_linter.py
```diff
@@ -14,8 +14,13 @@
 
 
 def extract_uris(file_name):
-    with open(file_name, "r") as file:
-        lines = file.readlines()
+    try:
+        with open(file_name, "r", encoding='utf8',) as file:
+            lines = file.readlines()
+    except UnicodeDecodeError as e:
```
vespian: We should probably do this validation in `get_files_to_check_for_uris` so that we keep things DRY, but this would require some more refactoring.

elezar: It seems as if we could just add the following function:
```python
def readlines_if_text_file(filename: str) -> List[str]:
    try:
        with open(filename, "r", encoding="utf8",) as fp:
            return file.readlines()
    except UnicodeDecodeError as e:
        print("The error message from below")
        return [] 
```

we can then safely call this function in `exctract_uris` and in the `for file_name in files:` loop in `validate_images` as:
```python
lines = readlines_if_text_file(file_name)
```
as the code in both cases seems to deal with the resultant empty list case correctly.

tools/build_go_exe.sh
```diff
@@ -89,7 +89,7 @@ cd $GOPATH_EXE_DIR
 # previous native build. if the sha1 matches, then we can skip the rebuild.
 NATIVE_FILENAME=".native-${EXE_BASE_NAME}"
 NATIVE_SHA1SUM_FILENAME="${NATIVE_FILENAME}.sha1sum"
-go build -o $NATIVE_FILENAME
+go build -a -ldflags "${GO_LDFLAGS}" -o $NATIVE_FILENAME
```
elezar: What about the case where `GO_LDFLAGS` is not defined?

elezar: Is the SHA time-dependent, or does the addition of the `-a` flag now mean that we will always build the platform-specific versions below?

vespian: > What about the case where GO_LDFLAGS is not defined?

An empty string will be used instead. We are not using bash strict mode (i.e. '-u' option) in this script. None the less I have added variable initialization.

> Is the SHA time-dependent, or does the addition of the -a flag now mean that we will always build the platform-specific versions below?

Neither. We are just making sure that before skipping the rebuild for other platforms, **nothing** has changed. We achieve it by forcefully rebuilding both the CLI and its dependencies. Basing on my quick tests, SHA is not time-dependent.

vespian:
@elezar  Ready for another pass.

vespian:
@elezar Done, ready for another pass.

vespian:
@mpereira  Thank you!

PR #2534, opened 2018-06-19
Update kafka to 1.1.0

Gupta-Amrit:

PR #2531, opened 2018-06-15
[BACKPORT] Update Kafka base to version 1.1.0 (#2509)

samvantran:
This backports PR #2509 which upgrades Kafka to version 1.1.0

PR #2542, opened 2018-06-21
Update kafka-client image for Kafka tests

samvantran:
Resolves [DCOS-38359](https://jira.mesosphere.com/browse/DCOS-38359)

This PR updates the `kafka-client` used in the Kafka tests and will be necessary when upgrading Kafka to version 1.1.0. 

Commit [39b7994](https://github.com/mesosphere/dcos-commons/commit/39b79941874522adf100062f5808757abd48d1b1) will verify that the `kafka-client` is backwards compatible and if so, this PR will be merged with #2534.

### DIFF
--- a/frameworks/kafka/tests/client.py
+++ b/frameworks/kafka/tests/client.py
@@ -93,7 +93,7 @@ class KafkaClient:
             "container": {
                 "type": "MESOS",
                 "docker": {
-                    "image": "elezar/kafka-client:4b9c060",
+                    "image": "elezar/kafka-client:deca3d0",
                     "forcePullImage": True
                 }
             },


elezar:
@samvantran I see the first ACL test succeeds, but the second one fails on Kafka installation. I believe this is because we are maintaining some state in ZK between the two tests. 

I have added a call to clear the ACLs that we add before uninstalling Kafka in the two tests. Let's see if that has the desired effect.

frameworks/kafka/tests/client.py
```diff
@@ -93,7 +93,7 @@ def install(self, kerberos: sdk_auth.KerberosEnvironment=None) -> dict:
             "container": {
                 "type": "MESOS",
                 "docker": {
-                    "image": "elezar/kafka-client:4b9c060",
+                    "image": "elezar/kafka-client:deca3d0",
```
benclarkwood: Ya'll, I hate to be that guy but...

1. We need to move this dockerfile into the kafka directory of this repo
2. We need to host the docker image in mesopshere docker somewhere (ideally with a jenkins job to update it)

elezar: Yes, there is a ticket for it.

samvantran: This one? https://jira.mesosphere.com/browse/DCOS-29527

samvantran:
@elezar, doesn't look like cleaning up ZK state resolved the issue; plus previous master tests pass without having to do that. My guess is the kafka-client update and kafka upgrade to 1.1.0 must occur together (since all tests pass on #2531). I merged the changes from #2534 into this branch to test the theory.

frameworks/kafka/tests/client.py
benclarkwood: This seems like the perfect PR to fix this then :)

elezar:
@benclarkwood could we merge this to also close #2534. I will then address getting the Kafka client images into the `dcos-commons` repo in the remainder of this sprint.

frameworks/kafka/src/test/java/com/mesosphere/sdk/kafka/scheduler/ServiceTest.java
```diff
@@ -52,6 +52,6 @@ private ServiceTestRunner getDefaultRunner() {
 
         return new ServiceTestRunner()
                 .setPodEnv("kafka", map)
-                .setBuildTemplateParams("kafka-version", "2.11-1.0.0"); // set by build.sh/versions.sh
+                .setBuildTemplateParams("kafka-version", "2.12-1.1.0"); // set by build.sh/versions.sh
```
elezar: We could use a dummy version number here so that we don't need to keep updating it for every version bump. This was also done for elastic in #2536.

frameworks/kafka/tests/client.py
elezar: I would prefer to address this in a separate PR.

elezar: (The reason for doing this in a separate PR is that the build tooling would also need to be updated, so there are a number of moving parts)

frameworks/kafka/src/test/java/com/mesosphere/sdk/kafka/scheduler/ServiceTest.java
samvantran: sounds good to me

samvantran: fixed in commit [1f9fe95](https://github.com/mesosphere/dcos-commons/pull/2542/commits/1f9fe955e4ed6f60b4b03f56bb80313341b3dbce)

PR #2544, opened 2018-06-26
[DCOS-38469] Call suppress again on receiving an offer when we are suppressed

takirala:
Address [DCOS-38469](https://jira.mesosphere.com/browse/DCOS-38469)


nickbp:
Going to just merge since the change is an additional unit test, and the unit tests passed.

PR #2546, opened 2018-06-26
DCOS-38473 Log and exit if exception is about to reach Mesos code

nickbp:
If an exception is passed up into the Mesos HTTP adapter, it could result in a zombie scheduler. Instead, log the exception and restart the scheduler process.

Also cleans up some HTTP resource-related warnings that the scheduler was emitting in stderr during startup.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
@@ -115,52 +115,64 @@ public class FrameworkScheduler implements Scheduler {
 
     @Override
     public void registered(SchedulerDriver driver, Protos.FrameworkID frameworkId, Protos.MasterInfo masterInfo) {
-        if (registerCalled.getAndSet(true)) {
-            // This may occur as the result of a master election.
-            LOGGER.info("Already registered, calling reregistered()");
-            reregistered(driver, masterInfo);
-            return;
-        }
-
-        LOGGER.info("Registered framework with frameworkId: {}", frameworkId.getValue());
         try {
-            frameworkStore.storeFrameworkId(frameworkId);
-        } catch (Exception e) {
-            LOGGER.error(String.format(
-                    "Unable to store registered framework ID '%s'", frameworkId.getValue()), e);
-            ProcessExit.exit(ProcessExit.REGISTRATION_FAILURE, e);
+            if (registerCalled.getAndSet(true)) {
+                // This may occur as the result of a master election.
+                LOGGER.info("Already registered, calling reregistered()");
+                reregistered(driver, masterInfo);
+                return;
+            }
+
+            LOGGER.info("Registered framework with frameworkId: {}", frameworkId.getValue());
+            try {
+                frameworkStore.storeFrameworkId(frameworkId);
+            } catch (Exception e) {
+                LOGGER.error(String.format(
+                        "Unable to store registered framework ID '%s'", frameworkId.getValue()), e);
+                ProcessExit.exit(ProcessExit.REGISTRATION_FAILURE, e);
+            }
+
+            updateDriverAndDomain(driver, masterInfo);
+            mesosEventClient.registered(false);
+
+            // Start background threads:
+            offerProcessor.start();
+            implicitReconciler.start();
+        } catch (Throwable e) {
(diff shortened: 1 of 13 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/http/endpoints/PlansResource.java
```diff
@@ -19,6 +21,7 @@
 /**
  * API for management of Plan(s).
  */
+@Singleton
```
kvish: whats the purpose of this?

nickbp: Fixes one of the stderr warnings, according to the warning it was defaulting to singleton anyway.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
```diff
@@ -211,68 +223,108 @@ public void resourceOffers(SchedulerDriver driver, List<Protos.Offer> offers) {
...
+    /**
+     * Logs an exception that we threw in response to a call from the Mesos adapter, and then kills the scheduler
+     * process so that it can be restarted. We carefully avoid the possibility of any exceptions reaching the Mesos
+     * adapter, because doing so can leave the Scheduler process in a zombie state: no offers/statuses being processed,
+     * but HTTP (on its own thread) still looking fine.
+     *
+     * @param e the exception that was thrown
+     */
+    private static void logExceptionAndExit(Throwable e) {
+        LOGGER.error("Got exception when invoked by Mesos, shutting down");
```
porridge: Please add a trailing dot.

nickbp:
As of the CI run with this fix, the same issue appeared in two Cassandra tests where an s3-backup was being performed. Offer arrives, then silence:

```
INFO  2018-06-27 01:11:21,723 [pool-6-thread-1] ReviveManager:reviveIfRequested(134): Reviving offers
INFO  2018-06-27 01:11:21,726 [Thread-103] com.mesosphere.mesos.HTTPAdapter.MesosToSchedulerDriverAdapter:received(181): Received event of type: OFFERS
INFO  2018-06-27 01:11:21,728 [Thread-103] OfferProcessor:enqueue(145): Enqueuing 1 offer. Updated offers in progress: [21b3d4fc-6f4f-4ef0-8ac7-19e0fc8c4731-O778]
```

Looks like it's indeed a deadlock rather than an outright throw.

nickbp:
OK, I think the issue is that ReviveManager's able to enter a deadlock, and it needs better lock handling. At the moment all its methods are synchronized. This can lead to problems like the following:
- We send a revive call via ReviveManager.reviveIfRequested(), which blocks (not sure why, but it's the best theory I can think of)
- We get offers, which immediately calls ReviveManager.notifyOffersReceived(). This gets deadlocked because reviveIfRequested refuses to return.

I'd have expected the revive call to not block like this, but it definitely looks like what's happening. I'd added a log following the call to ReviveManager.notifyOffersReceived(), which wasn't printed.

benclarkwood:
https://github.com/mesosphere/dcos-commons/pull/2549 should make this diagnosable once I add an automatic fetch on test failure for the scheduler threads

nickbp:
My current theory is that when we tell the Mesos client to revive, it immediately passes us an offer within that call. As a result we effectively end up calling back into ourselves and hit a double lock against the synchronized calls in ReviveManager.

The solution that I'm trying is fairly straightforward:
- Move the notifyOffersReceived() call into the main offer processing thread, after offers are taken off the queue rather than before. This breaks the cycle if Mesos does a blocking offer send (or whatever it's doing).
- Now that the ReviveManager is solely accessed from the main offer processing thread, its functions no longer need to be marked as synchronized. (Not directly required to fix the issue but still nice to have)

nickbp:
I've got another PR in the pipeline for a different bug fix, I'll put that change into that PR. There's also a temporary logging statement to remove. But for now I'm gonna bask in this green glow.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkScheduler.java
nickbp: FYI, addressed in #2550

PR #2548, opened 2018-06-27
[DCOS-38592] Scheduler DNS is not resolvable when the first tasks launch

benclarkwood:
This adds a check that the scheduler DNS entry is correct before completing the API server bring-up.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ApiServer.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ApiServer.java
@@ -2,7 +2,6 @@ package com.mesosphere.sdk.framework;
 
 import com.codahale.metrics.jetty9.InstrumentedHandler;
 import com.google.common.annotations.VisibleForTesting;
-import com.mesosphere.sdk.http.EndpointUtils;
 import com.mesosphere.sdk.offer.LoggingUtils;
 import com.mesosphere.sdk.scheduler.Metrics;
 import com.mesosphere.sdk.scheduler.SchedulerConfig;
@@ -19,14 +18,14 @@ import org.glassfish.jersey.servlet.ServletContainer;
 import org.slf4j.Logger;
 
 import javax.ws.rs.core.UriBuilder;
+import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
 import java.time.Duration;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Timer;
-import java.util.TimerTask;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * The {@link ApiServer} runs the Jetty {@link Server} that exposes the Scheduler's API.
@@ -42,7 +41,7 @@ public class ApiServer {
     private Thread serverThread;
 
     public static ApiServer start(
-            String serviceName,
+            String schedulerHostname,
             SchedulerConfig schedulerConfig,
             Collection<Object> resources,
             Runnable startedCallback) {
@@ -50,18 +49,19 @@ public class ApiServer {
         apiServer.start(new AbstractLifeCycle.AbstractLifeCycleListener() {
             @Override
             public void lifeCycleStarted(LifeCycle event) {
-                awaitSchedulerDns(serviceName, schedulerConfig);
+                awaitSchedulerDns(schedulerHostname, schedulerConfig);
                 startedCallback.run();
             }
         });
         return apiServer;
     }
 
-    private static void awaitSchedulerDns(String schedulerName, SchedulerConfig schedulerConfig) {
-        final String schedulerHostname = EndpointUtils.toSchedulerAutoIpHostname(schedulerName, schedulerConfig);
+    private static void awaitSchedulerDns(String schedulerHostname, SchedulerConfig schedulerConfig) {
         final String expectedAddress = schedulerConfig.getSchedulerIP();
+        final Duration waitTime = Duration.ofSeconds(5);
 
-        LOGGER.info("Waiting for the scheduler host {} to resolve to {}", schedulerHostname, expectedAddress);
+        LOGGER.info("Waiting for the scheduler host {} to resolve to {}. Wait will timeout after {}ms"
+                , schedulerHostname, expectedAddress, schedulerConfig.getApiServerInitTimeout().toMillis());
 
         final Timer resolutionTimer = new Timer("API-resolution-timeout");
         resolutionTimer.schedule(new TimerTask() {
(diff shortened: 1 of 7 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ApiServer.java
```diff
@@ -41,17 +42,71 @@
...
+                        schedulerHostname, schedulerConfig.getApiServerInitTimeout().toMillis());
+                ProcessExit.exit(ProcessExit.API_SERVER_ERROR);
+            }
+        }, schedulerConfig.getApiServerInitTimeout().toMillis());
+
+        while (true) {
+            try {
+                InetAddress[] addresses = InetAddress.getAllByName(schedulerHostname);
+
+                if (addresses.length != 1) {
```
nickbp: Does it make sense to check the multiple addresses for any matching entry?

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ApiServer.java
```diff
@@ -41,17 +42,71 @@
...
+                                schedulerHostname, resolvedAddress, expectedAddress);
+                        continue;
+                    }
+
+                    LOGGER.info("Resolved scheduler host {} to expected address {}",
+                            schedulerHostname, expectedAddress);
+                    resolutionTimer.cancel();
+                    return;
+                }
+            } catch (UnknownHostException uhe){
```
nickbp: Missing a space

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ApiServer.java
```diff
@@ -41,17 +42,71 @@
...
             }
         });
         return apiServer;
     }
 
+    private static void awaitSchedulerDns(String schedulerName, SchedulerConfig schedulerConfig) {
+        final String schedulerHostname = EndpointUtils.toSchedulerAutoIpHostname(schedulerName, schedulerConfig);
+        final String expectedAddress = schedulerConfig.getSchedulerIP();
+
+        LOGGER.info("Waiting for the scheduler host {} to resolve to {}", schedulerHostname, expectedAddress);
```
nickbp: Could include wait duration in this log

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ApiServer.java
```diff
@@ -41,17 +42,71 @@
...
             @Override
             public void lifeCycleStarted(LifeCycle event) {
+                awaitSchedulerDns(serviceName, schedulerConfig);
                 startedCallback.run();
             }
         });
         return apiServer;
     }
 
+    private static void awaitSchedulerDns(String schedulerName, SchedulerConfig schedulerConfig) {
```
nickbp: `frameworkName` imo, just to keep consistent with elsewhere

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ApiServer.java
```diff
@@ -41,17 +42,71 @@
     private Thread serverThread;
 
     public static ApiServer start(
-            SchedulerConfig schedulerConfig, Collection<Object> resources, Runnable startedCallback) {
+            String serviceName,
```
nickbp: `frameworkName`. serviceName is something else

porridge:
AFAICT this makes the scheduler resolve its own address right? Just wondering whether that guarantees in any way that other processes elsewhere in the cluster will be able to do the same yet... Or do we hope they will, by the time they start up?

bmcstdio:
@benclarkwood thanks! Looks like it is producing the desired effect, and we didn't get the failed tasks:

```
INFO  2018-06-28 09:15:08,629 [Thread-3] ApiServer:awaitSchedulerDns(64): Waiting for the scheduler host kubernetes.marathon.autoip.dcos.thisdcos.directory to resolve to 10.138.0.6
WARN  2018-06-28 09:15:08,725 [Thread-3] ApiServer:awaitSchedulerDns(98): Unable to look up scheduler host kubernetes.marathon.autoip.dcos.thisdcos.directory
WARN  2018-06-28 09:15:13,725 [Thread-3] ApiServer:awaitSchedulerDns(98): Unable to look up scheduler host kubernetes.marathon.autoip.dcos.thisdcos.directory
WARN  2018-06-28 09:15:18,817 [Thread-3] ApiServer:awaitSchedulerDns(98): Unable to look up scheduler host kubernetes.marathon.autoip.dcos.thisdcos.directory
WARN  2018-06-28 09:15:23,817 [Thread-3] ApiServer:awaitSchedulerDns(98): Unable to look up scheduler host kubernetes.marathon.autoip.dcos.thisdcos.directory
WARN  2018-06-28 09:15:28,907 [Thread-3] ApiServer:awaitSchedulerDns(98): Unable to look up scheduler host kubernetes.marathon.autoip.dcos.thisdcos.directory
WARN  2018-06-28 09:15:33,907 [Thread-3] ApiServer:awaitSchedulerDns(98): Unable to look up scheduler host kubernetes.marathon.autoip.dcos.thisdcos.directory
INFO  2018-06-28 09:15:38,909 [Thread-3] ApiServer:awaitSchedulerDns(92): Resolved scheduler host kubernetes.marathon.autoip.dcos.thisdcos.directory to expected address 10.138.0.6
```

In any case, I will run a couple more times to make extra sure!

bmcstdio:
Oops sorry for closing! I clicked the wrong button by mistake! 🤦‍♀️

benclarkwood:
@bmcstdio glad it seems to be a fix. We had to do something similar in bootstrap awhile ago.

@porridge Yes, this relies on the task being able to self-resolve as a test for whether the task DNS is "live". I cannot remember how far up the chain it goes to get the answer to that question. I believe that it ultimately is querying up to DC/OS DNS on the masters (if the answer was already cached in the agent it wouldn't go upward) but I could be wrong. That's a big part of why it also checks that it's the _right_ IP address to make sure it isn't just a record for an old copy of the service.

@jeid64 can you comment on the reliability of this method / speed of propagation? My rough knowledge of it is: navstar/spartan/et al slow to create DNS entry, very fast to propagate across cluster.

(FWIW, we've been successfully using this technique via the bootstrap utility in task bring up)

jeid64:
Autoip/containerip addresses are sourced from masters state.json. Every 30 seconds the masters state.json is polled and then the address is created, and then is pushed to all agents DNS resolvers very quickly. There is a chance that if dcos-dns is broken on an agent, that the scheduler on one agent will be able to resolve itself but tasks on another agent won't be able to, but that's pretty much out of our control.

VIP addresses are polled every 2 seconds by agents, so they're created a lot faster.

nickbp:
The main reason for waiting on the ApiServer to start is that it hosts the config artifact endpoints that are queried by the mesos fetcher as tasks are brought up.

So there are two dependencies before we should bring up tasks:
- DNS for the scheduler needs to be resolvable at the tasks (previously VIP, now autoip)
- scheduler itself needs to be running its API server

In practice the API server takes ~1s to actually load, but that's been enough of a delay to cause problems before

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -420,4 +425,10 @@ public long getImplicitReconcilePeriodMs() {
     public boolean isRegionAwarenessEnabled() {
         return envStore.getOptionalBoolean(ALLOW_REGION_AWARENESS_ENV, false);
     }
+
+    /**
+     * Returns the IP of the scheduler task's container. Note, this is dependent on the fact we are using the command
+     * executor.
+     */
+    public String getSchedulerIP() { return envStore.getRequired(LIBPROCESS_IP_ENVVAR); }
```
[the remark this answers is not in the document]
benclarkwood: This will need to change if we ever switch the scheduler to run under CNI.

We do not currently support CNI.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ApiServer.java
benclarkwood: I believe that if there are multiple addresses present, then the DNS is not yet correct.

WDYT, @jeid64?

PR #2532, opened 2018-06-15
[TOOLS][TESTS] Add tooling for running flake8 as a pre-commit hook and in CI

elezar:
This PR was a bit of train work from my side, and is a start at getting some basic syntax / error checking into our CI for our Python files. I started with `flake8`, but would like to add `pylint -E`, `black`, and `pyre` at some point.

There have been some errors in recent PRs that increased the time to merge and at least a subset of these would have been detected by these tools.

The PR makes the following specific changes:
* Adds `flake8` as a requirement in the docker image
* Adds `pylint` as a requirement in the docker image
* Adds a collection of scripts to drive these kinds of checks
* adds a `pre-commit-config.yaml` file for enabling checks locally
* Make some change to Python files in `testing/` while checking the commit hooks and scripts.

Note:
* Only modified files are checked. Meaning that we can gradually improve the codebase without having to check in large diffs.

I am also testing a basic TeamCity jobs:

* Intentional failure: https://teamcity.mesosphere.io/viewLog.html?buildId=1102872&buildTypeId=DcosIo_DcosCommons_PrJobs_TestCodeChecksPr&tab=buildLog
* Intentional success: https://teamcity.mesosphere.io/viewLog.html?buildId=1102873&buildTypeId=DcosIo_DcosCommons_PrJobs_TestCodeChecksPr&tab=buildLog
* A template configuration here: https://teamcity.mesosphere.io/viewType.html?buildTypeId=DcosIo_DcosCommons_PrJobs_RunCodeChecks

PR #2549, opened 2018-06-27
Add thread dumps to diagnostics

benclarkwood:
1. Adds an API to get a thread dump on-demand
2. Adds a thread dump to any Process.exit() call.
3. Grab thread dump of scheduler when test fails

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ProcessExit.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ProcessExit.java
@@ -1,5 +1,9 @@
 package com.mesosphere.sdk.framework;
 
+import com.codahale.metrics.jvm.ThreadDump;
+
+import java.lang.management.ManagementFactory;
+
 /**
  * This class handles exiting the Scheduler process, after completed teardown or after a fatal error.
  */
@@ -29,6 +33,8 @@ public class ProcessExit {
         String message = String.format("Process exiting immediately with code: %s[%d]", code, code.getValue());
         System.err.println(message);
         System.out.println(message);
+        System.err.println("Printing final thread state");
+        new ThreadDump(ManagementFactory.getThreadMXBean()).dump(System.err);
         System.exit(code.getValue());
     }
 
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ProcessExit.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ProcessExit.java
@@ -33,7 +33,7 @@ public class ProcessExit {
         String message = String.format("Process exiting immediately with code: %s[%d]", code, code.getValue());
         System.err.println(message);
         System.out.println(message);
-        System.err.println("Printing final thread state");
+        System.err.println("Printing final thread state...");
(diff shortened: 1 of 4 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/ProcessExit.java
```diff
@@ -29,6 +33,8 @@ public static void exit(Code code) {
         String message = String.format("Process exiting immediately with code: %s[%d]", code, code.getValue());
         System.err.println(message);
         System.out.println(message);
+        System.err.println("Printing final thread state");
```
kvish: nit: trailing dots to indicate action in progress to keep in line with other log statements

testing/sdk_diag.py
```diff
@@ -162,6 +167,15 @@ def _dump_plans(item: pytest.Item, service_name: str):
             f.write('\n') # ... and a trailing newline
 
 
+def _dump_threads(item: pytest.Item, service_name: str):
+    threads = sdk_cmd.service_request('GET', 'v1/debug/threads')
+    out_path = _setup_artifact_path(item, 'threads_{}.json'.format(service_name.replace('/', '_')))
```
elezar: The threads endpoint doesn't output JSON though, correct? The filename seems to indicate this.

Also, would replacing / with __ not be more consistent?

benclarkwood: Good catch on the first point. I'm claiming "copying prior art" on the second one :).

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/endpoints/DebugResource.java
```diff
@@ -24,11 +32,24 @@ public OfferOutcomeResource(OfferOutcomeTracker offerOutcomeTracker) {
...
             return ResponseUtils.htmlOkResponse(offerOutcomeTracker.toHtml());
         }
     }
+
+    /**
+     * Renders the current threads and thread state of the scheduler.
+     */
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    @Path("threads")
```
nickbp: Could add this to the API doc: https://github.com/mesosphere/dcos-commons/blob/master/docs/reference/swagger-api/swagger-spec.yaml#L644

testing/sdk_diag.py
```diff
@@ -162,6 +167,15 @@ def _dump_plans(item: pytest.Item, service_name: str):
             f.write('\n') # ... and a trailing newline
 
 
+def _dump_threads(item: pytest.Item, service_name: str):
+    threads = sdk_cmd.service_request('GET', 'v1/debug/threads')
```
elezar: There is a missing parameter here.

The `pylint -E testing` output:
```
************* Module sdk_diag
E:172,14: No value for argument 'service_path' in function call (no-value-for-parameter)
```
This should be:
```python
threads = sdk_cmd.service_request('GET', service_name, 'v1/debug/threads')
```

I have corrected this in #2532.

porridge:
FTR, this does not work: shakedown insists on json by default, and server responds with a `406`.

PR #2550, opened 2018-06-29
DCOS-38586 Don't wait for restart to store "finished-deploy" bit in ZK

nickbp:
Before this change, we only checked whether the service finished deployment following a scheduler restart. In that situation, we can effectively end up comparing the OLD persisted task states against the NEW deploy plan (not the old plan, because plans are not persisted). If, for example, new steps had been added to the deploy plan in the meantime, this can lead to us thinking that deployment didn't complete before, when in reality it did.

To avoid this problem, we can simply store the fact that we've finished deployment at the moment that it happens, rather than waiting for the next scheduler restart and hoping that the plans vs state happen to line up.

In the meantime, the previous 'after restart' check is left as-is, as users upgrading from prior SDK releases may only be caught by that check.


nickbp:
There isn't an update plan integration test yet, tests are just checking that the deployed bit had been set when expected. I'll write something for update plans as a service test in a follow-up PR

PR #2545, opened 2018-06-26
[DCOS-38379] Wait for scheduler task to be relaunched to fix flaky multi-service scheduler tests

kvish:

PR #2551, opened 2018-06-29
[DCOS-38379] Correctly check for a relaunched scheduler task

elezar:
Looking at #2545, I think the check that we are performing for a relaunched scheduler is incorrect. 

When we initially get the task ID of the scheduler (a `marathon` task) we use `.startswith(SERVICE_NAME)` as a match condition, but in the current `check_task_relaunched` function, we check `== SERVICE_NAME` which does not seem to yield the desired result.

This PR adds a *local* check function to the test file that is affected by this error. If this passes, we can move the function to `sdk_tasks` in a separate PR.

Note that this is also blocking the DC/OS smoke tests: https://teamcity.mesosphere.io/viewLog.html?buildId=1112832&buildTypeId=DcOs_Enterprise_Test_FrameworkTests_DcOsCommonsSmokeTest&tab=buildResultsDiv

### DIFF
--- a/frameworks/helloworld/tests/test_multiservice_dynamic.py
+++ b/frameworks/helloworld/tests/test_multiservice_dynamic.py
@@ -84,6 +84,7 @@ def test_add_deploy_restart_remove():
 
     # restart and check that service is recovered:
     sdk_marathon.restart_app(config.SERVICE_NAME)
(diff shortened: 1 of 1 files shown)


elezar:
The strict mode failures are a known issue.

nickbp:
The sole test edited here has already passed, so going to just merge.

PR #2555, opened 2018-07-03
DCOS-38149 Cleanup salvaged from PodInstanceRequirement breakup

nickbp:
While working towards footprint-up-front, an initial pass was made to break up `PodInstanceRequirement`. A better strategy has since been worked out, but it'd make sense to retain some of the minor cleanup from that initial effort:
- `Step.start()` now returns void, instead of returning the content of getPodInstanceRequirement(). This greatly clarifies the expected behavior of the `start()` function.
- Remove duplicate code between `TaskUtils.isSamePodInstance(TaskInfo, PodInstance)` and `PlacementUtils.areEquivalent(TaskInfo, PodInstance)`
- Various other minor changes like cleaning up log statements and adding javadoc where missing.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
@@ -12,6 +12,7 @@ import com.mesosphere.sdk.state.StateStore;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.mesos.Protos;
+import org.apache.mesos.Protos.TaskInfo;
 import org.slf4j.Logger;
 
 import java.util.*;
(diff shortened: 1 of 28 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -93,32 +94,23 @@ public static boolean hasTasksWithTLS(ServiceSpec serviceSpec) {
...
-                })
+                .filter(taskInfo -> areEquivalent(taskInfo, podInstance))
                 .collect(Collectors.toList());
     }
 
     /**
-     * Returns whether the provided {@link TaskInfo} (representing a launched task) and {@link PodInstance} (from the
-     * {@link ServiceSpec}) are both effectively for the same pod instance.
+     * Returns whether the provided {@link TaskInfo}, representing a previously-launched task,
+     * is in the same provided pod provided in the {@link PodInstance}.
```
benclarkwood: provided twice

nickbp:
Will solve typo in follow-up PR -- save some CI time

PR #2556, opened 2018-07-03
[COPS-3222][DCOS-38688] Add pod replace test to HDFS TLS

elezar:
This PR addresses an issue where we request multiple mounts of the same secret on TLS enabled data services. This currently only affects HDFS, but could be an issue for any service running multiple tasks with TLS per pod.

The changes included are:
* Modify `TLSEvaluationStage` to not add duplicate volumes to the `PodInfoBuilder`
* Modify `OfferEvaluator` to not add duplicate `TLSEvaluationStages` to the offer evaluation pipeline
* Add `pod replace` integration tests for HDFS, Kafka, Elastic, and Cassandra

(The tests were added first to confirm that this was only affecting HDFS -- after which the fix was implemented).

@nickbp while debugging this issue, I noted that an additional `ResourceEvaluation` stage (with `cpu=1.0` seems to be added between the two `LaunchEvaluationStage`s. Is this the desired behaviour? Does this have something to do with the resources allocated for the default executor?

*Note*: I had originally tried to also run this through a `ServiceTestRunner`, but found that the TLS-handling (especially the Secrets Client) is a bit tightly coupled and I did not want to refactor this further.

### Bugfix
* Fix an error with duplicate mount operations on TLS-enabled HDFS


frameworks/hdfs/tests/test_tls.py
```diff
@@ -1,16 +1,23 @@
...
+import sdk_tasks
 import sdk_utils
-import retrying
+
 
 from security import transport_encryption
 
 from tests import config
 
+pytestmark = pytest.mark.skipif(sdk_utils.dcos_version_less_than("1.10"),
```
porridge: Why assign to a value if it's unused?

frameworks/hdfs/tests/test_tls.py
```diff
@@ -106,6 +110,31 @@ def fn():
...
+
+        recovery_timeout_s = 25 * 60
+        sdk_plan.wait_for_kicked_off_recovery(config.SERVICE_NAME, recovery_timeout_s)
+        sdk_plan.wait_for_completed_recovery(config.SERVICE_NAME, recovery_timeout_s)
+
+        sdk_tasks.check_tasks_updated(config.SERVICE_NAME, pod_name, inital_task_id)
+
+        # TODO: Add checks for non-updated tasks
+    finally:
+        hdfs_service_tls(service_account)
```
porridge: I don't get this. Why do it in the finally block?

frameworks/hdfs/tests/test_tls.py
elezar: This is the way pytest handles global marks. They have to be called `pytestmark`. There is an error here though, they should be a list and not individually assigned.

elezar: (I also messed up the enterprise conditional 😞)

porridge:
Thanks, for the explanations. I'm used to frameworks where the environment is set up in a separate (setUp) function, and the test functions themselves are free to leave a mess behind.

frameworks/hdfs/tests/test_tls.py
elezar: For the record, I have removed this.

elezar:
> Thanks, for the explanations. I'm used to frameworks where the environment is set up in a separate (setUp) function, and the test functions themselves are free to leave a mess behind.

The fixtures in Pytest do allow this, but these are typically run at a module level, and not per test -- mostly due to the time required to uninstall and reinstall the services.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TLSEvaluationStage.java
```diff
@@ -110,24 +117,46 @@ public EvaluationOutcome evaluate(MesosResourcePool mesosResourcePool, PodInfoBu
...
-        Collection<Protos.Volume> volumes = new ArrayList<>();
-        for (TLSArtifactPaths.Entry entry : tlsArtifactPaths.getPathsForType(spec.getType(), spec.getName())) {
-            volumes.add(getSecretVolume(entry));
-        }
-        return volumes;
+
+        Collection<TLSArtifactPaths.Entry> paths = tlsArtifactPaths.getPathsForType(spec.getType(), spec.getName());
+        return paths.stream()
+                .map(TLSEvaluationStage::getSecretVolume)
+                .collect(Collectors.toSet());
```
nickbp: Does this set need a custom Comparator implementation? Example of building a set with a custom comparator from a stream: https://github.com/mesosphere/dcos-commons/pull/2571/files#diff-d71effdf03d955bff6c0fe818ae25c0cR227

elezar: It seems as if the `Protos.Volume` objects being compared here behave well enough. Could be that the protobuf compiler already generates applicable comparators for equality.

If your asking whether we should *only* be comparing the volume paths, that's a fair point, but if we're trying to map different secrets to the same path, then that's also an issue.

nickbp: Oh, that makes sense. For some reason I forgot that the protobuf types included equality checks.

nickbp:
Re description question: It's fine if there are resource reservations between launches, as long as any resource reservations occur before the tasks that use them.

PR #2512, opened 2018-05-26
Update changes by config helper and bootstrap in the config file

kohend:
Ticket: https://jira.mesosphere.com/browse/DCOS_OSS-3555

PR #2558, opened 2018-07-03
[DCOS_OSS-3555] Use a valid boolean value for Kafka envvar

elezar:
Following on from #2512 this PR sets the `KAFKA_ADVERTISE_HOST` for use with the setup helper utility.

Using the value `"set"` was breaking legacy behaviour when `kafka.kafka_advertise_host_ip` was set to `true` in the configuration options.

@kohend your comments are welcome here.

### Bug Fixes
* Fix Kafka behaviour when specifying the `kafka.kafka_advertise_host_ip` configuration option.


kohend:
Looks good to me.

PR #2560, opened 2018-07-03
[DCOS-38519] fix flakiness in test_zzzrecovery.test_shutdown_host_teardown

kvish:
Increasing the count of world pods to minimize risk of tasks on shut down host. Adding install/uninstall helper with options as well for the tests


nickbp:
Specifically, we saw a case where the 2 world tasks were placed on the same agents as the scheduler and package registry. By increasing the number of world tasks to 3, that should ensure that at least one of them is not colocated with the scheduler or package registry (because the world tasks are each placed on unique agents)

PR #2566, opened 2018-07-03
[DCOS-38840] Drop mesos v0 support across frameworks

kvish:
drop option to use mesos v0 api. Seems to be causing CI issues and is no longer necessary for support

### DIFF
--- a/frameworks/cassandra/tests/test_sanity.py
+++ b/frameworks/cassandra/tests/test_sanity.py
@@ -41,17 +41,6 @@ def test_service_health():
     assert shakedown.service_healthy(config.get_foldered_service_name())
 
 
-# @pytest.mark.sanity
-# @pytest.mark.smoke
-# @pytest.mark.mesos_v0
-# def test_mesos_v0_api():
-#     service_name = config.get_foldered_service_name()
-#     prior_api_version = sdk_marathon.get_mesos_api_version(service_name)
-#     if prior_api_version is not "V0":
-#         sdk_marathon.set_mesos_api_version(service_name, "V0")
-#         sdk_marathon.set_mesos_api_version(service_name, prior_api_version)
-
-
 @pytest.mark.sanity
 def test_endpoints():
     # check that we can reach the scheduler via admin router, and that returned endpoints are sanitized:
--- a/frameworks/cassandra/tests/test_sanity.py
+++ b/frameworks/cassandra/tests/test_sanity.py
@@ -41,15 +41,15 @@ def test_service_health():
     assert shakedown.service_healthy(config.get_foldered_service_name())
 
 
-@pytest.mark.sanity
-@pytest.mark.smoke
-@pytest.mark.mesos_v0
-def test_mesos_v0_api():
-    service_name = config.get_foldered_service_name()
-    prior_api_version = sdk_marathon.get_mesos_api_version(service_name)
-    if prior_api_version is not "V0":
-        sdk_marathon.set_mesos_api_version(service_name, "V0")
-        sdk_marathon.set_mesos_api_version(service_name, prior_api_version)
+# @pytest.mark.sanity
+# @pytest.mark.smoke
+# @pytest.mark.mesos_v0
+# def test_mesos_v0_api():
+#     service_name = config.get_foldered_service_name()
+#     prior_api_version = sdk_marathon.get_mesos_api_version(service_name)
+#     if prior_api_version is not "V0":
+#         sdk_marathon.set_mesos_api_version(service_name, "V0")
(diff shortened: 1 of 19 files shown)


nickbp:
Could also remove these from `sdk_marathon.py`:

```
def get_mesos_api_version(service_name):
    ...

def set_mesos_api_version(service_name, api_version, timeout=600):
    ...
```

nickbp:
Actually, it looks like we use this value ourselves in `SchedulerConfig.java` (see mention of `MESOS_API_VERSION` in there). I thought the value was read directly by the Mesos adapter, rather than accessed by us. Given that this value is under our control, I think we could do the following:
- Also remove the `MESOS_API_VERSION` envvar from our marathon.jsons
- Change SchedulerConfig's `getRequired()` call to a getOptional() which defaults to "V1"
(In other words: move the V1 default directly into the Java code)

The reason I think this makes sense is because we do similar things for other manually configurable `knobs` in the scheduler (as listed in SchedulerConfig): Default to some reasonable value in SchedulerConfig, but allow the operator to override by manually adding a scheduler envvar

kvish:
Updated from comments

frameworks/cassandra/tests/test_sanity.py
```diff
@@ -41,15 +41,15 @@ def test_service_health():
...
-@pytest.mark.sanity
-@pytest.mark.smoke
-@pytest.mark.mesos_v0
-def test_mesos_v0_api():
-    service_name = config.get_foldered_service_name()
-    prior_api_version = sdk_marathon.get_mesos_api_version(service_name)
-    if prior_api_version is not "V0":
-        sdk_marathon.set_mesos_api_version(service_name, "V0")
-        sdk_marathon.set_mesos_api_version(service_name, prior_api_version)
+# @pytest.mark.sanity
```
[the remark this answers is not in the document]
elezar: Yes, complete removal is much better than commenting them out.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -387,7 +387,7 @@ public int getStatsdPort() {
      * Returns the Mesos API version.
      */
     public String getMesosApiVersion() {
-        return envStore.getRequired(MESOS_API_VERSION_ENV);
+        return envStore.getOptional(MESOS_API_VERSION_ENV, "V1");
```
elezar: This means that we *could* still be operating with `V0` even though we are removing the tests for this. Is there a way to *ensure* that we always use V1 -- even for existing installations after an update? cc @nickbp.

That is to say, if a customer is running a service using `V0`, and performs an update to a scheduler using `V1`, is there a reasonable update path?

testing/sdk_marathon.py
```diff
@@ -225,16 +225,3 @@ def bump_task_count_config(service_name, key_name, delta=1):
     updated_node_count = int(config['env'][key_name]) + delta
     config['env'][key_name] = str(updated_node_count)
     update_app(service_name, config)
-
-
-def get_mesos_api_version(service_name):
```
elezar: Note that these `testing/` files are used in a number of other repos -- including ones not controlled by us. Is this tooling change "safe" for all of those too? We may want to leave these in here for the time being and mark them as deprecated.

frameworks/cassandra/tests/test_sanity.py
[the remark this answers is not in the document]
kvish: + 1 will remove

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
kvish: yes this is intentional, this is meant as an escape hatch for anyone who needs to use V0, they can manually add the envar. AFAIK update from v0 to v1 should be seamless.

testing/sdk_marathon.py
kvish: ya that makes sense, we can keep the tooling functions, will update

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
nickbp: With the removal of the api version option in config.json, I think in practice the customer would see that the option was no longer valid when upgrading, and would then remove it. As a result they would effectively be switched over to v1 when the new scheduler starts.

To be clear I don't expect anybody to be using the v0 option, or at least not for any good reason. It was only originally added as a fallback if we had problems with migrating to v1

PR #2571, opened 2018-07-06
DCOS-29329 Uninstall: Group resources by hosting agent

nickbp:
Each resource to be unreserve gets its own step. At the moment, we have a single flat phase for all resources in the cluster to be unreserved. With this PR, we will now group the resources by agent hostname. This provides two benefits to the operator:
- Makes it obvious when resources tied to a given agent are all stuck, since they should all be grouped within a given phase.
- Makes it easier to bypass all resources on a given stuck agent by specifying the (per-agent) phase name when invoking the CLI command.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallPlanFactory.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallPlanFactory.java
@@ -3,8 +3,15 @@ package com.mesosphere.sdk.scheduler.uninstall;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 import org.apache.http.impl.client.LaxRedirectStrategy;
@@ -17,13 +24,17 @@ import com.mesosphere.sdk.dcos.clients.SecretsClient;
 import com.mesosphere.sdk.offer.Constants;
 import com.mesosphere.sdk.offer.LoggingUtils;
 import com.mesosphere.sdk.offer.ResourceUtils;
+import com.mesosphere.sdk.offer.TaskException;
 import com.mesosphere.sdk.offer.TaskUtils;
+import com.mesosphere.sdk.offer.taskdata.TaskLabelReader;
 import com.mesosphere.sdk.scheduler.SchedulerConfig;
 import com.mesosphere.sdk.scheduler.plan.DefaultPhase;
 import com.mesosphere.sdk.scheduler.plan.DefaultPlan;
 import com.mesosphere.sdk.scheduler.plan.Phase;
 import com.mesosphere.sdk.scheduler.plan.Plan;
 import com.mesosphere.sdk.scheduler.plan.Step;
+import com.mesosphere.sdk.scheduler.plan.strategy.DependencyStrategy;
+import com.mesosphere.sdk.scheduler.plan.strategy.DependencyStrategyHelper;
 import com.mesosphere.sdk.scheduler.plan.strategy.ParallelStrategy;
 import com.mesosphere.sdk.scheduler.plan.strategy.SerialStrategy;
 import com.mesosphere.sdk.scheduler.recovery.FailureUtils;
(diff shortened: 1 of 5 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallPlanFactory.java
```diff
@@ -155,4 +174,67 @@ Plan getPlan() {
...
+        }
+
+        LOGGER.info("Configuring resource cleanup of {}/{} tasks across {} agents",
+                tasksNotFailedAndErrored.size(), allTasks.size(), dedupedResourceIdsByAgent.size());
+
+        // Map the resource IDs into ResourceCleanupSteps
+        return dedupedResourceIdsByAgent.entrySet().stream().collect(Collectors.toMap(
+                entry -> entry.getKey(),
+                entry -> entry.getValue().stream()
+                        .map(resourceId -> new ResourceCleanupStep(resourceId, namespace))
```
elezar: I'm not a big fan fo inline constructors like this as they tend to make things pretty difficult to read. Won't block the PR on this though.

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallSchedulerTest.java
```diff
@@ -94,41 +115,46 @@ public void testEmptyOffers() throws Exception {
...
+        stateStore.storeTasks(Arrays.asList(TASK_B, TASK_C));
 
         UninstallScheduler uninstallScheduler = getUninstallScheduler();
-        Plan plan = uninstallScheduler.getPlanCoordinator().getPlanManagers().stream().findFirst().get().getPlan();
-        // 2 task kills + 4 unique resources + deregister step.
+        Plan plan = getUninstallPlan(uninstallScheduler);
         List<Status> expected = Arrays.asList(
-                Status.PENDING, Status.PENDING, Status.PENDING, Status.PENDING, Status.PENDING,
-                Status.PENDING, Status.PENDING);
+                Status.PENDING, Status.PENDING, Status.PENDING, // 3 task kills
```
elezar: If all we're doing is checking for `Status.PENDING` what about checking `.size()` and all equal?

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallSchedulerTest.java
```diff
@@ -393,6 +406,10 @@ private UninstallScheduler getUninstallScheduler(ServiceSpec serviceSpec, Uninst
         return uninstallScheduler;
     }
 
+    private static Plan getUninstallPlan(AbstractScheduler scheduler) {
+        return scheduler.getPlanCoordinator().getPlanManagers().stream().findFirst().get().getPlan();
```
elezar: Nit: The line is pretty long.

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallSchedulerTest.java
elezar: I see we test other conditions, so it would be nice if it were easier to map what we are expecting to our expectations.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/uninstall/UninstallPlanFactory.java
```diff
@@ -126,13 +128,30 @@
...
+                Collections.emptyList());
+        phases.add(deregisterPhase);
+
+        // We need to construct a custom strategy in order to allow the resource phases to operate in parallel.
+        // In other words, the dependencies should look like this:
+        // 1. Kill all tasks, unreserve all resources, and delete any automatically created TLS certs as needed
+        // 2. When all of the above is complete, deregister
+        DependencyStrategyHelper<Phase> uninstallPhaseDependencies = new DependencyStrategyHelper<>(phases);
+        for (Phase phase : phases) {
+            if (phase == deregisterPhase) {
```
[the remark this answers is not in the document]
nickbp: Good point, done

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallSchedulerTest.java
nickbp: Yeah at the moment it's effectively checking that the step statuses all match a provided snapshot

PR #2573, opened 2018-07-06
DCOS-38544 Add mono to multi zk data migration

takirala:
Address [DCOS-38691](https://jira.mesosphere.com/browse/DCOS-38691)


takirala:
At this point, I made sure the integration tests pass the setup phase and execution phase. The teardown won't be success until https://github.com/mesosphere/dcos-commons/pull/2580 is merged in to this PR.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -198,6 +198,94 @@ public static Builder newBuilder(String serviceName, String zookeeperHostPort) {
         }
     }
 
+    @Override
```
takirala: The only new method I added is the `recursiveCopy` method in this file. The other methods were only reordered to be consistent with the interface and visibility.

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -233,4 +238,47 @@ private static ServiceSpec createSampleServiceSpec(SchedulerConfig schedulerConf
...
+            int curVer = schemaVersionStore.getOrSetVersion(SchemaVersionStore.getSupportedSchemaVersionMultiService());
+            if (curVer == SchemaVersionStore.getSupportedSchemaVersionSingleService()) {
+                try {
+                    LOGGER.warn("Found old schema in ZK Storage that can be migrated to a new schema");
+                    PersisterUtils.backUpFrameworkZKData(persister);
+                    PersisterUtils.migrateMonoToMultiZKData(persister, frameworkConfig);
+                    schemaVersionStore.store(SchemaVersionStore.getSupportedSchemaVersionMultiService());
+                    LOGGER.info("Successfully migrated from old schema to new schema!!");
+                } catch (PersisterException e) {
+                    LOGGER.error("Unable to migrate ZK data : ", e.getMessage(), e);
```
nickbp: When an exception is being passed, we would need to do our own string formatting via `String.format()`. The call being used here is:
```
     * @param format the format string
     * @param arg1   the first argument
     * @param arg2   the second argument
     */
    public void error(String format, Object arg1, Object arg2);
```
However, the one you would want here is:
```
     * @param msg the message accompanying the exception
     * @param t   the exception (throwable) to log
     */
    public void error(String msg, Throwable t);
```

... but in this case, this could just be of this form: `LOGGER.error("Unable to migrate ZK data", e);`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -179,6 +180,11 @@
      */
     private static final String LIBPROCESS_IP_ENVVAR = "LIBPROCESS_IP";
 
+    /**
+     * Environment variable to disable zk data migration from mono service schema to multi service schema.
+     */
+    private static final String DISABLE_MONO_TO_MULTI_MIGRATION = "DISABLE_MONO_TO_MULTI_MIGRATION";
```
nickbp: I think this can be removed. It's only referenced within hello-world. In practice I think any additional gating on migration should be developer-owned anyway.

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -233,4 +238,47 @@ private static ServiceSpec createSampleServiceSpec(SchedulerConfig schedulerConf
                         .build())
                 .build();
     }
+
+    private static void checkAndMigrate(
```
nickbp: Could move the below notes comment to be part of a javadoc comment here

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
```diff
@@ -0,0 +1,67 @@
...
+    sdk_install.install(config.PACKAGE_NAME,
+                        config.SERVICE_NAME,
+                        3,
+                        additional_options={"service": {"yaml": "svc"}})
+    hello_task_id = sdk_tasks.get_task_ids(config.SERVICE_NAME, 'hello')
+    # Start update plan with options that have list of yaml files to make it launch in multi service mode
+    sdk_upgrade.update_service(config.PACKAGE_NAME,
+                               config.SERVICE_NAME,
+                               additional_options={"service": {"yamls": "svc,foobar_service_name"}})
+    # Ensure the old tasks do not relaunch and new tasks are launched
```
nickbp: Could switch these checks: Wait for `foobar` to finish deployment, _then_ verify that `hello-0-server` was not relaunched.

Otherwise, we may be checking too quickly that `hello-0-server` was not relaunched -- maybe it just wasn't relaunched _yet_.

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
```diff
@@ -0,0 +1,67 @@
...
+import sdk_install
+import sdk_plan
+import sdk_tasks
+import sdk_upgrade
+
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope='function', autouse=True)
+def configure_package(configure_security):
```
nickbp: could rename this function to be clear that it's running for each test case, and not once for the whole file

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -198,6 +198,94 @@ public static Builder newBuilder(String serviceName, String zookeeperHostPort) {
...
+        Map<String, byte[]> pathBytesMap = new TreeMap<>();
+        for (Map.Entry<String, byte[]> entry : unprefixedPathBytesMap.entrySet()) {
+            pathBytesMap.put(withFrameworkPrefix(entry.getKey()), entry.getValue());
+        }
+        LOGGER.debug("Updating {} entries: {}", pathBytesMap.size(), pathBytesMap.keySet());
+        runTransactionWithRetries(new SetTransactionFactory(pathBytesMap));
+    }
+
+    @Override
+    public void recursiveCopy(String src, String dest) throws PersisterException {
```
nickbp: It'd probably make sense to follow the pattern elsewhere in CuratorPersister: name these params `unprefixedSrc`/`unprefixedDest`, then start the function with `String src=withFrameworkPrefix(unprefixedSrc);`

The main reason for doing this is to avoid accidentally using the unprefixed version of a path when performing the copy.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/AbstractResourceMapper.java
```diff
@@ -0,0 +1,90 @@
+package com.mesosphere.sdk.offer.evaluate;
```
nickbp: Nice cleanup!

I think these could just be presented as static functions, in e.g. a `ResourceMapperUtils` class. The `resourceNamespace` member appears to only be used in one of the functions, anyway, and could just be passed in as a parameter. The SDK code has generally avoided inheritance.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/AbstractResourceMapper.java
```diff
@@ -0,0 +1,90 @@
...
+abstract class AbstractResourceMapper {
+    final Logger logger;
+    final Optional<String> resourceNamespace;
+
+    AbstractResourceMapper(Optional<String> resourceNamespace) {
+        logger = LoggingUtils.getLogger(getClass(), resourceNamespace);
+        this.resourceNamespace = resourceNamespace;
+    }
+
+    Optional<ResourceLabels> findMatchingDiskSpec(
```
nickbp: FWIW, so far most of the code goes with this style:
```
void func(A a, B b) {
```
wrapping to this if the above doesn't fit 120col:
```
void func(
        A a, B b) {
```
then finally this if that doesn't fit either:
```
void func(
        A a,
        B b) {
```
...but that said, I'm open to just picking a style and enforcing it via lint at some point. There hasn't really been an 'official style' yet.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -246,7 +247,7 @@ public OfferEvaluator(
                 .getExecutor().toBuilder()
                 .setExecutorId(Protos.ExecutorID.newBuilder().setValue(""))
                 .build();
-        logger.info("Using old executor: {}", TextFormat.shortDebugString(executorInfo));
+        logger.info("Using new executor: {}", TextFormat.shortDebugString(executorInfo));
```
nickbp: How about `Using new executor derived from task <taskInfo.getName()>: <executorDebugString>`? The retrieval at L246 could be broken into two steps: picking a TaskInfo (whose name would be logged here), and then creating the ExecutorInfo with empty id from that.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -491,25 +508,17 @@ public TaskStatusResponse taskStatus(Protos.TaskStatus status) {
...
     private static OfferResources getEntry(
-            Map<String, Map<Protos.OfferID, OfferResources>> map, String serviceName, Protos.Offer offer) {
-        Map<Protos.OfferID, OfferResources> serviceOffers = map.get(serviceName);
-        if (serviceOffers == null) {
-            serviceOffers = new HashMap<>();
-            map.put(serviceName, serviceOffers);
-        }
-        return getEntry(serviceOffers, offer);
+            Map<String, Map<Protos.OfferID, OfferResources>> map,
+            String serviceName, Protos.Offer offer
```
nickbp: Should break these up into three separate lines -- either all on one line or all on separate lines

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -491,25 +508,17 @@ public TaskStatusResponse taskStatus(Protos.TaskStatus status) {
...
      */
     private static OfferResources getEntry(Map<Protos.OfferID, OfferResources> map, Protos.Offer offer) {
-        OfferResources currentValue = map.get(offer.getId());
-        if (currentValue == null) {
-            // Initialize entry
-            currentValue = new OfferResources(offer);
-            map.put(offer.getId(), currentValue);
-        }
-        return currentValue;
+        // Initialize entry
```
nickbp: Could update comment to clarify `Initialize entry if absent`

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterUtils.java
```diff
@@ -190,4 +201,68 @@ public static void clearAllData(Persister persister) throws PersisterException {
             }
         }
     }
+
+    public static void backUpFrameworkZKData(Persister persister) throws PersisterException {
+        try {
+            persister.recursiveDelete(BACKUP_ROOT_NAME);
```
nickbp: Could put the backup data in a timestamped subdir, like `backup/2018-07-18-095146/` (or some other format). Then there's much less risk of conflict with prior backups, and it gives operators some indication of where the backup could have come from.

Could see us wanting to have backups like this in the future, so it doesn't hurt to figure out a schema for this now.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -433,4 +436,19 @@ public boolean isRegionAwarenessEnabled() {
     public String getSchedulerIP() {
         return envStore.getRequired(LIBPROCESS_IP_ENVVAR);
     }
+
+    public JSONObject getBuildInfo() {
+        JSONObject jsonObject = new JSONObject();
+        jsonObject.put(PACKAGE_NAME_ENV, getPackageName());
+        jsonObject.put(PACKAGE_VERSION_ENV, getPackageVersion());
+        jsonObject.put(PACKAGE_BUILD_TIME_EPOCH_MS_ENV, Instant.ofEpochMilli(getPackageBuildTimeMs()));
+        for (Field f : SDKBuildInfo.class.getDeclaredFields()) {
```
nickbp: Rather than using reflection, I think it's fine to just statically define some names here. I could see someone being suprised by these names changing if SDKBuildInfo members are renamed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/AbstractResourceMapper.java
takirala: Sorry, that was the scala-ey me. However I propose this : 

we always do this : 
```
void func(A a, B b) {
```
and if that doesn't fit in to 120C then we _always_ adhere to : 
```
void func(
        A a,
        B b
) {
```
I feel like this would be the most simplest and readable for someone new to codebase.
I volunteer to add this styling rule and reformat the codebase to stick to the rule.

nickbp: That sounds fine, I'd subjectively prefer this style:

```
void func(
        A a,
        B b) {
    ... code here ...
}
```

...but I could be convinced. This would be a separate PR in any case. Maybe we could do a poll...

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/AbstractResourceMapper.java
nickbp: I think it'd be fine to have a static logger in the utils class

takirala:
@jeid64 @nickbp this is ready for review

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterUtils.java
```diff
@@ -87,21 +93,24 @@ public static String getServiceNamespacedRootPath(String namespace, String pathN
...
-            return new StringBuilder(first).deleteCharAt(first.length() - 1).append(second).toString();
-        } else if (first.endsWith(PATH_DELIM_STR) || second.startsWith(PATH_DELIM_STR)) {
-            // "hello/" + "world", OR "hello" + "/world" => "hello/world"
-            return new StringBuilder(first).append(second).toString();
-        } else {
-            // "hello" + "world" => "hello/world"
-            return new StringBuilder(first).append(PATH_DELIM).append(second).toString();
-        }
+    public static String join(final String... paths) {
+        return Arrays.stream(paths).reduce("", (first, second) -> {
```
elezar: Question, what about using the Java `Paths.get()` method? Also, would it not also make sense to just strip the leading and trailing slashes and then join the strings.
* https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#strip-java.lang.String-java.lang.String-

testing/sdk_upgrade.py
```diff
@@ -147,6 +147,23 @@ def get_config(package_name, service_name):
     return target_config
 
 
+def update_service(package_name, service_name, additional_options=None, to_package_version=None):
```
elezar: Nice!

testing/sdk_upgrade.py
```diff
@@ -147,6 +147,23 @@ def get_config(package_name, service_name):
...
+        with tempfile.NamedTemporaryFile() as opts_f:
+            opts_f.write(json.dumps(additional_options).encode('utf-8'))
+            opts_f.flush()  # ensure json content is available for the CLI to read below
+            sdk_cmd.svc_cli(
+                package_name,
+                service_name,
+                '{} --options={}'.format(' '.join(update_cmd), opts_f.name)
+            )
+    else:
+        sdk_cmd.svc_cli(package_name, service_name, ' '.join(update_cmd), check=True)
```
elezar: There is a slight difference here between the two commands, ideally we'd want to execute just one. We could: Create the named temporary file and then add `--options=filename` to `update_cmd`, have a single call to `sdk_cmd.svc_cli` and then cleanup the named temporary in a `finally` block. I think a named temporary may be cleaned up when the variable holding it goes out of scope anyway. (I just confirmed in ipython).

testing/sdk_upgrade.py
```diff
@@ -147,6 +147,23 @@ def get_config(package_name, service_name):
     return target_config
 
 
+def update_service(package_name, service_name, additional_options=None, to_package_version=None):
+    update_cmd = ['update', 'start']
+    if to_package_version:
+        update_cmd.append('--package-version={}'.format(to_package_version))
+    if additional_options:
+        with tempfile.NamedTemporaryFile() as opts_f:
+            opts_f.write(json.dumps(additional_options).encode('utf-8'))
```
elezar: How about using `json.dump(additional_optsion, opts_f.file, indent=2)` instead?

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
```diff
@@ -0,0 +1,67 @@
...
+                        config.SERVICE_NAME,
+                        3,
+                        additional_options={"service": {"yaml": "svc"}})
+    hello_task_id = sdk_tasks.get_task_ids(config.SERVICE_NAME, 'hello')
+    # Start update plan with options that have list of yaml files to make it launch in multi service mode
+    sdk_upgrade.update_service(config.PACKAGE_NAME,
+                               config.SERVICE_NAME,
+                               additional_options={"service": {"yamls": "svc,foobar_service_name"}})
+    # Ensure new tasks are launched but the old tasks do not relaunch
+    sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME, multiservice_name='foobar')
```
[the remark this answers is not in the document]
takirala: I am not sure what is your suggestion here. The `foobar_service_name` is exactly what the YAML is. Unlike all other YAML files where we have a service name that is inherited, `foobar_service_name` is a yaml file has `foobar` as its `service_name`. Do you want me to rename the yaml to something else ? `different_service_name.yml` may be ?

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
```diff
@@ -0,0 +1,67 @@
...
+    sdk_tasks.check_task_not_relaunched(config.SERVICE_NAME,
```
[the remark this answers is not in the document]
takirala: i have a task id `hello_task_id.pop()` and i am checking that specific task is not relaunched. What other tasks are you referring to ?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/Constants.java
```diff
@@ -98,4 +98,9 @@
      * The duration in seconds to decline offers the scheduler does not need for a short time.
      */
     public static final int SHORT_DECLINE_SECONDS = 5;
+
+    /**
+     * Indent (# of spaces) to be used to pretty print json objects.
+     */
+    public static final int JSON_INDENT_FACTOR = 2;
```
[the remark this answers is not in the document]
nickbp: I think a good place for it would be in `ResponseUtils` since it is somewhat specific to that, with the one exception being the build info log in `SchedulerConfig`

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -233,4 +238,41 @@ private static ServiceSpec createSampleServiceSpec(SchedulerConfig schedulerConf
...
+     * 3. The migration MUST happen before the {@link SchedulerBuilder#build()} is called (i.e., before the
+     * scheduler talks to {@link Persister}.
+     */
+    private static void checkAndMigrate(FrameworkConfig frameworkConfig, Persister persister) {
+        LOGGER.info("Checking if migration is needed...");
+        SchemaVersionStore schemaVersionStore = new SchemaVersionStore(persister);
+        int curVer = schemaVersionStore.getOrSetVersion(SchemaVersionStore.getSupportedSchemaVersionMultiService());
+        if (curVer == SchemaVersionStore.getSupportedSchemaVersionSingleService()) {
+            try {
+                LOGGER.warn("Found old schema in ZK Storage that can be migrated to a new schema");
```
nickbp: - This could be an `info()`
- Could give more context on the migration. For example, how about something like `Found single-service schema in ZK storage, migrating to multi-service schema`

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -233,4 +238,41 @@ private static ServiceSpec createSampleServiceSpec(SchedulerConfig schedulerConf
...
+                PersisterUtils.backUpFrameworkZKData(persister);
+                PersisterUtils.migrateMonoToMultiZKData(persister, frameworkConfig);
+                schemaVersionStore.store(SchemaVersionStore.getSupportedSchemaVersionMultiService());
+                LOGGER.info("Successfully migrated from old schema to new schema!!");
```
nickbp: - No exclamation marks
- Similar to the above, could specify `Successfully migrated from single-service schema to multi-service schema`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/AbstractResourceMapper.java
nickbp: Ping for static utils instead of inheritance

sdk/scheduler/src/test/java/com/mesosphere/sdk/storage/MemPersisterTest.java
```diff
@@ -181,6 +181,59 @@ public void testDeleteRoot() throws Exception {
         testDeleteRootForPersister(CuratorPersister.newBuilder(mockServiceSpec).disableLock().build(), "/");
     }
 
+    @Test
```
nickbp: It would make sense to have this test exactly match the one in CuratorPersisterTest, just to validate that the two behave identically

takirala: The only difference is curator persiter works even if the data is null. Mempersister does not. This is same for other methods as well. Should that be the case ?

nickbp: If it was like that already, I think that's fine for now.

testing/sdk_upgrade.py
takirala: If I understand your comment correctly, we always create a file handler but choose only to use it if `additional_options` is not None ? Can you put what you mean as a psuedocode ?

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterCache.java
```diff
@@ -83,6 +83,18 @@ public void setMany(Map<String, byte[]> pathBytesMap) throws PersisterException
         }
     }
 
+    @Override
+    public void recursiveCopy(String srcPath, String destPath) throws PersisterException {
+        rwlock.lock();
+        try {
+            MemPersister cache = getCache();
+            persister.recursiveCopy(srcPath, destPath);
+            cache.recursiveCopy(srcPath, destPath);
```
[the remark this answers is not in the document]
takirala: I am not sure what you mean. This method should fail if the destination exists. why do we need to invalidate or repopulate?

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/Persister.java
```diff
@@ -60,6 +60,20 @@
...
+     *  2. Commit the transaction to write new nodes.
+     * After (1) is done and before the completion of (2) there may be updates to nodes.
+     * Currently, this is used only for migrating the data which occurs during bootstrap and there should not be any
+     * conflicts. Be careful when using this method for other runtime operations.
+     *
+     * @param srcPath The source path of the node to copy from.
+     * @param destPath The destination path of the node to copy to.
+     * @throws PersisterException if the source node is not present or the destination node is already present
+     */
+    void recursiveCopy(String srcPath, String destPath) throws PersisterException;
```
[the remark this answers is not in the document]
takirala: This is true for any other method as well. E.g.: `get` follows the same semantics. I think it is fair, considering `Persister` is just an interface and `CuratorPersister` is specific to ZK and the `frameworkPrefix` is something that is specific to how we do things in ZK and `CuratorPersister`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/AbstractResourceMapper.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param taskResourceNamespace This is the namespace label from the Mesos
+     * @return If the taskResourceNamespace is non-empty, it MUST match with the namespace the scheduler is in.
+     * If the taskResourceNamespace is empty, we should NOT add a label now.
+     * This is applicable only in the "UPDATE" flow. During creating of new resources, we use the Scheduler namespace.
+     */
+    Optional<String> getNamespaceLabel(
+            Optional<String> taskResourceNamespace
+    ) {
+        return taskResourceNamespace.flatMap(x -> {
+            if (!resourceNamespace.isPresent() || !resourceNamespace.get().equals(x)) {
```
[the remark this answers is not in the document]
takirala: this is definitely not the scala-esque way (it is highly non-functional to use a variable outside the scope of the method). But we are in a java world and I think it is fair to say that we need to access the class level variable. I can refactor the code to use `if else` semantics if you would prefer that but I assumed this is the java-esque way of using `flatMap`. Am I incorrect in assuming so?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/AbstractResourceMapper.java
```diff
@@ -0,0 +1,90 @@
...
+    AbstractResourceMapper(Optional<String> resourceNamespace) {
+        logger = LoggingUtils.getLogger(getClass(), resourceNamespace);
+        this.resourceNamespace = resourceNamespace;
+    }
+
+    Optional<ResourceLabels> findMatchingDiskSpec(
+            Protos.Resource taskResource,
+            Collection<ResourceSpec> resourceSpecs
+    ) {
+        for (ResourceSpec resourceSpec : resourceSpecs) {
```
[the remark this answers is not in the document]
takirala: I agree, I just copy pasted the methods from the previous *ResourceMapper classes. Refactored them with a `findFirst` now.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/Constants.java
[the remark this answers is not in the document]
takirala: My bad, i thought this file was `com.mesosphere.sdk.Constants` anyway, i just resorted to use 2 as we used to do before.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -198,6 +198,102 @@ public static Builder newBuilder(String serviceName, String zookeeperHostPort) {
...
+            }
+
+            LinkedList<String> toBeWalked = new LinkedList<>(Collections.singleton(unprefixedSrc));
+            Map<String, byte[]> toBeAdded = new HashMap<>();
+            while (!toBeWalked.isEmpty()) {
+                String nextNode = toBeWalked.poll();
+                // This assertion should never fail acc. to the logic around it.
+                assert (nextNode.startsWith(unprefixedSrc)) : String.format("Failed to match prefix." +
+                        " Src [%s] Dest [%s] Child [%s]", unprefixedSrc, unprefixedDest, nextNode);
+                toBeAdded.put(withFrameworkPrefix(nextNode.replace(unprefixedSrc, unprefixedDest)), get(nextNode));
```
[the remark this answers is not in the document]
takirala: I agree. But i also feel that `curNode.replace(unprefixedSrc, unprefixedDest)` is "ugly" by itself but a good way to make sure this "ugly" code works is this should be inside this loop where we _definitely_ know it would not fail because of the logic around it. anyway, i will just keep this as is and address in a future cleanup when we end up sharing some code between `recursiveCopy` and `recursiveDelete`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -198,6 +198,102 @@ public static Builder newBuilder(String serviceName, String zookeeperHostPort) {
...
+                CuratorLocker.LOCK_PATH_NAME,
+                unprefixedSrc,
+                unprefixedDest)).size() != 4) {
+            throw new IllegalArgumentException(String.format("Cannot copy from %s to %s",
+                    unprefixedSrc, unprefixedDest));
+        }
+
+        try {
+            LOGGER.debug("Copying {} (and any children) to {}", unprefixedSrc, unprefixedDest);
+            if (client.checkExists().forPath(withFrameworkPrefix(unprefixedSrc)) == null) {
```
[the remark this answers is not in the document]
takirala: Good catch. However the `forPath` throws `java.lang.Exception`, I've added an extra catch block to rethrow them.

sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -198,6 +198,102 @@ public static Builder newBuilder(String serviceName, String zookeeperHostPort) {
...
+        runTransactionWithRetries(new SetTransactionFactory(pathBytesMap));
+    }
+
+    @Override
+    public void recursiveCopy(String unprefixedSrc, String unprefixedDest) throws PersisterException {
+        if (new HashSet<>(Arrays.asList(
+                serviceRootPath,
+                CuratorLocker.LOCK_PATH_NAME,
+                unprefixedSrc,
+                unprefixedDest)).size() != 4) {
```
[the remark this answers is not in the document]
takirala: As this method is recursively copying the data, i thought we should be extra careful. The check ensures that the special paths like the `serviceRootPath, CuratorLocker.LOCK_PATH_NAME` are not included in any of the `unprefixedSrc, unprefixedDest` variables. Considering this, using equals would result in multiple checks.

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
```diff
@@ -0,0 +1,67 @@
...
+    # Install hello-world with single yaml
+    sdk_install.install(config.PACKAGE_NAME,
+                        config.SERVICE_NAME,
+                        3,
+                        additional_options={"service": {"yaml": "svc"}})
+    hello_task_id = sdk_tasks.get_task_ids(config.SERVICE_NAME, 'hello')
+    # Start update plan with options that have list of yaml files to make it launch in multi service mode
+    sdk_upgrade.update_service(config.PACKAGE_NAME,
+                               config.SERVICE_NAME,
+                               additional_options={"service": {"yamls": "svc,foobar_service_name"}})
```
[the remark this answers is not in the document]
takirala: Good catch. I think the `MultiServiceManager#putService()` was talking care of the duplicates as I passed "svc" yml twice and that is why the test was passing. Fixed this now by setting `"yaml": ""`.

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
```diff
@@ -0,0 +1,67 @@
...
+    # Ensure new tasks are launched but the old tasks do not relaunch
+    sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME, multiservice_name='foobar')
+    sdk_tasks.check_task_not_relaunched(config.SERVICE_NAME,
+                                        'hello-0-server',
+                                        hello_task_id.pop(),
+                                        multiservice_name=config.SERVICE_NAME)
+    assert len(sdk_tasks.get_task_ids(config.SERVICE_NAME, 'foo')) == 1
```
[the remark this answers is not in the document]
takirala: Hmm, just being more explicit. It would catch the case where the scheduler thinks the deployment is complete but the tasks may not be launched. This PR deals with Mono to Multi Migration and i think its no harm to be explicit in the test.

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
```diff
@@ -0,0 +1,67 @@
...
+    hello_task_id = sdk_tasks.get_task_ids(config.SERVICE_NAME, 'hello')
+    # Start update plan with options that have list of yaml files to make it launch in multi service mode
+    sdk_upgrade.update_service(config.PACKAGE_NAME,
+                               config.SERVICE_NAME,
+                               additional_options={"service": {"yamls": "svc,foobar_service_name"}})
+    # Ensure new tasks are launched but the old tasks do not relaunch
+    sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME, multiservice_name='foobar')
+    sdk_tasks.check_task_not_relaunched(config.SERVICE_NAME,
+                                        'hello-0-server',
+                                        hello_task_id.pop(),
```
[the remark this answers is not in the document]
takirala: The tests should fail if the list is empty. But I should have added an assertion to check the same; added that now.

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -233,4 +238,41 @@ private static ServiceSpec createSampleServiceSpec(SchedulerConfig schedulerConf
...
+            } catch (PersisterException e) {
+                LOGGER.error("Unable to migrate ZK data : ", e);
+                throw new RuntimeException(e);
+            }
+        } else if (curVer == SchemaVersionStore.getSupportedSchemaVersionMultiService()) {
```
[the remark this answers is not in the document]
takirala: `curVer` is a primitive int and i cannot use `.equals`.

`isMigrationEnabled()` was my initial approach but seeing as that method doesn't do much other than "two trivial" if condition checks, i think it is more readable i follow the `if elif else` approach here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterUtils.java
takirala: I had something like

takirala:
https://github.com/mesosphere/dcos-commons/pull/2573/commits/b7a31ef653d7eaf9eb20250af568e87bb05e2de6 is the latest set of changes

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ExecutorResourceMapper.java
```diff
@@ -14,8 +16,9 @@
  * Handles cross-referencing a preexisting {@link Protos.ExecutorInfo}'s current {@link Protos.Resource}s against a set
  * of expected {@link VolumeSpec}s for that Executor.
  */
-public class ExecutorResourceMapper extends AbstractResourceMapper {
+public class ExecutorResourceMapper {
 
+    final Logger logger;
```
nickbp: This could be `private`, but this can be fixed in a separate PR if no other changes are needed

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
```diff
@@ -14,8 +16,10 @@
  * Handles cross-referencing a preexisting {@link Protos.TaskInfo}'s current {@link Protos.Resource}s against a set
  * of expected {@link ResourceSpec}s for that task.
  */
-class TaskResourceMapper extends AbstractResourceMapper {
+class TaskResourceMapper {
 
+    final Logger logger;
```
nickbp: Could also be `private`, same comment as above.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
```diff
@@ -18,8 +18,7 @@
  */
 class TaskResourceMapper {
 
-    private final Logger logger;
-
+    final Logger logger;
```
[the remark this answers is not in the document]
takirala: I missed this from the refactoring of `AbstractResourceMapper` and then further refactoring it in to `ResourceMapperUtils`. Fixed this by retaining the new visibility for `ExecutorResourceMapper` and `TaskResourceMapper` (as package private access was enough) and reverted my changes on increasing the visibility of `logger` objects.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -446,15 +456,22 @@ public UnexpectedResourcesResponse getUnexpectedResources(Collection<Protos.Offe
...
-            // Unrecognized service. Status for old task?
-            LOGGER.info("Received status for unknown task {}: {}",
-                    status.getTaskId().getValue(), TextFormat.shortDebugString(status));
-            return TaskStatusResponse.unknownTask();
-        }
-        LOGGER.info("Received status for task {}: {}", status.getTaskId().getValue(), status.getState());
-        return service.get().taskStatus(status);
+        return multiServiceManager
+                .getMatchingService(status)
+                .map(x -> x.taskStatus(status))
```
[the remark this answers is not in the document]
takirala: The first call to `.map(x -> x.taskStatus(status))` is for mapping the matched `AbstractScheduler` to `TaskStatusResponse`. If a matching service was found the second `.map(x -> x.taskStatus(status))` would never be executed.

`.map(x -> x.taskStatus(status))` is just converting `Optional<AbstractScheduler>` to `Optional<TaskStatusResponse>`

If a matching service was not found, we assume that the task status was meant for default service.

`orElseGet` would only be executed when the `getMatchingService` returns and `empty` optional.

This is whats happening in the call. 
```
                .getMatchingService(status) // See if a matching service is found and return an Optional
                .map(x -> x.taskStatus(status)) // If a matching `AbstractScheduler` is found, map it to a `TaskStatusResponse` object. This would happen only if above optional is non empty.
                .orElseGet(() -> multiServiceManager // only enter this block if getMatchingService returned an empty optional.
                    ....
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/MemPersister.java
```diff
@@ -130,6 +132,35 @@ public void setMany(Map<String, byte[]> pathBytesMap) throws PersisterException
         }
     }
 
+    @Override
+    public void recursiveCopy(String srcPath, String destPath) throws PersisterException {
+        lockRW();
+        try {
+            if (getNode(root, srcPath, false) == null) {
```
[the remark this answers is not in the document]
takirala: For checking the equality between source and destination, it is already handled by the checks where we are making sure the source is non empty and destination is empty.

For the inconsistency between MemPersister and CuratorPersister (For e.g.: We use `lock` as a ZK node to simulate locking whereas in MemPersister we use a ReadWrite Lock) that has always been the case even before the PR.

This is what the CuratorPersister does now : 
```
Stream
  .of(serviceRootPath, CuratorLocker.LOCK_PATH_NAME)
  .anyMatch(x -> x.equals(unprefixedSrc) || x.equals(unprefixedDest))) 
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterCache.java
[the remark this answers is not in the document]
takirala: I see your point now. But it is the same behavior with other methods in this class and any changes would be out of this PRs scope. I can create a jira to clean up this class overall.

sdk/scheduler/src/main/java/com/mesosphere/sdk/storage/PersisterUtils.java
takirala: I am not sure what would break if I change. Just saw that we have a test case to check the same. I added a comment inside the renamed method adding an example of this.

sdk/scheduler/src/test/java/com/mesosphere/sdk/storage/MemPersisterTest.java
takirala: copy pasting my comment from earlier
we don't need to check if source path and destination path are equal because we check the following : 
- that source path is non emtpy
- that dest path is empty

testing/sdk_tasks.py
```diff
@@ -191,17 +191,19 @@ def fn():
     fn()
 
 
-def check_task_not_relaunched(service_name, task_name, old_task_id, timeout_seconds=DEFAULT_TIMEOUT_SECONDS):
-    sdk_plan.wait_for_completed_deployment(service_name)
-    sdk_plan.wait_for_completed_recovery(service_name)
+def check_task_not_relaunched(service_name, task_name, old_task_id, multiservice_name=None, with_completed=False):
```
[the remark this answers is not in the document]
takirala: I agree this SHOULD be cleaned up. Created https://jira.mesosphere.com/browse/DCOS-39743 to address this.

testing/sdk_upgrade.py
takirala: I am not sure on `tempfile` semantics but by the time the code executes `sdk_cmd.svc_cli.....` isn't the temp file out of scope ? I ruled out this option initially thinking the file would be deleted by then.

testing/sdk_upgrade.py
```diff
@@ -147,6 +147,23 @@ def get_config(package_name, service_name):
...
+        with tempfile.NamedTemporaryFile("w") as f:
+            json.dump(additional_options, f)
+            f.flush()  # ensure json content is available for the CLI to read below
+            sdk_cmd.svc_cli(
+                package_name,
+                service_name,
+                '{} --options={}'.format(' '.join(update_cmd), f.name)
+            )
+    else:
+        sdk_cmd.svc_cli(package_name, service_name, ' '.join(update_cmd), check=True)
```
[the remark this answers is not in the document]
takirala: As we have helper methods for waiting on deploy, we can ignore to wait so that the callers can explicitly choose to wait. Also, in case of a multi service we do not actually know if we should wait for all the services and what are the names of the services if any. We can address this in a overhaul of testing to support multi service better. Added this to https://jira.mesosphere.com/browse/DCOS-39743

takirala:
Thanks for detailed review @elezar Really helps me understand SDK semantics better. 

Regarding your comments : 

> The validity checks for taskResource in TaskResourceMapper and ResourceMapperUtils seem to be executed in a loop while they are invariant. It would be good to know what the motivation for this is.

This seems to be left off code likely from a previous refactoring. I overlooked them while refactoring them from for loops to streams. Cleaned up this check in both the *Mapper classes now.

> There seems to be a subtle difference between the MemPersister and CuratorPersister that may cause different behaviours between our tests and running services.

Addressed your comments in detail above but some of the behavior, although inconsistent, it is at-least consistent with the other methods. We can address this in a future clean up if needed.

> The expectations around running the update function for a service in our tests.

I created a jira for passing around multi service names in our pytest tooling. https://jira.mesosphere.com/browse/DCOS-39743

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -234,4 +239,47 @@ private static ServiceSpec createSampleServiceSpec(SchedulerConfig schedulerConf
...
+     * offers, it does not know what to do with them if we support migration. Hence we support migration only in
+     * static service mode.
+     * 2. Note that if the {{user}} field in the yml is not specified in the mono mode, we used to default to `root`
+     * In the multi mode however, the framework config takes precedence if a user is not mentioned. Caution must
+     * be taken to ensure the service user is same (mention it explicitly in both mono and multi for the sake of
+     * consistency) during the migration.
+     * 3. The migration MUST happen before the {@link SchedulerBuilder#build()} is called (i.e., before the
+     * scheduler talks to {@link Persister}.
+     */
+    private static void checkAndMigrate(FrameworkConfig frameworkConfig, Persister persister) {
```
nickbp: I assume individual services shouldn't need to make customizations to this logic? I think we could upstream this logic as a single static function in `PersisterUtils`. At that point the separate backup/copy/delete function contents could be flattened into here as well, as they aren't doing that much individually.

frameworks/helloworld/src/main/java/com/mesosphere/sdk/helloworld/scheduler/Main.java
```diff
@@ -234,4 +239,47 @@ private static ServiceSpec createSampleServiceSpec(SchedulerConfig schedulerConf
...
+                LOGGER.info("Successfully migrated from single-service schema to multi-service schema");
+            } catch (PersisterException e) {
+                LOGGER.error("Unable to migrate ZK data : ", e);
+                throw new RuntimeException(e);
+            }
+        } else if (curVer == SchemaVersionStore.getSupportedSchemaVersionMultiService()) {
+            LOGGER.info("Schema version matches that of multi service mode. Nothing to migrate.");
+        } else {
+            throw new IllegalStateException(String.format("Storage schema version %d is not supported by this " +
+                    "software (expected: %d)", curVer, SchemaVersionStore.getSupportedSchemaVersionMultiService()));
```
nickbp: This error could be upgraded to mention that the single-service version is also expected.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/SchemaVersionStore.java
```diff
@@ -35,6 +44,14 @@
 
     private final Persister persister;
 
+    public static int getSupportedSchemaVersionSingleService() {
```
nickbp: How about just `getSingleServiceVersion`? `Supported` is implied, and `Schema` is implied via this being in `SchemaVersionStore`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/SchemaVersionStore.java
```diff
@@ -35,6 +44,14 @@
...
+        return SUPPORTED_SCHEMA_VERSION_SINGLE_SERVICE;
+    }
+
+    public static int getSupportedSchemaVersionMultiService() {
```
nickbp: Similarly this could be `getMultiServiceVersion`

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/SchemaVersionStore.java
```diff
@@ -18,6 +17,16 @@
  */
 public class SchemaVersionStore {
 
+    /**
```
nickbp: Actually, how about defining a `SchemaVersion` enum, and then use that enum instead of passing around raw ints directly?

It could look something like this, where the enum type could even include associated utility methods:

```
public enum SchemaVersion {
  SINGLE_SERVICE,
  MULTI_SERVICE,
  UNKNOWN;

  public static SchemaVersion parseInt(int rawVersion) {
    switch (rawVersion) {
      case 1:
        return SINGLE_SERVICE;
      case 2:
        return MULTI_SERVICE;
      default:
        return UNKNOWN;
    }
  }

  public int toInt() {
    switch (this) {
      case SINGLE_SERVICE:
        return 1;
      case MULTI_SERVICE:
        return 2;
      default:
        throw new IllegalArgumentException(String.format("Unable to convert %s to int", this));
    }
  }
};
```

Then the SchemaVersionStore would use the enum type in its interfaces:
```
public void check(SchemaVersion expectedVersion) {
  ...
}

public SchemaVersion getOrSetVersion(SchemaVersion expectedVersion) {
  ...
}
```

Also, the `getSupported*()` calls wouldn't be needed -- upstream could compare `SchemaVersion` enums directly

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/SchemaVersionStore.java
```diff
@@ -53,6 +84,15 @@ public SchemaVersionStore(Persister persister) {
      * @throws IllegalStateException if a value is present which doesn't match the expected value
      */
     public void check(int expectedVersion) throws StateStoreException {
+        int currentVersion = getOrSetVersion(expectedVersion);
```
nickbp: Can these `check()` and `getOrSetVersion()` calls be updated to accept the enum values as well?

takirala: I tried to do that initially but then realized we won't be able to test by passing random integer values in our unit testing. I will refactor if we are happy with unit tests passing `UNKNOWN`.

sdk/scheduler/src/test/java/com/mesosphere/sdk/storage/MemPersisterTest.java
elezar: OK, I've had a look at the two examples again, and you're right that the current checks also fail in the case where the arguments are the same. With that said, the error messages shown are a little cryptic and it may make it easier to debug if we have clearer messages.

I would also add unit tests to explicitly check this case to ensure that this behaviour is maintained in future.

testing/sdk_upgrade.py
elezar: See the comment here: https://github.com/mesosphere/dcos-commons/pull/2573/files/4b23550c81cdee1457a559057b1be03c18412057#r204701945

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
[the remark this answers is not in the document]
elezar: This is probably OK as it is. Thanks.

frameworks/helloworld/tests/test_mono_to_multi_migrate.py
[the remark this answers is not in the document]
elezar: I thought I had already made a comment regarding the use of `hello_task_id.pop()` here:

Two things here:

It may be a good idea to first wait for a completed deployment for both services associated with the multi-service before checking for tasks being relaunched. (update triggers a deployment)
I don't think that hello_task_id.pop() is what we want here:
```
In [3]: foo = ["bar", "baz"]

In [4]: set(foo.pop())
Out[4]: {'a', 'b', 'z'}
```

elezar:
Sorry @takirala. Only just saw:
> I don't understand (2). Our intention in this test is to make sure there is only ONE task that is relaunched. For the tasks of the foo service, they are new and relaunched does not make sense for them.

We could continue using `pop`, but we would have to replace this with:
```
sdk_tasks.check_task_relaunched(config.SERVICE_NAME, 'hello-0-server', [hello_task_id.pop()])
```
to ensure that we are not comparing with a list of characters internally.

PR #2575, opened 2018-07-06
[DCOS-29527] add Kafka and HDFS testing client

elezar:
This PR updates / adds the Kafka and HDFS testing client Dockerfiles to be used with the new release tooling in mesosphere/data-services-release-tooling.

Specifically: https://github.com/mesosphere/data-services-release-tooling/blob/master/client_docker_image.py#L12

This effectively moves the files from: https://github.com/elezar/dcos-clients

- `kafka-testing-client`:
  - Jenkins build: https://jenkins.mesosphere.com/service/jenkins/view/Infinity/job/infinity-tools/job/release-tools/job/build-package/145/
  - Docker repo: https://hub.docker.com/r/mesosphere/kafka-testing-client/
- `hdfs-testing-client`
  - Jenkins build: https://jenkins.mesosphere.com/service/jenkins/view/Infinity/job/infinity-tools/job/release-tools/job/build-package/146/
  - https://hub.docker.com/r/mesosphere/hdfs-testing-client/


frameworks/kafka/client/Dockerfile
```diff
@@ -6,14 +6,14 @@ RUN apt-get update && \
 
 ARG JAVA_TAR_ARCHIVE="jre-8u152-linux-x64-jce-unlimited.tar.gz"
```
samvantran: Probably can do this in a different PR but since we updated the JRE to 172 for all packages in #2488 we may want to keep the testing client in step

frameworks/kafka/client/Dockerfile
```diff
@@ -6,14 +6,14 @@ RUN apt-get update && \
...
-    curl -O https://downloads.mesosphere.com/java/${JAVA_TAR_ARCHIVE} && \
+    curl -L -O https://downloads.mesosphere.com/java/${JAVA_TAR_ARCHIVE} && \
 	tar xzvf ${JAVA_TAR_ARCHIVE} && \
 	rm ${JAVA_TAR_ARCHIVE}
 
 ENV JAVA_HOME /opt/jre1.8.0_152
 
 RUN cd /opt && \
-    curl -O https://downloads.mesosphere.com/dcos-commons/artifacts/0.41.0/bootstrap.zip && \
+    curl -L -O https://downloads.mesosphere.com/dcos-commons/artifacts/0.41.0/bootstrap.zip && \
```
samvantran: same comment if we're moving to bootstrap 0.52.x

samvantran:
Do we want to update all the places in code that will use the new official testing-clients? 
eg 
https://github.com/mesosphere/dcos-commons/blob/master/frameworks/kafka/tests/test_active_directory_auth.py#L87

frameworks/kafka/client/Dockerfile
elezar: I didn't want to change the clients in this PR, just add them to the repo. We can follow up with an update PR.

elezar:
> Do we want to update all the places in code that will use the new official testing-clients?
eg
https://github.com/mesosphere/dcos-commons/blob/master/frameworks/kafka/tests/test_active_directory_auth.py#L87

Ideally, we would want the docker images to be built off a commit on the master branch and not a PR. This would ensure that we would aid in reproducibility.

With this said, it may be better to move the testing clients to their own repo -- I'm not sure whether it really makes sense to have them so tightly coupled to the `dcos-commons` codebase.

frameworks/kafka/client/Dockerfile
samvantran: sounds good... and sounds like we'll be moving this to a separate repo anyhow

elezar:
@porridge these clients seem to fit relatively well into the scope of the `dcos-commons-ci` repo. Should I create a PR to move these there?

porridge:
SGTM, modulo mismatch of public/private status of the repos.

elezar:
I don't think these *need* to be public. They just need to be somewhere.

elezar:
I've moved the client dockerfiles to mesosphere/dcos-commons-ci#4.

PR #2576, opened 2018-07-09
[COPS-3222][DCOS-38688] Cassandra pod replace performs rolling update

elezar:
This is a follow up for #2556 which was not triggered as part of the PR fro Cassandra as that specific PR job has `-k "sanity and not recovery"` selector. It failed the nightlies since a permanent node recovery results in a rolling update of other node tasks.

This PR:
* Removes the `recovery` mark from the test (it could be re-added once it is confirmed that this fix addresses the issue)
* Adds support to the `check_permanent_recovery` for updating multiple pods.

### DIFF
--- a/testing/sdk_recovery.py
+++ b/testing/sdk_recovery.py
@@ -14,7 +14,7 @@ def check_permanent_recovery(
     service_name: str,
     pod_name: str,
     recovery_timeout_s: int,
-    pods_whos_tasks_should_change: typing.List[str] = None,
+    pods_with_updated_tasks: typing.List[str] = None,
 ):
     """
     Perform a replace operation on a specified pod and check that it is replaced
@@ -28,7 +28,7 @@ def check_permanent_recovery(
 
     pod_list = set(sdk_cmd.svc_cli(package_name, service_name, "pod list", json=True))
 
-    pods_to_update = set(pods_whos_tasks_should_change + [pod_name, ])
+    pods_to_update = set(pods_with_updated_tasks + [pod_name])
 
     tasks_to_replace = {}
     for pod in pods_to_update:
--- a/testing/sdk_recovery.py
+++ b/testing/sdk_recovery.py
@@ -28,7 +28,9 @@ def check_permanent_recovery(
 
     pod_list = set(sdk_cmd.svc_cli(package_name, service_name, "pod list", json=True))
 
-    pods_to_update = set(pods_with_updated_tasks + [pod_name])
+    pods_to_update = set(
+        pods_with_updated_tasks if pods_with_updated_tasks else [] + [pod_name]
+    )
 
     tasks_to_replace = {}
     for pod in pods_to_update:
--- a/testing/sdk_recovery.py
+++ b/testing/sdk_recovery.py
@@ -1,4 +1,5 @@
 import logging
+import typing
 
 import sdk_cmd
 import sdk_plan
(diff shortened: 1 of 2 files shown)


testing/sdk_recovery.py
```diff
@@ -24,15 +26,18 @@ def check_permanent_recovery(
     sdk_plan.wait_for_completed_deployment(service_name)
     sdk_plan.wait_for_completed_recovery(service_name)
 
-    pod_list = sdk_cmd.svc_cli(package_name, service_name, "pod list", json=True)
+    pod_list = set(sdk_cmd.svc_cli(package_name, service_name, "pod list", json=True))
+
+    pods_to_update = set(pods_whos_tasks_should_change + [pod_name, ])
```
porridge: No need for the trailing comma here.

testing/sdk_recovery.py
```diff
@@ -43,7 +48,8 @@ def check_permanent_recovery(
     sdk_plan.wait_for_kicked_off_recovery(service_name, recovery_timeout_s)
     sdk_plan.wait_for_completed_recovery(service_name, recovery_timeout_s)
 
-    sdk_tasks.check_tasks_updated(service_name, pod_name, tasks_to_replace)
+    for pod, tasks in tasks_to_replace.items():
```
porridge: s/items/iteritems/?

testing/sdk_recovery.py
```diff
@@ -13,6 +14,7 @@ def check_permanent_recovery(
     service_name: str,
     pod_name: str,
     recovery_timeout_s: int,
+    pods_whos_tasks_should_change: typing.List[str] = None,
```
mpereira: `s/who/whose/`? 💩

testing/sdk_recovery.py
elezar: from: https://stackoverflow.com/questions/10458437/what-is-the-difference-between-dict-items-and-dict-iteritems
> One of Python 3’s changes is that  items() now return iterators, and a list is never fully built. The iteritems() method is also gone, since items() in Python 3 works like viewitems() in Python 2.7.

Also:
```python
✗ ipython
Python 3.6.4 (default, Mar  9 2018, 23:15:03)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: foo = {'a': 1, 'b': 2}

In [2]: for f, v in foo.iteritems():
   ...:     print(f, v)
   ...:
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-2c0163c99ec9> in <module>()
----> 1 for f, v in foo.iteritems():
      2     print(f, v)
      3
```

elezar:
Sorry @mpereira, I don't quite follow. What would the `pod_deploy_strategy` be in this case? I'm definitely open to improving this and making the intent clearer, but my immediate concern was fixing the CI regression that was caused by merging #2556 on Friday.

mpereira:
I was thinking `pod_deploy_strategy` would be either `'serial'` or `'parallel'`. If it's `'serial'`, `check_permanent_recovery` could just use the `pod_list` value for what is currently `pods_with_updated_tasks`, instead of relying on the caller to specify it.

elezar:
In an offline discussion with @mpereira we sorted out his concerns. I have updated the docstring to be more descriptive.

mpereira:
Thanks for the improved docstring @elezar!

PR #2579, opened 2018-07-10
[DCOS-38379][TESTS] Remove all multiservice dynamic tests from smoke

elezar:
This PR removes an additional `test_multiservice_dynamic` test from the set of smoke tests. The original was removed in a68699864f5e6da65f325993ba5098d8a1e1e05d.

### DIFF
--- a/frameworks/helloworld/tests/test_multiservice_dynamic.py
+++ b/frameworks/helloworld/tests/test_multiservice_dynamic.py
@@ -49,7 +49,7 @@ def check_scheduler_relaunched(service_name: str, old_scheduler_task_id: str,
     """
     @retrying.retry(
         wait_fixed=1000,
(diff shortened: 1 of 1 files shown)


frameworks/helloworld/tests/test_multiservice_dynamic.py
```diff
@@ -111,7 +111,6 @@ def test_add_deploy_restart_remove():
 
 
 @pytest.mark.sanity
-@pytest.mark.smoke
```
mpereira: Will this test be re-tagged with `smoke` in the context of DCOS-38379?

elezar: Yes. Let me add a comment to the issue.

PR #2580, opened 2018-07-10
[DCOS-38545] Adding default service support for unlabelled tasks and offers

kvish:


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/multi/MultiServiceEventClient.java
```diff
@@ -512,4 +526,13 @@ private static OfferResources getEntry(Map<Protos.OfferID, OfferResources> map,
         }
         return currentValue;
     }
+
+    /**
+     * Sets the default service name. When specified unknown task statuses and reservations will be forwarded to
+     * the specified service.
+     */
+    public void setDefaultServiceName(String defaultServiceName) {
```
takirala: Change this to boolean and the `defaultServiceName` should be `frameworkName`

PR #2581, opened 2018-07-11
DCOS-39121 Automatically renew creds with 'maws' when upload fails

nickbp:
If an upload to S3 fails, an attempt will be automatically made to reauth before reattempting the upload. If reauth fails, or if `maws` is not present in the local `PATH`, then it reverts to current behavior and just gives up.

Automatically determines the profile to renew via the `AWS_PROFILE` envvar, with a fallback of extracting the profile name from the credentials INI file.

### DIFF
--- a/tools/universe/s3_uploader.py
+++ b/tools/universe/s3_uploader.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import configparser
 import logging
 import os
 import os.path
@@ -12,11 +13,12 @@ logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 class S3Uploader(object):
     def __init__(self, s3_directory, dry_run=False):
         # check if aws cli tools are installed
-        if not subprocess.run("aws --version".split()).returncode == 0:
+        if subprocess.run('aws --version'.split()).returncode != 0:
             raise Exception('Required "aws" command is not installed.')
 
         self._s3_directory = s3_directory
         self._aws_region = os.environ.get('AWS_UPLOAD_REGION', '')
+        self._reauth_attempted = False
         self._dry_run = dry_run
 
     def get_s3_directory(self):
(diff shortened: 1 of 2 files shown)


tools/universe/s3_uploader.py
```diff
@@ -36,6 +38,55 @@ def upload(self, filepath, content_type=None):
...
             raise Exception('Failed to upload {} to {}'.format(filepath, dest_url))
+
+    def _renew_credentials(self):
+        if self._reauth_attempted:
+            # reauth was already tried once during this session. any failure must be from something else.
+            return False
+        self._reauth_attempted = True
+
+        # check that maws is installed before trying anything else
+        if subprocess.run('maws -h'.split(), stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL).returncode != 0:
```
elezar: Any reason why we don't use `which`?

tools/universe/s3_uploader.py
```diff
@@ -36,6 +38,55 @@ def upload(self, filepath, content_type=None):
...
+            log.error('Unable to renew credentials: Missing "maws" command in PATH')
+            return False
+
+        # renew creds using discovered aws profile
+        aws_profile = self._get_aws_profile()
+        log.info('Upload attempt failed, renewing credentials for AWS profile {}'.format(aws_profile))
+        ret = os.system('maws login {}'.format(aws_profile))
```
elezar: Newer versions of `maws` have a `-r` flag which automatically refreshes the credentials. This would have to be started in a separate thread.

nickbp: In theory, this path should only be hit if the user hasn't already set that up

tools/universe/s3_uploader.py
nickbp: Good point, I'll fix it in a follow-up PR

PR #2582, opened 2018-07-12
DCOS-38517 Remove dependency on hibernate for spec validation

nickbp:
TLDR: A sea of cleanup to ServiceSpec construction boilerplate

- Our use of validation annotations can lead to library collision issues when other projects try to include us as a library. Therefore, switch to performing validation locally instead. As a bonus, the validation error messages aren't garbage anymore!
- This PR also cleans up the construction of several spec types, adding dedicated Builders etc. There were a lot of inconsistencies across the board before.


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultTaskSpec.java
```diff
@@ -109,6 +82,32 @@ private DefaultTaskSpec(Builder builder) {
                 builder.discoverySpec,
                 builder.taskKillGracePeriodSeconds,
                 builder.transportEncryption);
+        ValidationUtils.nonEmpty(this, "name", name);
+        ValidationUtils.nonNull(this, "goalState", goalState);
+        ValidationUtils.nonNull(this, "essential", essential);
+        ValidationUtils.nonNull(this, "resourceSet", resourceSet);
+
+        // Check for duplicate names and paths:
+        ValidationUtils.atLeast(
```
elezar: A `isUnique` check may reduce the boiler plate a little (especially if it accepts a stream).

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultTransportEncryptionSpec.java
```diff
@@ -2,39 +2,33 @@
...
     }
 
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
     public DefaultTransportEncryptionSpec(Builder builder) {
         this(builder.name, builder.type);
-        ValidationUtils.validate(this);
+        ValidationUtils.nonEmpty(this, "name", name);
```
elezar: Just thinking about this, would a `nonBlank` check make sense too?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
```diff
@@ -3,32 +3,27 @@
...
-           @Pattern(regexp = "[a-zA-Z0-9]+([a-zA-Z0-9_-]*[/\\\\]*)*")
-                someDir/anotherDir in SandBox is not a valid path
+    /**
+     * Disallow slashes in the container path. If a slash is used, Mesos will silently ignore the mount operation for
+     * the persistent volume. See also:
+     *       mesos:src/slave/containerizer/mesos/isolators/filesystem/linux.cpp#L628
+     *       mesos:src/slave/containerizer/docker.cpp#L473
      */
-    @Pattern(regexp = "[a-zA-Z0-9]+([a-zA-Z0-9_-]*)*")
+    private static final Pattern VALID_CONTAINER_PATH_PATTERN = Pattern.compile("[a-zA-Z0-9]+([a-zA-Z0-9_-]*)*");
```
elezar: Is a regexp really required here. Is a `contains` not sufficient? Alternatively, the comment should be updated to state that there are other characters that are also not allowed.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/NamedVIPSpec.java
```diff
@@ -80,4 +94,48 @@ public boolean equals(Object o) {
...
+        }
+
+        public Builder vipPort(Integer vipPort) {
+            this.vipPort = vipPort;
+            return this;
+        }
+
+        public NamedVIPSpec build() {
+            PortSpec portSpec = super.build();
+            return new NamedVIPSpec(
```
elezar: This means we don't call the builder constructor, correct? Does this mean that we miss the validation?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -601,30 +604,26 @@ private static DefaultNetworkSpec convertNetwork(
...
-                portSpecs.add(new PortSpec(
-                        portValueBuilder.build(),
-                        role,
-                        preReservedRole,
-                        principal,
-                        rawPort.getEnvKey(),
-                        name,
-                        visibility,
-                        networkNames));
+                portSpecBuilder = PortSpec.newBuilder();
```
elezar: It may make it a little easier to read if we negate the conditional and add the shorter branch at the start.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -601,30 +604,26 @@ private static DefaultNetworkSpec convertNetwork(
...
             }
+            portSpecBuilder
```
elezar: Nice cleanup!

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallSchedulerTest.java
```diff
@@ -262,22 +262,22 @@ public void testAllButDeregisteredPlanCompletes() throws Exception {
         // Starts with a near-empty plan with only the deregistered call incomplete
         Plan plan = getUninstallPlan(uninstallScheduler);
 
-        List<Status> expected = Arrays.asList(Status.PENDING);
+        Collection<Status> expected = Collections.singletonList(Status.PENDING);
```
elezar: Is `singletonList` strictly required?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultTaskSpec.java
nickbp: Created, SO had a way to do it efficiently from a stream as well

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultTaskSpec.java
```diff
@@ -109,6 +82,32 @@ private DefaultTaskSpec(Builder builder) {
...
+                this,
+                "configFiles",
+                configFiles.stream()
+                        .map(config -> config.getName())
+                        .distinct()
+                        .count(),
+                configFiles.size());
+        ValidationUtils.atLeast(
+                this,
+                "configFiles",
```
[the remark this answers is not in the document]
nickbp: Done

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultTransportEncryptionSpec.java
nickbp: Yes. Done. Also used it for PodSpec.type.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
nickbp: Updated the comment

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/NamedVIPSpec.java
nickbp: Yes, it got weird with these inherited types. I broke out the validation into separate functions to be explicitly called in the subtypes.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
nickbp: Good point, done

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/uninstall/UninstallSchedulerTest.java
nickbp: Yes. It's being compared against the return value of `getStepStatuses()` which is a list. For whatever reason, a List with one element never matches a Set with one element, even when that element is the same.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultResourceSpec.java
```diff
@@ -44,14 +31,20 @@ public DefaultResourceSpec(
...
         this.principal = principal;
+
+        ValidationUtils.nonEmpty(this, "name", name);
+        ValidationUtils.nonNull(this, "value", value);
+        ValidationUtils.nonEmpty(this, "role", role);
+        ValidationUtils.nonEmpty(this, "principal", principal);
+
+        if (value.hasScalar() && value.getScalar().getValue() <= 0) {
+            throw new IllegalArgumentException(
+                    String.format("Scalar resource value must be greater than zero: %s", this));
```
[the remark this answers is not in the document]
elezar: > I figured it could be left ouf of ValidationUtils since it's somewhat specific to resource validation
Yes, by the end of the review, I had come to a similar conclusion.

PR #2584, opened 2018-07-12
[TOOLS] Default upgrades_from to an empty string.

mpereira:
When this script is run from `make-stub-permanent` the ENV variable is an empty string. When it's run from `release-framework-to-universe` the ENV variable isn't set, so `os.environ.get('UPGRADES_FROM')` returns `None`.


mpereira:
Thanks @samvantran.

The issue was introduced with https://github.com/mesosphere/dcos-commons/pull/2583

PR #2335, opened 2018-03-02
Tooling/fix new standalone framework 0

mpereira:

PR #2586, opened 2018-07-12
DCOS-39268 Remove unused template project, fix lint

nickbp:
This project was originally added for creating new services from. In practice it's never been used and has gone stale -- at this point it just confuses new users. Remove it for now, until we've got a better story for new users.

Also some python lint-related fixes:
- In the lint checker, don't attempt to check a 'changed' python file that was deleted.
- Fix all current lint errors for python files in testing/ and tools/

Closes #2329
Closes #2335


elezar:
> it looks like self._s3_directory was never set? Do we make use of universe artifacts hosted via HTTP?

I don't think we use it for our own purposes, but someone else might.

testing/security/cipher_suites.py
```diff
@@ -183,9 +183,7 @@
     'EXP-KRB5-RC2-CBC-SHA': 'TLS_KRB5_EXPORT_WITH_RC2_CBC_40_SHA',
     'EXP-KRB5-RC4-MD5': 'TLS_KRB5_EXPORT_WITH_RC4_40_MD5',
     'EXP-KRB5-RC4-SHA': 'TLS_KRB5_EXPORT_WITH_RC4_40_SHA',
-    'EXP-RC2-CBC-MD5': 'SSL_CK_RC2_128_CBC_EXPORT40_WITH_MD5',
```
elezar: This addresses https://jira.mesosphere.com/browse/DCOS-37593

tools/ci/checks/get_applicable_changes.py
```diff
@@ -86,14 +96,16 @@ def main():
     else:
         files = flatten_file_list(args.files)
 
-    filtered_files = filter_build_files_and_folders(
-        ignore_extensions(
-            filter_extensions(files, args.extensions), args.ignore_extensions
-        ),
-        args.only_build_files,
-    )
+    if args.extensions:
```
elezar: Thanks @nickbp. This is much cleaner.

tools/enable_mount_volumes.py
```diff
@@ -16,7 +16,6 @@
 import logging
 import os
 import os.path
-#import pprint
 import sys
 import time
 import uuid
```
elezar: The `fabric` imports below are not part of our standard tooling. As far as I can tell, this is a deprecated API in any case.

This is the reason for the additional `pylint` failure.

frameworks/template/src/test/java/com/mesosphere/sdk/template/scheduler/ServiceTest.java
```diff
@@ -1,79 +0,0 @@
...
-    }
-
-    @Test(expected = IllegalStateException.class)
-    public void testInvalidPlacementConstraint() throws Exception {
-        new ServiceTestRunner().setSchedulerEnv("NODE_PLACEMENT", INVALID_HOSTNAME_CONSTRAINT).run(getDeploymentTicks());
-    }
-
-    @Test
-    public void testSwitchToInvalidPlacementConstraint() throws Exception {
-        ServiceTestResult initial = new ServiceTestRunner().setSchedulerEnv("NODE_PLACEMENT", VALID_HOSTNAME_CONSTRAINT).run(getDeploymentTicks());
```
elezar: One note, I don't think that these are replicated elsewhere -- and as such we should move these to `helloworld` or an alternate location.

elezar:
@nickbp I just noticed that we may not have updated the `developer-guide.md` (i.e. removed the instructions).

mpereira:
@elezar I pushed a PR doing that here: https://github.com/mesosphere/dcos-commons/pull/2590

PR #2587, opened 2018-07-13
[DCOS-38379] fixing flaky multi service test

kvish:
task name will be the same, it is the id that will be different. Also the old task id will always show up if we query for completed tasks.

### DIFF
--- a/frameworks/helloworld/tests/test_multiservice_dynamic.py
+++ b/frameworks/helloworld/tests/test_multiservice_dynamic.py
@@ -53,7 +53,7 @@ def check_scheduler_relaunched(service_name: str, old_scheduler_task_id: str,
         retry_on_result=lambda res: not res)
     def fn():
         try:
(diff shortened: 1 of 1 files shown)


frameworks/helloworld/tests/test_multiservice_dynamic.py
```diff
@@ -53,7 +53,7 @@ def check_scheduler_relaunched(service_name: str, old_scheduler_task_id: str,
         retry_on_result=lambda res: not res)
     def fn():
         try:
-            task_ids = set([t['id'] for t in shakedown.get_tasks(completed=True) if t['name'].startswith(service_name)])
+            task_ids = set([t['id'] for t in shakedown.get_tasks(completed=False) if t['name'] == service_name])
```
elezar: This only works for the new services that *always* prepend the service name? Also, this will not work with a foldered service names, correct?

PR #2588, opened 2018-07-13
[TOOLS] Add -e and --envfile flags to test.sh to pass environment

elezar:
This PR adds two options to test.sh for better integration into other tooling (such as that used by EdgeLB). This was after a discussion with @vespian today.

The options are:
- `-e ENV_PASSTHROUGH`
- `--envfile ENV_FILE`

Where `ENV_PASSTHROUGH` is a comma-separated list of environment variables and `ENV_FILE` is a path to a local environment file.

In both cases, these options are passed through *in addition* to those already handled by the script. The mechanism for passing the environment to `docker run` command remains a temporary environment file.

Additional improvements:
* A timestamp has also been added to the names of the temporary envfile and credentials file created by the script to allow multiple containers to be run simultaneously.
* Don't require a framework to be specified if `--interactive` mode is selected.
* Allow the docker image to be specified using the `DOCKER_IMAGE` environment variable.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -58,6 +58,7 @@ headless="false"
 interactive="false"
 package_registry="false"
 docker_command=${DOCKER_COMMAND:="bash /build-tools/test_runner.sh $WORK_DIR"}
+docker_image=${DOCKER_IMAGE:-"mesosphere/dcos-commons:latest"}
 env_passthrough=
 envfile_input=
 
@@ -368,7 +369,7 @@ ${docker_interactive_arg} \
 --env-file $envfile \
 ${volume_args} \
 -w $WORK_DIR \
-mesosphere/dcos-commons:latest \
+${docker_image} \
 ${docker_command}"
 
 echo "==="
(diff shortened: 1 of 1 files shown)


vespian:
@elezar  Thank you so much! When do you think it will be released? Which version is it going to be?

elezar:
> @elezar Thank you so much! When do you think it will be released? Which version is it going to be?

If the changes look good, I can backport this to the `sdk-0.40` branch which is probably what you are pulling in to edge-lb in any case?

vespian:
@elezar We are using 0.42.1 ATM, but it should not be a problem for us to update to a later version as long as it is a stable release.

elezar:
> @elezar We are using 0.42.1 ATM, but it should not be a problem for us to update to a later version as long as it is a stable release.

Thanks @vespian. Let's merge this then and backport to `sdk-0.40` we also have a pending bugfix that would make it worthwhile to cut a `0.42.2` release sometime soon and you can switch to that.

PR #2591, opened 2018-07-18
[DCOS-38688] Add custom rack upgrade test

elezar:
This PR adds an upgrade test for custom racks in Cassandra.

@nickbp we could probably rather add a Unit Test that ensures that the property is present in the Cassandra racks configuration file, but that would not really test upgrade.


elezar:
> Do you think making sdk_install.install merge its service_name argument into additional_options would be too much "magic"?

No, that was my thought: https://jira.mesosphere.com/browse/DCOS-39479 Are you happy to address that in a separate ticket?

mpereira:
Of course, that's why I approved 🙂

PR #2592, opened 2018-07-18
DCOS-39389 Support customizing marathon name and VIP TLD

nickbp:
Allow the operator to manually specify a custom name for marathon. This is used when creating URLs which point back to the scheduler, and may need to be customized if the scheduler is running in a MoM.

Similarly, we currently allow manual customization of the autoip TLD, but didn't yet support the same for the VIP TLD. Add support for that.


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/security/TLSArtifact.java
```diff
@@ -56,8 +58,7 @@ public String getDescription() {
     public String getSecretStoreName(String sansHash, String taskInstanceName, String transportEncryptionName) {
         String fullName = Arrays.asList(sansHash, taskInstanceName, transportEncryptionName, name)
                 .stream()
-                .filter(item -> item != null)
-                .filter(item -> !item.equals(""))
+                .filter(item -> !StringUtils.isEmpty(item))
```
elezar: `isBlank()`? or even just. cc @takirala is it possible to write:
```java
.filter(!StringUtils::isBlank)
```
instead?

sdk/scheduler/src/test/java/com/mesosphere/sdk/framework/FrameworkRunnerTest.java
```diff
@@ -37,6 +37,8 @@
     public void beforeEach() {
         MockitoAnnotations.initMocks(this);
         Capabilities.overrideCapabilities(mockCapabilities);
+
+        when(mockSchedulerConfig.getMarathonName()).thenReturn("test-marathon");
```
elezar: It's a little strange that this is needed here since none of the associated tests were changed. Does this indicate some kind of dependency that is not immediately apparent?

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/EndpointUtilsTest.java
```diff
@@ -33,49 +35,73 @@ public void testToEndpoint() {
...
+                EndpointUtils.toAutoIpEndpoint("path/to/svc", "task", 5, SchedulerConfigTestUtils.getTestSchedulerConfig()));
+        assertEquals("task.pathtosvc-with-dots.autoip.tld:5",
+                EndpointUtils.toAutoIpEndpoint("path/to/svc.with.dots", "task", 5, SchedulerConfigTestUtils.getTestSchedulerConfig()));
     }
 
     @Test
     public void testToAutoIpEndpointCustomTLD() {
         SchedulerConfig mockSchedulerConfig = SchedulerConfigTestUtils.getTestSchedulerConfig();
-        when(mockSchedulerConfig.getServiceTLD()).thenReturn("what.a.fun.test.tld");
+        when(mockSchedulerConfig.getAutoipTLD()).thenReturn("what.a.fun.test.tld");
```
elezar: This seems like a duplication of the previous `testToAutoIpEndpoint` test and I don't really see it adding additional functionality considering that we're mocking `getAutoipTLD()` in both cases. We are still not checking the effect of the envvar.

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/EndpointUtilsTest.java
```diff
@@ -33,49 +35,73 @@ public void testToEndpoint() {
...
+        assertEquals("vip.pathtosvc.vip.tld:5",
+                EndpointUtils.toVipEndpoint("path/to/svc", mockSchedulerConfig, new VipInfo("vip", 5)));
+        assertEquals("vip.pathtosvc.with.dots.vip.tld:5",
+                EndpointUtils.toVipEndpoint("path/to/svc.with.dots", mockSchedulerConfig, new VipInfo("vip", 5)));
 
-        assertEquals("vip.svc.l4lb.thisdcos.directory:5", EndpointUtils.toVipEndpoint("svc", new VipInfo("/vip", 5)));
-        assertEquals("vip.pathtosvc.l4lb.thisdcos.directory:5", EndpointUtils.toVipEndpoint("/path/to/svc", new VipInfo("/vip", 5)));
-        assertEquals("vip.pathtosvc.l4lb.thisdcos.directory:5", EndpointUtils.toVipEndpoint("path/to/svc", new VipInfo("/vip", 5)));
-        assertEquals("vip.pathtosvc.with.dots.l4lb.thisdcos.directory:5", EndpointUtils.toVipEndpoint("path/to/svc.with.dots", new VipInfo("/vip", 5)));
+        assertEquals("vip.svc.vip.tld:5",
```
elezar: The intent of the test cases from here on are not really clear. I mean, basically we're checking that the VIP endpoint doesn't care about the leading slash in `VipInfo`. Would it not make sense to just break this out into a separate test?

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/EndpointUtilsTest.java
```diff
@@ -33,49 +35,73 @@ public void testToEndpoint() {
...
     public void testToSchedulerAutoIpHostname() {
-        assertEquals("svc.marathon.some.tld", EndpointUtils.toSchedulerAutoIpHostname("svc", mockSchedulerConfig));
-        assertEquals("svc-to-path.marathon.some.tld", EndpointUtils.toSchedulerAutoIpHostname("/path/to/svc", mockSchedulerConfig));
-        assertEquals("svc-to-path.marathon.some.tld", EndpointUtils.toSchedulerAutoIpHostname("path/to/svc", mockSchedulerConfig));
-        assertEquals("svc-with-dots-to-path.marathon.some.tld", EndpointUtils.toSchedulerAutoIpHostname("path/to/svc.with.dots", mockSchedulerConfig));
+        assertEquals("svc.test-marathon.autoip.tld",
+                EndpointUtils.toSchedulerAutoIpHostname("svc", mockSchedulerConfig));
+        assertEquals("svc-to-path.test-marathon.autoip.tld",
+                EndpointUtils.toSchedulerAutoIpHostname("/path/to/svc", mockSchedulerConfig));
+        assertEquals("svc-to-path.test-marathon.autoip.tld",
```
elezar: Would be nice to have this in a separate test that the leading slash doesn't affect the hostname.

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/endpoints/ArtifactResourceTest.java
```diff
@@ -14,24 +14,25 @@
...
-    @Mock SchedulerConfig mockSchedulerConfig;
+    @Mock private SchedulerConfig mockSchedulerConfig;
 
     @Before
     public void beforeEach() {
         MockitoAnnotations.initMocks(this);
         when(mockSchedulerConfig.getApiServerPort()).thenReturn(1234);
-        when(mockSchedulerConfig.getServiceTLD()).thenReturn("some.tld");
+        when(mockSchedulerConfig.getAutoipTLD()).thenReturn("some.tld");
+        when(mockSchedulerConfig.getMarathonName()).thenReturn("test-marathon");
```
elezar: We have now removed the tests for the default and no longer access the ENVVAR.

sdk/scheduler/src/test/java/com/mesosphere/sdk/http/queries/EndpointsQueriesTest.java
```diff
@@ -180,7 +179,7 @@ private void testEndpoint(String expectedHostname) throws ConfigStoreException {
         assertEquals(200, response.getStatus());
         JSONObject json = new JSONObject((String) response.getEntity());
         assertEquals(json.toString(), 3, json.length());
-        assertEquals("vip1.svc-name.l4lb.thisdcos.directory:5432", json.get("vip"));
+        assertEquals("vip1.svc-name.vip.tld:5432", json.get("vip"));
```
elezar: There seems to be a dependence between this test and some other setup since both the autoIP and VIP TLDs had to change.

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/security/CertificateNamesGeneratorTest.java
```diff
@@ -83,8 +98,12 @@ public void testGetSANs() throws Exception {
         Assert.assertTrue(names.contains(taskDnsName(TestConstants.TASK_NAME, TestConstants.SERVICE_NAME)));
         Assert.assertFalse(names.contains(taskDnsName("*", TestConstants.SERVICE_NAME)));
         Assert.assertFalse(names.contains(taskVipName("*", TestConstants.SERVICE_NAME)));
-        // echo -n "some-pod-test-task-name.service-name.autoip.dcos.thisdcos.directory" | sha1sum
-        Assert.assertEquals("a22fd2735aae7c55e47bece5f6c10612866583bf", certificateNamesGenerator.getSANsHash());
+
+        // Expected sha1sum of the hostname:
+        byte[] digest = SHA1_HASHER.digest("some-pod-test-task-name.service-name.autoip.tld".getBytes(StandardCharsets.UTF_8));
```
elezar: This is repeated here quite a bit. Consider breaking out into a helper.

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/security/CertificateNamesGeneratorTest.java
```diff
@@ -11,18 +10,33 @@
...
+    private static final MessageDigest SHA1_HASHER;
+    static {
+        try {
+            SHA1_HASHER = MessageDigest.getInstance("SHA-1");
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private static final String AUTOIP_TLD = "autoip.tld";
```
elezar: If these were changed to the defaults, then we would not need the other changes in this test?

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/security/CertificateNamesGeneratorTest.java
nickbp: Done

nickbp:
Does this look good at this point?

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
```diff
@@ -74,35 +73,36 @@ public static String toAutoIpEndpoint(
...
-        return toAutoIpHostname("marathon", serviceName, schedulerConfig);
+        return toAutoIpHostname(schedulerConfig.getMarathonName(), serviceName, schedulerConfig);
     }
 
     /**
      * Returns the correct DNS hostname:port endpoint for accessing the Scheduler API given the provided service name.
      */
     public static String toSchedulerAutoIpEndpoint(String serviceName, SchedulerConfig schedulerConfig) {
-        return toAutoIpEndpoint("marathon", serviceName, schedulerConfig.getApiServerPort(), schedulerConfig);
+        return toAutoIpEndpoint(
```
elezar: Quick question. Why doesn't this call `toSchedulerAutoIpHostname` internally?

sdk/scheduler/src/test/java/com/mesosphere/sdk/scheduler/SchedulerConfigTest.java
```diff
@@ -50,6 +50,24 @@ public void testRegionAwareness() {
...
 
+    @Test
+    public void testHostnameCustomizations() {
+        Map<String, String> confMap = getMinimalMap();
+        SchedulerConfig schedulerConfig = SchedulerConfig.fromEnvStore(EnvStore.fromMap(confMap));
+        Assert.assertEquals("autoip.dcos.thisdcos.directory", schedulerConfig.getAutoipTLD());
+        Assert.assertEquals("l4lb.thisdcos.directory", schedulerConfig.getVipTLD());
+        Assert.assertEquals("marathon", schedulerConfig.getMarathonName());
+
+        confMap.put("SERVICE_TLD", "test.autoip.tld");
```
elezar: Nit: Should be two tests.

sdk/scheduler/src/main/java/com/mesosphere/sdk/http/EndpointUtils.java
nickbp: Looks like the dependency chain is:
```
toAutoIpDomain
    toAutoIpHostname
        toSchedulerAutoIpHostname
        toAutoIpEndpoint
            toSchedulerAutoIpEndpoint
```

PR #2593, opened 2018-07-19
Add a new yml to utilize the ports

takirala:


frameworks/helloworld/tests/test_multiple_ports.py
```diff
@@ -0,0 +1,29 @@
...
+        yield  # let the test session execute
+    finally:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.mark.sanity
+def test_install():
+    sdk_install.install(config.PACKAGE_NAME,
+                        config.SERVICE_NAME,
+                        0,
```
elezar: Is the number of tasks not equal to `{{HELLO_COUNT}}` on https://github.com/mesosphere/dcos-commons/pull/2593/files#diff-4119b477652218ce29e7fb630f9c13b0R7?

frameworks/helloworld/tests/test_multiple_ports.py
```diff
@@ -0,0 +1,29 @@
...
+                        additional_options={"service": {"yaml": "multiport"}})
+    sdk_plan.wait_for_completed_deployment(config.SERVICE_NAME)
```
elezar: `sdk_install.install` aready waits for completed deployment.

frameworks/helloworld/universe/config.json
```diff
@@ -165,6 +166,21 @@
           "description": "The number of seconds of grace to await a clean shutdown following SIGTERM before sending SIGKILL, default: `0`",
           "type": "integer",
           "default": 0
+        },
+        "port_one": {
+          "description": "any arbitrary port can be used",
+          "type": "integer",
+          "default": 1729
```
elezar: Would it not be better to hardcode this in the YAML itself? Alternatively, we could have one hard-coded port, and one auto port and expose this using an `env_key`.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
```diff
@@ -122,15 +122,15 @@ private static DefaultServiceSpec convertServiceSpec(
     private static void verifyDistinctDiscoveryPrefixes(Collection<RawPod> rawPods) {
         Map<String, Long> dnsPrefixCounts = rawPods.stream()
                 .flatMap(p -> p.getTasks().values().stream()
-                        .map(t -> t.getDiscovery())
-                        .filter(d -> d != null)
-                        .map(d -> d.getPrefix())
-                        .filter(prefix -> prefix != null)
+                        .map(RawTask::getDiscovery)
```
elezar: Although this is a good change, keep in mind that this would trigger the entire CI and may be better suited to a follow-up PR. I'm happy to leave it in though.

frameworks/helloworld/src/main/dist/multiport.yml
```diff
@@ -0,0 +1,65 @@
...
+            port: {{HELLO_PORT_TWO}}
+          port_three:
+            port: {{HELLO_PORT_THREE}}
+    tasks:
+      server:
+        goal: ONCE
+        resource-set: multi-port-resources
+        cmd: |
+          networktool=netcat
+          if hash ncat 2>/dev/null; then
```
elezar: What is the advantage of using `hash` over, say `which`?

frameworks/helloworld/src/main/dist/multiport.yml
```diff
@@ -0,0 +1,65 @@
...
+  principal: {{FRAMEWORK_PRINCIPAL}}
+  user: {{FRAMEWORK_USER}}
+pods:
+  multiport:
+    count: {{HELLO_COUNT}}
+    resource-sets:
+      multi-port-resources:
+        ports:
+          port_one:
+            port: {{HELLO_PORT_ONE}}
```
elezar: See my comment on simply hard-coding the ports here, leaving them to be automatically assigned, or both.

frameworks/helloworld/src/main/dist/multiport.yml
```diff
@@ -0,0 +1,65 @@
...
+          echo "Using networking tool : ${networktool}"
+
+          ${networktool} -l {{HELLO_PORT_ONE}} &
+          pid_1=$!
+          ${networktool} -l {{HELLO_PORT_TWO}} &
+          pid_2=$!
+          ${networktool} -l {{HELLO_PORT_THREE}} &
+          pid_3=$!
+
+          $(${networktool} -v multiport-0-server.$FRAMEWORK_HOST {{HELLO_PORT_ONE}})
```
elezar: Is this intended to only communicate with itself or between pods?

frameworks/helloworld/tests/test_multiple_ports.py
```diff
@@ -0,0 +1,29 @@
...
+def configure_package(configure_security):
+    try:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+        yield  # let the test session execute
+    finally:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.mark.sanity
+def test_install():
```
elezar: Let's give this test a more meaningful name. It will make things clearer if we ever add additional tests.

frameworks/helloworld/src/main/dist/multiport.yml
```diff
@@ -0,0 +1,65 @@
...
+  multiport:
+    count: {{HELLO_COUNT}}
+    resource-sets:
+      multi-port-resources:
+        ports:
+          port_one:
+            port: {{HELLO_PORT_ONE}}
+          port_two:
+            port: {{HELLO_PORT_TWO}}
+          port_three:
```
elezar: Does one need three ports, or are two sufficient?

frameworks/helloworld/src/main/dist/multiport.yml
```diff
@@ -0,0 +1,65 @@
...
+          ${networktool} -l {{HELLO_PORT_ONE}} &
+          pid_1=$!
+          ${networktool} -l {{HELLO_PORT_TWO}} &
+          pid_2=$!
+          ${networktool} -l {{HELLO_PORT_THREE}} &
+          pid_3=$!
+
+          $(${networktool} -v multiport-0-server.$FRAMEWORK_HOST {{HELLO_PORT_ONE}})
+          exit_1=$?
+          $(${networktool} -v multiport-0-server.$FRAMEWORK_HOST {{HELLO_PORT_TWO}})
```
elezar: We could use `$TASK_NAME` instead of `multiport-0-server` here which would make it work for multiple pods too.

frameworks/helloworld/tests/test_multiple_ports.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+import logging
+from tests import config
```
elezar: Convention in other files is to have local from imports after the `sdk_` imports. If you're using tooling to sort the imports, then please let me know so that we can look at how to integrate this into our workflow.

frameworks/helloworld/tests/test_multiple_ports.py
takirala: The task is not a long running task, by the time our tests see the TASK_RUNNING, it has already moved to TASK_FINISHED and that is why I used 0.

frameworks/helloworld/universe/config.json
takirala: I would say lets keep them in config.json so that we can control if we need to play with some ports in the future. I see no harm in doing that. 

otoh, how do i get them to be automatically assigned ? i do not know about "auto port"

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/yaml/YAMLToInternalMappers.java
takirala: Assuming we control the trigger vs non-trigger of the CI on teamcity VCS triggers based on file extensions, since this CI already has some diff in *.py files, it would already trigger the entire CI right? Or am I missing something?

frameworks/helloworld/src/main/dist/multiport.yml
takirala: Two should be enough. The PR is still a `wip` and i was testing something else.

frameworks/helloworld/tests/test_multiple_ports.py
takirala: I did not use any tooling, I will use the ^ convention from now on.

takirala:
Thanks for the comments @elezar 
I have made one to be a dynamic port and the other one static so that we would be testing multiple code paths.

PR #2596, opened 2018-07-20
[DCOS-37995] bump snapshot version to 0.42.3 snapshot

elezar:
Part of the 0.42.2 release.


gradle.properties
```diff
@@ -1,2 +1,2 @@
 org.gradle.daemon=false
-org.gradle.parallel=true
+org.gradle.parallel=false
```
elezar: This is to try and get the SDK builds on this branch a bit more stable.

elezar: It didn't work. Hence: 092971c

PR #2597, opened 2018-07-21
Enable black formatting checks in code checks

elezar:
This PR updates the code checks as follows:

* Adds black formatting checks (and adds the relevant flake8 config)
* Moves the checks to pre-commit instead of running these in individual docker steps for better portability.

TODO:
- [x] Add docs on adding the pre-commit hooks
- [x] Add docs on running black


.flake8
```diff
@@ -1,3 +1,5 @@
 [flake8]
-max-line-length = 120
-ignore = E501
+ignore = E203, E266, E501, W503
```
nickbp: FWIW it looks like this file supports `# comments`. Could specify what the `ignored` and `select` values mean

See examples in http://flake8.pycqa.org/en/latest/user/configuration.html

elezar: Thanks. Will do.

elezar: 6360452

.flake8
```diff
@@ -1,3 +1,12 @@
 [flake8]
-max-line-length = 120
-ignore = E501
+# The following are ignored:
+#   E203 whitespace before ':' (https://github.com/ambv/black#line-breaks--binary-operators)
+#   E501 line too long         (https://github.com/ambv/black#line-length)
+#   W503 line break before binary operator
+ignore = E203, E501, W503
+max-line-length = 88
```
porridge: Noooo... could we set black to make it at least 100 please? (Personally I'd even vote for keeping 120, but the rationale seems https://github.com/ambv/black#line-length kind-of convinces me...)

.flake8
```diff
@@ -1,3 +1,12 @@
 [flake8]
-max-line-length = 120
-ignore = E501
+# The following are ignored:
+#   E203 whitespace before ':' (https://github.com/ambv/black#line-breaks--binary-operators)
+#   E501 line too long         (https://github.com/ambv/black#line-length)
```
porridge: What's the point of both setting the line length below *and* ignoring the warnings?

.flake8
```diff
@@ -1,3 +1,12 @@
 [flake8]
-max-line-length = 120
-ignore = E501
+# The following are ignored:
```
porridge: It's not obvious right away why are these settings in `[flake8]` section but the links in comments point to `black`. Perhaps add ` to work better with "black"` before the colon.

.flake8
```diff
@@ -1,3 +1,12 @@
...
+#   E203 whitespace before ':' (https://github.com/ambv/black#line-breaks--binary-operators)
```
porridge: Change the link to https://github.com/ambv/black#slices

.flake8
```diff
@@ -1,3 +1,12 @@
...
+#   E501 line too long         (https://github.com/ambv/black#line-length)
+#   W503 line break before binary operator
```
porridge: Link to https://github.com/ambv/black#line-breaks--binary-operators

tools/ci/checks/run_black_checks.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+DOCKER_TAG=${DOCKER_TAG:-latest}
+DOCKER_IMAGE=${DOCKER_IMAGE:-mesosphere/dcos-commons:${DOCKER_TAG}}
+
+if [ x"$1" == x"--format" ]; then
```
porridge: Please drop the x, or drop the quotes and use ``[[ ]]` better yet.

.flake8
elezar: Hmmmm. Black doesn't always change the line length. Some examples I've seen is long doc-strings and string parameters:
```
✗ black testing; flake8 testing | wc -l
All done! ✨ 🍰 ✨
22 files left unchanged.
     157
```

tools/ci/checks/run_black_checks.sh
elezar: I will drop the file.

.flake8
elezar: Ok. I was going with the defaults for black, but 100 is a good compromise.

elezar:
Thanks @nickbp and @porridge.

nickbp:
FWIW unless anyone else has comments I think this is ready to go. The cassandra failure is due to a known regression in metronome and shouldn't block merging.

elezar:
> FWIW unless anyone else has comments I think this is ready to go. The cassandra failure is due to a known regression in metronome and shouldn't block merging.

Thanks. I've just updated with Master and will merge as soon as all except Cassandra are green.

elezar:
I'm merging this. As discussed, the Cassandra issues are known, and there are no changes in this PR that would affect any framework.

PR #2602, opened 2018-07-26
DCOS-29474 Detect leftover frameworks after uninstalls

nickbp:
We're occasionally seeing flakes in `test_srv_records` which were traced down to frameworks not being correctly unregistered. After uninstall, validate that the framework was correctly unregistered, and throw an exception (triggering log collection) if the check fails.

Other misc fixes:
- Also throw if dangling resources are found. We were previously logging them but not throwing, so it's not clear if this is actually a thing. We should check this though, as any issues here imply an uninstall bug.
- Fix error produced by linux `mktemp` when running `test.sh`, and prettify the timestamp used: `mktemp: too few X's in template`
- Misc python lint fixes, and cleanup of lightly-used functions in `sdk_utils`.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -10,11 +10,11 @@
 # Exit immediately on errors
 set -e
 
-timestamp="$(date +%d%m%y%H%M%s)"
+timestamp="$(date +%y%m%d-%H%M%S)"
 # Create a temp file for docker env.
 # When the script exits (successfully or otherwise), clean up the file automatically.
-credsfile="$(mktemp /tmp/sdk-test-creds-${timestamp}.tmp)"
-envfile="$(mktemp /tmp/sdk-test-env-${timestamp}.tmp)"
+credsfile="$(mktemp /tmp/sdk-test-creds-${timestamp}-XXXX.tmp)"
+envfile="$(mktemp /tmp/sdk-test-env-${timestamp}-XXXX.tmp)"
 function cleanup {
     rm -f ${credsfile}
     rm -f ${envfile}
--- a/testing/sdk_cmd.py
+++ b/testing/sdk_cmd.py
@@ -336,7 +336,7 @@ def _task_exec(task_id_prefix: str, cmd: str) -> tuple:
     return run_raw_cli("task exec {} {}".format(task_id_prefix, cmd))
 
 
-def resolve_hosts(marathon_task_name: str, hosts: list, bootstrap_cmd: str='./bootstrap') -> bool:
+def resolve_hosts(marathon_task_name: str, hosts: list, bootstrap_cmd: str = './bootstrap') -> bool:
     """
     Use bootstrap to resolve the specified list of hosts
     """
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -29,6 +29,10 @@ TIMEOUT_SECONDS = 15 * 60
 Used by post-test diagnostics to retrieve stuff from currently running services.'''
 _installed_service_names = set([])
 
+'''List of dead agents which should be ignored when checking for orphaned resources.
+Used by uninstall when validating that an uninstall completed successfully.'''
+_dead_agent_hosts = set([])
+
 
 def get_installed_service_names() -> set:
     '''Returns the a set of service names which had been installed via sdk_install in this session.'''
@@ -159,20 +163,33 @@ def _verify_completed_uninstall(service_name):
 
     # There should be no orphaned resources in the state summary (DCOS-30314)
     orphaned_resources = 0
+    ignored_orphaned_resources = 0
     service_role = sdk_utils.get_role(service_name)
     for agent in state_summary['slaves']:
         # resources should be grouped by role. check for any resources in our expected role:
         matching_reserved_resources = agent['reserved_resources'].get(service_role)
         if matching_reserved_resources:
-            log.error('Orphaned resources on agent {}/{}: {}'.format(
-                agent['id'], agent['hostname'], matching_reserved_resources))
-            orphaned_resources += len(matching_reserved_resources)
+            if agent['hostname'] in _dead_agent_hosts:
+                # The test told us ahead of time to expect orphaned resources on this host.
+                log.info('Ignoring orphaned resources on agent {}/{}: {}'.format(
+                    agent['id'], agent['hostname'], matching_reserved_resources))
+                ignored_orphaned_resources += len(matching_reserved_resources)
+            else:
+                log.error('Orphaned resources on agent {}/{}: {}'.format(
+                    agent['id'], agent['hostname'], matching_reserved_resources))
+                orphaned_resources += len(matching_reserved_resources)
     if orphaned_resources:
-        log.error('{} orphaned resources after uninstall of {}'.format(orphaned_resources, service_name))
+        log.error('{} orphaned resources (plus {} ignored) after uninstall of {}'.format(
+            orphaned_resources, ignored_orphaned_resources, service_name))
         log.error(state_summary)
-        raise Exception('Found {} orphaned resources after uninstall of {}'.format(
-            orphaned_resources, service_name))
-    log.info('No orphaned resources for role {} were found'.format(service_role))
+        raise Exception('Found {} orphaned resources (plus {} ignored) after uninstall of {}'.format(
+            orphaned_resources, ignored_orphaned_resources, service_name))
+    elif ignored_orphaned_resources:
+        log.info('Ignoring {} orphaned resources after uninstall of {}'.format(
+            ignored_orphaned_resources, service_name))
+        log.info(state_summary)
+    else:
+        log.info('No orphaned resources for role {} were found'.format(service_role))
 
     # There should be no framework entry for this service in the state summary (DCOS-29474)
     orphaned_frameworks = [fwk for fwk in state_summary['frameworks'] if fwk['name'] == service_name]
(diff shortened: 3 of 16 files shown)


testing/sdk_cmd.py
```diff
@@ -336,7 +336,7 @@ def _task_exec(task_id_prefix: str, cmd: str) -> tuple:
     return run_raw_cli("task exec {} {}".format(task_id_prefix, cmd))
 
 
-def resolve_hosts(marathon_task_name: str, hosts: list, bootstrap_cmd: str='./bootstrap') -> bool:
+def resolve_hosts(marathon_task_name: str, hosts: list, bootstrap_cmd: str = './bootstrap') -> bool:
```
elezar: I always thought that spaces around `kwarg` defaults were not recommended?

test.sh
```diff
@@ -10,11 +10,11 @@
 # Exit immediately on errors
 set -e
 
-timestamp="$(date +%d%m%y%H%M%s)"
+timestamp="$(date +%y%m%d-%H%M%S)"
 # Create a temp file for docker env.
 # When the script exits (successfully or otherwise), clean up the file automatically.
-credsfile="$(mktemp /tmp/sdk-test-creds-${timestamp}.tmp)"
-envfile="$(mktemp /tmp/sdk-test-env-${timestamp}.tmp)"
+credsfile="$(mktemp /tmp/sdk-test-creds-${timestamp}-XXXX.tmp)"
```
elezar: Just for my info, what is the motivation for the `XXXX`?

testing/sdk_install.py
```diff
@@ -176,41 +175,29 @@ def uninstall(
...
-
     log.info('Uninstalling {}'.format(service_name))
 
     try:
         retried_uninstall_package_and_wait(package_name, service_name=service_name)
     except Exception:
-        log.info('Got exception when uninstalling {}'.format(service_name))
-        log.info(traceback.format_exc())
+        log.exception('Got exception when uninstalling {}'.format(service_name))
         raise
```
elezar: Does this raise the last thrown exception? Would it make sense to change this to:
```
except Exception as e
    ... 
    raise e
```

testing/sdk_install.py
```diff
@@ -221,6 +208,37 @@ def uninstall(
             shakedown.pretty_duration(finish - cleanup_start),
             shakedown.pretty_duration(finish - start)))
 
+    state_summary = sdk_cmd.cluster_request('GET', '/mesos/state-summary').json()
+
+    # Sanity check: There should be no dangling resources in the state summary (DCOS-30314)
```
elezar: Consider moving this to a function? (or functions)

testing/sdk_install.py
```diff
@@ -176,41 +175,29 @@ def uninstall(
...
         else:
+            # 1.10 and later: Wait for uninstall scheduler to finish and be removed by Cosmos
             log.info('Waiting for Marathon app to be removed {}'.format(service_name))
             sdk_marathon.retried_wait_for_deployment_and_app_removal(
                 sdk_marathon.get_app_id(service_name), timeout=TIMEOUT_SECONDS)
     except Exception:
-        log.info('Got exception when cleaning up {}'.format(service_name))
-        log.info(traceback.format_exc())
+        log.exception('Got exception when cleaning up {}'.format(service_name))
         raise
```
elezar: Same comment as above? Also, the removal of the `finally` blocks in both cases changes behaviour subtly in that the resource list is *not* printed if there was an exception.

testing/sdk_install.py
```diff
@@ -221,6 +208,37 @@ def uninstall(
...
+    if found_dangling_resources:
+        log.error(state_summary)
+        raise dcos.errors.DCOSException('Found dangling resources after uninstall of {}'.format(service_name))
+    log.info('No dangling resources for role {} were found'.format(service_role))
+
+    # Sanity check: There should be no framework entry for this service in the state summary (DCOS-29474)
+    dangling_frameworks = [fwk for fwk in state_summary['frameworks'] if fwk['name'] == service_name]
+    if len(dangling_frameworks) > 0:
+        log.error(state_summary)
+        raise dcos.errors.DCOSException('Found dangling frameworks after uninstall of {}'.format(service_name))
```
elezar: Any reason for using a `DCOSException` here? Is it due to the retries around the uninstall functions? Remember that the `dcos` python package is being deprecated, and as such as regular Exception may be better here.

testing/sdk_install.py
```diff
@@ -221,6 +208,37 @@ def uninstall(
...
 
+    state_summary = sdk_cmd.cluster_request('GET', '/mesos/state-summary').json()
+
+    # Sanity check: There should be no dangling resources in the state summary (DCOS-30314)
+    found_dangling_resources = False
+    service_role = '{}-role'.format(service_name)
+    for agent in state_summary['slaves']:
+        # resources should be grouped by role. check for any resources in our expected role:
+        matching_reserved_resources = agent['reserved_resources'].get(service_role, {})
+        if len(matching_reserved_resources) > 0:
```
elezar: We could just check `if matching_reserved_resources:` (it's possibly a little more pythonic). Then, we could also change the line above to `matching_reserved_resources = agent['reserved_resources'].get(service_role)` which may be a small optimization.

testing/sdk_install.py
```diff
@@ -221,6 +208,37 @@ def uninstall(
...
+            log.error('Dangling resources on agent {}/{}: {}'.format(
+                agent['id'], agent['hostname'], matching_reserved_resources))
+            found_dangling_resources = True
+    if found_dangling_resources:
```
elezar: Would it be useful to have a count?

testing/sdk_install.py
```diff
@@ -221,6 +208,37 @@ def uninstall(
             shakedown.pretty_duration(finish - cleanup_start),
             shakedown.pretty_duration(finish - start)))
 
+    state_summary = sdk_cmd.cluster_request('GET', '/mesos/state-summary').json()
+
+    # Sanity check: There should be no dangling resources in the state summary (DCOS-30314)
+    found_dangling_resources = False
+    service_role = '{}-role'.format(service_name)
```
elezar: This doesn't work for foldered service names.

testing/sdk_install.py
```diff
@@ -221,6 +208,37 @@ def uninstall(
...
+            found_dangling_resources = True
+    if found_dangling_resources:
+        log.error(state_summary)
+        raise dcos.errors.DCOSException('Found dangling resources after uninstall of {}'.format(service_name))
+    log.info('No dangling resources for role {} were found'.format(service_role))
+
+    # Sanity check: There should be no framework entry for this service in the state summary (DCOS-29474)
+    dangling_frameworks = [fwk for fwk in state_summary['frameworks'] if fwk['name'] == service_name]
+    if len(dangling_frameworks) > 0:
+        log.error(state_summary)
```
elezar: Would it make sense to log the dangling frameworks found here? Also, we could just check `if dangling_frameworks:`

testing/sdk_cmd.py
nickbp: I was getting `flake8` errors on this: `missing whitespace around parameter equals [E252]`

test.sh
nickbp: `mktemp` uses it to insert a random token into the string:
```
$ mktemp foo-XXXX.tmp
foo-SXIq.tmp
```

testing/sdk_install.py
nickbp: It rethrows the original exception without messing with the stack of the original: https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement, https://stackoverflow.com/questions/18001721/raise-with-no-argument

testing/sdk_install.py
nickbp: The previous `list_reserved_resources()` functionality was moved to the following check, where it now throws if there are leftover resources (implies an uninstall bug).

If an exception is thrown here, there isn't much value in checking for leftover resources -- it's expected that there would be some if the uninstall trigger itself failed.

testing/sdk_install.py
nickbp: Done

testing/sdk_install.py
nickbp: `sdk_install` was already doing this correctly elsewhere, "upstreamed" that into `sdk_utils` and switched users to calling that.

testing/sdk_install.py
nickbp: Done

testing/sdk_install.py
nickbp: Sounds good, switched to a count and included it in the error log

testing/sdk_install.py
nickbp: Done

testing/sdk_install.py
nickbp: Switched to `Exception` throughout. I think it was just to align with `_retried_install_impl()`, which calls into shakedown a ton and thereby could be returning `DCOSException`s in a lot of cases.

test.sh
nickbp: Both are useful -- sometimes you can end up with a left over file if e.g. the build is aborted, so it doesn't hurt to include an origin timestamp in there so that old ones are apparent when listing the directory.

testing/sdk_install.py
```diff
@@ -189,7 +204,7 @@ def uninstall(
         if sdk_utils.dcos_version_less_than('1.10'):
             # 1.9 and earlier: Run janitor to unreserve resources
             log.info('Janitoring {}'.format(service_name))
-            retried_run_janitor(service_name, role, service_account, zk)
+            _retried_run_janitor(service_name)
```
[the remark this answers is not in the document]
nickbp: I checked and didn't find any. I don't remember any historical cases either. I think this was a "just in case" that never came to pass.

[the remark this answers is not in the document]
elezar: (Was just looking for the YAGNI emoji!)

nickbp:
The detection works!

test_zzzrecovery.test_shutdown_host_teardown:
```
E           Exception: Found 4 orphaned resources after uninstall of hello-world
```

Will add a way to not fail the test when remaining resources are expected. Leftover resources are expected in the mesos master's accounting when an agent has recently died.

PR #2611, opened 2018-08-01
DCOS-39979 Fix test_mono_to_multi_upgrade failures in 1.10 Open Cluster

takirala:
Address [DCOS-39979](https://jira.mesosphere.com/browse/DCOS-39979)

### DIFF
--- a/testing/sdk_upgrade.py
+++ b/testing/sdk_upgrade.py
@@ -171,8 +171,8 @@ def update_or_upgrade_or_downgrade(
 ):
     initial_config = get_config(package_name, service_name)
     task_ids = sdk_tasks.get_task_ids(service_name, "")
-    if (to_package_version and not sdk_utils.is_cli_supports_service_version_upgrade()) or (
-        additional_options and not sdk_utils.is_cli_supports_service_options_update()
+    if (to_package_version and not is_cli_supports_service_version_upgrade()) or (
+        additional_options and not is_cli_supports_service_options_update()
     ):
         log.info(
             "Using marathon flow to upgrade to package name : {} version : [{}]".format(
@@ -206,10 +206,10 @@ def _update_service_with_cli(
     )
     update_cmd = ["update", "start"]
     if to_package_version:
-        sdk_utils.ensure_cli_supports_service_version_upgrade()
+        ensure_cli_supports_service_version_upgrade()
         update_cmd.append("--package-version={}".format(to_package_version))
     if additional_options:
-        sdk_utils.ensure_cli_supports_service_options_update()
+        ensure_cli_supports_service_options_update()
         options_file = tempfile.NamedTemporaryFile("w")
         json.dump(additional_options, options_file)
         options_file.flush()  # ensure json content is available for the CLI to read below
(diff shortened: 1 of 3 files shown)


testing/sdk_upgrade.py
```diff
@@ -29,7 +29,7 @@
 def test_upgrade(
         package_name,
         service_name,
-        running_task_count,
+        expected_running_tasks,
```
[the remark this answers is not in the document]
elezar: (I see that it is also used in `update_service`, but it is not a `kwarg` and as such changing the name there would be much simpler).

testing/sdk_upgrade.py
```diff
@@ -147,49 +147,61 @@ def get_config(package_name, service_name):
...
-        service_name,
-        running_task_count,
-        additional_options,
-        timeout_seconds,
-        wait_for_deployment):
-
-    initial_config = get_config(package_name, service_name)
-    task_ids = sdk_tasks.get_task_ids(service_name, '')
-
+def update_service(package_name,
```
elezar: We have made this interface more complex by adding options that are not used at all.

I would:
1. Remove the 1.9 cluster specifics from here.
2. Remove the additional options from the `update_service` function.
3. Still call `update_service` from `_upgrade_or_downgrade`
4. Add a check with a visible error message when `update_service` is called on a 1.9 cluster.

We could drop the check for `1.9` clusters if we set `expected_running_tasks=0` and `wait_for_deployment=False` when calling `sdk_install.install` and then perform the equivalent checks in `_upgrade_or_downgrade`.

takirala: The only place this method is used needs it work for 1.9 as well. If i remove that then this method would be unused. I can remove this method and increase the visibility of `upgrade_or_downgrade`.

testing/sdk_upgrade.py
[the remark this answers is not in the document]
takirala: The same name is used in `sdk.install(...` and i wanted to keep things consistent. Also `expected_running_tasks` sounds like a list and `running_task_count` sounds like an integer which we want.

testing/sdk_upgrade.py
elezar: For clarification, the issue was not that this was not running correctly against 1.9, it was that this was not running correctly against **1.10 Open**.

testing/sdk_utils.py
```diff
@@ -88,37 +89,57 @@ def dcos_version_at_least(version):
...
 
 
 def is_strict_mode():
-    '''Determine if the tests are being run on a strict mode cluster.'''
-    return os.environ.get('SECURITY', '') == 'strict'
+    """Determine if the tests are being run on a strict mode cluster."""
+    return os.environ.get("SECURITY", "") == "strict"
+
+
+def is_cli_supports_service_version_upgrade():
```
nickbp: TBH I'd just put these checks for CLI support in `sdk_upgrade.py` for now. They're specific to the upgrade/update operation itself, and ideally nobody outside of `sdk_upgrade` should need to know about the different variants, they'd just be handled by `sdk_upgrade` internally.

testing/sdk_upgrade.py
```diff
@@ -89,131 +98,158 @@ def test_upgrade(
...
+    package_name, service_name, to_package_version=None, additional_options=None
+):
+    log.info(
+        "Using CLI to update to package name : {} version : [{}]".format(
+            package_name, to_package_version
+        )
+    )
+    update_cmd = ["update", "start"]
+    if to_package_version:
+        sdk_utils.ensure_cli_supports_service_version_upgrade()
```
nickbp: Would failing this check effectively imply a bug in `sdk_upgrade`, or should the caller have done something differently? If it's an internal error I think it'd make sense to mention something like "INTERNAL ERROR: ..." in the assert message (once that assert is moved here from `sdk_utils` per comment there)

(Same question for L212)

takirala: The caller should have done something differently. Since we have `is_cli_supports_*` methods that return boolean, the caller can use that to decide what to do "differently". The assertion error messages are to let the caller know what needs to be done differently.

testing/sdk_utils.py
```diff
@@ -18,6 +18,7 @@
 
 log = logging.getLogger(__name__)
 
+TIMEOUT_SECONDS = 15 * 60
```
nickbp: This change could be reverted

takirala: I am using it here https://github.com/mesosphere/dcos-commons/pull/2611/files#diff-6c805b3ab91ef126953b020997763e91R170 ; we can probably use it in someother classes as well in future, so just going to keep it `sdk_utils`.

PR #2613, opened 2018-08-03
Update kafka base tech to 1.1.1

Gupta-Amrit:


Gupta-Amrit:
kindly provide the kafka binary src link .. as currently I am using s3 bucket

PR #2614, opened 2018-08-03
DCOS-40149 Fix compatibility with gracePeriod field

nickbp:
SDK 0.52.0 switched to `grace-period`, but didn't include compatibility with the prior `gracePeriod`.

Add dual-read and dual-write support to `DefaultHealthCheckSpec` to support upgrade and downgrade compatibility across this divide. After a few months (and a few SDK releases), the `gracePeriod` handling can then be removed.

### DIFF
--- a/sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultHealthCheckSpecTest.java
+++ b/sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultHealthCheckSpecTest.java
@@ -1,7 +1,11 @@
 package com.mesosphere.sdk.specification;
 
+import java.io.IOException;
+
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultHealthCheckSpecTest.java
```diff
@@ -30,4 +34,94 @@ public void invalidHealthCheckSpec() {
                 .timeout(-1)
                 .build();
     }
+
+    @Test
+    public void upgradeFromOldHealthCheckSpec() throws IOException {
+        // Old version with just gracePeriod
+        String oldHealthCheckSpecJson = "{\n" +
```
takirala: Java doesn't have multiline strings :(

PR #2615, opened 2018-08-06
DCOS-40251 [TOOLS] Reformat python files according to new linter rules

takirala:
Recent changes to tooling will result in multiple changes to files. Instead of spending some time on formatting for every PR in near future (until all files are touched) I decided to format the entire repository in one PR so that we don't have to make any unexpected changes in the future PRs.

FYI @jeid64 @elezar 

This is what I have done : 

```
# Renamed [flake8] to [pep8] in the first line
autopep8  --global-config=.flake8 -r --in-place . 

# For black
black frameworks
black testing
black tools
black *.py
```

- [x] TODO : Fix the pylint [failing teamcity CI](https://teamcity.mesosphere.io/viewLog.html?buildId=1154080&buildTypeId=DcosIo_DcosCommons_PrJobs_RunCodeChecks&tab=buildLog&_focus=501#_state=348) : Unable to figure out a good fix for this. Disabled `invalid-sequence-index`, `no-name-in-module`, and `no-member` checks temporarily. Also the more ideal fix may need some code changes and this PR is very long for having any non-trivial code changes.


takirala:
@elezar i can hold off this until we are completely done with the linting changes. 

Ping @nickbp @kvish @jeid64 Lets use this PR to discuss any styling concerns.

takirala:
I have added `--disable=invalid-sequence-index, --disable=no-member, --disable=no-name-in-module` to pylint args to disable the checks temporarily. We can :
 - Figure out a way to make pylint understand they are all valid use cases and not errors (this has not been trivial) OR
 - Ignore them per occurrence (inline) instead of disabling globally.

But for now, I think it is fair to say that we can merge this PR so that the linting process would be much simpler.

takirala:
This PR is not being squashed. Instead, I made three commits - one for each of the linter changes (black, flake8, and pylint) thus making it easy if we ever need to revert each linter changes independently.

PR #2617, opened 2018-08-07
[DCOS-40189] Fix HDFS test: test_kill_scheduler

samvantran:
Resolves [DCOS-40189](https://jira.mesosphere.com/browse/DCOS-40189)

We ran into a CI failure where test `test_kill_scheduler` "passed" prematurely and resulted in the following test `kill_all_journalnodes` to fail due to the scheduler still being restarted.

This PR adds a check for the scheduler to be restarted successfully.


frameworks/hdfs/tests/test_sanity.py
```diff
@@ -135,9 +135,13 @@ def test_kill_data_node():
...
 @pytest.mark.recovery
 def test_kill_scheduler():
-    sdk_cmd.kill_task_with_pattern('hdfs.scheduler.Main', shakedown.get_service_ips('marathon').pop())
-    config.check_healthy(service_name=sdk_utils.get_foldered_name(config.SERVICE_NAME))
+    foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
+    task_id_service_name = sdk_utils.get_task_id_service_name(foldered_name)
+    old_scheduler_id = sdk_tasks.get_task_ids(foldered_name, task_id_service_name)[0]
 
+    sdk_cmd.kill_task_with_pattern('hdfs.scheduler.Main', shakedown.get_service_ips('marathon').pop())
+    config.expect_recovery(service_name=foldered_name)
```
elezar: Killing the scheduler shouldn't trigger a recovery as recovery implies that something is being done on the nodes themselves.

When killing the scheduler, we would expect:
* The scheduler task is restarted
* The other tasks are **not** restarted
* There is no in-progress deployment or recovery

testing/sdk_tasks.py
```diff
@@ -175,7 +175,27 @@ def get_completed_task_id(task_name):
...
+        try:
+            task_ids = set([t['id'] for t in shakedown.get_tasks(completed=False) if t['name'] == service_name])
+            log.info('found the following task ids {}'.format(task_ids))
+        except dcos.errors.DCOSHTTPException:
+            log.info('Failed to get task ids. service_name=%s', service_name)
+            task_ids = set([])
 
+        return len(task_ids) > 0 and (old_scheduler_task_id not in task_ids or len(task_ids) > 1)
+
+    fn()
```
elezar: Newlines?

testing/sdk_tasks.py
```diff
@@ -175,7 +175,27 @@ def get_completed_task_id(task_name):
 
     return tasks[0] if tasks else None
 
+def check_scheduler_relaunched(service_name: str, old_scheduler_task_id: str,
```
elezar: Since the PR that changed the way that `check_task_relaunched` is implemented, this function has become out of date. Would it be possible to reuse `check_task_relaunched` in some way instead?

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -148,10 +148,21 @@ def test_kill_data_node():
...
+    scheduler_id = sdk_tasks.get_task_ids(foldered_name, scheduler_task_name)
+    journal_ids = sdk_tasks.get_task_ids(foldered_name, "journal")
+    name_ids = sdk_tasks.get_task_ids(foldered_name, "name")
+    data_ids = sdk_tasks.get_task_ids(foldered_name, "data")
+
     sdk_cmd.kill_task_with_pattern(
         "hdfs.scheduler.Main", shakedown.get_service_ips("marathon").pop()
     )
-    config.check_healthy(service_name=sdk_utils.get_foldered_name(config.SERVICE_NAME))
+    config.check_healthy(service_name=foldered_name)
```
elezar: We can't call `check` healthy here. There is not scheduler running at this point. (Since you're out, I'll push a change)

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -148,10 +148,23 @@ def test_kill_data_node():
...
+    data_ids = sdk_tasks.get_task_ids(foldered_name, "data")
+
     sdk_cmd.kill_task_with_pattern(
         "hdfs.scheduler.Main", shakedown.get_service_ips("marathon").pop()
     )
-    config.check_healthy(service_name=sdk_utils.get_foldered_name(config.SERVICE_NAME))
+
+    sdk_marathon.wait_for_deployment_and_app_running(foldered_name, 15 * 60)
+
+    sdk_tasks.check_task_relaunched(scheduler_task_name, scheduler_id)
```
elezar: Looking at the logs we seem to be getting:
```
6|sdk_tasks-check_task_relaunched(210)|INFO]: Checking (task_name:test.integration.hdfs) (old_task_id:[]) with (ensure_new_task_not_completed:True) is relaunched
```
Where the old-task ID is `[]`. Not sure if this is causing things to fail.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -148,10 +148,23 @@ def test_kill_data_node():
 @pytest.mark.sanity
 @pytest.mark.recovery
 def test_kill_scheduler():
+    foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
+    scheduler_name = sdk_utils.get_scheduler_name(foldered_name)
+    scheduler_id = sdk_tasks.get_task_ids(foldered_name, scheduler_name)
```
elezar: Looking at the test output I see:
```
[2018-08-15 11:29:22,8
02|sdk_tasks-check_task_relaunched(210)|INFO]: Checking (task_name:hdfs.integration.test) (old_task_id:[]) with (ensure_new_task_not_completed:True) is relaunched
[2018-08-15 11:29:22,804|sdk_cmd-cluster_request(82)|INFO]: (HTTP GET) /mesos/tasks
[2018-08-15 11:29:22,874|sdk_cmd-fn(99)|INFO]: Got 200 for GET /mesos/tasks
[2018-08-15 11:29:22,879|sdk_cmd-cluster_request(82)|INFO]: (HTTP GET) /mesos/slaves
[2018-08-15 11:29:22,899|sdk_cmd-fn(99)|INFO]: Got 200 for GET /mesos/slaves
[2018-08-15 11:29:22,916|sdk_tasks-get_summary(149)|INFO]: Task summary (with_completed=True) (task_name=[hdfs.integration.test]):
- Task[name="hdfs.integration.test"	state=TASK_RUNNING	id=test_integration_hdfs.723eed03-a07e-11e8-8d87-0209f1235b3f	host=10.0.3.178	framework_id=74e0d766-a764-4b55-88a2-8d8a24e4c716-0001	agent=74e0d766-a764-4b55-88a2-8d8a24e4c716-S7]
- Task[name="hdfs.integration.test"	state=TASK_FAILED	id=test_integration_hdfs.213d89e2-a07c-11e8-8d87-0209f1235b3f	host=10.0.3.178	framework_id=74e0d766-a764-4b55-88a2-8d8a24e4c716-0001	agent=74e0d766-a764-4b55-88a2-8d8a24e4c716-S7]
- Task[name="hdfs.integration.test"	state=TASK_KILLED	id=test_integration_hdfs.015613f1-a07b-11e8-8d87-0209f1235b3f	host=10.0.3.235	framework_id=74e0d766-a764-4b55-88a2-8d8a24e4c716-0001	agent=74e0d766-a764-4b55-88a2-8d8a24e4c716-S9]
- Task[name="hdfs.integration.test"	state=TASK_KILLED	id=test_integration_hdfs.66228ccc-a078-11e8-8d87-0209f1235b3f	host=10.0.1.127	framework_id=74e0d766-a764-4b55-88a2-8d8a24e4c716-0001	agent=74e0d766-a764-4b55-88a2-8d8a24e4c716-S6]
- Task[name="hdfs.integration.test"	state=TASK_KILLED	id=test_integration_hdfs.9c4f7c6a-a076-11e8-8d87-0209f1235b3f	host=10.0.2.13	framework_id=74e0d766-a764-4b55-88a2-8d8a24e4c716-0001	agent=74e0d766-a764-4b55-88a2-8d8a24e4c716-S5]
```
Note the first line which contains: `(old_task_id:[])`. This seems to indicate that `scheduler_id` is not set as expected. Note, the Marathon output earlier in the log shows both the previous and the current task ID for the scheduler.

Looking at the code for `check_task_relaunched`, it seems as if `old_task_id` is supposed to be a string, and not a list that is being returned by `get_task_ids`.

So, we need:
* To return the valid scheduler ID (or just get it from the marathon app state) from `get_task_ids`
* Pass a **single** string to `check_task_relaunched `.

nickbp: May have ended up fixing this test via the shakedown removal PR? May be able to do something like this: https://github.com/mesosphere/dcos-commons/pull/2616/files#diff-a0ec6572c50515f49ce94fae88201609R146

testing/sdk_utils.py
```diff
@@ -48,6 +48,15 @@ def get_task_id_service_name(service_name):
     return service_name.lstrip("/").replace("/", ".")
 
 
+def get_scheduler_name(service_name):
```
samvantran: this will likely change when Nick merges #2616 but it's essentially the same thing as `get_scheduler_task_prefix`; for now this satisfies the ticket and fixing the broken test

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -150,21 +150,21 @@ def test_kill_data_node():
...
     )
 
-    sdk_marathon.wait_for_deployment_and_app_running(foldered_name, 15 * 60)
-
-    sdk_tasks.check_task_relaunched(scheduler_name, scheduler_id)
+    sdk_tasks.check_tasks_updated("marathon", scheduler_name, scheduler_id)
     sdk_tasks.check_tasks_not_updated(foldered_name, "journal", journal_ids)
     sdk_tasks.check_tasks_not_updated(foldered_name, "name", name_ids)
     sdk_tasks.check_tasks_not_updated(foldered_name, "data", data_ids)
+    config.check_healthy(service_name=foldered_name)
```
elezar: Would it not be better to `check_healthy` **before** we check that none of the tasks have completed. This will give an unexpected deployment or recovery time to completed (and change the tasks) before we check that they have not changed.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -148,10 +147,23 @@ def test_kill_data_node():
...
+        "./hdfs-scheduler/bin/hdfs",
+        sdk_marathon.get_scheduler_host(foldered_name),
     )
-    config.check_healthy(service_name=sdk_utils.get_foldered_name(config.SERVICE_NAME))
+
+    sdk_tasks.check_tasks_updated("marathon", scheduler_name, scheduler_id)
+    sdk_tasks.check_tasks_not_updated(foldered_name, "journal", journal_ids)
+    sdk_tasks.check_tasks_not_updated(foldered_name, "name", name_ids)
+    sdk_tasks.check_tasks_not_updated(foldered_name, "data", data_ids)
+    config.check_healthy(service_name=foldered_name)
```
elezar: Would it not be better to `check_healthy` **before** we check that none of the tasks have completed. This will give an unexpected deployment or recovery time to completed (and change the tasks) before we check that they have not changed.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -148,10 +147,23 @@ def test_kill_data_node():
 @pytest.mark.sanity
 @pytest.mark.recovery
 def test_kill_scheduler():
+    foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
+    scheduler_name = sdk_utils.get_scheduler_name(foldered_name)
+    scheduler_id = sdk_tasks.get_task_ids("marathon", scheduler_name)
```
elezar: Strictly speaking, `scheduler_id` is a list of tasks, and we should use `scheduler_ids` as a variable name. I was initially confused because I read . the call to `check_tasks_updated` as passing in a single string, which would have been an error.

testing/sdk_utils.py
elezar: Would you mind linking the location in @nickbp's PR where you expect the changes to be made.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -148,10 +147,23 @@ def test_kill_data_node():
...
+    journal_ids = sdk_tasks.get_task_ids(foldered_name, "journal")
```
nickbp: Rather than breaking this up into three checks for each of the three types of nodes, this could just do a single non-prefixed retrieval of all nodes in the service, followed by a single check_tasks_not_updated below.

I think each call to check_tasks_not_updated involves a 30s wait to verify that no change occurs, so the test should finish a bit faster if only one check was performed. It would ultimately be equivalent to the three granular checks anyway.

elezar: @nickbp remind me. The scheduler task doesn't show up in the list because it's a `marathon` task, correct?

nickbp: Yes

elezar: Let's go with a single check then. Otherwise this looks good. Sorry for the back and forth Sam.

testing/sdk_utils.py
samvantran: https://github.com/mesosphere/dcos-commons/pull/2616/files#diff-e912dd60c30385c2608f3a5bed229478R276

Nick edited the get_scheduler_host method to create `get_scheduler_task_prefix`. It's just a different name but the logic is identical

frameworks/hdfs/tests/test_sanity.py
samvantran: This was in Nick's PR so I copied it over but I'm not sure I follow you. `check_healthy` is just a bunch of gets against the deploy/recovery plan for a status `COMPLETE` and a check that the task is in state `TASK_RUNNING`. I don't think it changes any of the tasks?

To me it makes sense to check that the service is healthy after restarting the scheduler to verify that everything is still in working order?

frameworks/hdfs/tests/test_sanity.py
samvantran: For awhile I was wondering why `get_task_ids(foldered_name, scheduler_name)` wasn't returning an ID even though I saw it in the UI. Now it makes sense since it's a `marathon` task.

Anyway, I'm +1 for making it a single check.

PR #2621, opened 2018-08-09
Add .pre-commit-config.yaml to Docker image

elezar:
This adds `.pre-commit-config.yaml` to the Docker image to allow the same checks to be easily used in other projects.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,3 +9,4 @@
 !test.sh
 !TESTING.md
 !/tools/**/*.py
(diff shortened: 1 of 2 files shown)


elezar:
Thanks for the question @nickbp 
> Just in case -- I assume this doesn't immediately take effect for other projects, that they need to enable it in their build tooling?

No, this is all opt-in. This just ensures that the file is available in the non-mono-repo project when updating the tooling.

PR #2628, opened 2018-08-15
[DCOS-40457] Improve install failure detection

mockingbird2:


testing/sdk_metrics.py
```diff
@@ -24,6 +25,22 @@ def get_scheduler_metrics(service_name, timeout_seconds=15 * 60):
     return sdk_cmd.service_request("GET", service_name, "/v1/metrics").json()
 
 
+def get_failure_metrics(service_name: str) -> typing.Dict:
+    history = sdk_cmd.history_request().json()
+    service_history = [h for h in history["frameworks"] if h["name"] == service_name]
+    if not service_history:
+        return dict()
+
+    def collect():
```
elezar: A couple of things:
* We assume a length of the `service_history` list. Should we assert if we find multiple services with the same name?
* Then, we each of the `frameworks` that are in `service_history` have a known structure. Would it not make sense to just call `get()` on the desired failure keys?
* Lastly, what about the other failure modes: `TASK_LOST`, `TASK_UNREACHABLE`, or `TASK_KILLED`?

testing/sdk_cmd.py
```diff
@@ -53,10 +53,14 @@ def service_request(
         log_args=log_args,
         log_response=log_response,
         timeout_seconds=timeout_seconds,
-        **kwargs
+        **kwargs,
     )
 
 
+def history_request(retry=False) -> requests.Response:
```
elezar: I'm not sold on having this in `sdk_cmd`. It seems to be here just because it calls `cluster_request`. We could rather just call this directly from the `get_failure_metrics` function.

testing/sdk_metrics.py
```diff
@@ -24,6 +25,22 @@ def get_scheduler_metrics(service_name, timeout_seconds=15 * 60):
     return sdk_cmd.service_request("GET", service_name, "/v1/metrics").json()
 
 
+def get_failure_metrics(service_name: str) -> typing.Dict:
+    history = sdk_cmd.history_request().json()
+    service_history = [h for h in history["frameworks"] if h["name"] == service_name]
+    if not service_history:
```
elezar: Is not finding the service an acceptable state?

testing/sdk_metrics.py
```diff
@@ -24,6 +25,22 @@ def get_scheduler_metrics(service_name, timeout_seconds=15 * 60):
     return sdk_cmd.service_request("GET", service_name, "/v1/metrics").json()
 
 
+def get_failure_metrics(service_name: str) -> typing.Dict:
+    history = sdk_cmd.history_request().json()
```
elezar: Should we call `resp.raise_for_status()` here?

testing/sdk_metrics.py
```diff
@@ -24,6 +25,22 @@ def get_scheduler_metrics(service_name, timeout_seconds=15 * 60):
...
+        return dict()
+
+    def collect():
+        failure_keys = ["TASK_FAILED", "TASK_ERROR"]
+        for k, v in service_history[0].items():
+            if k not in failure_keys:
+                continue
+            yield k, v
+
+    return dict(collect())
```
elezar: Since we're never using the individual values for the failure modes, could we not just return the `sum()` of the counts here? This would make its usage at the call-sites much simpler.

testing/sdk_metrics.py
```diff
@@ -24,6 +25,22 @@ def get_scheduler_metrics(service_name, timeout_seconds=15 * 60):
     return sdk_cmd.service_request("GET", service_name, "/v1/metrics").json()
 
 
+def get_failure_metrics(service_name: str) -> typing.Dict:
```
elezar: If we change this to return the sum, we could change the function name.

testing/sdk_plan.py
```diff
@@ -136,10 +143,20 @@ def wait_for_plan_status(
...
+        wait_fixed=1000,
+        stop_max_delay=timeout_seconds * 1000,
+        retry_on_result=lambda res: not res,
+        retry_on_exception=lambda e: not isinstance(e, FailuresExceededError),
     )
     def fn():
+        failure_metrics = sdk_metrics.get_failure_metrics(service_name)
+        failures = sum(failure_metrics.values())
+        if failures - initial_failures > ALLOWED_FAILURES:
+            raise FailuresExceededError("Service not recoverable: {}".format(service_name))
```
elezar: I would add more context to the exception message, or at the very least add an `.error` log line.

testing/sdk_plan.py
```diff
@@ -10,9 +10,16 @@
...
 
 import sdk_cmd
+import sdk_metrics
 
 TIMEOUT_SECONDS = 15 * 60
 SHORT_TIMEOUT_SECONDS = 30
+ALLOWED_FAILURES = 2
+
+
+class FailuresExceededError(Exception):
```
elezar: I would say: `TaskFailuresExceededException`.

testing/sdk_plan.py
```diff
@@ -10,9 +10,16 @@
 import retrying
 
 import sdk_cmd
+import sdk_metrics
 
 TIMEOUT_SECONDS = 15 * 60
 SHORT_TIMEOUT_SECONDS = 30
+ALLOWED_FAILURES = 2
```
elezar: `ALLOWED_FAILURES` is a little bit misleading here, since it controls the allowed INCREASE in task failures while waiting for a plan status to change.

testing/sdk_metrics.py
mockingbird2: I experienced that the entry in the history server might not have been created yet when querying for the plan status. Thus I am not sure how to deal with the not found case.

testing/sdk_metrics.py
mockingbird2: The `.json()` raises an exception on `204` or invalid JSON types. What are additional checks you think of?

testing/sdk_plan.py
mockingbird2: True, I forgot about it after introducing the error diff.

testing/sdk_metrics.py
mockingbird2: - Assert for only one service sounds good
- The only assumed structure currently is the `name` field. I'll add a `get` with default to empty then it behaves like the service is not found and does not crash. 
- I am not completely familiar with what kind of status is raised when. Maybe you can give an update on that.

testing/sdk_metrics.py
elezar: This is a HTTP request, and can have any number of return values that indicate some issue. Using `raise_for_status()` just helps to show us that these errors occurred instead of having to debug the reason for the `.json()` call failing.

testing/sdk_metrics.py
elezar: One of the `"frameworks"` looks as follows:
```json
{
"id": "850f4418-2fc0-429f-9b08-e8530e42d8f4-0001",
"name": "marathon",
"pid": "scheduler-e851e3a7-2c86-497a-970b-b1a36766b06f@10.0.7.148:15101",
"used_resources": {
"disk": 0,
"mem": 0,
"gpus": 0,
"cpus": 0
},
"offered_resources": {
"disk": 0,
"mem": 0,
"gpus": 0,
"cpus": 0
},
"capabilities": [
"TASK_KILLING_STATE",
"GPU_RESOURCES",
"PARTITION_AWARE",
"REGION_AWARE"
],
"hostname": "10.0.7.148",
"webui_url": "https://10.0.7.148:8443",
"active": true,
"connected": true,
"recovered": false,
"TASK_STAGING": 0,
"TASK_STARTING": 0,
"TASK_RUNNING": 0,
"TASK_KILLING": 0,
"TASK_FINISHED": 0,
"TASK_KILLED": 0,
"TASK_FAILED": 0,
"TASK_LOST": 0,
"TASK_ERROR": 0,
"TASK_UNREACHABLE": 0,
"slave_ids": []
}
```
I would then explicitly do the following (or using an explicit `for` loop with `yield` and `collect`):
```python
framework_history = service_history[0]
fields_of_interest = ["TASK_FAILED", "TASK_ERROR"]
return dict([f, framework_history.get(f, 0) for f in fields_of_interest])
```
To me this is easier to read, and I would be willing to wage that this is faster as we are not iterating over a large number of key-value pairs just to find the ones of interest, but using the fast lookups in the dict.

testing/sdk_metrics.py
elezar: If at some point we need the additional functionality, we can refactor this (maybe for #2638), but I would say that we should look for the simplest solution here.

testing/sdk_metrics.py
elezar: An overview of the mesos states and their reasons can be found here: http://mesos.apache.org/documentation/latest/task-state-reasons/#reasons-for-terminal-status-updates

testing/sdk_metrics.py
elezar: I'm also not really sure about having this in `sdk_metrics`. In our context metrics mean something different -- they are generated by the scheduler or the underlying framework. Here we are only interested in the number of tasks in a (failed) terminal state. I would argue that `sdk_tasks` is a better module to implement this in.

testing/sdk_plan.py
```diff
@@ -136,20 +143,31 @@ def wait_for_plan_status(
...
-        wait_fixed=1000, stop_max_delay=timeout_seconds * 1000, retry_on_result=lambda res: not res
+        wait_fixed=1000,
+        stop_max_delay=timeout_seconds * 1000,
+        retry_on_result=lambda res: not res,
+        retry_on_exception=lambda e: not isinstance(e, TaskFailuresExceededException),
     )
     def fn():
+        failures = sdk_metrics.sum_service_failures(service_name)
+        if failures - initial_failures > ALLOWED_FAILURES_INCREASE:
+            log.error(
```
elezar: Let's not use `.format()` in logging strings as these support `%s` followed by a list of variables to interpolate.

testing/sdk_plan.py
```diff
@@ -136,20 +143,31 @@ def wait_for_plan_status(
...
             service_name,
             plan_name,
             timeout_seconds=SHORT_TIMEOUT_SECONDS,
             multiservice_name=multiservice_name,
         )
         log.info(
-            "Waiting for {} {} plan:\n{}".format(
-                status, plan_name, plan_string(plan_name, plan)
-            )
+            "Waiting for {} {} plan:\n{}".format(status, plan_name, plan_string(plan_name, plan))
```
elezar: We could change that for this one here too.

testing/sdk_metrics.py
mockingbird2: I now have used this reference: https://github.com/apache/mesos/blob/73d1da0893bab8cb7c5d540f115c5106f3e2039e/include/mesos/mesos.proto#L2212 
to compile the following state list 
```
FATAL_TERMINAL_TASK_STATES = set(
    ["TASK_FAILED", "TASK_ERROR", "TASK_KILLED", "TASK_DROPPED", "TASK_GONE"]
)
```

mockingbird2:
Thanks, @elezar for reviewing. I'll have a look at #2638.

testing/sdk_tasks.py
```diff
@@ -16,22 +16,23 @@
 
 DEFAULT_TIMEOUT_SECONDS = 30 * 60
 
+FATAL_TERMINAL_TASK_STATES = set(
+    ["TASK_FAILED", "TASK_ERROR", "TASK_KILLED", "TASK_DROPPED", "TASK_GONE"]
+)
+
```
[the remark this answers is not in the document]
mockingbird2: It is complaining but it is supposed to be two.

testing/sdk_tasks.py
```diff
@@ -138,6 +139,25 @@ def get_all_status_history(task_name: str, with_completed_tasks=True) -> list:
     return history
 
 
+def get_task_failures_sum(service_name) -> int:
+    history_response = sdk_cmd.cluster_request(
+        "GET", "/dcos-history-service/history/last", retry=False
```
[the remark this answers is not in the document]
mockingbird2: Tried to be a bit conservative but the initial failure count might fail now. I will make it optional and use it initially.

testing/sdk_tasks.py
```diff
@@ -138,6 +141,21 @@ def get_all_status_history(task_name: str, with_completed_tasks=True) -> list:
     return history
 
 
+def get_task_failed_count(service_name: str, retry: bool = False) -> int:
```
elezar: Nit: `get_failed_task_count` reads better.

PR #2629, opened 2018-08-15
[DCOS-40522] Add endpoint wait test util

mockingbird2:


testing/sdk_networks.py
```diff
@@ -44,16 +48,28 @@ def check_task_network(task_name, expected_network_name="dcos"):
                     )
 
 
+@retrying.retry(
+    wait_fixed=1000, stop_max_delay=5 * 1000, retry_on_result=lambda res: not res
+)
+def wait_for_endpoint_info(
+    package_name: str, service_name: str, endpoint_name
```
elezar: `endpint_name` should be annotated as a `str` too.

testing/sdk_networks.py
```diff
@@ -44,16 +48,28 @@ def check_task_network(task_name, expected_network_name="dcos"):
...
     Checks that there is the correct number of endpoints"""
-    endpoints = sdk_cmd.svc_cli(
-        package_name, service_name, "endpoints {}".format(endpoint_to_get), json=True
-    )
-    assert len(endpoints) == correct_count, "Wrong number of endpoints, got {} should be {}".format(
+    endpoints = wait_for_endpoint_info(package_name, service_name, endpoint_to_get)
+    assert (
+        len(endpoints) == correct_count
+    ), "Wrong number of endpoints, got {} should be {}".format(
         len(endpoints), correct_count
```
elezar: While we're at it, we may want to print the list of endpoints in case of failure too.

testing/sdk_networks.py
```diff
@@ -44,16 +48,28 @@ def check_task_network(task_name, expected_network_name="dcos"):
                     )
 
 
+@retrying.retry(
+    wait_fixed=1000, stop_max_delay=5 * 1000, retry_on_result=lambda res: not res
+)
+def wait_for_endpoint_info(
```
elezar: Are there tests in `dcos-commons` that could use this?

testing/sdk_networks.py
```diff
@@ -44,17 +48,29 @@ def check_task_network(task_name, expected_network_name="dcos"):
...
+    package_name: str, service_name: str, endpoint_name: str, json: bool = True
+) -> typing.Dict:
```
elezar: The addition of the `json` arg means that this return type is no longer valid. Is there a better way to handle this than overloading the function?

frameworks/cassandra/tests/test_sanity.py
```diff
@@ -24,7 +24,9 @@ def configure_package(configure_security):
             config.PACKAGE_NAME,
             config.get_foldered_service_name(),
             config.DEFAULT_TASK_COUNT,
-            additional_options={"service": {"name": config.get_foldered_service_name()}},
+            additional_options={
```
elezar: A lot of these changes seem to be related to black formatting. In the config in the project root, we set the line length to 100. So currently:
```
✗ black --diff frameworks/cassandra/tests/test_sanity.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

mockingbird2: Oh, I might have missed this. I will revert these changes.

frameworks/hdfs/tests/test_overlay.py
```diff
@@ -58,8 +58,8 @@ def test_endpoints_on_overlay():
     expected_endpoints = ("hdfs-site.xml", "core-site.xml")
```
elezar: I know this wasn't part of your changes, but this should ideally be a list and not a tuple.

frameworks/kafka/tests/test_tls.py
```diff
@@ -13,7 +13,9 @@
 from tests import config
 
 pytestmark = [
-    pytest.mark.skipif(sdk_utils.is_open_dcos(), reason="Feature only supported in DC/OS EE"),
+    pytest.mark.skipif(
```
elezar: This file still contains a lot of formatting change that could be reverted.

testing/sdk_networks.py
mockingbird2: For now, I could set the return type to `Union[Dict, str]`. Unfortunately not every service endpoint is json serializable at the moment.  Fixing this behaviour would probably include changing existing endpoint definition of services.

testing/sdk_networks.py
```diff
@@ -44,17 +46,27 @@ def check_task_network(task_name, expected_network_name="dcos"):
                     )
 
 
+@retrying.retry(wait_fixed=1000, stop_max_delay=5 * 1000, retry_on_result=lambda res: not res)
+def wait_for_endpoint_info(
```
elezar: My thought here is as follows: Consider splitting this into two functions:
* `wait_for_endpoint_json()`:
** returns a `typing.Dict`
** Could be left as `wait_for_endpoint_info`
* `wait_for_endpoint_string()`:
** returns a `str`
** returns `output.strip()` since we are capturing stdout, and this adds additional whitespace.

This makes the distinction at the call-site more explicit.

What are your thoughts on this?

mockingbird2: Definitely sounds reasonable. I'll also try then to clean up the unnecessary used `strip()` calls which are used currently.

frameworks/hdfs/tests/test_overlay.py
mockingbird2: What is the problem with having this immutable?

elezar: It's not a question of immutability, but rather that tuples are prone to errors when removing items. In this case, if we remove one of the elements then we'll end up iterating over the individual characters of the remaining string instead. 

This isn't critical as part of these changes though, and we can leave it as is.

frameworks/kafka/tests/test_sanity.py
```diff
@@ -47,20 +47,10 @@ def test_service_health():
 @pytest.mark.sanity
 def test_endpoints_address():
     foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
+    endpoints = sdk_networks.wait_for_endpoint_info(
```
elezar: Could we do more to align this with the TensorFlow test that you added?

frameworks/kafka/tests/test_sanity.py
```diff
@@ -47,20 +47,10 @@ def test_service_health():
...
+        config.PACKAGE_NAME, foldered_name, config.DEFAULT_TASK_NAME
```
elezar: I feel that using `"broker"` here instead of `config.DEFAULT_TASK_NAME` is clearer.

frameworks/kafka/tests/test_sanity.py
```diff
@@ -105,7 +97,9 @@ def test_custom_zookeeper():
     # wait for brokers to finish registering
     test_utils.broker_count_check(config.DEFAULT_BROKER_COUNT, service_name=foldered_name)
 
-    zookeeper = sdk_cmd.svc_cli(config.PACKAGE_NAME, foldered_name, "endpoints zookeeper")
+    zookeeper = sdk_networks.wait_for_endpoint_info_string(
+        config.PACKAGE_NAME, foldered_name, "zookeeper"
+    )
     assert zookeeper.rstrip("\n") == zk_path
```
elezar: It is my understanding that `strip()` would also remove `\n` and as such this `rstrip()` is now redundant.

frameworks/kafka/tests/test_sanity.py
```diff
@@ -75,7 +65,9 @@ def wait():
 @pytest.mark.sanity
 def test_endpoints_zookeeper_default():
     foldered_name = sdk_utils.get_foldered_name(config.SERVICE_NAME)
-    zookeeper = sdk_cmd.svc_cli(config.PACKAGE_NAME, foldered_name, "endpoints zookeeper")
+    zookeeper = sdk_networks.wait_for_endpoint_info_string(
+        config.PACKAGE_NAME, foldered_name, "zookeeper"
+    )
     assert zookeeper.rstrip("\n") == "master.mesos:2181/{}".format(
```
elezar: See comment below regarding `rstrip()`.

testing/sdk_networks.py
```diff
@@ -44,17 +46,35 @@ def check_task_network(task_name, expected_network_name="dcos"):
                     )
 
 
+@retrying.retry(wait_fixed=1000, stop_max_delay=5 * 1000, retry_on_result=lambda res: not res)
+def _get_endpoint_info(
```
elezar: Nice work!

testing/sdk_networks.py
```diff
@@ -7,6 +7,8 @@
 import logging
 import shakedown
 import sdk_cmd
```
elezar: I know it's not documented anywhere, but let's first import standard library packages, and then the `testing` modules (`sdk_*`) and then the local test modules (if any).

frameworks/hdfs/tests/test_overlay.py
mockingbird2: Is there even another way of removing an element from a tuple than casting to `list`, remove and cast back?

frameworks/kafka/tests/test_sanity.py
mockingbird2: True

frameworks/kafka/tests/test_sanity.py
mockingbird2: Can Kafka have any other task names in general? Otherwise, I would replace all the occurrences and remove the variable in `config.py`.

frameworks/kafka/tests/test_sanity.py
mockingbird2: I think most of the data-services do a count check for the endpoints. Thus it might be applicable for more than the Kafka service. In this case, it already looks quite similar because I started the Tensorflow implementation based on this. I'd rather have a look at this in a follow-up PR where I try to reuse more of the existing Tensorflow endpoint test util.

frameworks/hdfs/tests/test_overlay.py
mockingbird2: I think I see now the struggle: In the future, it might happen that someone changes the initialization to `("hdfs-site.xml")` and misses the comma at the end... So the expression is evaluated to a string.

frameworks/kafka/tests/test_sanity.py
elezar: Kafka only has `broker` tasks: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/kafka/src/main/dist/svc.yml#L31

Feel free to remove this variable.

testing/sdk_networks.py
elezar: After stdlib imports, but before local imports.

PR #2631, opened 2018-08-20
[DCOS-40636][TOOLS] Add interactive testing note to TESTING.md

elezar:
This PR:
* Adds a section on interactive testing to `TESTING.md`
* Improves interactive mode for `test.sh` by using an existing AWS credentials file if it exists. This allows credentials to be refreshed without having to restart the docker container.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -79,6 +79,49 @@ The detected frameworks would be:
 
 If no `frameworks` folder exists, the current folder is assumed to be the root of a single framework. Note that the name of the framework displayed in this case (and above) is purely cosmetic, and the package name for the framework is determined from the catalog definition files.
 
+#### Running tests interactively
+
+The default when invoking `test.sh` is to run all tests defined for a given framework. It may, however be desirable to run a subset of tests interactively while developing a feature or the related tests themselves.
+
+The following command:
+```bash
+CLUSTER_URL=$(dcos config show core.dcos_url) ./test.sh -i
+```
+starts an interactive shell in a Docker container with the requirements for running the tests installed. The folder from which `./test.sh` was started is mounted into the container as `/build` and the default (or specified) SSH key is mounted at `/ssh/key`.
+
+**Note:** In order to run tests against a strict mode cluster and additional `-s` flag is required when starting the container.
+
+##### Setting up
+Before running test we perform the following steps inside the container:
+* Connect to the cluster: `dcos cluster setup ${CLUSTER_URL}`
+* Add the SSH key to the SSH agent: `eval $(ssh-agent -s); ssh-add /ssh/key`
+* Set the `PYTHONPATH` environment variable: `export PYTHONPATH=testing`
+
+Most system integration tests rely on a stub-universe being set up allowing us to install the version of the framework under test. Thus it is required to run:
+```bash
+export STUB_UNIVERSE_URL="https://universe-converter.mesosphere.com/transform?url=https://infinity-artifacts.s3.amazonaws.com/autodelete7d/my-framework/20180806-125351-sdEdQ7mRfHFXQzmT/stub-universe-my-framework.json
+```
+The stub-universe URL can be optained from an earlier build, or by building the framework (`my-framework` in this case) in the container by running the following command:
+```bash
+./frameworks/my-framwork/build.sh aws
+```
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -271,8 +271,10 @@ EOF
         echo "Missing AWS credentials file (${aws_creds_path}) and AWS env (AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY)"
         exit 1
     fi
+    chmod 500 "${credsfile}"
```
porridge: This is racy, it should probably be set at creation time done by something like `(umask 077 && mktemp ...)` instead.

test.sh
```diff
@@ -252,16 +252,16 @@ fi
 
 # Write the AWS credential file (deleted on script exit)
 if [ -f "${aws_creds_path}" ]; then
-    cat $aws_creds_path > $credsfile
+    aws_credential_file_path="${aws_creds_path}"
```
porridge: Perhaps this is out of scope for this PR, but now we have:
- `aws_creds_path`
- `aws_credential_file_path`
- `credsfile`

and it is not obvious from their names how their purpose differs.

TESTING.md
```diff
@@ -79,6 +79,49 @@ The detected frameworks would be:
 
 If no `frameworks` folder exists, the current folder is assumed to be the root of a single framework. Note that the name of the framework displayed in this case (and above) is purely cosmetic, and the package name for the framework is determined from the catalog definition files.
 
+#### Running tests interactively
+
+The default when invoking `test.sh` is to run all tests defined for a given framework. It may, however be desirable to run a subset of tests interactively while developing a feature or the related tests themselves.
+
+The following command:
+```bash
+CLUSTER_URL=$(dcos config show core.dcos_url) ./test.sh -i
```
porridge: I wonder whether it would make sense that when `-i` is specified, then if CLUSTER_URL is unset then the script should default to using current dcos-attached cluster rather than create one automatically?

TESTING.md
```diff
@@ -79,6 +79,49 @@ The detected frameworks would be:
...
+```
+starts an interactive shell in a Docker container with the requirements for running the tests installed. The folder from which `./test.sh` was started is mounted into the container as `/build` and the default (or specified) SSH key is mounted at `/ssh/key`.
+
+**Note:** In order to run tests against a strict mode cluster and additional `-s` flag is required when starting the container.
+
+##### Setting up
+Before running test we perform the following steps inside the container:
+* Connect to the cluster: `dcos cluster setup ${CLUSTER_URL}`
+* Add the SSH key to the SSH agent: `eval $(ssh-agent -s); ssh-add /ssh/key`
+* Set the `PYTHONPATH` environment variable: `export PYTHONPATH=testing`
```
porridge: Um, a relative `PYTHONPATH` entry seems wrong...

porridge: Also, I cannot see where it's being set in the script?

elezar: It's not set in the script, but is set in `__init__.py` and in `test_runner.sh`, but yes, this could be `/build/testing` in this case.

test.sh
elezar: I have removed the setting of the permissions. This is mounted as `:ro` into the container in any case.

TESTING.md
elezar: d95ac60

test.sh
elezar: I have renamed the variables in d95ac60 and 4a3dda1 to make the intent a bit clearer.

elezar:
Thanks for the review @porridge. I think I have addressed your concerns. Please have a look and let me know.

TESTING.md
porridge: Doesn't it? Is https://github.com/mesosphere/dcos-commons/blob/6538663/test.sh#L99-L100 stale?

test.sh
porridge: I thought the point was to prevent others on the host from reading the file?

TESTING.md
porridge: Can you please drop the `CLUSTER_URL=$(dcos config show core.dcos_url) ` part from this line now that the script does it automagically?

elezar:
Done! 
49ff6b2

PR #2632, opened 2018-08-22
Local Profile Mount Volume Support.

chhsia0:
Epic: [DCOS-30035](https://jira.mesosphere.com/browse/DCOS-30035)

This PR resolves the following tickets:
[DCOS-40328](https://jira.mesosphere.com/browse/DCOS-40328) Bump Mesos Proto to 1.5+ in SDK.
[DCOS-40329](https://jira.mesosphere.com/browse/DCOS-40329) Add a `profiles` field to `VolumeSpec` and `RawVolume`.
[DCOS-40330](https://jira.mesosphere.com/browse/DCOS-40330) Validate `profiles` for volume specifications.
[DCOS-40331](https://jira.mesosphere.com/browse/DCOS-40331) Volume profile support for offer evaluation.


pires:
cc @sreis @bmcstdio

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
```diff
@@ -25,33 +27,43 @@
...
+     * characters ([a-zA-Z0-9]), dashes (-), underscores (_) or dots(.), and should be non-empty and at most 128
+     * characters. See: https://jira.mesosphere.com/browse/DCOS-40365
+     */
+    private static final Pattern VALID_PROFILE_PATTERN = Pattern.compile("[a-zA-Z0-9_.-]{1,128}");
+
     private final Type type;
     private final String containerPath;
+    private final List<String> profiles;
 
     public DefaultVolumeSpec(
```
nickbp: Since there are now different sets of arguments between ROOT/PATH vs MOUNT volumes, how about having two static methods here, one for MOUNT (with a profiles list) and another for ROOT/PATH (omits profiles list)?

Something like:
```
public static DefaultVolumeSpec createMount(... common args ..., List<String> profiles);
public static DefaultVolumeSpec createRootOrMount(Type type, ... common args ...);
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
```diff
@@ -154,10 +163,14 @@ public Offer getOffer() {
             if (atomicResources == null) {
                 LOGGER.info("Offer lacks any atomic resources named {}", resourceName);
             } else {
+                String desired = spec.getValue().toString();
+                if (spec.getProfiles() != null) {
+                    desired += "(profiles: " + spec.getProfiles() + ")";
```
nickbp: Could do `desired += String.format(" (profiles: %s)", spec.getProfiles())` here (with a leading space to separate from the value)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
```diff
@@ -154,10 +163,14 @@ public Offer getOffer() {
             if (atomicResources == null) {
                 LOGGER.info("Offer lacks any atomic resources named {}", resourceName);
             } else {
+                String desired = spec.getValue().toString();
```
nickbp: This could use `TextFormat.shortDebugString(spec.getValue())` to avoid newlines in the log

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
```diff
@@ -154,10 +163,14 @@ public Offer getOffer() {
...
+                if (spec.getProfiles() != null) {
```
nickbp: `!spec.getProfiles().isEmpty()`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
```diff
@@ -271,6 +284,18 @@ private static boolean sufficientValue(Value desired, Value available) {
         return ValueUtils.compare(difference, ValueUtils.getZero(desired.getType())) <= 0;
     }
 
+    private static boolean matchAnyProfile(List<String> desired, Optional<String> actual) {
+        if (desired != null && actual.isPresent()) {
```
nickbp: Should be able to skip null checks here and at L296

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceBuilder.java
```diff
@@ -97,10 +111,17 @@ private static ResourceSpec getResourceSpec(Protos.Resource resource) {
     @SuppressWarnings("deprecation")
     private static VolumeSpec getVolumeSpec(Protos.Resource resource) {
         VolumeSpec.Type type = resource.getDisk().hasSource() ? VolumeSpec.Type.MOUNT : VolumeSpec.Type.ROOT;
+
+        // Profiles are only supported on MOUNT volumes. For ROOT volumes, we make it null.
+        List<String> profiles = resource.getDisk().getSource().hasProfile()
+                ? Arrays.asList(resource.getDisk().getSource().getProfile())
+                : null;
```
nickbp: `Collections.emptyList()` (and update comment at L115)

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
```diff
@@ -60,6 +72,7 @@ private DefaultVolumeSpec(
         super(name, value, role, preReservedRole, principal);
         this.type = type;
         this.containerPath = containerPath;
+        this.profiles = profiles;
```
nickbp: Even after the Java code isn't passing in `null` anymore, this code should still check for `null` because it's being invoked to read preexisting configs in ZK. It could be written as e.g. `this.profiles = profiles == null ? Collections.emptyList() : profiles;`. This pattern can be seen in other `Default*Spec` types where values may be unset in the serialized rendering

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
```diff
@@ -86,15 +105,33 @@ public int hashCode() {
...
+        validateResource();
+        ValidationUtils.matchesRegex(this, "containerPath", containerPath, VALID_CONTAINER_PATH_PATTERN);
+
+        if (type == Type.MOUNT) {
+            ValidationUtils.nonEmptyAllowNull(this, "profiles", profiles);
+        } else {
+            ValidationUtils.isNull(this, "profiles", profiles);
+        }
+
+        if (profiles != null) {
```
nickbp: `if (!profiles.isEmpty()) {`

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
```diff
@@ -86,15 +105,33 @@ public int hashCode() {
...
     }
+
+    private void validateVolume() {
+        validateResource();
+        ValidationUtils.matchesRegex(this, "containerPath", containerPath, VALID_CONTAINER_PATH_PATTERN);
+
+        if (type == Type.MOUNT) {
+            ValidationUtils.nonEmptyAllowNull(this, "profiles", profiles);
+        } else {
+            ValidationUtils.isNull(this, "profiles", profiles);
```
nickbp: `isEmpty()` (would need to add this to `ValidationUtils`)

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
```diff
@@ -86,15 +105,33 @@ public int hashCode() {
...
         builder.getScalarBuilder().setValue(value);
         return builder.build();
     }
+
+    private void validateVolume() {
+        validateResource();
+        ValidationUtils.matchesRegex(this, "containerPath", containerPath, VALID_CONTAINER_PATH_PATTERN);
+
+        if (type == Type.MOUNT) {
+            ValidationUtils.nonEmptyAllowNull(this, "profiles", profiles);
```
nickbp: `nonEmpty()`

nickbp: May make sense to also validate that the profiles are valid profile strings. Are there limitations on valid characters in profile names? could check `matchesRegex()` against that definition.

If there are no character limitations then it should probably at least check for `nonBlank()` for each of the provided profile strings.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/ValidationUtils.java
```diff
@@ -180,4 +180,13 @@ public static void nonNull(Object parent, String fieldName, Object field) {
             throw new IllegalArgumentException(String.format("%s cannot be null: %s", fieldName, parent));
         }
     }
+
+    /**
+     * Throws an exception if {@code field} is not {@code null}.
+     */
+    public static void isNull(Object parent, String fieldName, Object field) {
```
nickbp: Can remove this since it wouldn't be used after the above changes

sdk/scheduler/src/test/java/com/mesosphere/sdk/config/validate/TaskVolumesCannotChangeTest.java
```diff
@@ -25,20 +25,23 @@
             DISK_SIZE_MB,
             VolumeSpec.Type.MOUNT,
             "some_path",
+            null,
```
nickbp: `Collections.emptyList()` here and below

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/MesosResourcePoolTest.java
```diff
@@ -57,17 +60,98 @@ public void testMultipleUnreservedAtomicPool() {
...
     @Test
     public void testConsumeUnreservedAtomicResource() {
-        Resource offerResource = ResourceTestUtils.getUnreservedMountVolume(1000);
-        Protos.Value resourceValue = ValueUtils.getValue(offerResource);
+        Resource offerResource = ResourceTestUtils.getUnreservedMountVolume(1000, Optional.empty());
+        VolumeSpec spec = new DefaultVolumeSpec(
+                ValueUtils.getValue(offerResource).getScalar().getValue(),
+                VolumeSpec.Type.MOUNT,
+                TestConstants.CONTAINER_PATH,
+                null,
```
nickbp: `Collections.emptyList()`

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/MesosResourcePoolTest.java
```diff
@@ -57,17 +60,98 @@ public void testMultipleUnreservedAtomicPool() {
...
+    }
+
+    @Test
+    public void testConsumeUnreservedAtomicResourceUnmatchedProfile3() {
+        Resource offerResource = ResourceTestUtils.getUnreservedMountVolume(1000, Optional.of("bar"));
+        VolumeSpec spec = new DefaultVolumeSpec(
+                ValueUtils.getValue(offerResource).getScalar().getValue(),
+                VolumeSpec.Type.MOUNT,
+                TestConstants.CONTAINER_PATH,
+                null,
```
nickbp: `Collections.emptyList()`

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/ResourceBuilderTest.java
```diff
@@ -250,6 +252,7 @@ private static void testExistingFromRootVolumeSpec(Optional<String> namespace) {
                 10,
                 VolumeSpec.Type.ROOT,
                 TestConstants.CONTAINER_PATH,
+                null,
```
nickbp: `Collections.emptyList()`

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/ResourceBuilderTest.java
```diff
@@ -360,6 +363,7 @@ private static void testExistingFromMountVolumeSpec(Optional<String> namespace)
                 10,
                 VolumeSpec.Type.MOUNT,
                 TestConstants.CONTAINER_PATH,
+                null,
```
nickbp: `Collections.emptyList()`

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/ResourceBuilderTest.java
```diff
@@ -433,6 +435,7 @@ private static void testFromExistingRootVolume(Optional<String> namespace) {
                 10,
                 VolumeSpec.Type.ROOT,
                 TestConstants.CONTAINER_PATH,
+                null,
```
nickbp: `Collections.emptyList()`

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/ResourceBuilderTest.java
```diff
@@ -469,18 +473,19 @@ private static void testFromExistingMountVolume(Optional<String> namespace) {
                 10,
                 VolumeSpec.Type.MOUNT,
                 TestConstants.CONTAINER_PATH,
+                null,
```
nickbp: `Collections.emptyList()`

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorVolumesTest.java
```diff
@@ -174,10 +175,10 @@ public void testReserveLaunchScalarRefined() throws Exception {
     @Test
     public void testReserveCreateLaunchMountVolume() throws Exception {
         Resource offeredCpuResource = ResourceTestUtils.getUnreservedCpus(1.0);
-        Resource offeredDiskResource = ResourceTestUtils.getUnreservedMountVolume(2000);
+        Resource offeredDiskResource = ResourceTestUtils.getUnreservedMountVolume(2000, Optional.empty());
 
         List<OfferRecommendation> recommendations = evaluator.evaluate(
-                PodInstanceRequirementTestUtils.getMountVolumeRequirement(1.0, 1500),
+                PodInstanceRequirementTestUtils.getMountVolumeRequirement(1.0, 1500, null),
```
nickbp: `Collections.emptyList()`, or could have two overloaded versions of `getMountVolumeRequirement()` where one includes a `profiles` parameter and the other defaults to an empty list?

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorVolumesTest.java
```diff
@@ -223,10 +224,10 @@ public void testReserveCreateLaunchMountVolume() throws Exception {
     public void testExpectedMountVolume() throws Exception {
         // Launch for the first time.
         Resource offeredCpuResource = ResourceTestUtils.getUnreservedCpus(1.0);
-        Resource offeredDiskResource = ResourceTestUtils.getUnreservedMountVolume(2000);
+        Resource offeredDiskResource = ResourceTestUtils.getUnreservedMountVolume(2000, Optional.empty());
 
         PodInstanceRequirement podInstanceRequirement =
-                PodInstanceRequirementTestUtils.getMountVolumeRequirement(1.0, 1500);
+                PodInstanceRequirementTestUtils.getMountVolumeRequirement(1.0, 1500, null);
```
nickbp: `Collections.emptyList()` or use overloaded method without `profiles` parameter

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
chhsia0: These two functions look a bit confusing, and the names won't be future-proof if we're going o support new volume types, such as PATH or BLOCK. If we want to have these new methods, it seems to me that the following provides a cleaner API:
```java
public static DefaultVolumeSpec createRootVolume(...);
public static DefaultVolumeSpec createMountVolume(...);
public static DefaultVolumeSpec updateVolume(VolumeSpec original, double diskSize);
```
However, this also means we would have the following code snippets:
`ResourceBuilder.java`:
```java
return resource.getDisk.hasSource()
        ? DefaultVolumeSpec.createMountVolume(..., Array.asList(profile), ...)
        : DefaultVolumeSpec.createRootVolume(...);
```
`DefaultResourceSet.java`:
```java
public Builder addRootVolume(...);
public Builder addMountVolume(...);
```
`YAMLToInternalMappers.java`:
```java
return volumeTypeEnum == VolumeSpec.Type.MOUNT
        ? DefaultVolumeSpec.createMountVolume(...)
        : DefaultVolumeSpec.createRootVolume(...);

if (rawVolume.getType() == VolumeSpec.Type.MOUNT.name()) {
    resourceSetBuilder.addMountVolume(...);
} else {
    resourceSetBuilder.addRootVolume(...);
}
```
I'm not sure if this change would lead to a cleaner API. WDYT?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
chhsia0: This is actually why I went with `null` instead of an empty list. But since it's more consistent to go with an empty list, I'll do what you suggested.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
chhsia0: If we go with empty lists, this should be simply `nonNull()` instead since we do support MOUNT volumes with no profile.

chhsia0: I did in a couple lines after :)
```
ValidationUtils.matchesRegex(this, "profiles[" + index + "]", profile, VALID_PROFILE_PATTERN);
```

chhsia0:
This feature could be used in 1.11 w/ DSS. However, DSS in 1.11 is not compatible with 1.12.

sdk/scheduler/build.gradle
```diff
@@ -139,7 +139,7 @@ dependencies {
     compile "javax.el:javax.el-api:${elVer}"
     compile "org.glassfish.web:javax.el:${elVer}"
     compile 'com.mesosphere:mesos-http-adapter:0.4.1'
-    compile 'org.apache.mesos:mesos:1.4.0-rc1'
+    compile 'org.apache.mesos:mesos:1.5.0'
```
nickbp: Could make sense to have a comment here mentioning that this corresponds to the version of Mesos in DC/OS 1.11.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/ResourceLabels.java
```diff
@@ -1,7 +1,10 @@
 package com.mesosphere.sdk.offer.evaluate;
 
 import com.mesosphere.sdk.specification.ResourceSpec;
+
 import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.mesos.Protos.ResourceProviderID;
```
nickbp: Should just import `Protos` here and then refer to the types as `Protos.X` below

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/VolumeEvaluationStage.java
```diff
@@ -4,6 +4,7 @@
 import com.mesosphere.sdk.offer.*;
 import com.mesosphere.sdk.specification.VolumeSpec;
 import org.apache.mesos.Protos.Resource;
+import org.apache.mesos.Protos.ResourceProviderID;
```
nickbp: Now that this is using multiple `Protos` types, can just import `Protos` here and then refer to the types as `Protos.X` below.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultVolumeSpec.java
nickbp: What you suggested looks great to me! Given that each of these types ends up being evaluated differently it's probably better to have explicit constructors for each.

sdk/scheduler/src/test/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluatorTest.java
```diff
@@ -12,9 +12,16 @@
 import com.mesosphere.sdk.state.PersistentLaunchRecorder;
 import com.mesosphere.sdk.testutils.*;
 import org.apache.commons.lang3.tuple.Pair;
-import org.apache.mesos.Protos;
-import org.apache.mesos.Protos.*;
+import org.apache.mesos.Protos.Attribute;
```
nickbp: Could just import `Protos` here and refer to each type as `Proto.X` below.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/MesosResourcePool.java
```diff
@@ -285,15 +285,17 @@ private static boolean sufficientValue(Value desired, Value available) {
...
-        if (desired != null && actual.isPresent()) {
-            for (String profile : desired) {
-                if (profile.equals(actual.get())) {
-                    return true;
-                }
+        if (!actual.isPresent()) {
+            return desired.isEmpty();
+        }
+
+        for (String profile : desired) {
```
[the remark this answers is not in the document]
chhsia0: Apparently I don't speak Java fluently lol.

nickbp:
Ping

chhsia0:
@nickbp Addressed your comments in the last commit. PTAL!

pires:
The Kubernetes team tested this PR (manually) successfully.

elezar:
@nickbp @chhsia0 the merge of this PR has broken the SDK gradle builds. I have created #2706 to revert these changes as this blocks all our other PRs.

The only change to the gradle files was the mesos bump, so there could be some transitive dependency that is no longer working as expected.

PR #2634, opened 2018-08-24
DCOS-40888 Workaround readiness check VIP inaccessibility on 1.10/1.11

nickbp:
VIPs are not accessible from readiness checks in 1.10 and 1.11, resulting in failed nightlies on test_overlay.
As a workaround this switches to checking the server port directly, with a TODO to switch back to the VIP later.
We still check that the VIP is reachable from regular tasks via the later 'getter' task.


nickbp:
Planning on merging as soon as hello-world's test_overlay is shown to be passing, as that's the main thing being changed here. Looks like this may also be blocking 1.10/1.11 PRs (not a lot, but some).

PR #2635, opened 2018-08-24
DCOS-40885 Re-add pip dependencies required by external repos

nickbp:
These can be removed someday after external repos (spark-build, dcos-zookeeper, ...) no longer depend on them, but for now let's avoid breaking things.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,13 @@ RUN pip3 install awscli
 COPY test_requirements.txt test_requirements.txt
 RUN pip3 install --upgrade -r test_requirements.txt
 
+# Deprecated libraries used by external repos.
+# As of PR#2616, these libraries aren't used by dcos-commons anymore.
+# However external repos using this image currently still need them.
+# Remove these once external repos aren't depending on them anymore.
+RUN pip3 install git+https://github.com/dcos/dcos-cli.git@f1fd38c9e72e1d521cf8120fe4948a789fb40cbc && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -46,6 +46,13 @@ RUN pip3 install awscli
 COPY test_requirements.txt test_requirements.txt
 RUN pip3 install --upgrade -r test_requirements.txt
 
+# Deprecated libraries used by external repos.
+# As of PR#2616, these libraries aren't used by dcos-commons anymore.
+# However external repos using this image currently still need them.
+# Remove these once external repos aren't depending on them anymore.
+RUN pip3 install git+https://github.com/dcos/dcos-cli.git@f1fd38c9e72e1d521cf8120fe4948a789fb40cbc && \
```
elezar: Why add it here and not in `test_requirements.txt`?

nickbp: Thought it made it clearer that it wasn't a requirement of local tests and was instead just being added to the docker image for use by others. However I'm finding that it doesn't work as expected when added as manual pip commands like this, so using test_requirements.txt may end up being required anyway.

nickbp: Yep, turns out pip pretty much ignored the install calls in Dockerfile for some reason, they need to go back in test_requirements.txt to work. Will have a new PR shortly after doing some more manual runs against sdk-0.40

elezar: Could it be that one needs an `--upgrade` argument?

PR #2637, opened 2018-08-27
DCOS-40885 Second attempt at fixing 'dcos' deps for extern/sdk-0.40

nickbp:
- The previous method of adding them directly to the Dockerfile doesn't work, they need to be in `test_requirements.txt` or else pip effectively ignores the specified versions.
- The dcoscli dependency is also needed to avoid breaking on an internal `servicemanager` dependency within the `dcos` library.


nickbp:
Looks like the docker construction + test runs are working fine in master, at least. Going to push this now to unblock other repos.

PR #2638, opened 2018-08-28
Wait for deployment to finish before checking taskcfg.

jeid64:
https://jira.mesosphere.com/browse/DCOS-39468
https://jira.mesosphere.com/browse/DCOS-40376
Previously we weren't waiting for deployment to finish before checking if tasks had finished running in the taskcfg tests. Since we added a possible 30 second delay to schedulers coming up due to DNS checks, we've introduced a race condition in the test where we no longer are fast enough to pass.

### DIFF
--- a/frameworks/helloworld/tests/test_taskcfg.py
+++ b/frameworks/helloworld/tests/test_taskcfg.py
@@ -20,7 +20,7 @@ def configure_package(configure_security):
             config.SERVICE_NAME,
             0,
             {"service": {"yaml": "taskcfg"}},
-            wait_for_deployment=False,
+            wait_for_deployment=True,
         )
 
         yield  # let the test session execute


frameworks/helloworld/tests/test_taskcfg.py
```diff
@@ -20,7 +20,7 @@ def configure_package(configure_security):
             config.SERVICE_NAME,
             0,
             {"service": {"yaml": "taskcfg"}},
-            wait_for_deployment=False,
```
nickbp: Could just remove the field since I think true is the default, but it's not worth running CI a second time over if it's green

mockingbird2:
@jeid64 it might help to have a look at this: https://github.com/mesosphere/dcos-commons/blob/06526d908a88cf4cdbeabfd642032bc6e4c7308c/frameworks/helloworld/tests/test_fast_failure.py#L13

In my PR #2628 I have added tracking of failed states during plan wait which seems pretty similar to this testing behavior.

nickbp:
There's currently an assumption of ordering in the returned state history here, I wonder if that's actually valid. May need to update sdk_tasks to explicitly order the state history by their timestamps?

Alternatively, this logic could be changed to checking that the total number of `TASK_FAILED` entries has gone up.

I think this particular test has been flaky for a while. Haven't really figured out the most consistent way of detecting that a task is crashlooping. The current state history lookup is a relatively recent attempt at that.

elezar:
> Alternatively, this logic could be changed to checking that the total number of TASK_FAILED entries has gone up.

Yes, this is essentially what @mockingbird2 has done for #2628.

> I think it'd also make sense to increase the current 30s timeout to say 5 minutes, in addition to the above explicit wait for the scheduler to come up. 30s feels short enough that the test could still fail if the task just happened to take too long to start

Yes, we should definitely increase the timeout too. What about `sdk_plan.wait_for_kicked_off_deployment()` to check that the scheduler is ready.

mockingbird2:
@jeid64 #2628 is merged now so feel free to use it.

jeid64:
Closing to do new approach.

PR #2639, opened 2018-08-29
[DCOS-40654] When listing reserved resources, fail test if we leaked some.

porridge:
This should let sdk_diag get the log for uninstall scheduler to help us
establish the cause for those leaks.

This is already implemented on master in a different way.

### DIFF
--- a/testing/sdk_utils.py
+++ b/testing/sdk_utils.py
@@ -25,18 +25,30 @@ def get_service_name(default: str) -> str:
     return os.environ.get("INTEGRATION_TEST__SERVICE_NAME") or default
 
 
-def list_reserved_resources():
-    '''Displays the currently reserved resources on all agents via state.json;
-       Currently for INFINITY-1881 where we believe uninstall may not be
-       always doing its job correctly.'''
-    state_json_slaveinfo = dcos.mesos.DCOSClient().get_state_summary()['slaves']
+def list_reserved_resources(service_name: str) -> bool:
+    """
+    Logs all reserved resources and returns a true value if there are some for this service's role.
 
-    for slave in state_json_slaveinfo:
+    The reservations are retrieved from state.json.
+    This is for debugging DCOS-40654.
+    """
+    service_role = service_name.strip("/").replace("/", "__") + "-role"
+    state_json_slave_info = dcos.mesos.DCOSClient().get_state_summary()['slaves']
+    reserved_resources_messages = []
+    found_for_this_service = False
+    msg = 'on slaveid=%s hostname=%s reserved resources: %s'
+    for slave in state_json_slave_info:
         reserved_resources = slave['reserved_resources']
-        if reserved_resources == {}:
+        if not reserved_resources:
             continue
-        msg = 'on slaveid=%s hostname=%s reserved resources: %s'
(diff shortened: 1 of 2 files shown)


testing/sdk_utils.py
```diff
@@ -25,18 +25,30 @@ def get_service_name(default: str) -> str:
...
-    state_json_slaveinfo = dcos.mesos.DCOSClient().get_state_summary()['slaves']
+def list_reserved_resources(service_name: str) -> bool:
+    """
+    Logs all reserved resources and returns a true value if there are some for this service's role.
 
-    for slave in state_json_slaveinfo:
+    The reservations are retrieved from state.json.
+    This is for debugging DCOS-40654.
+    """
+    service_role = service_name.strip("/").replace("/", "__") + "-role"
```
elezar: I think there is a `get_role()` function in `sdk_utils` now.

testing/sdk_utils.py
```diff
@@ -25,18 +25,30 @@ def get_service_name(default: str) -> str:
...
-        if reserved_resources == {}:
+        if not reserved_resources:
             continue
-        msg = 'on slaveid=%s hostname=%s reserved resources: %s'
-        log.info(msg % (slave['id'], slave['hostname'], reserved_resources))
+        reserved_resources_messages.append(msg % (slave['id'], slave['hostname'], reserved_resources))
+        if service_role in reserved_resources:
+            found_for_this_service = True
+    if reserved_resources_messages:
+        log.info('Found following reserved resources post uninstall.')
```
elezar: Would it be useful to print the name of the service we are uninstalling?

Nit: This function doesn't really have anything with uninstalling.

testing/sdk_utils.py
```diff
@@ -25,18 +25,30 @@ def get_service_name(default: str) -> str:
     return os.environ.get("INTEGRATION_TEST__SERVICE_NAME") or default
 
 
-def list_reserved_resources():
-    '''Displays the currently reserved resources on all agents via state.json;
-       Currently for INFINITY-1881 where we believe uninstall may not be
-       always doing its job correctly.'''
-    state_json_slaveinfo = dcos.mesos.DCOSClient().get_state_summary()['slaves']
+def list_reserved_resources(service_name: str) -> bool:
```
elezar: If we only call this on uninstall and with the service name, we could change the name of the function.

testing/sdk_utils.py
elezar: (scratch that ... that was in `master`)

porridge:
Yes, this is to make tests fail in the guilty test, and therefore gather useful logs.
Changes pushed, PTAL.

porridge:
FTR I've cherry-picked your CI fixes into a separate PR:
https://github.com/mesosphere/dcos-commons/pull/2644

Once that merges, I will rebase this one, to keep the changes separate.

porridge:
Rebased to see if we can get a successful run.

porridge:
I'm holding off on merging this until we fix the root cause, since it would probably just make all of our CI red, given how frequently the new code finds a leak.

kaiwalyajoshi:
Closing this PR as we're really trying to tombstone all frameworks on SDK 0.40.0 branches.

PR #2641, opened 2018-08-29
Update code checks

elezar:
The code checks also need to be updated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,19 @@
-FROM ubuntu:16.04
-RUN apt-get update && apt-get install -y \
-    git \
+FROM ubuntu:18.04
+
+ENV GO_VERSION=1.10.2
+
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,53 +1,78 @@
-FROM ubuntu:16.04
-RUN apt-get update && apt-get install -y \
-    git \
+FROM ubuntu:18.04
```
elezar: There is some git bug that means that these are shown in the diff even though it's already been merged ...

porridge:
FTR this can probably be dropped in favor of https://github.com/mesosphere/dcos-commons/pull/2644
BTW, It's interesting that the tests passed at c371110, i.e. before the code-checks fix commit?

elezar:
Closing in favour of #2644

PR #2643, opened 2018-08-30
[DCOS-39710] Investigate kafka acl test failures

mockingbird2:
Extracted the failure detection from #2636 to better track the changes.

### DIFF
--- a/frameworks/kafka/tests/auth.py
+++ b/frameworks/kafka/tests/auth.py
@@ -170,6 +170,10 @@ def write_to_topic(
             if "LEADER_NOT_AVAILABLE" in stderr:
                 LOG.warning("Write failed with LEADER_NOT_AVAILABLE")
                 return True
+            if not is_not_authorized(output):
+                LOG.warning("Write failed with AuthorizationException")
+                return False
+            raise Exception("Not yet captured error while sending")
 
         LOG.info("Write appears to have succeeded")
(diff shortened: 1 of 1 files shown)


frameworks/kafka/tests/auth.py
```diff
@@ -170,6 +170,10 @@ def write_failed(output) -> bool:
             if "LEADER_NOT_AVAILABLE" in stderr:
                 LOG.warning("Write failed with LEADER_NOT_AVAILABLE")
                 return True
+            if not is_not_authorized(output):
```
elezar: Note that `output` is a `tuple` and `is_not_authorized` accepts a string. *Yay for interpreted dynamic typed languages*

frameworks/kafka/tests/auth.py
```diff
@@ -170,6 +170,10 @@ def write_failed(output) -> bool:
...
+                LOG.warning("Write failed with AuthorizationException")
+                return False
+            raise Exception("Not yet captured error while sending")
```
elezar: What about `"Unexpected error while sending"`?

elezar: Does reading need similar changes?

mockingbird2: In the read method, all the conditional branches are handled.

PR #2644, opened 2018-08-30
[DCOS-41163] Short-term fixes to make PR tests against 0.40 work again.

porridge:


porridge:
@elezar does not look like that second commit actually fixes the code-checks problem?
`[13:06:19]Unable to find image 'dcos-commons:0be01f9' locally`

elezar:
Yes, I saw this and dug a little bit. We have the following issues.
1. The docker build step for the code checks actually fails, but because we don't have `set -e` in the configured script the build continues.
2. This means that the image that is supposed to be used is never build and thus doesn't exists.

The reason for 1. is that the step here: https://github.com/mesosphere/dcos-commons/pull/2644/files#diff-3254677a7917c6c01f55212f86c57fbfR60 *requires* a git repository in the current folder. Looking at the `.dockerignore` file we, however, have: https://github.com/mesosphere/dcos-commons/blob/sdk-0.40/.dockerignore#L5 (which was removed in `master`). I would *assume* that this is because of the function here: https://github.com/mesosphere/dcos-commons/blob/sdk-0.40/sdk/common/build.gradle#L25 which should be updated to handle the case where the current folder is *not* a git repository.

I have confirmed this locally by running:
```bash
➜  dcos-commons git:(master) mv .git not.git
➜  dcos-commons ./gradlew testClasses
Parallel execution is an incubating feature.

> Configure project :scheduler
fatal: Not a git repository (or any of the parent directories): .git


FAILURE: Build failed with an exception.

* Where:
Build file '/private/tmp/dcos-commons/sdk/scheduler/build.gradle' line: 9

* What went wrong:
A problem occurred evaluating project ':scheduler'.
> Process 'command 'git'' finished with non-zero exit value 128

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 4s
```

The immediate fix for this PR: Remove the `.*` line from `.dockerignore`.

Follow-up steps:
* Fix the `getGitHash` function.
* Add `set -e` to the builds config.

porridge:
Hm didn't help. https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=1186922&_focus=2761

Let me try different syntax.

porridge:
Actually, docker seems to obey that directive, the problem must be elsewhere:
```
porridge@beczulka:~/Pulpit/work/mesosphere/code/dcos-commons$ docker build .
Sending build context to Docker daemon    126MB
Step 1/34 : FROM ubuntu:18.04
 ---> 16508e5c265d
Step 2/34 : RUN mkdir /tmp/repo/
 ---> Running in 63bdac07e4ec
Removing intermediate container 63bdac07e4ec
 ---> 59ede682bd51
Step 3/34 : COPY / /tmp/repo/
 ---> 9d2fc6e232ce
Step 4/34 : RUN ls -al /tmp/repo && false
 ---> Running in 016aa08779c8
total 184
drwxr-xr-x 1 root root  4096 Aug 30 13:05 .
drwxrwxrwt 1 root root  4096 Aug 30 13:05 ..
drwxr-xr-x 8 root root  4096 Aug 30 12:51 .git
```

porridge:
It could also be that the docker client we use on TC does not understand `!`.

elezar:
> It could also be that the docker client we use on TC does not understand !.

Do we know what docker version TC is running?

porridge:
> > It could also be that the docker client we use on TC does not understand !.
> 
> Do we know what docker version TC is running?

I don't, and it's not visible in the build log.

porridge:
```
[15:22:31][Step 4/6] An error has occurred: FatalError: git failed. Is it installed, and are you in a Git repository directory?
[15:22:31][Step 4/6] Check the log at /root/.cache/pre-commit/pre-commit.log
[15:22:31][Step 4/6] The command '/bin/sh -c cd /tmp/repo/ &&     ./gradlew testClasses &&     pre-commit install-hooks &&     cd / &&     rm -rf /tmp/repo/' returned a non-zero code: 1
[15:22:31][Step 4/6] + echo '
[15:22:31][Step 4/6] ##teamcity[setParameter name='\''DOCKER_IMAGE'\'' value='\''dcos-commons:c5714ef'\'']
[15:22:31]
[Step 4/6] Incorrect property name.
Valid property list format is (name( )*=( )*'escaped_value'( )*)* where escape symbol is "|"
[15:22:31][Step 4/6] '
```

Dockerfile
```diff
@@ -1,53 +1,78 @@
...
 # make a dir for holding the SSH key in tests
 RUN mkdir /root/.ssh
 
+# Copy all of the repo into the image, then run some build/lint commands against the copy to heat up caches. Then delete the copy.
+RUN mkdir /tmp/repo/
+COPY / /tmp/repo/
+# gradlew: Heat up jar cache. pre-commit: Heat up lint tooling cache.
+RUN cd /tmp/repo/ && \
+    ./gradlew testClasses && \
+    pre-commit install-hooks && \
```
elezar: let's add the `git init && \` line here as with #2645 as it should unblock this.

__init__.py
```diff
@@ -1,6 +1,6 @@
 import sys
 import os.path
 
-# Add /testing/ to PYTHONPATH:
+# Add /testing/ directory to PYTHONPATH:
```
elezar: We could drop this change as we should still trigger the *build* of the docker image for the Check job: https://teamcity.mesosphere.io/admin/editTriggers.html?id=buildType:DcosIo_DcosCommons_PrJobs_RunCodeChecks which is what was causing the earlier failures.

porridge:
I have the changes ready. Waiting for https://github.com/mesosphere/dcos-commons/pull/2646 to merge before pushing to trigger tests again.

elezar:
> I have the changes ready. Waiting for #2646 to merge before pushing to trigger tests again.

I don't think that #2646 is going to pass without the additional `Dockerfile` change -- as is indicated by the failing Cassandra tests.

Note from the build there:
```
[10:25:43]	[Step 6/9] ++ egrep '^(Dockerfile|test_requirements.txt|.dockerignore)$'
[10:25:43]	[Step 6/9] ++ git diff --name-only origin/master
[10:25:43]	[Step 6/9] + matched_files=.dockerignore
[10:25:43]	[Step 6/9] + set -e
[10:25:43]	[Step 6/9] + [[ -n .dockerignore ]]
```
Meaning that we will *also* have to copy the `.dockerignore` file from `master` in this case (or remove `.dockerignore` from the list of files checked.

PR #2645, opened 2018-08-30
[DCOS-41166] Make getGitSHA robust for use in non-git folders

elezar:
* Use UNKNOWN as git sha for scheduler JAR in case of error
* Don't pass `.*` to Docker context by default

When executing in a non-git folder or when building the docker image we now see:
```bash
./gradlew testClasses
Parallel execution is an incubating feature.

> Configure project :scheduler
fatal: Not a git repository (or any of the parent directories): .git
An error occurred while calculating the git hash. Using %s
```

See related PRs:
* #2623 where the implicit requirement that this support a non-git repo folder was introduced.
* #2644 where this caused issues

PR #2646, opened 2018-08-30
[DCOS-41166] Use UNKNOWN as git sha for scheduler JAR in case of error

elezar:
Makes a change equivalent to the Gradle change in #2645 to the `sdk-0.40` branch. This is required to unblock #2644 which in-turn unblocks #2639


elezar:
I'm closing since #2644 is green.

PR #2647, opened 2018-08-30
Use UNKNOWN as git sha for scheduler JAR in case of error

elezar:
Makes a change equivalent to the Gradle change in #2645 to the sdk-0.40 branch. This is required to unblock #2644 which in-turn unblocks #2639

We would still need to make the change to the docker file for `pre-commit`.


elezar:
@porridge due to the Dockerfile differences to master, this isn't going to pass any of of the tests that aren't `code-checks`. Should we merge this into your branch and then get the issues resolved there?

PR #2650, opened 2018-08-31
DCOS-38256 Fixed bug where if Port env-key was defined, env based secrets were empty.

jeid64:
I added a port env-key to the secrets test to test this as an integration in the future, and wrote some unit tests for withEnvVar.

### DIFF
--- a/frameworks/helloworld/src/main/dist/secrets.yml
+++ b/frameworks/helloworld/src/main/dist/secrets.yml
@@ -26,6 +26,10 @@ pods:
                echo hello >> hello-container-path/output && sleep $SLEEP_DURATION
         cpus: {{HELLO_CPUS}}
         memory: {{HELLO_MEM}}
+          ports:
+            test:
+              port: 0
+              env-key: PORT_ENV_VAR
         volume:
           path: hello-container-path
           type: ROOT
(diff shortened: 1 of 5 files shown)


frameworks/helloworld/src/main/dist/secrets.yml
```diff
@@ -26,6 +26,10 @@ pods:
                echo hello >> hello-container-path/output && sleep $SLEEP_DURATION
         cpus: {{HELLO_CPUS}}
         memory: {{HELLO_MEM}}
+          ports:
```
nickbp: Indent is off

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/taskdata/EnvUtils.java
```diff
@@ -23,11 +23,11 @@ private EnvUtils() {
      * In the event of duplicate labels, the last duplicate wins.
      * This is the inverse of {@link #toProto(Map)}.
      */
-    public static Map<String, String> toMap(Environment environment) {
+    public static Map<String, Environment.Variable> toMap(Environment environment) {
```
nickbp: Maybe just fold this into withEnvVar? Looks like nothing else is calling it. The javadoc is also slightly inaccurate now as it's not quite the inverse of `toProto()` anymore.

jeid64: toMap does get called by a few tests that I changed to handle the new type passed in, so I was going to keep it. The tests need it anyway.

PR #2651, opened 2018-09-04
Wait for deploy plan to kick off before checking for task state.

jeid64:
https://jira.mesosphere.com/browse/DCOS-40376
https://jira.mesosphere.com/browse/DCOS-39468

Bumped up wait timeout to last through the schedulers DNS resolution. 
Also added sdk_plan.wait_for_kicked_off_deployment which should wait until the apiServer is up and responding.

### DIFF
--- a/frameworks/helloworld/tests/test_taskcfg.py
+++ b/frameworks/helloworld/tests/test_taskcfg.py
@@ -43,7 +43,7 @@ def test_deploy():
     # wait for new TASK_FAILEDs to appear:
     @retrying.retry(
         wait_fixed=1000,
-        stop_max_delay=1000*wait_time_in_seconds,
+        stop_max_delay=1000 * wait_time_in_seconds,
         retry_on_result=lambda res: not res)
     def wait_for_new_failures():
         new_state_history = _get_state_history(task_name)


frameworks/helloworld/tests/test_taskcfg.py
```diff
@@ -29,7 +30,8 @@ def configure_package(configure_security):
 
 @pytest.mark.sanity
 def test_deploy():
-    wait_time = 30
+    wait_time = 600
```
elezar: Minor comment: `wait_time_in_seconds`?

frameworks/helloworld/tests/test_taskcfg.py
```diff
@@ -29,7 +30,8 @@ def configure_package(configure_security):
...
+    sdk_plan.wait_for_kicked_off_deployment(config.SERVICE_NAME)
```
elezar: As @porridge points out in https://jira.mesosphere.com/browse/DCOS-40639 this may not be enough to ensure scheduler health, but that is an orthogonal problem.

jeid64: Yeah, for this use case wait_for_kicked_off_deployment should be fine. Would be good to revisit for a future PR.

frameworks/helloworld/tests/test_taskcfg.py
jeid64: Changed!

frameworks/helloworld/tests/test_taskcfg.py
```diff
@@ -30,18 +31,20 @@ def configure_package(configure_security):
 
 @pytest.mark.sanity
 def test_deploy():
-    wait_time = 30
+    wait_time_in_seconds = 600
+    sdk_plan.wait_for_kicked_off_deployment(config.SERVICE_NAME)
```
porridge: Shouldn't this be in a shared fixture?

PR #2652, opened 2018-09-04
[DCOS-41242] Download bootstrap URI from scheduler for DNS lookup for multiport test.

jeid64:
https://github.com/mesosphere/dcos-commons/pull/2627/files
This PR added bootstrap to cmd to wait for DNS resolution of the pod before trying to do networking. Unfortunately it didn't add the URI for the bootstrap binary, so ./bootstrap was unable to be found. This adds the URI so the binary is pulled into the sandbox, and should fix DCOS-41242.

### DIFF
--- a/frameworks/helloworld/src/main/dist/multiport.yml
+++ b/frameworks/helloworld/src/main/dist/multiport.yml
@@ -5,6 +5,8 @@ scheduler:
 pods:
   multiport:
     count: {{HELLO_COUNT}}
+    uris:
+      - {{BOOTSTRAP_URI}}
     resource-sets:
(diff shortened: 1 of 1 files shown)


porridge:
It really feels like `set -e` should be mandatory or at least default in `cmd`.
Do we have a wishlist of compatibility-breaking changes for sdk 1.0?
/cc @nickbp

nickbp:
If the developer added that to the start of their cmd, would it work as expected? If so I think it'd be fine to have the developer do it. There can be cases where -e behaves in surprising ways, and its wouldn't be the end of the world if the developer enabled it explicitly

PR #2653, opened 2018-09-04
DCOS-38256 Backport Fixed bug where if Port env-key was defined, env based secrets were empty.

jeid64:
Backport to 0.40

* DCOS-38256 Fixed bug where if Port env-key was defined, env based
secrets had an empty string value.


sdk/common/src/main/java/com/mesosphere/sdk/offer/taskdata/EnvUtils.java
```diff
@@ -40,8 +40,8 @@ public static Environment toProto(Map<String, String> environmentMap) {
         Environment.Builder envBuilder = Environment.newBuilder();
         for (Map.Entry<String, String> entry : environmentMap.entrySet()) {
             envBuilder.addVariablesBuilder()
-                .setName(entry.getKey())
-                .setValue(entry.getValue());
+                    .setName(entry.getKey())
```
kvish: misindent

PR #2656, opened 2018-09-06
[DCOS-40696] log if scheduler is about to hit a null reference

kvish:
Log if the scheduler is about to hit a nullref

this is to help us detect the bug in https://jira.mesosphere.com/browse/DCOS-40696

if we determine this is the correct null ref we can work backwords to detect why it was null in that case.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
@@ -2,6 +2,7 @@ package com.mesosphere.sdk.scheduler;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.TextFormat;
+import com.mesosphere.sdk.framework.ProcessExit;
 import com.mesosphere.sdk.http.types.EndpointProducer;
 import com.mesosphere.sdk.offer.LoggingUtils;
 import com.mesosphere.sdk.scheduler.plan.Plan;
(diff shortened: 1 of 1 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -127,6 +127,9 @@ public ClientStatusResponse getClientStatus() {
                     inProgressSteps.stream().map(step -> step.getMessage()).collect(Collectors.toList()));
         }
         activeWorkSet.addAll(inProgressSteps);
+        if (workSetTracker == null) {
```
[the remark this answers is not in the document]
kvish: trying to nail down why the scheduler hit a null ref outlined here https://jira.mesosphere.com/browse/DCOS-40696, workset tracker is initialized in the registered call ( public void registered(boolean reRegistered on line 107) the theory is that the scheduler hit this block without initializing workset tracker, if we see this log line it will indicate that this theory is true

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -127,10 +127,13 @@ public ClientStatusResponse getClientStatus() {
                     inProgressSteps.stream().map(step -> step.getMessage()).collect(Collectors.toList()));
         }
         activeWorkSet.addAll(inProgressSteps);
-        if (workSetTracker == null) {
-            logger.warn("workset tracker is uninitialized, scheduler will hit a null reference and exit.");
+
+        try {
+            workSetTracker.updateWorkSet(activeWorkSet);
+        } catch (NullPointerException e) {
+            logger.warn("workset tracker is uninitialized, scheduler has his a null reference and is exiting.");
```
mpereira: Won't this catch will prevent the scheduler from exiting?

kvish: yes, need to add an exit

PR #2657, opened 2018-09-06
DCOS-39567 Show SDK Plan before performing uninstall.

kaiwalyajoshi:
DCOS-39567
Display SDK deploy plan before performing uninstall. This is to help diagnose hanging uninstalls.

### DIFF
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -303,7 +303,7 @@ def uninstall(package_name, service_name):
     start = time.time()
 
     log.info("Uninstalling {}".format(service_name))
-   
+
     # Display current SDK Plan before uninstall, helps with debugging stuck uninstalls.
     log.info("Current plan status for {}".format(service_name))
     sdk_cmd.svc_cli(package_name, service_name, check=False)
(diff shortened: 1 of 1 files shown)


testing/sdk_install.py
```diff
@@ -303,6 +303,10 @@ def uninstall(package_name, service_name):
     start = time.time()
 
     log.info("Uninstalling {}".format(service_name))
+   
+    # Display current SDK Plan before uninstall, helps with debugging stuck uninstalls.
+    log.info("Current plan status for {}".format(service_name))
+    sdk_cmd.svc_cli(package_name, service_name, check=False)
```
takirala: i think you forgot to add the command it self. 
```
sdk_cmd.svc_cli(package_name, service_name, "plan deploy status", check=False)
```

takirala:
FYI, we have some [linter rules](https://github.com/mesosphere/dcos-commons/blob/master/TESTING.md#static-checking) which you will be able to hook in to `git commit` to catch errors early (if any).

elezar:
@kaiwalyajoshi we have disabled the metrics tests for the time being, so you should be able to trigger the tests again.

PR #2659, opened 2018-09-09
[DCOS-44176] More linting rules on java source set

takirala:
In the interest of upcoming SDK refactoring, I feel this is a good time to revisit our style checks as we refactor the codebase. I have added some style checks to start a discussion. THIS IS NOT FINAL. We can discuss in this PR and decide what works for all of us and then we can adapt those changes during the refactor. I have made sure to retain as many rules as I can from earlier `checkstyle.xml`. However, there was been few instance where I changed a rule. But for the most part of the diff, I have only added new rules and sorted them by sections (a comment consisting of `url` being the implicit header for that section). I have consciously avoided adding comments as I feel like many of the rule names are self explanatory and adding comments in a file of 400+ lines would decrease the readability. Also, this is not something we read on a regular basis. Many of the rules are just a google search away to read and understand what they do (And i have provided the urls in the comments).

Feel free to comment / commit to this PR as you wish.
 
I understand this PR may be a lot of decisions to make, so I have organized the changes by sections in http://checkstyle.sourceforge.net/checks.html in the `checkstyle.xml` as follows :

```
http://checkstyle.sourceforge.net/config_annotation.html
http://checkstyle.sourceforge.net/config_blocks.html
http://checkstyle.sourceforge.net/config_design.html
http://checkstyle.sourceforge.net/config_coding.html
http://checkstyle.sourceforge.net/config_header.html
http://checkstyle.sourceforge.net/config_imports.html
http://checkstyle.sourceforge.net/config_javadoc.html
http://checkstyle.sourceforge.net/config_metrics.html
http://checkstyle.sourceforge.net/config_misc.html
http://checkstyle.sourceforge.net/config_modifier.html
http://checkstyle.sourceforge.net/config_naming.html
http://checkstyle.sourceforge.net/config_regexp.html
http://checkstyle.sourceforge.net/config_sizes.html
http://checkstyle.sourceforge.net/config_whitespace.html
```

For where the rules needed a numerical value (For e.g.: Maximum method length), I have used sensible defaults but we can change them as we see fit. I feel that some of these style rules would make extending our code and testing easier. We can discuss this in slack too if you think it would be easier to communicate.


takirala:
Thanks for the reviews @elezar and @kaiwalyajoshi 

I am adding `TODO` in the `checkstyle.xml` for rules I don't want to fix in this PR. For example ideally, we should have a cyclomatic complexity of less than 10. But If i decide to breakdown every method that has cyclomatic complexity of > 10 it would bloat this PR's scope and may introduce unexpected logic changes as well if I am not careful enough. I want to add as many rules as possible with `TODO`s and then we can address each of them at our own pace. Once I am done, the diff of this PR would be large but i expected very less to none logic changes. I am planning for the diff to be only reformatting changes.

takirala:
- Fully formatted all the `frameworks/*` java files with new linting rules.
- All new violations would be reported as `errors`, we can change the severity if this turns out to be a pain (i don't expect it to).
- Added several TODO's and suppressed multiple rules in the source code. Feel free to fix them in future PRs whenever we modify the relevant files.
- Excluded the test files from linting rules. We can address this in future PRs.
- Also added a small guide on how the rules can be enforced. We don't need any new CI changes as one of the existing CI step (`gradle check`) already covers this.

@elezar @kaiwalyajoshi @kvish the diff is huge but I did not make any logic changes (The CI passed on [first run](https://teamcity.mesosphere.io/viewLog.html?buildId=1272118&buildTypeId=DcosIo_DcosCommons_Sdk_BuildAndTestPr&tab=dependencies#_expand=block_bt1289_bt992_bt993_bt994_bt995_bt996-1272120_1272124_1272127_1272130_1272133_1272136) :) ).

I really wanted to enforce the rule where we have only one parameter per line in method declaration but looks like `checkstyle.xml` does not have a provision for this ([said so by one of checkstyle contributors](https://stackoverflow.com/questions/44443336/what-rules-do-i-need-to-enforce-each-method-parameter-on-a-new-line-with-checkst)). I tried to include a [custom check](https://gist.github.com/pbludov/28f3cec9c9382e7eca0d79f43666cbf4) to enforce this but it looks like a new module/jar [is needed](http://checkstyle.sourceforge.net/writingchecks.html) to do that and it felt like an overkill to do this. May be we can look at [sevntu-checkstyle](https://github.com/sevntu-checkstyle/sevntu.checkstyle/) in future. The same would be nice to enforce for method chaining. This is especially helpful with java 8 lambdas and method chaining where intellij shows each method return type: 
screen shot 2018-10-26 at 4 12 31 pm
May be for now, we consciously enforce this style ourselves wherever we seem fit.

takirala:
Merging this. We can always revert/fix if there are any other unexpected implications.

PR #2662, opened 2018-09-11
[DCOS-41604, DCOS-41603] Fix cryptic message and relax failed tasks threshold.

porridge:
DCOS-41603 shows a case where a threshold of 2 failure is much too
strict.

### DIFF
--- a/testing/sdk_plan.py
+++ b/testing/sdk_plan.py
@@ -6,6 +6,7 @@ SHOULD ALSO BE APPLIED TO sdk_plan IN ANY OTHER PARTNER REPOS
 ************************************************************************
 """
 
+import datetime
 import logging
 import retrying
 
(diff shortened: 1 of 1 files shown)


testing/sdk_plan.py
```diff
@@ -155,8 +157,10 @@ def fn():
         failures = sdk_tasks.get_failed_task_count(service_name)
         if failures - initial_failures > MAX_NEW_TASK_FAILURES:
             log.error(
-                "Service %s exceeded failures increase while doing %s and trying to reach %s",
+                "Tasks in service %s failed %d times since starting %ds ago to wait for %s to reach %s, aborting.",
```
samvantran: you may want to append the letter `s` to `wait for %s to reach` else your message will look something like
`wait for 26.252 for deploy to reach COMPLETE`

samvantran: ah you're right, sorry about that

elezar:
@porridge I have kicked off the builds that had failing metrics tests as this is still causing some issues in the nightlies.

PR #2663, opened 2018-09-12
[TESTS] Disable metrics tests on 1.12

elezar:
This PR just disables the metrics tests (which are failing) on 1.12.

For the record, I also have #2664 to add a hello-world smoke test for this so that we can easily test alternative fixes.


elezar:
Thanks @samvantran. Please merge this once it's green and backport so that we can unblock our CI.

PR #2666, opened 2018-09-13
[DCOS-41803, DCOS-41812] Updated login_session to use new CLI features.

ArmandGrillet:


porridge:
This would break if we later change to using older CLIs with older clusters, correct?
/cc @takirala

ArmandGrillet:
@porridge This will work for DC/OS clusters >= 1.10, that's the main feature of the new CLI.

tools/dcos_login.py
```diff
@@ -202,16 +113,7 @@ def ignore_empty(envvar, default):
...
-    dcos_acs_token = os.environ.get("DCOS_ACS_TOKEN")
-    if not dcos_acs_token:
-        dcos_acs_token = login(
-            dcosurl=cluster_url,
-            username=dcos_login_username,
-            password=dcos_login_password,
-            is_enterprise=dcos_enterprise,
-        )
-    configure_cli(dcosurl=cluster_url, token=dcos_acs_token)
+    subprocess.run(["dcos", "cluster", "setup", cluster_url, "--username", dcos_login_username, "--password", dcos_login_password])
```
elezar: One thing to note is that this doesn't handle the case where we run this against a cluster that is already attached / setup.

bamarni: Under what conditions these tests can be run? With the new CLI the cluster setup command became idempotent, re-running it will just overwrite the previous config if any.

Otherwise it is also possible to run `dcos cluster list --attached --json` to see if there is any attached cluster and skip the setup based on this.

porridge:
Also apparently it will not work with 1.9 clusters, even with 1.12's CLI.

elezar:
I know that @jeschkies has recently spent quite a bit of time addressing this flow in `shakedown`. Would it make sense to consolidate this knowledge there so that this would work out the box for anyone that wants to use it -- otherwise the Marathon team will end up with the same issues as soon as they start using the new CLI.

tools/dcos_login.py
elezar: Thanks for the info. That will simplify things a lot. Then we only have to handle 1.9 clusters as an edge case.

ArmandGrillet:
Closing this as the first solution investigated to solve the issue is in review and will be easier to use on your side: https://github.com/dcos/dcos-cli/pull/1335

PR #2665, opened 2018-09-12
[TESTS] Backport Disable metrics tests on 1.12 (#2663)

samvantran:
Backports PR #2663 which disables the metrics tests (which are failing) on 1.12.

PR #2667, opened 2018-09-13
[DCOS-41808] Make flake8 and pylint happy

porridge:
- Add flake8 config from master
- Apply autopep8 -i to all python files
- Fix up remaning flake8 complaints manually


elezar:
I just pushed another fix for a `pytest.raises` syntax issue: The exception was a string and not a class: 🤦‍♂️

elezar:
@porridge looks like it's just the (expected) metrics tests falling. Merge?

porridge:
Merging as agreed to break the dependency loop with https://github.com/mesosphere/dcos-commons/pull/2665

PR #2668, opened 2018-09-13
DCOS_OSS-4126 Initial HDFS transparent encryption

wolf31o2:
This PR adds the configuration options for configuring an external
HDFS KMS provider, to enable HDFS transparent encryption.

Signed-off-by: Chris Gianelloni <cgianelloni@applause.com>


wolf31o2:
Looks like this property is already configurable in more recent versions of this. Closing.

PR #2670, opened 2018-09-15
[DCOS-41803] Add tooling to support to support the new plugin based CLI

takirala:
TODO : 

- [ ] Should we use the `../latest/..` binary of CLI?
- [ ] Test this works with Open/EE 1.9 and 1.12 clusters (May be run nightlies manually with this custom PR/branch)
- [ ] Test the script works on multiple runs of integration tests with same cluster.


tools/dcos_login.py
```diff
@@ -203,16 +174,80 @@ def ignore_empty(envvar, default):
     dcos_login_username = ignore_empty("DCOS_LOGIN_USERNAME", __CLI_LOGIN_EE_USERNAME)
```
bamarni: If this was called `DCOS_USERNAME` it'd be picked up automatically by `dcos cluster setup` without the need to pass a flag. The password can be read from `DCOS_PASSWORD`.

tools/dcos_login.py
```diff
@@ -203,16 +174,80 @@ def ignore_empty(envvar, default):
...
+            _run_cmd("dcos config set core.dcos_acs_token {}".format(dcos_acs_token))
+    else:
+        if dcos_enterprise:
+            _run_cmd("dcos cluster setup {} --username {} --password {} --insecure".format(
+                cluster_url,
+                dcos_login_username,
+                dcos_login_password))
+        else:
+            dcos_acs_token = os.environ.get("DCOS_ACS_TOKEN", dcos_acs_token)
+            _run_cmd("env DCOS_ACS_TOKEN={} dcos cluster setup {} --insecure".format(
```
bamarni: It should also be possible to pass `__CLI_LOGIN_OPEN_TOKEN` to stdin in order to avoid retrieving the acs token manually. The only thing is that it'd open a browser when run on desktop environments.

tools/dcos_login.py
takirala: Is this the behavior for 1.9 cli as well (https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.9/dcos) ?

tools/dcos_login.py
```diff
@@ -203,16 +143,100 @@ def ignore_empty(envvar, default):
...
+                "dcos cluster setup {} --username {} --password {} --insecure".format(
+                    cluster_url,
+                    dcos_login_username,
+                    dcos_login_password
+                ),
+                check=True
+            )
+        else:
+            _run_cmd(
+                "env DCOS_ACS_TOKEN={} dcos cluster setup {} --insecure".format(
```
takirala: @bamarni is this the correct usage of the command for 1.10+ open clusters with the latest cli ?

tools/dcos_login.py
bamarni: No it isn't, this is something new.

tools/dcos_login.py
bamarni: Yep :+1:

takirala:
Except for the 1.10+ open cluster case (which has to be fixed in CLI), all other cases have been tested.

bamarni:
> It should also be possible to pass __CLI_LOGIN_OPEN_TOKEN to stdin directly to the `cluster setup` command in order to avoid retrieving the acs token manually. The only thing is that it'd open a browser when run on desktop environments.

@tarun any comment on the above?

takirala:
@bamarni 
- given that it may open browser when run directly on a dev machine, i guess it would be alright to stick to the current behavior
- the current api behavior works (and is not discouraged / deprecated - right ?!)
- i do think its trivial to enter input in python (without using sth like `pexpect`) and the current behavior enables us to call the script directly from teamcity CI if needed.

bamarni:
I guess that's fine to rely on the `DCOS_ACS_TOKEN` env var, however this is something we quickly added to unblock the situation and didn't really investigated / audited what is done with the CLI in these tests.

So we will probably not document this env var for now, how about we look at it again later, after the 1.12 release? Maybe when dropping support for 1.9 and the old CLI. My main concern here is that `DCOS_ACS_TOKEN` will be passed to plugins in the future (and contain the auth token for the current cluster), so if plugins then call `cluster setup` for some reasons they might get strange behaviours (because of the default env var inheritance with child processes).

takirala:
(discussed in slack, copy pasting) I just realized that this needs to be backported to a sdk40 as well (and it needs to support 1.9 to master as well) and it already ended up a bigger diff than I previously imagined. I will account for the "entering auth token via stdin" refactor when we deprecate 1.9 flow for good in future where we have less things to worry about.

tools/dcos_login.py
```diff
@@ -203,16 +143,100 @@ def ignore_empty(envvar, default):
...
+                        is_enterprise=dcos_enterprise,
+                    )),
+                    cluster_url),
+                check=True
+            )
+    _run_cmd("dcos node --json", check=True)
+
+
+def check_and_install_native_cli(cluster_url: str) -> bool:
+    old_cli = "https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.9/dcos"
```
takirala: This may look ugly, but this is "to be deprecated".

takirala:
@kvish created https://jira.mesosphere.com/browse/DCOS-42159 to address the deprecation.

tools/dcos_login.py
```diff
@@ -187,32 +27,78 @@ def login_session() -> None:
...
+                dcos_login_password
+            ),
+            check=True)
+    else:
+        # This would try to use `xdg-open` and print a warning that it was not found in the PATH,
+        # but reads stdin and cluster will setup successfully.
+        rc, _, _ = _run_cmd(
+            "dcos cluster setup {} --insecure".format(cluster_url),
+            check=True,
+            cmd_input=bytes(os.environ.get(__DCOS_ACS_TOKEN_ENV, __CLI_LOGIN_OPEN_TOKEN), encoding="utf-8"))
```
bamarni: On DC/OS open the expected input token is the login token, it is then sent to the login API in order to get an ACS token back, so I think it should only be : `cmd_input=bytes(__CLI_LOGIN_OPEN_TOKEN, encoding="utf-8"))`

Otherwise does this mean you don't need the `DCOS_ACS_TOKEN` env variable support to skip login during cluster setup? If so I'd really like to remove support for it, or call it differently (`DCOS_CLUSTER_SETUP_ACS_TOKEN`), as the `DCOS_ACS_TOKEN` is already supported for other things and it might create strange behaviors.

tools/dcos_login.py
```diff
@@ -187,32 +27,78 @@ def login_session() -> None:
...
-        dcos_acs_token = login(
-            dcosurl=cluster_url,
-            username=dcos_login_username,
-            password=dcos_login_password,
-            is_enterprise=dcos_enterprise,
-        )
-    configure_cli(dcosurl=cluster_url, token=dcos_acs_token)
+    if ignore_empty("DCOS_ENTERPRISE", "true").lower() == "true":
+        _run_cmd(
+            "dcos cluster setup {} --username {} --password {} --insecure".format(
```
jeschkies: Just out of curiosity, why no log in via the environment variables? Checkout https://github.com/mesosphere/marathon/blob/master/tests/shakedown/shakedown/dcos/command.py#L206. I'd be more than happy to switch over to `dcos-commons`.

tools/dcos_login.py
```diff
@@ -187,32 +27,78 @@ def login_session() -> None:
...
+        # but reads stdin and cluster will setup successfully.
+        rc, _, _ = _run_cmd(
+            "dcos cluster setup {} --insecure".format(cluster_url),
+            check=True,
+            cmd_input=bytes(os.environ.get(__DCOS_ACS_TOKEN_ENV, __CLI_LOGIN_OPEN_TOKEN), encoding="utf-8"))
+    _run_cmd("dcos node --json", check=True)
+
+
+def _run_cmd(cmd, check=False, cmd_input=None):
+    log.info("[CMD] {}".format(cmd))
```
jeschkies: This might leak credentials since the command is called with `--username {}` and `--password {}`. I really would avoid this type of authentication. What's the reason not to use environment variables?

tools/dcos_login.py
jeschkies: IMHO, I'd not support 1.9 here. Are we not dropping support with 1.12 for 1.9 anyways?

takirala: Yes, I have removed this code now. But we still support 1.9 for sdk-40 branch (at least until 1.12 GA), so this is still relevant there.

takirala:
Thanks for the pointers @elezar and @jeschkies  I have addressed your concern of accessing username and password from the environment.

tools/dcos_login.py
bamarni: > Now, I want to use the same cluster to run the tests (which run inside a docker container so they have no knowledge of what happened previously). I am assuming i cannot use the __CLI_LOGIN_OPEN_TOKEN anymore.

Yes you can, actually it wouldn't work to pass the ACS token to stdin as it's expecting a login token.

The trick here is that `__CLI_LOGIN_OPEN_TOKEN` is a token which is always valid on Open clusters, it was created to make automated testing easier. see https://mesosphere.slack.com/archives/C07D97YFL/p1537271631000100

takirala:
@bamarni updated to use `DCOS_CLUSTER_SETUP_ACS_TOKEN` in accordance with https://github.com/dcos/dcos-cli/pull/1343/files

tools/dcos_login.py
```diff
@@ -187,32 +27,75 @@ def login_session() -> None:
...
-            return default
+        return value if value else default
+
+    if ignore_empty("DCOS_ENTERPRISE", "true").lower() == "true":
+        os.environ["DCOS_USERNAME"] = ignore_empty("DCOS_LOGIN_USERNAME", __CLI_LOGIN_EE_USERNAME)
+        os.environ["DCOS_PASSWORD"] = ignore_empty("DCOS_LOGIN_PASSWORD", __CLI_LOGIN_EE_PASSWORD)
+        _run_cmd("dcos cluster setup {} --insecure".format(cluster_url))
+    else:
+        # This would try to use `xdg-open` and print a warning that it was not found in the PATH,
+        # but reads stdin and cluster will setup successfully.
```
bamarni: this comment should be on top of the else condition block below.

PR #2679, opened 2018-09-20
DCOS-42191 Don't try unreserving pre-reserved resources, fix plan spam

nickbp:
- Check for a `resource_id` label when examining resources during uninstall. The immediate issue is a no-op where the scheduler tries to unreserve resources which are only statically pre-reserved, but it doesn't hurt to be conservative and always check for the `resource_id` label before unreserving. This should avoid the possibility of attempting to unreserve an offered resource that wasn't created by the SDK. Note that there are already upstream filters in FrameworkScheduler for filtering out Marathon volumes, so this check is effectively redundant in that case.
- Avoid (repeated) logging of the recovery plan when it's currently complete. It can be non-empty but complete in the case when prior recoveries had occurred within the lifespan of the scheduler process.

To be clear, neither of these issues should result in any problems for end users. The ultimate effect of this PR should just be a reduction in log noise and avoiding a Mesos noop.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
@@ -59,54 +59,42 @@ public class ResourceUtils {
         return new MesosResource(resource).getRole();
     }
 
-    public static Optional<String> getPrincipal(Protos.Resource resource) {
-        Optional<Protos.Resource.ReservationInfo> reservationInfo = getReservation(resource);
-
-        if (reservationInfo.isPresent()) {
-            return Optional.of(reservationInfo.get().getPrincipal());
-        } else {
-            return Optional.empty();
-        }
-    }
-
     public static Optional<Protos.Resource.ReservationInfo> getReservation(Protos.Resource resource) {
-        if (resource.getReservationsCount() > 0) {
-            return getRefinedReservation(resource);
-        } else {
-            return getLegacyReservation(resource);
-        }
-    }
-
-    private static Optional<Protos.Resource.ReservationInfo> getRefinedReservation(Protos.Resource resource) {
-        if (resource.getReservationsCount() == 0) {
-            return Optional.empty();
-        }
-
-        return Optional.of(resource.getReservations(resource.getReservationsCount() - 1));
-    }
-
-    private static Optional<Protos.Resource.ReservationInfo> getLegacyReservation(Protos.Resource resource) {
-        if (resource.hasReservation()) {
+        int count = resource.getReservationsCount();
+        if (count > 0) {
+            // This is a refined reservation against a pre-reserved resource. Reservation entries should in this order:
+            // 1. STATIC reservation for the pre-reserved role (e.g. slave_public)
+            // 2. DYNAMIC reservation for our refined role (e.g. slave_public/svc-role)
+            return Optional.of(resource.getReservations(count - 1));
+        } else if (resource.hasReservation()) {
+            // "Classic" reservation against a resource that isn't statically reserved.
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -59,54 +59,42 @@ public static String getRole(Protos.Resource resource) {
...
-        return Optional.of(resource.getReservations(resource.getReservationsCount() - 1));
     }
 
-    private static Optional<Protos.Resource.ReservationInfo> getLegacyReservation(Protos.Resource resource) {
-        if (resource.hasReservation()) {
-            return Optional.of(resource.getReservation());
-        } else {
-            return Optional.empty();
-        }
+    public static Optional<String> getPrincipal(Protos.Resource resource) {
```
takirala: (nit) can simplify this to
```
return getReservation(resource).map(Protos.Resource.ReservationInfo::getPrincipal);
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -59,54 +59,42 @@ public static String getRole(Protos.Resource resource) {
...
+                : Optional.empty();
     }
 
     public static Optional<String> getNamespace(Protos.Resource resource) {
         Optional<Protos.Resource.ReservationInfo> reservationInfo = getReservation(resource);
-        if (!reservationInfo.isPresent()) {
-            return Optional.empty();
-        }
-        return AuxLabelAccess.getResourceNamespace(reservationInfo.get());
+        return reservationInfo.isPresent()
```
takirala: (nit; duplicate comment)
```
return getReservation(resource).flatMap(AuxLabelAccess::getResourceNamespace);
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -59,54 +59,42 @@ public static String getRole(Protos.Resource resource) {
...
+                : Optional.empty();
     }
 
     public static Optional<String> getResourceId(Protos.Resource resource) {
         Optional<Protos.Resource.ReservationInfo> reservationInfo = getReservation(resource);
-        if (!reservationInfo.isPresent()) {
-            return Optional.empty();
-        }
-        return AuxLabelAccess.getResourceId(reservationInfo.get());
+        return reservationInfo.isPresent()
```
takirala: (nit; duplicate comment)
```
return getReservation(resource).flatMap(AuxLabelAccess::getResourceId);
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/recovery/DefaultRecoveryPlanManager.java
```diff
@@ -96,7 +96,8 @@ public void setPlan(Plan plan) {
     protected void setPlanInternal(Plan plan) {
         synchronized (planLock) {
             this.plan = plan;
-            if (!plan.getChildren().isEmpty()) {
+            // Avoid logging for non-empty completed plan, as is the case when a previous recovery had occurred:
+            if (!plan.getChildren().isEmpty() && !plan.isComplete()) {
```
takirala: 🎊 seen this spam myself before and got confused as well.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
nickbp: Aha, I was wondering what the shorthand version was. Fixed here and below. Thanks!

PR #2511, opened 2018-05-25
[DCOS-37660] Add initial diagnostics script.

mpereira:
**May 25:**
Initial WIP pushed:

**Jul 30:**
With the commits I just pushed I believe this is now a Good Enough™ diagnostics tool for SDK services.

What is done:

-   [X] Fetching of DC/OS version, Mesos tasks, Marathon apps, service configuration, pod status, plans statuses (deploy, recovery, backup-s3, etc.)
-   [X] Fetching of scheduler stderr/stdout logs
-   [X] Fetching of task stderr/stdout logs
-   [X] Fetching of task executor stderr/stdout logs
-   [X] Fetching of killed/failed task stderr/stdout logs (in theory, if Mesos hasn't GC'd the sandbox, the script should be able to do this)
-   [X] Fetching of base-tech specific diagnostics commands (Cassandra "nodetool status", Elastic "node stats", etc.)
-   [X] An easy way to extend base-tech specific diagnostics commands
-   [X] Retrying on all operations where it makes sense
-   [X] Temporary Docker image pushed to [mpereira/dcos-commons:diagnostics](https://hub.docker.com/r/mpereira/dcos-commons/)

I re-used some logic from sdk_diag for the fetching of log files. Some of the functions there were a bit too complected with tests (their original intent), so I couldn't use them verbatim. I believe some of the functions in the "create bundle" script could be moved to modules. I didn't really worry about that.

Possible future work:

-   [ ] Concatenate log files (e.g. stdout+stdout.1+stdout.2 -> stdout)
-   [ ] Create archive from output directory
-   [ ] Add interactive inputs and helpful output to make sure user is running diagnostics against the right cluster/service

Still want to do before merging the PR:

-   [ ] Provide command so that people can run something like (maybe provide a wrapper script?):

    ```bash
    docker run --rm mpereira/dcos-commons:diagnostics bash -c 'PYTHONPATH=/dcos-commons-dist/testing python3 /dcos-commons-dist/tools/create_service_bundle.py hdfs /data-services/hdfs'
    ```
-   [ ] Add docstrings to functions
-   [ ] Write usage documention
-   [ ] Use argparse

Example run:

```text
$ PYTHONPATH=testing ./tools/create_service_bundle.py hdfs /data-services/hdfs
...
(a bunch of output)
...
$ tree hdfs_data-services__hdfs_20180730222531
hdfs_data-services__hdfs_20180730222531
├── dcos_marathon_task_list.json
├── dcos_task.json
├── dcos_version.txt
[...]

PR #2680, opened 2018-09-20
[DCOS-42000] Add toolz as a test dependency.

mpereira:


frameworks/elastic/tests/test_rack_awareness.py
```diff
@@ -1,6 +1,7 @@
 import logging
-
 import pytest
+from toolz import get_in
```
elezar: Does this require a change to some `requirements.txt` file?

mpereira: [Yea](https://github.com/mesosphere/dcos-commons/pull/2680/files#diff-bdeaf6bb06566ad0d723cd60c75e563dR11).

mpereira:
Ping (CI green).

frameworks/elastic/tests/test_rack_awareness.py
elezar: Sorry, I missed it.

mpereira:
> Why we decided initially not to rely on this library and rather copy the method?

The first function (`get_in`) was a bit different than the one in toolz, and from then on, inertia.

frameworks/kafka/tests/client.py
```diff
@@ -2,14 +2,14 @@
 A collection of client utilites for Kafka.
 """
 import logging
+from toolz import merge as merge_dictionaries
```
elezar: Nit: We could call `toolz.merge` from `sdk_utils.merge_dictionaries` instead.

Alternatively: Since we're making the changes on every line anyway, we could just call `toolz.merge` directly too.

frameworks/elastic/tests/test_rack_awareness.py
elezar: Are you opposed to `import toolz` and then using `toolz.get_in`? The same number of lines will be changed.

elezar: (as with `merge_dictionaries` we could also call the `toolz` implementation from `sdk_utils` instead.

testing/sdk_utils.py
```diff
@@ -177,43 +175,3 @@ def pretty_duration(seconds):
 
 def random_string(length=8):
     return "".join(random.choice(string.ascii_lowercase + string.digits) for _ in range(length))
-
-
-# Pretty much https://github.com/pytoolz/toolz/blob/a8cd0adb5f12ec5b9541d6c2ef5a23072e1b11a3/toolz/dicttoolz.py#L279
-def get_in(keys, coll, default=None):
```
elezar: As discussed, if we call `toolz.get_in` here and `tools.merge` below, then we would not have to change any of the client code -- but I'm happy to reduce the scope of `sdk_utils` a bit!

frameworks/kafka/tests/client.py
mpereira: I thought we wanted to keep `sdk_utils` for DC/OS specifics? https://github.com/mesosphere/dcos-commons/pull/2511#discussion_r207285221

mpereira:
Thanks for the re-review @elezar.

> Not critical, but I prefer the use of toolz.get_in and toolz.merge explicitly. Not going to fight you on it though.

My personal preference (from other prog langs) for things that are stdlib-y (in this case Python data structure functions, i.e. `merge`, `get_in`) is to import them directly into modules. Since you're not up for a fight I'll leave them as-is, but feel free to change them in a future PR if you think it's better!

elezar:
I'm also all for using the external packages directly. I was just thinking of minimising changes. My review might also been a bit inconsistent

PR #2681, opened 2018-09-20
Update build.gradle to point to next SNAPSHOT

jeid64:
Releasing 0.42.3 today, bumping up snapshot.


jeid64:
I'm just following what the Wiki page here says https://wiki.mesosphere.com/display/ENGINEERING/Releasing+the+SDK
I have no preference either.

nickbp:
Updated the page, may as well go with bumping to 0.43.0-SNAPSHOT

samvantran:
+1 to 0.43-SNAPSHOT, it gives you room to work on features for the next release while being able to patch if necessary. If you update the patch number then you're kind of implying that youre fixing another bug. Then again, isn't sdk042 in maintenance mode or are there still features planned?

PR #2682, opened 2018-09-21
[DCOS-29502] Add default user auth mappings to resolve principal

mockingbird2:
This populates the `hadoop.security.auth_to_local` in the `core-site.xml` with auth user  auth mappings by default. Thus the Hdfs tasks can resolve its principal without the need of additional variables in the environment.


frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
```diff
@@ -55,28 +57,57 @@ private static SchedulerBuilder createSchedulerBuilder(File yamlSpecFile) throws
         File configDir = yamlSpecFile.getParentFile();
 
         SchedulerConfig schedulerConfig = SchedulerConfig.fromEnv();
+
+        Map<String, String> env = System.getenv();
+
+        String userAuthMapping;
+
+        if (Boolean.valueOf(env.get("TASKCFG_ALL_SECURITY_KERBEROS_ENABLED"))) {
+            String frameworkHostname = EndpointUtils.toAutoIpDomain(rawServiceSpec.getName(), schedulerConfig);
```
elezar: It might be useful to link this to `FRAMEWORK_HOST` in the variable name?

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
```diff
@@ -55,28 +57,57 @@ private static SchedulerBuilder createSchedulerBuilder(File yamlSpecFile) throws
...
+            HdfsUserAuthMapper authMapper = setupAuthMapper(env, frameworkHostname);
+            authMapper.addUserAuthMappingFromEnv(env, TASKCFG_ALL_AUTH_TO_LOCAL);
```
elezar: We could move the `TASKCFG_ALL_AUTH_TO_LOCAL` envvar specification to the `HdfsUserAuthMapper` class instead. Assuming it is not used elsewhere?

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
```diff
@@ -55,28 +57,57 @@ private static SchedulerBuilder createSchedulerBuilder(File yamlSpecFile) throws
...
+            authMapper.addDefaultUserAuthMapping(JOURNAL_POD_TYPE, "node", JOURNAL_NODE_COUNT);
```
elezar: From my point of view this interface could be simplified. Would it not make sense to have a single function call (or builder) to build the mapping?

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
```diff
@@ -55,28 +57,57 @@ private static SchedulerBuilder createSchedulerBuilder(File yamlSpecFile) throws
...
+        }
+
+
+
         DefaultServiceSpec serviceSpec = DefaultServiceSpec.newGenerator(rawServiceSpec, schedulerConfig, configDir)
                 // Used by 'zkfc' and 'zkfc-format' tasks within this pod:
                 .setPodEnv("name", SERVICE_ZK_ROOT_TASKENV, CuratorUtils.getServiceRootPath(rawServiceSpec.getName()))
-                .setAllPodsEnv(DECODED_AUTH_TO_LOCAL,
-                                getHDFSUserAuthMappings(System.getenv(), TASKCFG_ALL_AUTH_TO_LOCAL))
+                .setAllPodsEnv(DECODED_AUTH_TO_LOCAL, userAuthMapping)
```
elezar: I really like how this simplifies!

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
```diff
@@ -55,28 +57,57 @@ private static SchedulerBuilder createSchedulerBuilder(File yamlSpecFile) throws
...
                 .setEndpointProducer(CORE_SITE_XML, EndpointProducer.constant(
-                        renderTemplate(new File(configDir, CORE_SITE_XML), serviceSpec.getName(), schedulerConfig)))
+                        renderTemplate(new File(configDir, CORE_SITE_XML), serviceSpec.getName(), schedulerConfig,
+                                userAuthMapping)))
                 .setCustomConfigValidators(Arrays.asList(new HDFSZoneValidator()))
                 .withSingleRegionConstraint();
     }
 
+    private static HdfsUserAuthMapper setupAuthMapper(Map<String, String> env, String frameworkHostname) {
+        String realm = env.get("TASKCFG_ALL_SECURITY_KERBEROS_REALM");
```
elezar: Would it make sense to move this into the construction of the auth mapper instead?

frameworks/hdfs/src/test/java/com/mesosphere/sdk/hdfs/scheduler/HdfsUserAuthMapperTest.java
```diff
@@ -0,0 +1,111 @@
...
+        HdfsUserAuthMapper mapper = new HdfsUserAuthMapper(primary, frameworkHost, realm, frameworkUser);
+        mapper.addDefaultUserAuthMapping("data", "node", 2);
+
+        taskEnv.put("DECODED_AUTH_TO_LOCAL", mapper.getUserAuthMappingString());
+
+        String filledTemplate = TemplateUtils.renderMustacheThrowIfMissing(template.getName(), templateContent, taskEnv);
+        Assert.assertEquals(findAuthMappings(filledTemplate), mapper.getUserAuthMappingString());
+    }
+
+    private static String findAuthMappings(String config) throws IOException {
```
elezar: Although I don't have a problem with handling this here -- it's definitely faster, but we could check this here instead: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/hdfs/tests/test_sanity.py#L61. Would it be possible to use an XML-parser in the test instead? Not critical for this PR though. An alternative is would be to not use the full template in this test.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsUserAuthMapper.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.Map;
+
+/**
+ * This class helps to generate user auth mappings for the HDFS setup with kerberos.
+ */
+public class HdfsUserAuthMapper {
```
elezar: Should this be `HDFSUserAuthMapper`?

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsUserAuthMapper.java
```diff
@@ -0,0 +1,47 @@
...
+    private final ArrayList<String> authMappings = new ArrayList<>();
+
+    public HdfsUserAuthMapper(String primary, String frameworkHost, String realm, String frameworkUser) {
+        this.primary = primary;
+        this.frameworkHost = frameworkHost;
+        this.realm = realm;
+        this.frameworkUser = frameworkUser;
+    }
+
+    public void addUserAuthMappingFromEnv(Map<String, String> env, String envVarKeyName) {
```
elezar: Add a test for when the specified envvar doesn't exist.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HdfsUserAuthMapper.java
```diff
@@ -0,0 +1,47 @@
...
+        for (int i = 0; i < nodeCount; i++) {
+            String taskName = String.format("%s-%s-%s", hostPrefix, i, hostPostfix);
+            String authMapping = String.format(authToLocalDefaultTemplate, primary, taskName,
+                    frameworkHost, realm, frameworkUser);
+            authMappings.add(authMapping);
+        }
+    }
+
+    public String getUserAuthMappingString() {
+        return String.join("\n", authMappings);
```
elezar: Should we strip blank strings before joining?

frameworks/hdfs/src/test/java/com/mesosphere/sdk/hdfs/scheduler/HdfsUserAuthMapperTest.java
mockingbird2: Yes, indeed it would be nice to use a proper XML parsing tool. Unfortunately, I haven't found a good one now. Maybe someone with more Java experience might know a proper solution.

mockingbird2:
@elezar thanks for the nice discussion about improving this. I think I addressed most of the feedback regarding the java side. One last thing I have to do is adjusting the `frameworks/hdfs/tests/auth.py` because the default mappings are not needed anymore.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
mockingbird2: I am not sure what you by this. Isn't the variable name `frameworkHostname` self-explaining?

frameworks/hdfs/src/test/java/com/mesosphere/sdk/hdfs/scheduler/HDFSUserAuthMapperBuilderTest.java
```diff
@@ -60,6 +60,17 @@ public void testAddDefaultUserAuthMapping() {
...
 
+    @Test
+    public void testEnvAuthToLocalIsNotOverwritten() {
+        String expected = "authToLocal";
+        Map<String, String> env = getEnvWithEncodedAuthToLocal(expected);
+        String authMappings = new HDFSUserAuthMapperBuilder(env, frameworkHost)
+                .addUserAuthMappingFromEnv()
+                .addDefaultUserAuthMapping("data", "node", 1)
+                .build();
+        Assert.assertEquals(expected, authMappings.split("\n")[0]);
```
elezar: Minor point: This implies a particular ordering. We could do a search in the lines instead.

frameworks/hdfs/tests/auth.py
```diff
@@ -52,7 +52,7 @@ def get_principal_to_user_mapping() -> str:
     we need to create an appropriate mapping to test authorization functionality.
     :return: A base64-encoded string of principal->user mappings
     """
-    rules = ["RULE:[2:$1@$0](^hdfs@.*$)s/.*/hdfs/", "RULE:[1:$1@$0](^nobody@.*$)s/.*/nobody/"]
+    rules = ["RULE:[2:$1@$0](^hdfs@.*$)s/.*/hdfs/"]
```
elezar: Note that this rule effectively overrides the rules that we define since *ALL* our principals match this pattern.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HDFSAuthEnvContainer.java
```diff
@@ -0,0 +1,74 @@
...
+    public HDFSAuthEnvContainer(Map<String, String> env) {
+        checkEnvHasRequiredKeys(env);
+        this.envAuthMapping = getAuthMappingFromEnv(env);
+        this.primary = env.get(PRIMARY_ENV_KEY);
+        this.frameworkUser = env.get(FRAMEWORK_USER_ENV_KEY);
+        this.realm = env.get(REALM_ENV_KEY);
+    }
+
+    private String getAuthMappingFromEnv(Map<String, String> env) {
+        if (!env.containsKey(TASKCFG_ALL_AUTH_TO_LOCAL)) {
```
elezar: Minor: We could also handle the blank string here.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HDFSUserAuthMapperBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+            String taskName = String.format("%s-%s-%s", hostPrefix, i, hostPostfix);
+            String authMapping = String.format(authToLocalDefaultTemplate, envContainer.getPrimary(), taskName,
+                    frameworkHost, envContainer.getRealm(), envContainer.getFrameworkUser());
+            authMappings.add(authMapping);
+        }
+        return this;
+    }
+
+    public String build() {
+        authMappings.removeIf(String::isEmpty);
```
elezar: Let's use `isBlank` instead.

frameworks/hdfs/tests/auth.py
mockingbird2: When looking at the actual `core-site.xml` the added rules look like this `hdfs/journal-0-node.testintegrationhdfs.autoip.dcos.thisdcos.directory@LOCAL)s/.*/nobody/`. I already tried removing the other rule as well but then the tests start failing. So I think this rule is an addition.

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HDFSAuthEnvContainer.java
mockingbird2: I would rather keep it in the builder because the default case for this method is currently to return an empty string. This needs some adjustment if we move the black string logic here.

elezar: OK. That's fine.

For the record, I was thinking of something like:
```java
        if (StringUtils.isBlank(env.getOrDefault("TASKCFG_ALL_AUTH_TO_LOCAL", ""))) {
            return "";
        }
```

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/HDFSUserAuthMapperBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+            String taskName = String.format("%s-%s-%s", hostPrefix, i, hostPostfix);
+            String authMapping = String.format(authToLocalDefaultTemplate, envContainer.getPrimary(), taskName,
+                    frameworkHost, envContainer.getRealm(), envContainer.getFrameworkUser());
+            authMappings.add(authMapping);
+        }
+        return this;
+    }
+
+    public String build() {
+        authMappings.removeIf(StringUtils::isBlank);
```
elezar: minor: Indentation.

Also, are we modifying the state here. What about:
```java
        return authMappings.stream().filter(StringUtils::isBlank).collect(Collectors.joining("\n"));
```
as a single line?

frameworks/hdfs/src/main/java/com/mesosphere/sdk/hdfs/scheduler/Main.java
elezar: I mean that we use `frameworkHost` instead of `frameworkHostname`.

frameworks/hdfs/tests/auth.py
elezar: If the rules we add with `nobody` don't work, then we need to ensure that our rules look like:
```
RULE:[2:$1/$2@$0](hdfs/journal-0-node.testintegrationhdfs.autoip.dcos.thisdcos.directory@LOCAL)s/.*/hdfs/
```
Remember that this has to map to the HDFS super user which is `hdfs`...

mockingbird2: Is there anything I have to change? Since the integration test passes without the need for the superuser rule so it is not required for the initial setup at least.

elezar: It may be good to check that this passes on a strict cluster. I'll start a build tomorrow.

elezar:
@mockingbird2 the strict nightly build has passed at least once. Let's merge this and get it backported!

PR #2686, opened 2018-09-25
Enable whitelisting of tasks and services in sdk_diag via pytest markers

jimmidyson:
As discussed on slack with @mpereira.

As background, in the Kubernetes package we install the package only once for the whole test suite (with the exception of testing installs and upgrades obviously). In this case, that means that sdk_diag does not collect logs, plans, bundles, etc for tasks created by Kubernetes package because it is running before tests run. Adding whitelist in this PR allows us to collect diagnostic data that we need when tests fail.

### DIFF
--- a/testing/sdk_diag.py
+++ b/testing/sdk_diag.py
@@ -12,6 +12,7 @@ import os.path
 import re
 import shutil
 import time
+import inspect
 
 import pytest
 import retrying
@@ -63,6 +64,11 @@ def get_test_suite_name(item: pytest.Item):
     """Returns the test suite name to use for a given test."""
     # frameworks/template/tests/test_sanity.py => test_sanity_py
     # tests/test_sanity.py => test_sanity_py
+
+    # use the class name as the suite name if item is a method
+    if inspect.ismethod(item.obj):
+        return os.path.basename(item.getparent(pytest.Class).name).replace(".", "_")
+
     return os.path.basename(item.parent.name).replace(".", "_")
 
 
--- a/testing/sdk_diag.py
+++ b/testing/sdk_diag.py
@@ -119,8 +119,9 @@ def _task_whitelist_callback(item: pytest.Item):
     to be able to hierarchically configure the whitelist.
     """
     def _callback(task):
+        # get_closest_marker is only available in latest pytest versions
         if item.get_closest_marker(name='diag_task_whitelist') is None:
-            return True
+            return False
 
         for mark in item.iter_markers(name='diag_task_whitelist'):
             if mark.args[0](task):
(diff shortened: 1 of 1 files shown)


pires:
Re-triggered failed checks as all of them failed for the same reason: timeout downloading some JAR artifacts.

jimmidyson:
This is a regular problem with bintray - can we switch to Maven Central?

elezar:
> This is a regular problem with bintray - can we switch to Maven Central?

@jimmidyson is this something we can do in this repo, or is it a TeamCity setting. Could you create a ticket for tools infra if it's the latter?

testing/sdk_diag.py
```diff
@@ -82,7 +82,11 @@ def handle_test_setup(item: pytest.Item):
         _testlogs_current_test_suite = test_suite
         global _testlogs_ignored_task_ids
         _testlogs_ignored_task_ids = _testlogs_ignored_task_ids.union(
-            [task.id for task in sdk_tasks.get_summary(with_completed=True)]
+            [
+                task.id
+                for task in sdk_tasks.get_summary(with_completed=True)
+                if not _task_whitelist_callback(task)
```
elezar: The function `_task_whitelist_callback()` returns a callback and it's not sure how this is evaluated here.

testing/sdk_diag.py
```diff
@@ -101,6 +105,32 @@ def handle_test_setup(item: pytest.Item):
...
+
+    @pytest.mark.diag_task_whitelist.with_args(whitelist_fn)
+    def your_test_here(): ...
+
+    Note that the diag_task_whitelist marker can be used on function, class, or module
+    to be able to hierarchically configure the whitelist.
+    """
+    def _callback(task):
+        if item.get_closest_marker(name='diag_task_whitelist') is None:
+            return True
```
elezar: Does this mean that we whitelist tasks by default?

hectorj2f: I changed that, we shouldn't whitelist tasks by default

testing/sdk_diag.py
hectorj2f: this is working as expected, what do you mean by that ?

testing/sdk_diag.py
```diff
@@ -63,8 +64,12 @@ def get_test_suite_name(item: pytest.Item):
     """Returns the test suite name to use for a given test."""
     # frameworks/template/tests/test_sanity.py => test_sanity_py
     # tests/test_sanity.py => test_sanity_py
-    return os.path.basename(item.parent.name).replace(".", "_")
 
+    # use the class name as the suite name if item is a method
+    if inspect.ismethod(item.obj):
```
hectorj2f: When tests are part of a class we should use the class name as suite name

elezar: Sorry @hectorj2f does this mean that there is still work required for this PR. Please re-open if this is the case.

hectorj2f: Nope, we are good with the merged code :)

PR #2687, opened 2018-09-26
[DCOS-38928] Backport sdk_tasks.get_summary() from master.

porridge:
This makes it capable of returning more than 100 tasks.
This changed the exposed API, but I audited the usage of get_summary()
in this repo and just a couple of lines needed changing.

PR #2688, opened 2018-09-27
Reproduce breakage in another PR

porridge:
Just trying to see whether this effectively no-op PR faces the same failure as https://github.com/mesosphere/dcos-commons/pull/2687 on 1.9.


porridge:
I used the wrong destination branch and only realized I change change my choice after I created another PR :man_facepalming:.

PR #2690, opened 2018-09-27
Reproduce breakage in another PR

porridge:
Just trying to see whether this effectively no-op PR faces the same failure as #2687 on 1.9.
Supersedes https://github.com/mesosphere/dcos-commons/pull/2689 where TC seems to have become confused, retrying with a different branch name.


porridge:
*Sigh* it thinks it's being smart.

PR #2698, opened 2018-09-28
Revert "[DCOS-41803] Overwrite the CLI in docker image with an old CL…

elezar:
…I for sdk 40 (#2676)"

This reverts commit 997fcbf38877f72c1fa07313bc70f0bd6c1aedac.

This change breaks non-mono-repo builds including `spark-build` and `kafka-zookeeper`.

We would only expect the CI for this PR to be green after #2697 has been merged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ COPY test_requirements.txt test_requirements.txt
 RUN pip3 install --upgrade -r test_requirements.txt
 
 # Get DC/OS CLI
-RUN curl -O https://downloads.dcos.io/binaries/cli/linux/x86-64/latest/dcos && \
+RUN curl -O https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.12/dcos && \
     chmod +x dcos && \
     mv dcos /usr/local/bin && \
     dcos --version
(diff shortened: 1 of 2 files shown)


elezar:
For the record. The failures here are the same as those for non-mono-repo builds.

elezar:
The issue is that the `sdk-0.40` PR removes the local tools changes that are required for the CLI which is still present in the docker image. Since the teamcity job in the `sdk-0.40` branch still compares to `master` to rebuild the docker image it happens that on some of the build agents, the `origin/master` git reference doesn't exist and the docker image is *not* rebuilt and `mesosphere/dcos-commons:latest` is used instead. (e.g. Cassandra). This is effectively what happens for the non-mono-repo builds. For others, we are able to detect the differences in `Dockerfile`, rebuild a local docker image and run successfully: e.g. Elastic.

elezar:
There is a new docker images available off #2697 and I have retriggered all tests.

ArmandGrillet:
Please ping @bamarni and I if the CI doesn't pass, we're keeping our eyes on DCOS-41803. Sorry for the inconvenience due to the link change.

elezar:
> Please ping @bamarni and I if the CI doesn't pass, we're keeping our eyes on DCOS-41803. Sorry for the inconvenience due to the link change.

Thanks @ArmandGrillet! Will do. This is on us too as we have our own dependencies to deal with!

PR #2696, opened 2018-09-28
Fix bugs in test.sh for edge-lb

elezar:
This PR contains the changes from #2693 for the `master` branch in order to allow `dcos-commons` to be used as a submodule when building edge-lb.

Important changes:
* Allow `REPO_ROOT_DIR` to be overridden.
* Don't require a framework to be specified if a custom `DOCKER_COMMAND` is specified.

PR #2693, opened 2018-09-27
test.sh: bugfixes required for integrating the dcos-commons as git submodule into other repo

vespian:
### This PR:

This PR fixes few bugs that I found while integrating `test.sh` into edge-lb's CI. Namely:
* mktemp on some distros/Unixes requires XXX+ placeholder to work. On others it will simply be ignored I believe.
* providing framework names for testing should not be required when running `test.sh` with `DOCKER_COMMAND` env var defined
* repository location should not be hardcoded as this prevents other repos (like e.g. edge) to re-use dcos-commons contents and very tightly binds dcos-commons repo's frameworks to it's test tools.

### Related Jiras:
This change is required by Edge-LB's CI refactoring effort in order to integrate dcos-commons tooling into the edge-lb's one and have an easy way to bump dcos-commons without having to commit the tooling into the repo itself (i.e. via git submodules). See:
* https://jira.mesosphere.com/browse/DCOS-38216 `Simplifying the build scripts`
* https://jira.mesosphere.com/browse/DCOS-38215 `Switch edgelb CI to Jenkins Pipelines-based`
* https://jira.mesosphere.com/browse/DCOS-38546 `create a common development container for edgelb`

PR #2699, opened 2018-09-28
Fix bugs in test.sh for edge-lb (#2696)

elezar:
This PR updates `test.sh` with the one from `master` to fix integration with `edge-lb`.

This replaces #2693

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -10,18 +10,18 @@
 # Exit immediately on errors
 set -e
 
-timestamp="$(date +%d%m%y%H%M%s)"
+timestamp="$(date +%y%m%d-%H%M%S)"
 # Create a temp file for docker env.
 # When the script exits (successfully or otherwise), clean up the file automatically.
-credsfile="$(mktemp /tmp/sdk-test-creds-${timestamp}.tmp)"
-envfile="$(mktemp /tmp/sdk-test-env-${timestamp}.tmp)"
+tmp_aws_creds_path="$(mktemp /tmp/sdk-test-creds-${timestamp}-XXXX.tmp)"
+envfile="$(mktemp /tmp/sdk-test-env-${timestamp}-XXXX.tmp)"
 function cleanup {
-    rm -f ${credsfile}
+    rm -f ${tmp_aws_creds_path}
     rm -f ${envfile}
 }
 trap cleanup EXIT
 
-REPO_ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_ROOT_DIR=${REPO_ROOT_DIR:="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"}
 WORK_DIR="/build" # where REPO_ROOT_DIR is mounted within the image
 
 # Find out what framework(s) are available.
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -214,6 +224,11 @@ fi
 
 volume_args="-v ${REPO_ROOT_DIR}:$WORK_DIR"
 
+if [ -z "$CLUSTER_URL" -a x"$interactive" == x"true" ]; then
+    CLUSTER_URL="$(dcos config show core.dcos_url)"
+    echo "CLUSTER_URL not specified. Using attached cluster ${CLUSTER_URL} in interactive mode"
```
vespian: What if I do not have *any* cluster configured, neither via `CLUSTER_URL` nor via `dcos config...`? What if I want to use test.sh just to see if I can build stuff?

I believe this is one of the things that the discussion we had yesterday touches:
```
TBH I wonder if we should be using test.sh here at all - basing on the changes that nick did it is used for... running tests. The params we already added made the test a bit more "hairy". I wonder if exposing a way to just run a plain docker container be enough for use cases like mine. I get a feeling that test.sh is still evolving.
```

What exactly is `test.sh` meant to do? Is it a test runner or a build environment?

elezar: We use `test.sh` to create a reliable environment to run our tests in. The *default* behaviour is to invoke our `test_runner.sh` script that runs the tests of the specified framework (or all frameworks).

We (some of us) use it to set up an interactive environment for running a SUBSET of tests -- usually while developing the tests themselves. We do also copy this between our different repos as an attempt to provide some uniformity with how we run tests locally without having to remember the exact docker arguments to create the environment and trigger the tests.

Furthermore, it mirrors what we do in CI -- although there we don't always invoke the script directly.

You're correct in stating that you may not need this script at all and invoking the relevant `docker run` command directly may be more beneficial to you -- and decouple you somewhat from changes to the `dcos-commons` repo.

With regards to:
> What if I do not have any cluster configured, neither via CLUSTER_URL nor via dcos config...? What if I want to use test.sh just to see if I can build stuff?

specifically, in this case it will start a container based on the `dcos-commons` docker image with the `CLUSTER_URL` envvar set to empty.

vespian: @elezar Thanks for the lengthy answer!

If I was a community member developing with SDK and asked you the following question:

`How can I reliably build and test my SDK-based framework, without having to worry about build dependencies changing with every SDK version?`

What would be the answer?

vespian: I also did a quick experiment - I have removed all the configured clusters, and tried to build SDK
```
# vespian @ zara in ~edge on git:prozlach/add-pr-template o [12:40:59]
$ REPO_ROOT_DIR=`pwd` ./framework/dcos-commons/test.sh -p /home/vespian/work/meso.shared.ssh_key -i --aws-profile 273854932432_Mesosphere-PowerUser

Property 'core.dcos_url' doesn't exist
```
My thinking here is that if I just want to build, the script should be OK with not having `dcos_url` configured. Same I think should apply when I want to just publish to AWS.

elezar: I did run a quick sanity check, but it was *without* the `-i` flag. I agree that we should handle the case a bit more cleanly, but as you have pointed out, it's also not clear what this script really does. (I will think on and answer you question above).

As a workaround you can try:
```bash
CLUSTER_URL=https://no-clulister REPO_ROOT_DIR=`pwd` ./framework/dcos-commons/test.sh -p /home/vespian/work/meso.shared.ssh_key -i --aws-profile 273854932432_Mesosphere-PowerUser
```
to get you an interactive shell for building in this case.

PR #2664, opened 2018-09-12
[DCOS-41744] add hello world metrics smoke test

elezar:
This adds a metrics smoke test to Hello-world and should fail with the current master build.

TODO:
- [ ] Ensure that the list of metrics are sane.
- [ ] Switch to getting the scheduler metrics here including those adding in #2538

PR #2701, opened 2018-09-28
[DCOS-41744] add hello world metrics smoke test

elezar:
I'm recreating #2664 which:
* adds a metrics smoke test to Hello-world and should fail with the current master build.

This is related to: https://jira.mesosphere.com/browse/DCOS-41416


kvish:
landed my change in master, let me rebase this against master and get it merged in

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -19,21 +21,65 @@
 
 
 @pytest.fixture(scope="module", autouse=True)
-def configure_package(configure_security):
+def helloworld_service(configure_security):
```
takirala: we are using `configure_package` (just grepped and it has 39 occurences) and these methods usually assume `install` of the package. any strong reason for changing this to `helloworld_service` now?

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -19,21 +21,65 @@
...
+@pytest.mark.smoke
+def test_install():
+    config.check_running(sdk_utils.get_foldered_name(config.SERVICE_NAME))
+
+
+@pytest.mark.sanity
+@pytest.mark.metrics
+@pytest.mark.smoke
+@pytest.mark.dcos_min_version("1.9")
+def test_metrics_cli_for_scheduler_metrics(helloworld_service):
```
takirala: Since `helloworld_service` has `autouse=true`, we don't need to add this fixture explicitly right?

testing/sdk_metrics.py
```diff
@@ -136,7 +163,19 @@ def get_metrics(package_name, service_name, pod_name, task_name):
...
+
+    if "dimensions" not in app_json:
+        log.error("Expected key '%s' not found in app metrics: %s", "dimensions", app_json)
+        raise Exception("Expected key 'dimensions' not found in app metrics: {}")
+
+    if "task_name" not in app_json["dimensions"]:
+        log.error(
+            "Expected key '%s' not found in app metrics: %s", "dimensions.task_name", app_json
+        )
+        raise Exception("Expected key 'dimensions' not found in app metrics: {}")
```
takirala: Change error messages. Expected key `task_name` ..... instead of `dimensions`.

frameworks/helloworld/tests/test_sanity.py
takirala: actually, i am not sure about the behavior of explicitly importing a module scoped fixture in each of the tests.

frameworks/helloworld/tests/test_sanity.py
elezar: No real reason. I thought it's more communicative with respect to the configuration that is being passed through. I will revert this change for the time being.

frameworks/helloworld/tests/test_sanity.py
elezar: I would also have to check exactly what the implications of this is, but I do prefer to make the dependency explicit.

testing/sdk_metrics.py
elezar: 3fcf4de

frameworks/helloworld/tests/test_sanity.py
elezar: 3fcf4de

elezar:
@takirala @kvish since a lot has happened since you last reviewed, would either of you mind giving this a look again?

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -21,19 +23,69 @@
...
 
 
+@pytest.mark.sanity
+@pytest.mark.smoke
+def test_install():
+    config.check_running(sdk_utils.get_foldered_name(config.SERVICE_NAME))
+
+
+@pytest.mark.sanity
+@pytest.mark.metrics
```
takirala: Any reason for retaining this `metrics` mark ? Once this PR is merged, `dcos-enterprise` smoke test would run these tests and that should be good enough, am i missing something on the usage of `metrics` mark ?!

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -21,19 +23,69 @@
...
+@pytest.mark.smoke
+@pytest.mark.dcos_min_version("1.9")
+def test_metrics_for_task_metrics(configure_package):
+
+    metric_name = "test.metrics.CamelCase"
+    bash_command = sdk_cmd.get_bash_command(
+        'echo \\"{}:1|c\\" | ncat -w 1 -u \\$STATSD_UDP_HOST \\$STATSD_UDP_PORT'.format(
+            metric_name
```
takirala: (nit) may be add a note here on what the bash command is doing here.

frameworks/helloworld/tests/test_sanity.py
```diff
@@ -217,15 +269,13 @@ def test_config_cli():
...
-            "debug config show {}".format(configs[0]),
-            print_output=False,  # noisy output
-        )[0]
+    rc, _, _ = sdk_cmd.svc_cli(
+        config.PACKAGE_NAME,
+        foldered_name,
+        "debug config show {}".format(configs[0]),
+        print_output=False,  # noisy output
     )
+    assert rc == 0
```
takirala: instead of `assert rc == 0` here, how about passing `check=True` on previous call to `svc_cli`?

takirala: same comment for the rest of the similar occurrences in this file.

takirala: on a closer look, it looks like `check=True` may cause the logging to be skipped. One alternative is to do the check ourself (instead of passing it to `subprocess.run` [here](https://github.com/mesosphere/dcos-commons/blob/master/testing/sdk_cmd.py#L199)) like 

```
def _run_cmd(cmd, print_output, check, timeout_seconds=None):
..
..
..
..
  if check:
    assert not result.returncode, stdout + stderr

```
but this may be out of scope and i leave this to you.

frameworks/helloworld/tests/test_sanity.py
elezar: I took the `metrics` mark from the other tests, but I agree that it's not required. Will remove it.

frameworks/helloworld/tests/test_sanity.py
elezar: I did not want to functionally change the code but just make it clearer that we are checking the `rc` value. This also made the formatting a bit cleaner. Switching to `check=True` is out of scope for this PR.

PR #2703, opened 2018-10-01
[DCOS-42770] Wait until no deployments are queued after deletion

mockingbird2:
This PR ensures that the `sdk_marathon.destroy_app` call waits until the marathon app is completely removed. Before we only checked if it is still registered.

### DIFF
--- a/testing/sdk_marathon.py
+++ b/testing/sdk_marathon.py
@@ -118,7 +118,13 @@ def wait_for_deployment(app_name: str, timeout: int, expected_version: str) -> N
             len(deployments),
             log_extra,
         )
-        return staged == 0 and unhealthy == 0 and len(deployments) == 0 and instances_check and extra_check
+        return (
+            staged == 0
+            and unhealthy == 0
+            and len(deployments) == 0
+            and instances_check
+            and extra_check
+        )
 
     if expected_version:
         log.info(
@@ -225,14 +231,14 @@ def update_app(
 
 def destroy_app(app_name: str, timeout=TIMEOUT_SECONDS) -> None:
     @retrying.retry(stop_max_delay=timeout * 1000, wait_fixed=2000)
-    def _destroy():
+    def _destroy() -> dict:
         response = sdk_cmd.cluster_request(
             "DELETE", _api_url("apps/{}".format(app_name)), params={"force": "true"}
         )
-        return _handle_marathon_deployment_response(response)
+        return response.json()
 
     result = _destroy()
-    result.raise_on_error()
+    deployment_id = result["deploymentId"]
 
     # This check is different from the other deployment checks.
     # When it's complete, the app is gone entirely.
@@ -240,7 +246,14 @@ def destroy_app(app_name: str, timeout=TIMEOUT_SECONDS) -> None:
         stop_max_delay=timeout * 1000, wait_fixed=2000, retry_on_result=lambda result: not result
     )
     def _wait_for_app_destroyed():
-        return not app_exists(app_name, timeout)
+        running_deployments = sdk_cmd.cluster_request("GET", _api_url("deployments")).json()
+        log.info(
+            "While waiting to delete %s, currently running marathon deployments: %s",
+            deployment_id,
+            running_deployments,
+        )
+        running_id = next((d for d in running_deployments if d["id"] != deployment_id), None)
+        return running_id is None
 
     log.info("Waiting for {} to be removed...".format(app_name))
     _wait_for_app_destroyed()
--- a/testing/sdk_marathon.py
+++ b/testing/sdk_marathon.py
@@ -166,7 +166,6 @@ def install_app(app_definition: dict, timeout=TIMEOUT_SECONDS) -> None:
         try:
             deployment_response = MarathonDeploymentResponse(response)
         except requests.HTTPError as e:
-            log.error("HTTP error response code while installing: %s", response.json()["message"])
             if e.response.status_code == 409:
                 # App exists already, left over from previous run? Delete and try again.
                 destroy_app(app_name, timeout=timeout)
--- a/testing/sdk_marathon.py
+++ b/testing/sdk_marathon.py
@@ -88,7 +88,7 @@ class MarathonDeploymentResponse:
             deployment_id = response_json["deploymentId"]
         self._apps = [MarathonDeploymentResponse.App(version, deployment_id)]
 
-    def get_apps(self) -> typing.List[App]:
+    def get_apps(self) -> typing.List["App"]:
         return self._apps
 
 
--- a/testing/sdk_marathon.py
+++ b/testing/sdk_marathon.py
@@ -8,6 +8,7 @@ SHOULD ALSO BE APPLIED TO sdk_marathon IN ANY OTHER PARTNER REPOS
 import logging
 import retrying
 import requests
+import typing
 
 import sdk_cmd
 import sdk_tasks
@@ -52,29 +53,54 @@ def get_config(app_name, timeout=TIMEOUT_SECONDS):
 
 
 class MarathonDeploymentResponse:
+    class App:
+        def __init__(self, version: str, deployment_id: str) -> None:
+            self._version = version
+            self._deployment_id = deployment_id
+
+        def get_deployment_id(self) -> str:
+            return self._deployment_id
+
+        def get_version(self) -> str:
+            return self._version
+
     def __init__(self, response: requests.Response) -> None:
         try:
-            response_json = response.json()
+            self._response = response
+            response_json = self._response.json()
         except ValueError:
-            log.error("Failed to parse marathon response as JSON: %s", response.text)
+            log.error("Failed to parse marathon response as JSON: %s", self._response.text)
             raise
-        log.info("Marathon deployment response JSON: %s", response_json)
-        response.raise_for_status()
+        log.info(
+            "Requested %s and received Marathon deployment response JSON: %s",
+            self._response.url,
+            response_json,
+        )
+        self._response.raise_for_status()
 
-        self._version = response_json["version"]
-        if response.status_code == 201:
-            self._deployment_id = self._parse_deployment_id_on_app_creation(response_json)
+        self._parse_app_information(response_json)
+
+    def _parse_app_information(self, response_json: dict) -> None:
+        version = response_json["version"]
+        if self._response.status_code == 201:
+            deployment_id = response_json["deployments"][0]["id"]
         else:
-            self._deployment_id = response_json["deploymentId"]
+            deployment_id = response_json["deploymentId"]
+        self._apps = [MarathonDeploymentResponse.App(version, deployment_id)]
 
-    def _parse_deployment_id_on_app_creation(self, response_json: dict) -> str:
-        return response_json["deployments"][0]["id"]
+    def get_apps(self) -> typing.List[App]:
+        return self._apps
 
-    def get_deployment_id(self) -> str:
-        return self._deployment_id
 
-    def get_version(self) -> str:
-        return self._version
+class MarathonDeploymentsResponse(MarathonDeploymentResponse):
+    def _parse_app_information(self, response_json: dict) -> None:
+        def _parse(deployments) -> typing.Iterable[MarathonDeploymentResponse.App]:
+            for deployment in deployments:
+                version = deployment["version"]
+                deployment_id = deployment["id"]
+                yield MarathonDeploymentResponse.App(version, deployment_id)
+
+        self._apps = list(_parse(response_json))
 
 
 def wait_for_deployment(app_name: str, timeout: int, expected_version: str) -> None:
@@ -172,7 +198,7 @@ def install_app(app_definition: dict, timeout=TIMEOUT_SECONDS) -> None:
             raise e
         return deployment_response
 
-    result = _install()
+    result = _install().get_apps()[0]
     wait_for_deployment(app_name, timeout, result.get_version())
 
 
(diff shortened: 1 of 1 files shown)


testing/sdk_marathon.py
```diff
@@ -225,22 +231,29 @@ def _update():
 
 def destroy_app(app_name: str, timeout=TIMEOUT_SECONDS) -> None:
     @retrying.retry(stop_max_delay=timeout * 1000, wait_fixed=2000)
-    def _destroy():
+    def _destroy() -> dict:
         response = sdk_cmd.cluster_request(
             "DELETE", _api_url("apps/{}".format(app_name)), params={"force": "true"}
         )
-        return _handle_marathon_deployment_response(response)
+        return response.json()
```
elezar: Would it not make sense to rather include the `deploymentId` in the `_deployment_result` class than handling it directly here.

testing/sdk_marathon.py
```diff
@@ -225,22 +231,29 @@ def _update():
...
 
     result = _destroy()
-    result.raise_on_error()
+    deployment_id = result["deploymentId"]
```
elezar: How does not raising on error here change the behaviour?

testing/sdk_marathon.py
```diff
@@ -225,22 +231,29 @@ def _update():
...
 
     # This check is different from the other deployment checks.
     # When it's complete, the app is gone entirely.
     @retrying.retry(
         stop_max_delay=timeout * 1000, wait_fixed=2000, retry_on_result=lambda result: not result
     )
     def _wait_for_app_destroyed():
-        return not app_exists(app_name, timeout)
+        running_deployments = sdk_cmd.cluster_request("GET", _api_url("deployments")).json()
```
elezar: Should this check not include *both* a `not app_exists` call as well as a wait for empty deployments?

testing/sdk_marathon.py
```diff
@@ -225,22 +231,29 @@ def _update():
...
+        log.info(
+            "While waiting to delete %s, currently running marathon deployments: %s",
+            deployment_id,
+            running_deployments,
+        )
+        running_id = next((d for d in running_deployments if d["id"] != deployment_id), None)
```
elezar: In my opinion, this check could be simplified if we use something like:
```python
return deployment_id not in [d["id"] for d in running_deployments]
```

testing/sdk_marathon.py
mockingbird2: I don't think `app_exists` is very helpful here because when Marathon receives the event, it immediately removes the `/apps/{name}` entry because the desired state is `removed`.

testing/sdk_marathon.py
mockingbird2: The problem is that the `_deployment_result` is also used for `GET`  requests on `apps/{name}` to parse the response. Hereby the `deploymentId` is not included. Mixing both didn't seem like a good idea.

testing/sdk_marathon.py
mockingbird2: I would assume if the `deploymentId` is in the response the `DELETE` was successful otherwise it fails but maybe it's helpful to log the response message here.

testing/sdk_marathon.py
mockingbird2: Discussed in person and will be addressed.

testing/sdk_marathon.py
elezar: Here: https://github.com/mesosphere/dcos-commons/blob/DCOS-42770-wait-until-marathon-app-deletion-finished/testing/sdk_marathon.py#L141

The deploymentID is explicitly mentioned, so it would make sense to include it there too.

testing/sdk_marathon.py
elezar: OK. I see what this is -- this is `_deployment_result.raise_on_error()` which is different to `response.raise_for_status()` which is what I was thinking of. We could follow-up to simplify things a little here, but I don't think it's too critical.

testing/sdk_marathon.py
mockingbird2: Yes, it is included in the comment but the key is not in the response json. I think someone had the idea of building a serialization with this class but then discovered the `deploymentId` is not always present.

testing/sdk_marathon.py
mockingbird2: This refers probably to the complete way how we handle the marathon response. An improvement here might include looking into other projects handling the parsing the responses correctly.

mockingbird2:
@elezar thank you for the fruitful feedback and comments. I hope I addressed all the concerning points.

mockingbird2:
Consecutively restarting the CI didn't fix the flaky `test_sanity.test_scheduler` test. I created [https://jira.mesosphere.com/browse/DCOS-43034](https://jira.mesosphere.com/browse/DCOS-43034) to track the issue. This PR shouldn't have an effect on this test since the new code is not executed. I'd like to merge this soon because it currently blocks #2692.

testing/sdk_marathon.py
```diff
@@ -52,19 +53,29 @@ def get_config(app_name, timeout=TIMEOUT_SECONDS):
...
+            response_json = response.json()
+        except ValueError:
+            log.error("Failed to parse marathon response as JSON: %s", response.text)
+            raise
+        log.info("Marathon deployment response JSON: %s", response_json)
+        response.raise_for_status()
+
+        self._version = response_json["version"]
+        if response.status_code == 201:
+            self.deployment_id = self._parse_deployment_id_on_app_creation(response_json)
```
elezar: There's a mismatch in the member names here. This should probably be `_deployment_id`.

testing/sdk_marathon.py
```diff
@@ -52,19 +53,29 @@ def get_config(app_name, timeout=TIMEOUT_SECONDS):
...
-        self._errmsg = errmsg
+class MarathonDeploymentResponse:
+    def __init__(self, response: requests.Response) -> None:
+        try:
+            response_json = response.json()
+        except ValueError:
+            log.error("Failed to parse marathon response as JSON: %s", response.text)
+            raise
+        log.info("Marathon deployment response JSON: %s", response_json)
+        response.raise_for_status()
```
elezar: Should we `raise_for_status` *before* trying to decode the JSON?

testing/sdk_marathon.py
```diff
@@ -174,21 +161,22 @@ def install_app(app_definition: dict, timeout=TIMEOUT_SECONDS) -> None:
...
-    def _install():
+    def _install() -> MarathonDeploymentResponse:
         response = sdk_cmd.cluster_request(
             "POST", _api_url("apps"), json=app_definition, log_args=False, raise_on_error=False
         )
-        return _handle_marathon_deployment_response(response)
+        try:
+            deployment_response = MarathonDeploymentResponse(response)
+        except requests.HTTPError as e:
+            log.error("HTTP error response code while installing: %s", response.json()["message"])
```
elezar: Should we really call `response.json()` here?

Why not have `MarathonDeploymentResponse` wrap an instance `requests.Response`. We could then encapsulate the logic in this error handling in methods of the class.

Something like:
```python
deployment_response = MarathonDeploymentResponse(response)
if deployment_response.app_exists():
    destroy_app(app_name, timeout)
deployment_response.raise_for_status()  # This should explicitly ignore the 409 case that was already checked.
return deployment_response
```

elezar: (ok, this breaks the other usages of this response a little, but we could then include similar checks in these two places)

testing/sdk_marathon.py
```diff
@@ -52,19 +53,29 @@ def get_config(app_name, timeout=TIMEOUT_SECONDS):
...
+        if response.status_code == 201:
+            self.deployment_id = self._parse_deployment_id_on_app_creation(response_json)
+        else:
+            self._deployment_id = response_json["deploymentId"]
 
-    def raise_on_error(self) -> None:
-        if self._errmsg is not None:
-            raise Exception("Operation failed with error message: {}".format(self._errmsg))
+    def _parse_deployment_id_on_app_creation(self, response_json: dict) -> str:
+        return response_json["deployments"][0]["id"]
```
elezar: Is there always only one deployment?

testing/sdk_marathon.py
```diff
@@ -212,52 +200,58 @@ def _update():
             log_args=False,
             raise_on_error=False,
         )
-        return _handle_marathon_deployment_response(response)
+        return MarathonDeploymentResponse(response)
 
     result = _update()
-    result.raise_on_error()
```
elezar: There is still a subtle difference here. We don't check for error messages.

testing/sdk_marathon.py
```diff
@@ -52,19 +53,29 @@ def get_config(app_name, timeout=TIMEOUT_SECONDS):
...
+        log.info("Marathon deployment response JSON: %s", response_json)
+        response.raise_for_status()
+
+        self._version = response_json["version"]
+        if response.status_code == 201:
+            self.deployment_id = self._parse_deployment_id_on_app_creation(response_json)
+        else:
+            self._deployment_id = response_json["deploymentId"]
 
-    def raise_on_error(self) -> None:
```
elezar: I don't see `raise_on_error` or `_errmsg` captured in the new class.

testing/sdk_marathon.py
mockingbird2: The idea behind it was to log the error message provided by Marathon before raising the potential exception. `response_json` includes during the error state a `message` field.

testing/sdk_marathon.py
mockingbird2: Yes, since this method is only used if the status code is `201` saying the app was created.

testing/sdk_marathon.py
mockingbird2: This should be now handled by the `MarathonDeploymentResponse` class during initialization. It always serializes the JSON first and logs it. Thus the error message is shown. Afterwards, it raises the HTTP error.

testing/sdk_marathon.py
mockingbird2: I'd rather keep `MarathonDeploymentResponse` as small as possible. Handling different response codes seems pretty dependent on the endpoint. It's not necessary to call `response.json()` here again because it was already done during `MarathonDeploymentResponse` init. I'll remove it.

testing/sdk_marathon.py
mockingbird2: Hopefully, this comment is covered by the responses before.

testing/sdk_marathon.py
elezar: Yes, but the original method raised an error if the error string in the message body was non-empty, or am I missing something?

mockingbird2: I see this wasn't clear from my side. Marathon only adds a message string in the response if the status code is not `2xx`. In this case the error is thrown immediately during initialization now. Therefore I don't think it is necessary anymore to explicitly check for the message since it should have the same meaning as the HTTP error code. The log statement showing the response JSON should in an error case now include the message string to help debugging.

elezar: Ok. That's fine then! Thanks for the clarification.

elezar: (do we log the error message as it would have been to add enough context?)

testing/sdk_marathon.py
```diff
@@ -212,52 +197,58 @@ def _update():
...
     # This check is different from the other deployment checks.
     # When it's complete, the app is gone entirely.
     @retrying.retry(
         stop_max_delay=timeout * 1000, wait_fixed=2000, retry_on_result=lambda result: not result
     )
     def _wait_for_app_destroyed():
-        return not app_exists(app_name, timeout)
+        if app_exists(app_name, timeout):
+            return False
+        running_deployments = sdk_cmd.cluster_request("GET", _api_url("deployments")).json()
```
porridge: It's still not obvious to me why we aren't wrapping this in `MarathonDeploymentResponse` for consistency?

mockingbird2: I fully understand the concern. Unfortunately, `MarathonDeploymentResponse` should ideally only deal with JSONs in the following format `{"id": xxxx, "version": xxxx}` and we already have one exception during install of the app which is not optimal. Querying all the deployments would result in parsing a completely different format, in addition, might need an exposure of more members than `id` and `version`. One thing I can propose is changing the name of the class to `MarathonAppStatusResponse` which should be a better description since it does not really have anything to do with deployments. Thoughts?

porridge: Ah, I get it now. Sounds like it's because `MarathonDeploymentResponse` has two responsibilities:
1. handling error response codes
1. parsing the content

So it sounds like one way would be to extract the error handling into a parent class and then create a sister `MarathonDeploymentsResponse` (note the `s`).

FWIW, the format is not that radically different, it's a matter of one entity vs a collection of them. Maybe there is a nice way to exploit this fact to keep the code simple?

If marathon supported querying for a single deployment, another way would be to just fetch the single deployment using `deployment_id`, rather than the full collection. But [AFAICT](http://mesosphere.github.io/marathon/api-console/index.html) it does not.

mockingbird2:
@porridge Thanks for your suggestions. I also parse the deployment list response now with the new API and hopefully have addressed your issue with that.

porridge:
Looks like this is supserseded by #2699.

testing/sdk_marathon.py
```diff
@@ -52,29 +53,54 @@ def get_config(app_name, timeout=TIMEOUT_SECONDS):
...
+        self._response.raise_for_status()
+
+        self._parse_app_information(response_json)
 
-        self._version = response_json["version"]
-        if response.status_code == 201:
-            self._deployment_id = self._parse_deployment_id_on_app_creation(response_json)
+    def _parse_app_information(self, response_json: dict) -> None:
+        version = response_json["version"]
+        if self._response.status_code == 201:
```
porridge: Can you add a comment here? It's not obvious why this is so.

PR #2704, opened 2018-10-01
[DCOS-41824] Enforce data service version under test

mockingbird2:
This PR sets the installed version of a package during testing to `stub-universe`. It ensures that we always test the newly built version and not by accident the universe version.

### DIFF
--- a/testing/sdk_install.py
+++ b/testing/sdk_install.py
@@ -10,6 +10,7 @@ import logging
 import time
 import retrying
 import tempfile
+from enum import Enum
 
 import sdk_cmd
 import sdk_marathon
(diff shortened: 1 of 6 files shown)


testing/sdk_install.py
```diff
@@ -86,7 +86,7 @@ def install(
     service_name,
     expected_running_tasks,
     additional_options={},
-    package_version=None,
+    package_version="stub-universe",
```
elezar: Should we make `stub-universe` a module-level constant? Not critical though.

mockingbird2:
@elezar I have integrated an enum with two different options: `STUB_UNIVERSE`, `LATEST_UNIVERSE` to make the version more explicit. I didn't completely understand the idea behind `ANY_VERSION` since the version is always translated to a string. 
One thing we could consider would be to change the API in a follow-up because most of the versions we want to install are either the latest from universe or stub-universe. Unfortunately, allowing only those two might also break other things I am currently not aware of.

PR #2711, opened 2018-10-04
[DCOS-40332] Integration tests for local profile volume support.

chhsia0:
This PR adds an integration test for #2707. Changes include:
* Added a `hello.volume_profile` option to the Hello World service.
* Added an example YAML to test profile mount volumes.
* Created an additional MOUNT volume with profile "xfs" on every agent for the test.

### DIFF
--- a/frameworks/helloworld/tests/test_profile_mount_volumes.py
+++ b/frameworks/helloworld/tests/test_profile_mount_volumes.py
@@ -1,0 +1,32 @@
+import logging
+import pytest
+
+import sdk_install
+import sdk_tasks
+from tests import config
+
+log = logging.getLogger(__name__)
+
+NUM_HELLO = 2
+
+
+@pytest.fixture(scope="module", autouse=True)
+def configure_package(configure_security):
+    try:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+        options = {
+            "service": {"yaml": "pod-profile-mount-volume"},
+            "hello": {"volume_profile": "xfs"}  # hardcoded in `tools/create_testing_volumes.py`.
+        }
+
+        sdk_install.install(config.PACKAGE_NAME, config.SERVICE_NAME, NUM_HELLO, additional_options=options)
+
+        yield  # let the test session execute
+    finally:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.mark.sanity
+@pytest.mark.dcos_min_version("1.12")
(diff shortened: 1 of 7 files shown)


chhsia0:
@nickbp Removed the WIP tag. The test passed locally. PTAL!

frameworks/helloworld/tests/test_profile_mount_volumes.py
```diff
@@ -0,0 +1,32 @@
...
+
+        yield  # let the test session execute
+    finally:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.mark.sanity
+@pytest.mark.dcos_min_version("1.12")
+def test_profile_mount_volumes():
+    sdk_tasks.check_running(config.SERVICE_NAME, NUM_HELLO)
```
nickbp: Would it be possible to verify here that the correct `xfs` volumes were indeed consumed? e.g. by fetching and looking at state.json or slaves.json. For some examples of accessing mesos jsons, see `sdk_tasks.py`, e.g.: `sdk_cmd.cluster_request("GET", "/mesos/tasks").json()["tasks"]`

tools/create_testing_volumes.py
```diff
@@ -58,11 +61,41 @@ def mount_volumes():
             dcos_mount="/dcos/volume{}".format(i),
             loop_dev="/dev/loop{}".format(i),
             loop_file="/root/volume{}.img".format(i),
+            fs_type=p or "ext4"
         )
 
+    # To create profile mount volumes, we manually run `make_disk_resources.py`
+    # to generate disk resources, then use `jq` to parse the result and set the
```
nickbp: Comment out of date: not using `jq` after all

frameworks/helloworld/tests/test_profile_mount_volumes.py
chhsia0: The task itself already checked the filesystem (`df -t $VOLUME_PROFILE`). Wouldn't that be sufficient?

nickbp: Oh I missed that. I think that's fine.

chhsia0:
Added a commit to address the comments and fix testing issues:
1. Use `/opt/mesosphere/bin/python` instead of `python` which may not exist on the cluster (The CI nodes have `python3` instead).
2. Updated the example YAML to use different container paths for task volumes. Otherwise the latter one will overwrite the first one and will make the test flaky.
3. Updated the example YAML to test the filesystem of the pod volume as well.

PR #2716, opened 2018-10-10
DCOS-43083 : Decommision Task Kills

kaiwalyajoshi:
k8s uses custom decommissioners to perform cleanup. Currently on a scheduler restart, we're prematurely killing off the decommission related tasks.
    
In this patch:
- Remove global kills of tasks with GoalStateOverride of PENDING as default behaviour.
- Move current semantics into legacyKillUnneededTasks() and use environment variable USE_LEGACY_KILL_UNNEEDED_TASKS to toggle previous functionality at runtime.

Update hello-world unit test ServiceTest.testWorldDecommission to reflect the following changes
- Initial task kill has been removed, tasks will be killed during TriggerDecommisionStep.
- Update counts to reflect that world-1-server and world-0-server are only killed once.

### DIFF
--- a/frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
+++ b/frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
@@ -389,10 +389,6 @@ public class ServiceTest {
         ticks.add(Send.taskStatus("world-0-server", Protos.TaskState.TASK_RUNNING).setReadinessCheckExitCode(0).build());
         ticks.add(Send.taskStatus("world-1-server", Protos.TaskState.TASK_RUNNING).setReadinessCheckExitCode(0).build());
 
-        // An initial kill is sent early on for the decommissioned tasks:
-        ticks.add(Expect.taskNameKilled("world-0-server", 1));
-        ticks.add(Expect.taskNameKilled("world-1-server", 1));
-
         // Now, we expect there to be the following plan state:
         // - a deploy plan that's COMPLETE, with only hello-0 (empty world phase)
         // - a recovery plan that's COMPLETE
@@ -413,7 +409,7 @@ public class ServiceTest {
         // Check plan state after an offer came through: world-1-server killed
         ticks.add(new ExpectDecommissionPlanProgress(Arrays.asList(
                 new StepCount("world-1", stepCount - 1, 0, 1), new StepCount("world-0", stepCount, 0, 0))));
-        ticks.add(Expect.taskNameKilled("world-1-server", 2));
+        ticks.add(Expect.taskNameKilled("world-1-server", 1));
 
         // Offer world-0 resources and check that nothing happens (haven't gotten there yet):
         ticks.add(Send.offerBuilder("world").setPodIndexToReoffer(0).build());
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -270,23 +293,36 @@ private static void killUnneededTasks(StateStore stateStore, Set<String> taskToD
                 .collect(Collectors.toSet());
 
         // Clear the TaskIDs from the TaskInfos so we drop all future TaskStatus Messages
+        //TODO: Investigate why this is needed anymore.
```
bmcstdio: If this is still needed, should it make its way to the new `killIUnneededTasks` method?

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -270,23 +293,36 @@ private static void killUnneededTasks(StateStore stateStore, Set<String> taskToD
...
         Set<Protos.TaskInfo> cleanedTaskInfos = unneededTaskInfos.stream()
                 .map(taskInfo -> taskInfo.toBuilder())
                 .map(builder -> builder.setTaskId(Protos.TaskID.newBuilder().setValue("")).build())
                 .collect(Collectors.toSet());
 
         // Remove both TaskInfo and TaskStatus, then store the cleaned TaskInfo one at a time to limit damage in the
-        // event of an untimely scheduler crash
+        // event of an untimely scheduler crash.
+        //TODO: Investigate why this is needed anymore.
```
bmcstdio: _Idem_.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
```diff
@@ -182,7 +182,12 @@ protected void registeredWithMesos() {
             activeTasks.addAll(decommissionedTasks);
         }
 
-        killUnneededTasks(stateStore, activeTasks);
+        if (schedulerConfig.useLegacyUnneededTaskKills()) {
+            //This case is turned off by default.
+            legacyKillUnneededTasks(stateStore, activeTasks, logger);
```
nickbp: Maybe add an expiration date here in a comment, like "Remove this on or after Dec 2018" (for example).

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/DefaultScheduler.java
nickbp: At this point my best guess is that this code would be run if, for example, an entire class of pods or tasks was removed from the service spec, as part of e.g. an upgrade. It isn't something that we've been exercising, so it's not even clear if this would work as intended.

I'm sorta on the fence about whether it even makes sense to keep around behind an envvar, but at the same time it doesn't really hurt anything to do so.

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/ServiceTest.java
```diff
@@ -389,10 +389,6 @@ public void testWorldDecommission() throws Exception {
         ticks.add(Send.taskStatus("world-0-server", Protos.TaskState.TASK_RUNNING).setReadinessCheckExitCode(0).build());
         ticks.add(Send.taskStatus("world-1-server", Protos.TaskState.TASK_RUNNING).setReadinessCheckExitCode(0).build());
 
-        // An initial kill is sent early on for the decommissioned tasks:
```
nickbp: Going by how the test was already "verifying" the bad behavior during decommission, it looks like this has been going on for a while. I think it's a good sign that this is the only noticeable side effect of cleaning up the old code.

PR #2717, opened 2018-10-11
[DCOS-41674] Add test to check envvars under restart (#2677)

elezar:
This is a backport of #2677

* Add tests for Cassandra and HelloWorld to check envvars under restart

* Update docker-specific files to skip rebuild of docker image in CI.

### DIFF
--- a/frameworks/cassandra/tests/test_sanity.py
+++ b/frameworks/cassandra/tests/test_sanity.py
@@ -1,4 +1,6 @@
 import pytest
+import logging
+
 import sdk_cmd
 import sdk_hosts
 import sdk_install
@@ -13,7 +15,10 @@ import sdk_utils
 from tests import config
 
 
-@pytest.fixture(scope='module', autouse=True)
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="module", autouse=True)
 def configure_package(configure_security):
     test_jobs = []
     try:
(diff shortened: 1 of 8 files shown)


frameworks/helloworld/tests/test_sidecar.py
```diff
@@ -32,6 +34,42 @@ def configure_package(configure_security):
...
+        sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, " ".join(cmd_list))
+
+        sdk_plan.wait_for_kicked_off_recovery(config.SERVICE_NAME)
+        sdk_plan.wait_for_completed_recovery(config.SERVICE_NAME)
+
+        _, stdout, _ = sdk_cmd.service_task_exec(config.SERVICE_NAME, "hello-0-server", "env")
+
+        envvar = "CONFIG_SLEEP_DURATION="
+        envvar_pos = stdout.find(envvar)
+        if envvar_pos < 0:
```
porridge: Can we do something a bit more high level, such as:
```
env = dict(l.rstrip().split("=", 1) for l in stdout.readlines())
val = env.get("CONFIG_SLEEP_DURATION", "absent")
if val != str(sleep_duration):
  raise(...)
```

frameworks/helloworld/tests/test_sidecar.py
```diff
@@ -32,6 +34,42 @@ def configure_package(configure_security):
...
+        config.SERVICE_NAME,
+        to_package_version=None,
+        additional_options={
+            "service": {"name": config.SERVICE_NAME, "sleep": sleep_duration, "yaml": "sidecar"}
+        },
+        expected_running_tasks=2,
+        wait_for_deployment=True,
+    )
+
+    for attempt in range(3):
```
porridge: Why 3 times? Please add a comment.

frameworks/helloworld/src/main/dist/sidecar.yml
```diff
@@ -25,6 +25,7 @@ pods:
           # Note, this is very high so that the test test_sidecar.test_toxic_sidecar_doesnt_trigger_recovery
           # will always have an empty recovery plan.
           SLEEP_DURATION: 10000
+          CONFIG_SLEEP_DURATION: {{SLEEP_DURATION}}
```
porridge: Wow, now I'm really confused. Does the `SLEEP_DURATION` on line 27 have anything to do with the `{{SLEEP_DURATION}}` on 28? Because it looks as if we allow setting it in service options, while actually don't, because we hardcode it to 10000 on line 27?

frameworks/cassandra/tests/test_sanity.py
```diff
@@ -123,3 +128,50 @@ def expected_metrics_exist(emitted_metrics):
         config.DEFAULT_CASSANDRA_TIMEOUT,
         expected_metrics_exist
     )
+
+
+@pytest.mark.sanity
+def test_config_update_across_restart():
```
porridge: Same comments apply here as in the HW test.

porridge:
Also, if this is a backport, please mention the original PR.

elezar:
Thanks for the review @porridge. You're right that this is a bit hacky. I'm going to close this and address your concerns in `master` instead.

PR #2720, opened 2018-10-12
DCOS-42539 Fix how TaskInfos are updated with sidecar tasks and resource sets

nickbp:
Updates the task refresh logic to explicitly update the StateStore TaskInfos for ALL tasks sharing a ResourceSet when any task in that set is updated. This avoids the possibility of hitting mismatches between TaskInfos, and makes it easier to track down what resources a given task is expected to be using at any given time, and avoids the current problems around TaskInfo data (config IDs) potentially being out of date in sidecar tasks. The new logic explicitly treats resource evaluation as per-ResourceSet, and then applies the resulting changes to the paired TaskInfo(s). See comments in OfferEvaluator for more info.

Other unrelated fixes for this last PR:
- Allow overriding the executor resource "overhead" via scheduler envvars
- Remove the long-deprecated FINISHED goalstate, which was long ago renamed to ONCE. YAML files will raise an error at build time, while serialzed ServiceSpecs will automatically convert FINISHED to ONCE to ensure compatibility during upgrade/downgrade.
- Clarify and add more information to evaluation logs, in particular log when resource ids are generated and be much clearer about when they are required.
- In recovery logic, avoid repeatedly refetching the full list of TaskInfos from the StateStore, which would involve reparsing large protobufs each time. Instead fetch the list only once and reuse that copy where needed.
- Add internal support for the added mesos task states (breaking up TASK_LOST). Note that these won't effectively be enabled until the scheduler explicitly opts in by setting the PARTITION_AWARE flag in the FrameworkInfo (in SchedulerDriverFactory)

### DIFF
--- a/frameworks/cassandra/tests/test_sanity.py
+++ b/frameworks/cassandra/tests/test_sanity.py
@@ -9,7 +9,6 @@ import sdk_metrics
 import sdk_networks
 import sdk_plan
 import sdk_upgrade
-import sdk_utils
 
 from tests import config
 
@@ -79,12 +78,7 @@ def test_repair_cleanup_plans_complete():
 
 
 @pytest.mark.sanity
-@pytest.mark.metrics
 @pytest.mark.dcos_min_version("1.9")
-@pytest.mark.skipif(
-    sdk_utils.dcos_version_at_least("1.12"),
-    reason="Metrics are not working on 1.12. Reenable once this is fixed",
-)
 def test_metrics():
     expected_metrics = [
         "org.apache.cassandra.metrics.Table.CoordinatorReadLatency.system.hints.p999",
--- a/frameworks/cassandra/universe/marathon.json.mustache
+++ b/frameworks/cassandra/universe/marathon.json.mustache
@@ -15,6 +15,15 @@
     "DCOS_SERVICE_SCHEME": "http"
   },
   {{#service.service_account_secret}}
+  "container": {
+    "type": "MESOS",
+    "volumes": [
+      {
+        "containerPath": "secrets/service-account.json",
+        "secret": "serviceCredential"
+      }
+    ]
+  },
   "secrets": {
     "serviceCredential": {
       "source": "{{service.service_account_secret}}"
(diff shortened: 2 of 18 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/DefaultServiceSpec.java
```diff
@@ -468,7 +438,8 @@ public GoalState deserialize(
                 String value = ((TextNode) p.getCodec().readTree(p)).textValue();
 
                 if (value.equals("FINISHED") || value.equals("ONCE")) {
```
takirala: Can add a TODO here to remove the `||`

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/TaskUtils.java
```diff
@@ -459,49 +465,86 @@ private static PodSpec getPodSpec(ConfigStore<ServiceSpec> configStore, Protos.T
...
+                continue;
+            }
+
+            Optional<TaskSpec> taskSpec = getTaskSpec(configStore, info);
+            if (!taskSpec.isPresent()) {
+                throw new TaskException("Failed to determine TaskSpec from TaskInfo: " + info);
+            }
+
+            boolean markedFailed = FailureUtils.isPermanentlyFailed(info);
+            boolean isPermanentlyFailed = markedFailed && taskSpec.get().getGoal() == GoalState.RUNNING;
```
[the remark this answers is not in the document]
nickbp: The idea is that tasks which are ONCE or FINISH are instead only handled by the deploy plan. If they failed, then it'd be the deploy plan dealing with relaunching them until they have finished successfully.

I cleaned up this logic and added some comments. There were a few different places in the same file all checking for a RUNNING GoalState for the same reason

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/OfferEvaluator.java
```diff
@@ -353,92 +341,100 @@ static void logOutcome(StringBuilder stringBuilder, EvaluationOutcome outcome, S
...
+            String taskSpecName = taskEntry.getKey();
+            ResourceSet resourceSet = taskEntry.getValue();
+            List<ResourceSpec> resourceSpecs = getOrderedResourceSpecs(resourceSet);
+
+            if (!addedExecutorResources) {
+                // The default executor needs a fixed amount of "overhead" resources, to be added once per pod.
+                // For consistency, let's put this before the per-task/resourceset RESERVE calls below.
+                addedExecutorResources = true;
+
+                getExecutorResourceSpecs(
```
elezar: Since `getExecutorResourceSpecs` is private, we could change that to rather return the list of `ResourceEvaluationStage`s instead of performing the map here.

kvish:
taking ownership of this PR

kvish:
resolved comments and fixed linting rules

kvish:
Pr resolves the following issues:
DCOS-43924
DCOS-43925
DCOS-43926
DCOS-43927
DCOS-43928
DCOS-43929

kvish:
resolved bugs

PR #2721, opened 2018-10-12
[DCOS-28837] Validate that steps refer to tasks existing in the pod

kvish:

### DIFF
--- a/frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/CustomStepsTest.java
+++ b/frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/CustomStepsTest.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 
 import org.apache.mesos.Protos;
+import org.junit.Assert;
 import org.junit.Test;
 
 import com.mesosphere.sdk.testing.Expect;
@@ -120,6 +121,16 @@ public class CustomStepsTest {
                 .setSchedulerEnv(
                         "DEPLOY_STRATEGY", "serial",
                         "DEPLOY_STEPS", "[[first, second]]")
+                .run(ticks);
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void testInvalidStep() throws Exception {
+        Collection<SimulationTick> ticks = new ArrayList<>();
+        new ServiceTestRunner("custom_steps.yml")
+                .setSchedulerEnv(
+                        "DEPLOY_STRATEGY", "serial",
+                        "DEPLOY_STEPS", "[[foo, bar]]")
                 .run(ticks);
     }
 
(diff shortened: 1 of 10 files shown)


frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/CustomStepsTest.java
```diff
@@ -123,6 +123,16 @@ public void testSerialStrategyParallelSteps() throws Exception {
...
     }
 
+    @Test(expected = IllegalStateException.class)
+    public void testInvalidStep() throws Exception {
+        Collection<SimulationTick> ticks = new ArrayList<>();
+        new ServiceTestRunner("custom_steps.yml")
+                .setSchedulerEnv(
+                        "DEPLOY_STRATEGY", "serial",
+                        "DEPLOY_STEPS", "[[foo, bar]]")
+                .run(ticks);
```
takirala: can replace this with `.run(ticks);` with `.run(new ArrayList<>());` and then delete line 128.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/PlanGenerator.java
```diff
@@ -191,6 +195,16 @@ private Step generateStep(PodInstance podInstance, List<String> tasksToLaunch) {
         }
     }
 
+    /**
+     * Validates that all tasks in a step are defined in the pod.
+     */
+    private boolean isValidStep(PodSpec podSpec, List<String> tasksToLaunch) {
```
nickbp: Function could be static

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/PlanGenerator.java
```diff
@@ -191,6 +195,16 @@ private Step generateStep(PodInstance podInstance, List<String> tasksToLaunch) {
...
+        List<String> allTaskNames = podSpec.getTasks().stream()
+                .map(taskSpec -> taskSpec.getName())
+                .collect(Collectors.toList());
```
nickbp: Super nitpick: allTaskNames could be a Set (with Collectors.toSet()), and then the following containsAll would be O(n)

frameworks/helloworld/src/test/java/com/mesosphere/sdk/helloworld/scheduler/CustomStepsTest.java
nickbp: - Could use `Collections.emptyList()` instead of creating a new arraylist (one difference is that the collections one is immutable)
- Should be able to just call the `run()` overload

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/PlanGenerator.java
kvish: resolved

kvish:
removed uneeded files

PR #2722, opened 2018-10-12
Disable metrics test due to changed container metrics structure

mockingbird2:
The metrics container structure probably has changed slightly so even with the newer changes the test fail. As follow-up https://jira.mesosphere.com/browse/DCOS-43546 is created.


elezar:
@kvish the nightlies still failed. Let's merge this and then ensure that #2701 can be merged as it adds more context to the error messages.

PR #2724, opened 2018-10-17
kafka: Add volume profile support

jieyu:
This patch adds the volume profile support for kafka brokers. This
allows operator to pick disks based on profile names (see DC/OS storage
service doc: https://docs.mesosphere.com/services/beta-storage/)

### DIFF
--- a/frameworks/kafka/src/main/dist/svc.yml
+++ b/frameworks/kafka/src/main/dist/svc.yml
@@ -69,6 +69,10 @@ pods:
           path: {{BROKER_DISK_PATH}}
           type: {{BROKER_DISK_TYPE}}
           size: {{BROKER_DISK_SIZE}}
+          {{#BROKER_DISK_PROFILE_ENABLED}}
+          profiles:
+            - {{BROKER_DISK_PROFILE}}
+          {{/BROKER_DISK_PROFILE_ENABLED}}
         env:
           KAFKA_DISK_PATH: "{{BROKER_DISK_PATH}}"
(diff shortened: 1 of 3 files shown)


jieyu:
The package can be tested using this stub universe:
```
dcos package repo remove kafka-aws
dcos package repo add --index=0 kafka-aws https://universe-converter.mesosphere.com/transform?url=https://infinity-artifacts.s3.amazonaws.com/autodelete7d/kafka/20181017-150108-L
Hqlav2L0vFRTsw6/stub-universe-kafka.json
dcos package install --yes kafka
```

frameworks/kafka/src/main/dist/svc.yml
```diff
@@ -69,6 +69,10 @@ pods:
           path: {{BROKER_DISK_PATH}}
           type: {{BROKER_DISK_TYPE}}
           size: {{BROKER_DISK_SIZE}}
+          {{#BROKER_DISK_PROFILE_ENABLED}}
```
elezar: Does the service behave well when we toggle this setting?

frameworks/kafka/universe/config.json
```diff
@@ -263,6 +263,10 @@
           "description": "Disk type to be used for storing broker data. See documentation. [ROOT, MOUNT]",
           "default": "ROOT"
         },
+        "disk_profile": {
```
elezar: Just for information, how does this setting interact with the disk type? Are there combinations that are not supported?

jieyu: yes, ROOT volume with profile is not supported.

How can I do some validation here?

elezar: I think there are ways to do it in the config schema itself, but at the very least we could add a validator for the scheduler.

PR #2725, opened 2018-10-18
[DCOS-41837] add scale test tooling for hello world 

kvish:
can be run as follows from the root of dcos-commons:
`export PYTEST_ARGS="--scenario=normal --count=5"`
`./test.sh -m scale helloworld`


frameworks/helloworld/tests/conftest.py
```diff
@@ -6,3 +6,21 @@
 @pytest.fixture(scope="session")
 def configure_security(configure_universe):
     yield from sdk_security.security_session(config.SERVICE_NAME)
+
+
+# pytest fixtures to run hello-world scale test located in frameworks/helloworld/tests/scale
+def pytest_addoption(parser):
```
elezar: Definitely going to use this myself!

frameworks/helloworld/tests/scale/test_scale.py
```diff
@@ -0,0 +1,72 @@
...
+        config.PACKAGE_NAME,
+        service_name,
+        config.DEFAULT_TASK_COUNT,
+        additional_options=options,
+        wait_for_deployment=False,
+        wait_for_all_conditions=False
+    )
+
+
+def _create_service_account(service_name, security=None):
```
elezar: How about:
```suggestion
def _create_service_account(service_name, security=None):
    if security == sdk_dcos.DCOS_SECURITY.strict:
        try:
            log.info("Creating service accounts for '{}'"
                     .format(service_name))

            sa_name = "{}-principal".format(service_name)
            sa_secret = "{}-secret".format(service_name)
            return sdk_security.setup_security(
                service_name,
                linux_user="nobody",
                service_account=sa_name,
                service_account_secret=sa_secret,
            )

        except Exception as e:
            log.warning("Error encountered while creating service account: {}".format(e))
            raise e

    return None
```

frameworks/helloworld/tests/scale/test_scale.py
```diff
@@ -0,0 +1,72 @@
...
+    """Launch a load test scenario. This does not verify the results
+    of the test, but does ensure the instances and jobs were created.
+
+    Args:
+        count: number of helloworld services to install
+        scenario: yaml scenario to run helloworld with (normal, crashloop) are added for this case
+    """
+    security_mode = sdk_dcos.get_security_mode()
+    # TODO: parallelize account creation and installation if time is an issue in scale tests
+    for index in range(service_count):
```
elezar:
```suggestion
    for index in range(service_count):
        service_name = "{}-{}-{}".format(config.PACKAGE_NAME, scenario, index)
        security_info = _create_service_account(service_name,
                                                security_mode)
        _install_service(service_name,
                         scenario,
                         security_info)
```

frameworks/helloworld/tests/scale/test_scale.py
```diff
@@ -0,0 +1,72 @@
...
+        service_name = "{}-{}-{}".format(config.PACKAGE_NAME, scenario, index)
+        _create_service_account(service_name,
+                                security_mode)
+        _install_service(service_name,
+                         scenario,
+                         security_mode)
+
+
+def _install_service(service_name, scenario, security=None):
```
elezar: With the other changes:
```suggestion
def _install_service(service_name, scenario, security_info):
    # do not wait for deploy plan to complete, all tasks to launch or marathon app deployment
    # supports rapid deployments in scale test scenario
    options = {"service": {"name": service_name, "yaml": scenario}}
    if security_info:
        options["service"]["service_account"] = security_info["name"]
        options["service"]["service_account_secret"] = security_info["secret"]

    sdk_install.install(
        config.PACKAGE_NAME,
        service_name,
        config.DEFAULT_TASK_COUNT,
        additional_options=options,
        wait_for_deployment=False,
        wait_for_all_conditions=False
    )
```

testing/sdk_dcos.py
```diff
@@ -0,0 +1,28 @@
...
+FOR THE TIME BEING WHATEVER MODIFICATIONS ARE APPLIED TO THIS FILE
+SHOULD ALSO BE APPLIED TO sdk_dcos IN ANY OTHER PARTNER REPOS
+************************************************************************
+'''
+from enum import Enum
+
+import sdk_cmd
+
+
+class DCOS_SECURITY(Enum):
```
elezar: Question: Does this add any advantage over `sdk_utils.is_strict()`? for the current usage?

frameworks/helloworld/tests/scale/test_scale.py
```diff
@@ -0,0 +1,72 @@
+import pytest
+import logging
+import sdk_install
+import sdk_security
+import sdk_dcos
+from tests import config
+
+log = logging.getLogger(__name__)
+
+ACCOUNTS = {}
```
elezar: Do we need this global?

frameworks/helloworld/tests/scale/test_scale.py
elezar: Looking at this, it may even make sense to change the keys for the `security_info` dict in `sdk_security` to `service_account` and `service_account_secret` so that the key mapping isn't even required.

frameworks/helloworld/src/main/dist/normal.yml
```diff
@@ -0,0 +1,15 @@
+# This service will launch tasks that will run as long as indicated by the sleep value.
```
elezar: This comment is not valid. Also, can one not reuse an existing definition for this?

testing/sdk_dcos.py
kvish: nope this is the same thing will replace it

kvish: seems to be the same thing will tremove this file and use existing methods

frameworks/helloworld/src/main/dist/normal.yml
kvish: yes this seems to be mostly a repeat of simple.yml, no need to reinvent it

frameworks/helloworld/tests/scale/test_scale.py
kvish: This is not needed, was going to multi thread the creation of service accounts like we do in jenkins but for now until we run this test and get more info there is no need to do this. I like the suggestions you outlined

testing/sdk_dcos.py
takirala: FYI : the reason sdk method takes from environment is because the endpoint `dcos-metadata/bootstrap-config.json` is relatively new. might worth adding a check on cluster version (there should be a helper to do this already) before assuming `mode = r['security']` works.

testing/sdk_install.py
```diff
@@ -122,6 +125,7 @@ def install(
         options,
         package_version,
         timeout_seconds,
+        wait_for_all_conditions
```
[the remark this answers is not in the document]
kvish: there exists tests which do not expect this behavior when passing wait_for_deployment.

testing/sdk_dcos.py
kvish: since this is manually triggered, it may be easier to use the old method and just set the mode in the enviornment, although using this method makes it easier for the operator

testing/sdk_install.py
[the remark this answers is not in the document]
takirala: How about renaming `wait_for_all_conditions` to `wait_for_running_tasks` (defaults to `True`) and then use the `wait_for_deployment` for https://github.com/mesosphere/dcos-commons/pull/2725/files#diff-a4edd5b45d2ef4fda8fd5a8e90348736R83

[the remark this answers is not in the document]
kvish: sdk_marathon.wait_for_deployment will wait for all health checks to be empty, people who set it to false today, assume they are just not watiing for deploy plan to complete, this would proceed a test without waiting for tasks to be healthy which can change test behavior for those who are opted into it now

frameworks/helloworld/tests/scale/test_scale.py
```diff
@@ -0,0 +1,67 @@
...
+log = logging.getLogger(__name__)
+
+
+@pytest.mark.scale
+def test_scaling_load(service_count,
+                      scenario) -> None:
+    """Launch a load test scenario. This does not verify the results
+    of the test, but does ensure the instances and jobs were created.
+
+    Args:
```
takirala: arg name mismatch `count` vs `service_count` in docs.

kvish:
renamed args for clarity

PR #2727, opened 2018-10-22
[DCOS-43889] Freeze all pip depencies to recover from build breakage.

porridge:
I'm aware this is horrible. See the jira ticket for why simpler approaches did not work.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,6 +7,7 @@
 /logs
 
 !test_requirements.txt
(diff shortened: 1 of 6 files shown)


porridge:
Thanks @elezar !

porridge:
Force-merging since there were *some* successes, and the failures are due to an unrelated problem: DCOS-43953 (CLI/cluster mismatch in most PR jobs).

PR #2729, opened 2018-10-23
[DCOS-43953] Use a stable snapshot of dcos CLI.

porridge:
This should prevent build breakage whenever a broken dcos binary happens to be
picked up by the image build.


porridge:
It's not really versioned, I mentioned the downloaded-on date in the commit message, I'm afraid a comment in a different file will get rotten easily.

elezar:
Thought: We could download with a SHA and then create a symlink?

PR #2730, opened 2018-10-23
hdfs: Add volume profile support

jieyu:
This patch adds volume profile support for HDFS framework. Users are now
allowed to pick a volume profile for journal, name or data node.


jieyu:
Rebased to master

PR #2732, opened 2018-10-24
[DCOS-43947] Update JRE to 8u192 to address CVEs (#2728)

elezar:
This PR is a backport of #2728 :

* Updates the JRE version to `8u192`
* Updates the SDK version string in `package_builder.py` to match that in `build.gradle` (Cosmetic change)


vespian:
Are we going to do a release in the 0.40 branch, or should I bump the dependency to the tip of the sdk-0.40 branch?

PR #2733, opened 2018-10-24
[DCOS_OSS-4343] account for recovery plan completing quickly

kvish:
when waiting for a kicked off recovery plan, plan could already have completed

### DIFF
--- a/testing/sdk_plan.py
+++ b/testing/sdk_plan.py
@@ -123,7 +123,7 @@ def wait_for_completed_step(
 
 def wait_for_kicked_off_plan(service_name, plan_name, timeout_seconds=TIMEOUT_SECONDS):
     return wait_for_plan_status(
-        service_name, plan_name, ["PENDING", "STARTING", "IN_PROGRESS"], timeout_seconds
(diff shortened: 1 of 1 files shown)


testing/sdk_plan.py
```diff
@@ -123,7 +123,7 @@ def wait_for_completed_step(
 
 def wait_for_kicked_off_plan(service_name, plan_name, timeout_seconds=TIMEOUT_SECONDS):
     return wait_for_plan_status(
-        service_name, plan_name, ["PENDING", "STARTING", "IN_PROGRESS"], timeout_seconds
+        service_name, plan_name, ["PENDING", "STARTING", "IN_PROGRESS", "COMPLETE"], timeout_seconds
```
elezar: This is going to cause the exact problems that the `wait_for_kicked_off_plan` was supposed solve. The default recovery plan is empty and *always* `COMPLETE`.

kvish:
ack makes sense @elezar this may be an actual issue then where recovery plan isnt getting kicked off, I assumed that the recovery plan had completed

PR #2737, opened 2018-10-30
Whitelisting LIBPROCESS_IP in airgap_linter

akirillov:
**Change description**
When `MESOS_CONTAINER_IP` is not present in the environment, `LIBPROCESS_IP` can be used instead. The main use case for this is to replace `MESOS_HTTP` health check when a process within a container is bound to an interface different from `localhost` and `0.0.0.0`. 

**Related Spark package PR**
[Spark Dispatcher and History Server health check replacement for MESOS_HTTP](https://github.com/mesosphere/spark-build/pull/437)

**Related Apache Mesos issue**
[[MESOS-6517] Health checking only on 127.0.0.1 is limiting](https://issues.apache.org/jira/browse/MESOS-6517)


elezar:
@akirillov feel free to merge this whenever you're ready.

PR #2743, opened 2018-11-02
Handle case where service name does *not* omit the first slash.

mpereira:
Apparently DCOS foldered service names may or may not omit the first slash. This script was looking for the service based on its name with the first slash omitted.

For example: in soak112s a service named `/data-services/cassandra` has `data-services/cassandra` as the service `name` field. On the cluster related to https://jira.mesosphere.com/browse/COPS-4058 a service named `/green/cassandra-green7` has `/green/cassandra-green7` as the service `name` field.

### DIFF
--- a/tools/diagnostics/full_bundle.py
+++ b/tools/diagnostics/full_bundle.py
@@ -30,17 +30,14 @@ def get_dcos_services() -> (bool, str):
         return (True, stdout)
 
 
-def to_dcos_service_name(service_name: str) -> str:
-    """DC/OS service names don't contain the first slash.
-    i.e.: |     SDK service name     |   DC/OS service name    |
+def is_service_named(service_name: str, service: dict) -> bool:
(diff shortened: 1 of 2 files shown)


tools/diagnostics/full_bundle.py
```diff
@@ -63,9 +60,7 @@ def is_service_scheduler_task(package_name: str, service_name: str, task: dict)
     dcos_service_name = next(
         iter([l.get("value") for l in labels if l.get("key") == "DCOS_SERVICE_NAME"]), ""
     )
-    return dcos_package_name == package_name and dcos_service_name == to_dcos_service_name(
-        service_name
-    )
+    return dcos_package_name == package_name and dcos_service_name == service_name
```
elezar: Should this not also use the same logic as for `is_service_named`?

mpereira: This value seems to always have the leading slash, even though in the Mesos state the service name might not.

PR #2746, opened 2018-11-06
[DCOS-43952] [TOOLS] Install readwrite registry from bootstrap registry

takirala:
Should be good to merge once https://github.com/mesosphere/dcos-enterprise/pull/3776 is merged.

### DIFF
--- a/gradle/checkstyle/README.md
+++ b/gradle/checkstyle/README.md
@@ -8,11 +8,13 @@ Most of the linting rules are taken from google style guide with some customizat
 
 This plugin does not support all the rules that we use out of box. Following needs to be done manually if using this plugin:
 
+- Install the `Checkstyle-IDEA` plugin.
 - Navigate to `Settings  > Editor > Code Style > Java`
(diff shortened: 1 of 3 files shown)


kvish:
is there ever a case where an override path to pass your own stub is useful? by default rw registry can come from bootstrap registry, but might be useful to allow us to set a stub via envvar if we wanted to test package registry changes, with this change we can only test the version bundled with bootstrap registry. (if we wont be making lots of package registry changes maybe this isnt super useful)

takirala:
We added the stub mechanism to test package registry when we had 1.11 as master. By still using the stub we were essentially testing a stale version of package registry. By installing package registry from bootstrap registry we would be testing the package registry as a moving target. Having stub also skips the test of installation of rw registry from bootstrap registry, fwiw. We can bring back the env var if/when such a need arises (but `dcos-commons` is not the most ideal place to test it though).

takirala:
Once this is merged, i will remove the package registry stub related environment parameters in teamcity CI

PR #2748, opened 2018-11-08
[DCOS-44605] Filter sensitive environment keys and add file based secret support for frameworks

takirala:
- Introduce regex based blacklisting of log-sensitive environment variables
- Add file based secret handling support for SDK. We still support environment based secrets for backwards compatibility.
- Update frameworks mustache files with file based secrets.

### DIFF
--- a/sdk/bootstrap/main.go
+++ b/sdk/bootstrap/main.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
@@ -25,13 +26,19 @@ import (
 const (
 	configTemplatePrefix = "CONFIG_TEMPLATE_"
 	resolveRetryDelay    = time.Duration(1) * time.Second
+	hiddenEnvKeyValue    = "********"
 )
 
 var verbose = false
 
+var blackListedEnvironKeys = regexp.MustCompile(`(?i)(DCOS_SERVICE_ACCOUNT_CREDENTIAL|credential|password|secret|token)`)
+
 type args struct {
 	// Whether to print the container environment up-front
 	printEnvEnabled bool
+
+	// Whether to print the process environment including all the sensitive keys
+	printEnvInsecureEnabled bool
 
 	// Whether to enable host resolution at startup
 	resolveEnabled bool
@@ -61,6 +68,9 @@ func parseArgs() args {
 
 	flag.BoolVar(&args.printEnvEnabled, "print-env", true,
 		"Whether to print the process environment.")
+
+	flag.BoolVar(&args.printEnvInsecureEnabled, "insecure", false,
+		"Whether to print the process environment including ALL THE SENSITIVE KEYS. This flag needs to be passed along with print-env flag and is meant only for diagnostic purposes")
 
 	flag.BoolVar(&args.resolveEnabled, "resolve", true,
 		"Whether to enable the step of waiting for hosts to resolve. "+
@@ -93,7 +103,7 @@ func parseArgs() args {
 		// the user didn't provide hosts of their own.
 		taskHost, err := parseTaskHost()
 		if err != nil {
-			printEnv()
+			printEnv(args.printEnvInsecureEnabled)
 			log.Fatalf("Missing required envvar(s) to build default -resolve-hosts value. " +
 				"Either specify -resolve-hosts or provide these envvars: TASK_NAME, FRAMEWORK_HOST.")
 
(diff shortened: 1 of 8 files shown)


sdk/bootstrap/main.go
```diff
@@ -130,13 +135,28 @@ func splitAndClean(s string, sep string) []string {
 }
 
 // env print
-
 func printEnv() {
-	env := os.Environ()
+	env := logSensitiveEnviron()
```
elezar: Nit: Since the function is `printEnv` we could use `logSensitiveEnv` instead?

sdk/bootstrap/main.go
```diff
@@ -130,13 +135,28 @@ func splitAndClean(s string, sep string) []string {
...
 	sort.Strings(env)
 	log.Printf("Bootstrapping with environment:\n%s", strings.Join(env, "\n"))
 }
 
+func logSensitiveEnviron() (ret []string) {
```
elezar: The name of this function is misleading. It is not logging anything, but is filtering the environment.

sdk/bootstrap/main.go
```diff
@@ -130,13 +135,28 @@ func splitAndClean(s string, sep string) []string {
...
+	var filteredEnviron []string
+	for _, keyValue := range os.Environ() {
+		// Each string is in format key=value; value may be empty.
+		s := strings.Split(keyValue, "=")
+		key := s[0]
+		value := s[1]
+		if val, ok := blackListedEnvironKeys[key]; ok && len(value) > 0 {
```
elezar: We should still blacklist even if a key is empty as even this leaks information.

sdk/bootstrap/main.go
```diff
@@ -29,6 +29,11 @@ const (
 
 var verbose = false
 
+// The following replaces the keys with the values when logging.
+var blackListedEnvironKeys = map[string]string{
+	"DCOS_SERVICE_ACCOUNT_CREDENTIAL": "********",
```
elezar: We technically don't need the second part of the map. This will be the same for all entries.

sdk/bootstrap/main.go
```diff
@@ -29,6 +29,11 @@ const (
 
 var verbose = false
 
+// The following replaces the keys with the values when logging.
+var blackListedEnvironKeys = map[string]string{
```
elezar: It would be great if we could have a test for this so as to prevent regressions in the future.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -422,6 +427,19 @@ public TokenProvider getDcosAuthTokenProvider() throws IOException {
...
   }
 
+  private JSONObject loadFileOrEnvSecret() throws IOException {
+    String content = envStore.getRequired(SIDECHANNEL_AUTH_ENV);
+    JSONObject serviceAccountObject;
+    if (Files.isRegularFile(Paths.get(content))) {
+      LOGGER.info("Reading file {} to load secrets", content);
+      serviceAccountObject = new JSONObject(IOUtils.toString(new FileInputStream(content)));
+    } else {
+      LOGGER.info("{} is not a valid path. Parsing as JSON content.", SIDECHANNEL_AUTH_ENV);
```
elezar: I don't think that the `is not a valid path` is required here. What about:
```suggestion
      LOGGER.info("Reading service account information from {}.", SIDECHANNEL_AUTH_ENV);
```
As this does not give any information as to the nature of the data.

sdk/bootstrap/main.go
takirala: It will be same, but, I added it to understand if `DCOS_SERVICE_ACCOUNT_CREDENTIAL` is present (or not). Should we not retain that information? fwiw, the name of those environment variables will be quite obvious by looking at the the publicly available `config.json` and `marathon.json.mustache` but the presence (or absence) is not - and I am not sure if we should have this information in the logs to make it easier for debugging.

elezar: I mean we could still check for presence in the map, it's just that we wouldn't store or use the `value`. So here we would have:
```golang
var blackListedEnvironKeys = map[string]bool{
	"DCOS_SERVICE_ACCOUNT_CREDENTIAL": true,
}
```
and then where we use it we would then have:
```golang
if _, ok := blackListedEnvironKeys[key]; ok {
    filteredEnviron = append(filteredEnviron, fmt.Sprintf("%s=**********", key))
}
```

takirala:
Thanks for review @elezar and @kaiwalyajoshi 
Backporting (cherry-picked _only_ the first commit from this PR) via #2751

PR #2749, opened 2018-11-08
[DCOS-41674][TESTS] Address review comments on restart tests for envvars

elezar:
This PR addresses review comments from #2717 which itself was a backport of #2677.

The changes made are:
* Remove the test from Cassandra since it is already covered by the hello-world test
* clean up how we check the envvars
* Be more explicit about the envvar values before update, after update and after restart


frameworks/helloworld/tests/test_sidecar.py
```diff
@@ -31,7 +31,32 @@ def configure_package(configure_security):
...
+    def assert_envvar_has_value(envvar: str, expected_value: str):
+        _, stdout, _ = sdk_cmd.service_task_exec(config.SERVICE_NAME, "hello-0-server", "env")
+        env = dict(l.strip().split("=", 1) for l in stdout.strip().split('\n'))
+        val = env.get(envvar, "absent")
+
+        if val == "absent":
+            raise ConfigException("Required envvar not found")
+
+        if val != expected_value:
+            log.error("Looking for %s=%d but found: %s", envvar, sleep_duration, val)
```
samvantran: surprised `sleep_duration` doesn't err with `value referenced before assignment`
but anyway shouldn't `sleep_duration` be `expected_value`?

elezar: Good catch! Python won't catch this until the line is actually executed.

PR #2750, opened 2018-11-08
DCOS-40763 - Enable GLOG_v=1 in hello-world scheduler by default.

kaiwalyajoshi:
Config setting service.verbose_mesos_logging = 0 can be used to disable verbose Mesos prints.

### DIFF
--- a/frameworks/helloworld/universe/config.json
+++ b/frameworks/helloworld/universe/config.json
@@ -134,6 +134,7 @@
         "verbose_mesos_logging": {
           "description": "Sets the value of GLOG_v env-var used by Mesos. Enabled by default, set to 0 to disable.",
           "type": "integer",
+          "enum": [0, 1],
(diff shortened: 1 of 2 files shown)


frameworks/helloworld/universe/config.json
```diff
@@ -130,6 +130,11 @@
           },
           "description": "Placement constraint for the hello-world scheduler",
           "default": []
+        },
+        "verbose_mesos_logging": {
+          "description": "Sets the value of GLOG_v env-var used by Mesos. Enabled by default, set to 0 to disable.",
+          "type": "integer",
```
takirala: Is the list of possible values exhaustive? If yes, how about making this enum?!

kaiwalyajoshi: Good point, pushing up patch in next commit.

PR #2761, opened 2018-11-13
[DCOS-29208] switch to mesosphere client images

elezar:
This PR switches the testing client Docker images to ones in the mesosphere docker org.

These are hosted at: https://github.com/mesosphere/dcos-commons-ci/tree/master/docker/testing-client

### DIFF
--- a/frameworks/hdfs/tests/config.py
+++ b/frameworks/hdfs/tests/config.py
@@ -24,8 +24,7 @@ HDFS_POD_TYPES = {"journal", "name", "data"}
 KEYTAB = "hdfs.keytab"
 HADOOP_VERSION = "hadoop-2.6.0-cdh5.9.1"
 
-# TODO: This should be replaced once the tests pass.
-DOCKER_IMAGE_NAME = "elezar/hdfs-testing-client:dev"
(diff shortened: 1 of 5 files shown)


frameworks/hdfs/tests/config.py
```diff
@@ -24,7 +24,7 @@
 KEYTAB = "hdfs.keytab"
 HADOOP_VERSION = "hadoop-2.6.0-cdh5.9.1"
 
-DOCKER_IMAGE_NAME = "elezar/hdfs-client:dev"
+DOCKER_IMAGE_NAME = "mesosphere/hdfs-testing-client:30ff636fdf1795a7c3f443cfa77898720a3e6fd3"
```
samvantran: Nice, I can also update the client we're using [here](https://github.com/mesosphere/soak-cluster-configurations/blob/master/infinity/packages/hdfsclient-kerberos.json#L8) now

elezar:
I found the issue with the HDFS image and corrected it in mesosphere/dcos-commons-ci#7

frameworks/hdfs/tests/config.py
elezar: @samvantran the client here: `mesosphere/hdfs-testing-client:6972ea3833c9449111aceaa998e3e093a9c8dcee` works as expected!

PR #2762, opened 2018-11-13
[DCOS-41496] Use Elastic and Kibana resources from downloads.mesosphere.com.

mpereira:
I uploaded the resources using the [Jenkins job](https://jenkins.mesosphere.com/service/jenkins/view/Infinity/job/infinity-tools/job/production-asset-tools/job/upload-production-asset/) and tested on a local cluster. Apparently the `svg` Kibana icon images lose the `image/svg` content type when uploaded, so they don't show up in the DC/OS UI. I still need to look into that.

### DIFF
--- a/frameworks/elastic/universe-kibana/resource.json
+++ b/frameworks/elastic/universe-kibana/resource.json
@@ -1,13 +1,13 @@
 {
   "assets": {
     "uris": {
-      "kibana-tar-gz": "https://artifacts.elastic.co/downloads/kibana/kibana-{{elastic-version}}-linux-x86_64.tar.gz",
-      "xpack-zip": "https://artifacts.elastic.co/downloads/packs/x-pack/x-pack-{{elastic-version}}.zip"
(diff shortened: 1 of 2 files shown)


mpereira:
I was able to have the `svg` icon uploaded with the correct content type by changing the Jenkins job.

The `.com` URLs seems to be served from a different Cloudfront instance than `.io` URLs. We can see now that the `.io` URL returns the correct headers:

```bash
$ curl -I https://downloads.mesosphere.io/elastic/assets/icon-kibana-bb.svg | grep -i '\(content-type\|x-cache\|last-modified\)'
Content-Type: image/svg+xml
X-Cache: Miss from cloudfront
Last-Modified: Wed, 14 Nov 2018 20:04:46 GMT
```
```bash
$ curl -I https://downloads.mesosphere.com/elastic/assets/icon-kibana-bb.svg | grep -i '\(content-type\|x-cache\|last-modified\)'
Content-Type: binary/octet-stream
X-Cache: Hit from cloudfront
Last-Modified: Tue, 13 Nov 2018 15:48:17 GMT
```

I don't have permissions to force a cache refresh, but I guess it's ok to wait for it to expire?

PR #2771, opened 2018-11-14
[WIP] Remove 'build_dir_path' from 'tools/pip/setup.py'

klueska:
This variable is not referenced anywhere, and causes the setup.py script
to fail when it attempts to remove 'build_dir_path' since it was never
created in the first place.

### DIFF
--- a/tools/pip/setup.py
+++ b/tools/pip/setup.py
@@ -89,7 +89,7 @@ def main():
     )
 
     # clean up build detritus:
-    shutil.rmtree(os.path.join(script_dir, "{}.egg-info".format(output_name)))
+    shutil.rmtree(os.path.join(script_dir, "pip-egg-info", "{}.egg-info".format(output_name)))
     shutil.rmtree(output_dir_path)
     # move whl file into script dir:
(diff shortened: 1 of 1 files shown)


tools/pip/setup.py
```diff
@@ -92,8 +89,7 @@ def main():
     )
 
     # clean up build detritus:
-    shutil.rmtree(build_dir_path)
-    shutil.rmtree(os.path.join(script_dir, "{}.egg-info".format(output_name)))
+    shutil.rmtree(os.path.join(script_dir, "pip-egg-info", "{}.egg-info".format(output_name)))
```
elezar: I don't see where the original path is set. Is this added when building the egg?

klueska:
I'm still playing around with this and I'm not sure this is the full, right way to do things yet. I'll ping you once it's more stable. Is there a way to block CI from being triggered when I push an update to the branch?

mpereira:
> Is there a way to block CI from being triggered when I push an update to the branch?

I think there's no way to do that right now, unfortunately.

klueska:
I just decided to closse it for now to avoid triggering it. I'll create a new PR when I have something more credible.

PR #2776, opened 2018-11-14
[DCOS-42966] Fix potential segfault on empty response in CLI

kvish:


cli/client/response.go
```diff
@@ -46,6 +46,8 @@ HTTP %s Query for %s failed: %s
 For more syntax information`,
 				response.Request.Method, response.Request.URL, err)
 		}
+	} else if response == nil {
+		return nil, fmt.Errorf("Encountered an empty response, with error: %s", err)
```
takirala: is it possible that `err != nil && response == nil` ? If yes, when/why can that happen? 

The earlier switch conditions in this code return helpful messages to user listing few things that can go wrong and how the user can fix them. We should do the same thing here for the sake of being consistent.

kvish: The case that this applies to would be as follows:

'''
URL [http://greg-1-12-elasticl-1955tfrjvlcam-1562868946.us-west-2.elb.amazonaws.com/package/uninstall] is unreachable.
'''
In this case you have an empty response with a non empty error. Perhaps our error message could be to retrigger the operation?

PR #2777, opened 2018-11-14
[DCOS-44873] Add file name to ClassicRPCAuthenticatee module

takirala:
- ClassicRPCAuthenticatee is used when using V0 API. We use V0 in `1.10` clusters.
- HttpAuthenticatee is used when using V1 API.

Going forward, i think we can introduce helloworld in `1.10`, `1.11`, `1.12` (both permissive and strict) for our PR CI in dcos-commons. This would also be useful when we remove the frameworks from dcos-commons repo.

### DIFF
--- a/frameworks/cassandra/universe/marathon.json.mustache
+++ b/frameworks/cassandra/universe/marathon.json.mustache
@@ -44,7 +44,7 @@
 
     {{#service.service_account_secret}}
     "DCOS_SERVICE_ACCOUNT_CREDENTIAL": "secrets/service-account.json",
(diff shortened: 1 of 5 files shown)


kvish:
lgtm pending CI

takirala:
Tested this manually by installing kafka from a stub generated from this PR and then using the stub to install kafka in 1.10 strict, 1.11 strict and 1.12 strict clusters. Merging this while a CI is still running to get this tested in nightlies. We can always revert if something doesn't work.

PR #2780, opened 2018-11-15
[DCOS-44001] Fix GitHub Security Warnings

kaiwalyajoshi:
Update the following packages:
- cryptography to 2.3.1
- requests to 2.20.0

Remove pycrypto-2.6.1 by removing python3-crypto
Replace pycrypto-2.6.1 with pycryptodome-3.7.0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 # not be updating this dependency.  See
 # https://mesosphere.slack.com/archives/C4E91G0CX/p1541505296001800 for more
 # background.
-FROM mesosphere/dcos-commons-base:af8469627a8cc4c823fac826a5e02f0acaf07869
+FROM mesosphere/dcos-commons-base@sha256:076d1fdf4033ccd25fcedc3402a23cf017285672181d94e2f6fb11ace48c310e
 
 ENV GO_VERSION=1.10.2
 ENV PATH=$PATH:/usr/local/go/bin
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 # not be updating this dependency.  See
 # https://mesosphere.slack.com/archives/C4E91G0CX/p1541505296001800 for more
 # background.
-FROM mesosphere/dcos-commons-base@sha256:e9f67d72aa1431f4a15544981285f028decc7507538f8e5aa8b0888a94d820c0
+FROM mesosphere/dcos-commons-base@sha256:af8469627a8cc4c823fac826a5e02f0acaf07869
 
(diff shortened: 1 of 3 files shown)


kaiwalyajoshi:
Seems mesosphere/dcos-commons-base will need to be updated, checking with team on update procedures.

kaiwalyajoshi:
Breaking this up into multiple fixes.

https://github.com/mesosphere/dcos-commons/pull/2784 for dcos-commons-base fix.

Dockerfile
```diff
@@ -4,7 +4,7 @@
 # not be updating this dependency.  See
 # https://mesosphere.slack.com/archives/C4E91G0CX/p1541505296001800 for more
 # background.
-FROM mesosphere/dcos-commons-base@sha256:e9f67d72aa1431f4a15544981285f028decc7507538f8e5aa8b0888a94d820c0
+FROM mesosphere/dcos-commons-base:af8469627a8cc4c823fac826a5e02f0acaf07869
```
elezar: Is this not a step in the wrong direction with regards to managing these dependencies?

elezar: Does this meant that you built and pushed this image separately?

frozen_requirements.txt
```diff
@@ -85,7 +85,7 @@ pyasn1==0.4.4
 pyasn1-modules==0.2.2
 pycodestyle==2.3.1
 pycparser==2.19
-pycrypto==2.6.1
+pycryptodome==3.7.0
```
elezar: Would this not require changes to `test_requirements.txt` too?

Dockerfile
kaiwalyajoshi: @elezar that was the strategy used here, this PR builds the dcos-commons-base image: https://github.com/mesosphere/dcos-commons/pull/2784

Here we are pointing to the built image, with the other updates.

I've pushed up a commit to use the sha256 hash as opposed to the git-hash which was used before for pulling the correct version of the image.

frozen_requirements.txt
kaiwalyajoshi: I'm not fully clear on how the security vulnerabilities apply to test_requirements.txt. test_requirements.txt doesn't have any mentions of installing via the base-image, specifically python3-crypto. I could use help here if I'm missing something.

elezar: I mean that we currently have two different "requirements" files that could now be out of sync due to the switch from `pycrypto` to `pycryptodome`. (but looking at the file there isn't anything specified there)

PR #2784, opened 2018-11-16
[DCOS-44001] Remove pycrypto-2.6.1 which has a security vulnerability.

kaiwalyajoshi:
First of two fixes. This one fixes the dcos-commons-base image.


kaiwalyajoshi:
This approach doesn't fix the dcos-commons-base image location in the Docker file. Closing PR.

PR #2786, opened 2018-11-22
Update fixes to 'sdk_cmd.{_ssh(), _scp()}'

klueska:
This PR updates some of the functionality of the `sdk_cmd._ssh()` and `sdk_cmd._scp()` functions.

First, it adds a missing `-W` parameter to the `sdk_cmd._scp()` function. Without this, the underlying 'scp' with a proxy through SSH would never have worked. I'm guessing this function is no longer used at the moment or else someone would have run into this.

Second, it adds the ability to only send a single key to both `sdk_cmd._ssh()` and `sdk_cmd._scp()`. Previously, there was no way to specifically set the identity file for use by the underlying 'ssh' and 'scp' commands. Now there is.

### DIFF
--- a/testing/sdk_cmd.py
+++ b/testing/sdk_cmd.py
@@ -24,6 +24,7 @@ log = logging.getLogger(__name__)
 
 DEFAULT_TIMEOUT_SECONDS = 30 * 60
 SSH_USERNAME = os.environ.get("DCOS_SSH_USERNAME", "core")
+SSH_KEY_FILE = os.environ.get("DCOS_SSH_KEY_FILE", "")
 
 # Silence this warning. We expect certs to be self-signed:
 # /usr/local/lib/python3.6/dist-packages/urllib3/connectionpool.py:857:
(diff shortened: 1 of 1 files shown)


klueska:
It appears the two failing tests here are a result of the fix proposed in: https://github.com/mesosphere/dcos-commons/pull/2788

Is there a protocol for overriding these failures and merging despite them? Or should I wait till that fix lands, rebase and try again?

testing/sdk_cmd.py
```diff
@@ -365,14 +366,28 @@ def _ssh(cmd: str, host: str, timeout_seconds: int, print_output: bool, check: b
         ]
     )
 
+    direct_args = " ".join(
```
elezar: nit: should we skip joining the empty string if `SSH_KEY_FILE` is not specified?

testing/sdk_cmd.py
```diff
@@ -397,6 +412,8 @@ def _scp(
             # -oConnectTimeout=#: Limit the duration for the connection to be created.
             #                     We also configure a timeout for the command itself to run once connected, see below.
             "-oConnectTimeout={}".format(timeout_seconds),
+            # -i <identity_file>: The identity file to use for login
+            "-i {}".format(SSH_KEY_FILE) if SSH_KEY_FILE else "",
```
elezar: Nit: There seems to be some overlap between these args and those used in `_ssh`. Could we have a function that returns them.

klueska: I would prefer to do this in a separate cleanup commit / PR, but I can do it here if you think it's necessary.

testing/sdk_cmd.py
klueska: I personally like the way it is now because the code is cleaner. Doing what you suggest will require an extra if statement here, for the same result.

elezar: you mean:
```suggestion
    direct_args = " ".join(str(s) if s for s in
```
But yes, it's not critical.

testing/sdk_cmd.py
elezar: A separate one or a "note-to-self" is fine.

klueska: After looking more closely at this, there are only 2 flags that scp and ssh share, and I'm not sure how I'd actually split them out to have a helper function without it convoluting the code. Sometimes its preferable to duplicate if it means maintenance and verification of correctness are more isolated.

testing/sdk_cmd.py
elezar: (the first level of `[ ]` are not needed), but this is a bit verbose for no real additional value. Let's leve it as is.

testing/sdk_cmd.py
elezar: Yes. There isn't too much overlap, so the advantage is minimal.

elezar:
(for the record, I updated this branch off `master` to re-trigger the tests)

PR #2789, opened 2018-11-23
[DCOS-45209] Remove Kafka framework from the mono-repo

elezar:
This PR is a follow up to mesosphere/dcos-kafka-service#382 which updates the `dcos-kafka-service` repo to be equivalent to a `master` build of Kafka.

Here, the `frameworks/kafka` folder is removed and references to `kafka` in the documentation and examples is minimised.

**Note**: It is expected that the Kafka tests fail with this change.


cli/README.md
```diff
@@ -98,6 +98,6 @@ When installed, the module is executed as a regular binary by the DC/OS CLI, wit
 $ <exename> <modulename> [args]
 ```
 
-So for example, if someone called `dcos kafka broker list`, the `dcos-service-cli` CLI module would be run as `dcos-service-cli kafka broker list` by the DC/OS CLI.
+So for example, if someone called `dcos hello-world pod list`, the `dcos-service-cli` CLI module would be run as `dcos-service-cli hello-world pod list` by the DC/OS CLI.
 
-The inclusion of `modulename` (`kafka` in this example) as an argument allows reuse of a single CLI module build across multiple packages. In particular this is used by the default CLI build which is shared across most packages.
+The inclusion of `modulename` (`hello-workd` in this example) as an argument allows reuse of a single CLI module build across multiple packages. In particular this is used by the default CLI build which is shared across most packages.
```
kaiwalyajoshi: (nit) spelling on `hello-workd`

elezar: Thanks. I've corrected this and also replaced `hello-world` with `helloworld` in this doc. It should be the package name.

PR #2790, opened 2018-11-26
[TOOLS] Add a package version envvar to force an update if there is n…

elezar:
…o difference between the universe and non-universe version (#2788)

This is a backport of #2788 to fix the sdk-0.40 nightlies.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,8 +3,11 @@
 /**/__pycache__
 *.pyc
 .*
+/service-diagnostic-bundles
+/logs
 
 !test_requirements.txt
+!frozen_requirements.txt
 !test_runner.sh
 !/tools/ci
 !test.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,50 +1,28 @@
-FROM ubuntu:18.04
+# See Dockerfile.base for instructions on how to update this base image.
+# TODO: change to mesosphere/dcos-commons-base:latest@sha256:somechecksum once
+# the docker on TeamCity is recent enough. Until it happens, renovatebot will
+# not be updating this dependency.  See
+# https://mesosphere.slack.com/archives/C4E91G0CX/p1541505296001800 for more
+# background.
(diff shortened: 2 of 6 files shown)


samvantran:
This PR (and original PR) both failed the `hello-world` tests and `strict` tests. Is that expected?

elezar:
No, that's not expected. There seem to be issues with the 0.40 branch following the release fo the `3.0.0-0.54.3` hello-world version. This would indicate that **DOWNGRADES** are not supported.

takirala:
@samvantran if by original PR, you mean #2788 looks like it passed all the CI checks.

@elezar I don't think we have any documented or official notice on breaking upgradability/downgradability. If this is not possible, it _may_ be a bug in either the SDK or the testing utilities. I found this in logs from one of the failing CIs: 

```
[2018-11-27 22:22:55,897|sdk_install|INFO]: Installed package=hello-world service=/test/integration/hello-world after 33.5s
[2018-11-27 22:22:56,306|sdk_upgrade|INFO]: Waiting for test build version of hello-world to appear: version != 3.0.0-0.54.3
[2018-11-27 22:22:57,424|sdk_upgrade|INFO]: Current version of hello-world is: stub-universe
[2018-11-27 22:22:57,424|sdk_upgrade|INFO]: Upgrading hello-world: 3.0.0-0.54.3 => stub-universe
[2018-11-27 22:22:58,001|sdk_cmd|WARNING]: stderr for command 'hello-world --name=/test/integration/hello-world config target' is non-empty: dcos-hello-world: error: This command is deprecated, use 'dcos hello-world debug config target' instead, try --help
[2018-11-27 22:22:58,001|sdk_cmd|WARNING]: Error converting stdout to json:
```

I think this error is because of the testing codebase assumed sdk-40 hw (and thus sdk 40 CLI) but the universe hw is of sdk-50 (and thus the CLI is from master). I think we can either ignore the failure (as it happens only for hw and not in other frameworks) or fix the test to use a sdk 40 version of hw.

elezar:
Thanks for digging @takirala. I will have a look at how much effort is required in cleaning this up and then we can make a call as to what is the best way forward.

elezar:
Closing this PR.

PR #2791, opened 2018-11-27
[DCOS-44701] adding interface for debug endpoints

kvish:
Add common debug endpoint interface that all debug endpoints will implement


takirala:
Is it intentional to use a new top level `sdk/debug` directory? Does it work with out gradle build file ?

takirala:
> we discussed putting the debug related code under sdk/debug ...

I must have misunderstood this. I thought the code would be just a new package inside sdk/scheduler/src/main/java/com/mesosphere/**sdk** I was thinking of sdk in the existing src code. That being said, do we gain anything by creating a new top level directory?

> Did you have any concerns with landing these commits under sdk/debug

Being somewhat unfamiliar with gradle, i will say this : I _think_ without registering this as a subproject in gradle settings [here](https://github.com/mesosphere/dcos-commons/blob/master/settings.gradle) we are creating alienated source code that is difficult to manage (inherit dependencies, apply linting rule).

kvish:
going to go ahead and merge this, since its just a first step in landing other debug code. I will look into required gradle changes on my taskInfo PR.

PR #2792, opened 2018-11-27
[COPS-4148] fixes failing quorum on replacing HDFS journalnode

soumasish:
### **What changes were proposed in this pull request?**
Resolves [COPS-4148](https://jira.mesosphere.com/browse/COPS-4148)
This fixes the issue of quorum failing on HDFS when replacing a HDFS Journalnode and adds an integration test.
https://jira.mesosphere.com/browse/COPS-4148
### **How were these changes tested?**
Integration tests.

### DIFF
--- a/frameworks/hdfs/tests/test_sanity.py
+++ b/frameworks/hdfs/tests/test_sanity.py
@@ -13,6 +13,7 @@ import sdk_recovery
 import sdk_tasks
 import sdk_upgrade
 import sdk_utils
+import retrying
 
 from tests import config
 
--- a/frameworks/hdfs/tests/test_sanity.py
+++ b/frameworks/hdfs/tests/test_sanity.py
@@ -407,7 +407,19 @@ def test_namenodes_acheive_quorum_after_journalnode_replace():
     This test aims to check that namenodes recover after a journalnode failure.
     It checks the fix to this issue works: https://jira.apache.org/jira/browse/HDFS-10659
     """
-    pod_list = ["journal-0", "journal-1", "journal-2"]
-    for pod in pod_list:
-        sdk_recovery.check_permanent_recovery(config.PACKAGE_NAME, foldered_name, pod, recovery_timeout_s=25 * 60,
-                                              pods_with_updated_tasks=pod_list + ["name-0", "name-1"])
+
+    @retrying.retry(stop_max_delay=150 * 1000,
+                    wait_fixed=1000,
+                    retry_on_result=lambda res: res == 0)
+    def wait_for_failed_tasks():
+        return sdk_tasks.get_failed_task_count(service_name=foldered_name)
+
+    sdk_cmd.svc_cli(config.PACKAGE_NAME, foldered_name, "pod replace journal-0")
+    config.expect_recovery(service_name=foldered_name)
+
+    sdk_cmd.svc_cli(config.PACKAGE_NAME, foldered_name, "pod restart journal-1")
+    config.expect_recovery(service_name=foldered_name)
+
+    assertEqual(0, wait_for_failed_tasks)
+
+
--- a/frameworks/hdfs/tests/test_sanity.py
+++ b/frameworks/hdfs/tests/test_sanity.py
@@ -407,10 +407,7 @@ def test_namenodes_acheive_quorum_after_journalnode_replace():
     This test aims to check that namenodes recover after a journalnode failure.
     It checks the fix to this issue works: https://jira.apache.org/jira/browse/HDFS-10659
     """
-
-    journal_task = sdk_tasks.get_service_tasks(foldered_name, "journal-0")[0]
-    sdk_cmd.kill_task_with_pattern("journalnode", "nobody", agent_host=journal_task.host)
-
-    pod_list = ["name-0", "name-1"]
+    pod_list = ["journal-0", "journal-1", "journal-2"]
     for pod in pod_list:
-        sdk_recovery.check_permanent_recovery(config.PACKAGE_NAME, foldered_name, pod, recovery_timeout_s=25 * 60)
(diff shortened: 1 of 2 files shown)


frameworks/hdfs/tests/test_sanity.py
```diff
@@ -398,3 +398,19 @@ def test_permanently_replace_journalnodes():
...
+def test_namenodes_acheive_quorum_after_journalnode_replace():
+    config.check_healthy(service_name=foldered_name)
+    name_0_ids = sdk_tasks.get_task_ids(foldered_name, "name-0")
+    name_1_ids = sdk_tasks.get_task_ids(foldered_name, "name-1")
+    pod_list = ["journal-0", "journal-1", "journal-2"]
+    for pod in pod_list:
+        sdk_recovery.check_permanent_recovery(
+            config.PACKAGE_NAME, foldered_name, pod, recovery_timeout_s=25 * 60
+        )
+    config.expect_recovery(service_name=foldered_name)
```
mpereira: Why is this this call to `config.expect_recovery` required?

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -398,3 +398,19 @@ def test_permanently_replace_journalnodes():
...
+    sdk_tasks.check_tasks_updated(foldered_name, "name-0", name_0_ids)
+    sdk_tasks.check_tasks_updated(foldered_name, "name-1", name_1_ids)
```
mpereira: Isn't `sdk_recovery.check_permanent_recovery` checking for the opposite condition in https://github.com/mesosphere/dcos-commons/blob/6989eeea796a9f8cb5b04c17b09b748db2c19c7a/testing/sdk_recovery.py#L69? Since CI still didn't test this PR, did this test pass on a manual run?

frameworks/hdfs/tests/test_sanity.py
soumasish: Just an additional layer of security. The idea of the test is to check the namenodes have recovered. But yes we can remove that.

frameworks/hdfs/tests/test_sanity.py
soumasish: Yes its still on a manual run locally.

soumasish: In all honesty, ` check_healthy(service_name=service_name, count=DEFAULT_TASK_COUNT, recovery_expected=True)` should be enough to tell the service is healthy and the namenodes are fine, but I was being more explicit about it.

akirillov:
thanks for the fix @soumasish, it would be great to include a short description of the test logic somewhere. It's not quite clear right away why we need to check for NameNode tasks updates after replacing JournalNode tasks. 

Also, it would be great to add a link to the original HDFS issue to provide more context.

akirillov:
thanks for the fix @soumasish, a few minor comments:
- it would be great to add a link to the original HDFS issue to provide more context
- integration test cleanup based on comments

Apart from this, the PR looks good, let's wait for CI build results.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -398,3 +398,19 @@ def test_permanently_replace_journalnodes():
...
+
+@pytest.mark.sanity
+@pytest.mark.recovery
+def test_namenodes_acheive_quorum_after_journalnode_replace():
+    config.check_healthy(service_name=foldered_name)
+    name_0_ids = sdk_tasks.get_task_ids(foldered_name, "name-0")
+    name_1_ids = sdk_tasks.get_task_ids(foldered_name, "name-1")
+    pod_list = ["journal-0", "journal-1", "journal-2"]
+    for pod in pod_list:
+        sdk_recovery.check_permanent_recovery(
```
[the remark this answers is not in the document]
soumasish: Making this change.

frameworks/hdfs/tests/test_sanity.py
```diff
@@ -398,3 +398,16 @@ def test_permanently_replace_journalnodes():
...
+@pytest.mark.sanity
+@pytest.mark.recovery
+def test_namenodes_acheive_quorum_after_journalnode_replace():
+    """
+    This test aims to check that namenodes recover after a journalnode failure.
+    It checks the fix to this issue works: https://jira.apache.org/jira/browse/HDFS-10659
+    """
+    pod_list = ["journal-0", "journal-1", "journal-2"]
+    for pod in pod_list:
+        sdk_recovery.check_permanent_recovery(config.PACKAGE_NAME, foldered_name, pod, recovery_timeout_s=25 * 60,
```
akirillov: LGTM, thanks @soumasish. Let's wait for CI tests to complete successfully

soumasish:
Retagged @samvantran @akirillov @mpereira the build passes now.

frameworks/hdfs/tests/test_sanity.py
samvantran: Sorry @soumasish I don't think this test does what you think it does. It could be a bug in the `sdk_recovery` tooling but if you search `0510d1b2-78cf-48e3-b01d-db8e35ada60` in the full build log, you'll see that for every pod you replace, eg `journal-0`, the replaced tasks all get appended with the same ID and what ends up happening is the test checks all the new tasks should have new IDs but... it'll always be true bc they had the wrong IDs to start with.

And isn't this a test for namenode finding quorum ie not failing to find the `paxos` directory? Should we not instead check that namenodes are healthy (not replaced) after JN failures _and_ there was no exception thrown in the logs regarding missing paxos?

wavesoft:
@akirillov , compared to `cops-4148-test-failure` I see that you have an additional test part there, is there any reason why this isn't here too?

```
    recovery_plan = sdk_plan.get_plan(foldered_name, "recovery")
    recovered_pods = [phase['name'] for phase in recovery_plan["phases"]]
     for pod in recovered_pods:
        assert "name-" not in pod, "Name Node pods shouldn't be recovered when Journal Nodes are"
```

PR #2795, opened 2018-11-29
[DCOS-44702] Implement Plans Debug Endpoint

kaiwalyajoshi:
- Shows a unified view of all plans (including custom-plans)
- Shows the currently active plans.
- Shows the current state the scheduler is in which can be on of (DEPLOYING, DEPLOYED, RECOVERING and DECOMMISIONING)


sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,297 @@
...
+
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Response;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+// SUPPRESS CHECKSTYLE MultipleStringLiteralsCheck
```
kensipe: why do you have this suppressed at the class level and repeated throughout the code?
In general the amount of checkstyle suppression seems aggressive IMO

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,297 @@
...
+    this.planCoordinator = planCoordinator;
+    this.stateStore = stateStore;
+  }
+
+  private JSONObject generateServiceStatus(String filterPlan,
+                                           String filterPhase,
+                                           String filterStep,
+                                           boolean allowFiltering,
+                                           boolean skipDetails)
+  {
```
kensipe: inconsistent with the `{` opening... all previous to this are end of line... this is beginning of line.   can we have all end of line?

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,297 @@
...
+    for (PlanManager planManager : planCoordinator.getPlanManagers()) {
+      Plan plan = planManager.getPlan();
+
+      //Filter down to a plan if specified.
+      if (allowFiltering && filterPlan != null && !plan.getName().equalsIgnoreCase(filterPlan)) {
+        continue;
+      }
+
+      JSONObject planObject = new JSONObject();
+      JSONArray phaseArray = new JSONArray();
```
kensipe: IMO... I would rather create a POJO of type we are interested in... and have a separate step what transforms to JSON..  there should be libraries that given nested POJOs will generate the output we are looking for.   This seems to conflate the gather of the info with the format we need as an output.

kensipe: perhaps this is that conversation class... I'm new to this code base..

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,297 @@
...
+
+        phaseObject.put("name", phase.getName());
+        if (!skipDetails) {
+          phaseObject.put("status", phase.getStatus());
+          phaseObject.put("strategy", phase.getStrategy().getName());
+        }
+        phaseObject.put("steps", stepArray);
+
+        //Iterate over steps.
+        // SUPPRESS CHECKSTYLE NestedForDepthCheck
```
kensipe: lets refactor and remove this suppress

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,297 @@
...
+        phaseArray.put(phaseObject);
+      }
+
+      plansArray.put(planObject);
+    }
+    outcome.put("plans", plansArray);
+    return outcome;
+  }
+
+  // SUPPRESS CHECKSTYLE ReturnCountCheck
```
kensipe: I would love to not have this suppression!    I'm not a fan of this many returns... or put another way... I love this checkstyle :)

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,297 @@
...
+      Plan plan = planManager.getPlan();
+      planMap.put(plan.getName(), plan);
+    }
+
+    //Set scheduler to one of DEPLOYING, DEPLOYED, RECOVERING or DECOMISSIONING.
+    //If unknown, set to generic status to RUNNING when running custom-plans.
+    String schedulerStatus = "RUNNING";
+    if (planMap.containsKey(Constants.DEPLOY_PLAN_NAME) &&
+        planMap.get(Constants.DEPLOY_PLAN_NAME).isRunning())
+    {
```
kensipe: this style is inconsistent `{` again... it's different all over the file.

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,297 @@
...
+
+      if (!skipDetails) {
+        //Get a rollup aggregation of the steps.
+        int totalSteps = plan.getChildren().stream()
+            .flatMap(phase -> phase.getChildren().stream())
+            .collect(Collectors.toSet())
+            .size();
+        int completedSteps = plan.getChildren().stream()
+            .flatMap(phase -> phase.getChildren().stream())
+            .filter(step -> step.isComplete())
```
kensipe: this could be `.filter(Element::isComplete)` which seems easier read... but isn't a big deal

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,297 @@
...
+        true));
+
+    response.put("service-status", generateServiceStatus(filterPlan,
+        filterPhase,
+        filterStep,
+        true,
+        false));
+
+    //Retrieve the latest Plans from PlanCoordinator. Some plans i.e recovery do change over time.
+    HashMap<String, Plan> planMap = new HashMap<String, Plan>();
```
kensipe: `HashMap<String, Plan> planMap = new HashMap<String, Plan>(` could be `HashMap<String, Plan> planMap = new HashMap<>(`

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
kaiwalyajoshi: IIRC it was on the same line before, but checkstyle didn't like that. Same goes for many other ones below.

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
kaiwalyajoshi: Yea, this class is intended for converting between the internal representation to JSON. I'll work on creating simplified POJOs, flooding them the desired state and serializing that as JSON. Lets see if that simplifies the logic here.

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
takirala: This is a [rule](http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/blocks/LeftCurlyOption.html#NLOW) enforced by checkstyle that we use

kaiwalyajoshi:
@kensipe please have a look at the new version which makes use of POJO's to render the final payload. Keeping this PR in WIP as I need to mock the Plans & PlanCoordinators to generate the tests for this class.

sdk/scheduler/src/test/java/com/mesosphere/sdk/debug/PlansTrackerTest.java
```diff
@@ -0,0 +1,382 @@
...
+    SerializeElement world2StepElement = worldPhaseElement.getChildren().get().get(1);
+    Assert.assertEquals("world-step-2", world2StepElement.getName());
+    Assert.assertEquals("step", world2StepElement.getType());
+    Assert.assertNotNull(world2StepElement.getChildren());
+    Assert.assertEquals(false, world2StepElement.getChildren().isPresent());
+
+    //Done testing filtered plans.
+  }
+
+}
```
kvish: nit: newline

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/PlansTracker.java
```diff
@@ -0,0 +1,492 @@
...
+    }
+
+    //If a phase is defined ensure parent plan is also provided.
+    if (!invalidInputReason.isPresent() && filterPhase != null && filterPlan == null) {
+      invalidInputReason = Optional.of("Phase specified without parent Plan.");
+    }
+
+    //Ensure correct ownership. Start with the plan.
+
+    //Plan is specified, ensure it exists within our list of plans.
```
kvish: this comment is also on line 88, comments should be cleaned up here

PR #2796, opened 2018-11-29
[WIP][DCOS-44703] Implement TaskStatus Endpoint

kvish:


sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/TaskStatusesTracker.java
```diff
@@ -0,0 +1,98 @@
...
+    this.planCoordinator = planCoordinator;
+    this.stateStore = stateStore;
+  }
+
+  //CHECKSTYLE:OFF NestedForDepth
+
+  private JSONArray getTaskStatuses(String filterPlan, String filterPhase, String filterStep) {
+    JSONArray planArray = new JSONArray();
+    for (PlanManager planManager : planCoordinator.getPlanManagers()) {
+      if (filterPlan != null && !planManager.getPlan().getName().equalsIgnoreCase(filterPlan)) {
```
timcharper: what are your guys' thoughts on adopting Java 8's Optional type?

https://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/TaskStatusesTracker.java
```diff
@@ -0,0 +1,98 @@
...
+        continue;
+      }
+      Plan plan = planManager.getPlan();
+      JSONObject planObject = new JSONObject();
+      planObject.put("plan", plan.getName());
+      for (Phase phase : plan.getChildren()) {
+        if (filterPhase != null && !phase.getName().equalsIgnoreCase(filterPhase)) {
```
timcharper: seems like `phase.getName` shouldn't ever return null, so `!phase.getName().equalsIgnoreCase(filterPhase)` should be sufficient here? `"asdf".equalsIgnoreCase(null)` returns false

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/TaskStatusesTracker.java
```diff
@@ -0,0 +1,98 @@
...
+          continue;
+        }
+        JSONObject phaseObject = new JSONObject();
+        for (Step step : phase.getChildren()) {
+          JSONArray taskArray = new JSONArray();
+          if (filterStep != null && !step.getName().equalsIgnoreCase(filterStep)) {
```
timcharper: ditto ☝️

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/TaskStatusesTracker.java
timcharper: (I actually see that it's already used down below; if it is valid to provide a null filterPhase, then perhaps the param should be changed to Optional?)

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/TaskStatusesTracker.java
```diff
@@ -0,0 +1,98 @@
...
+            continue;
+          }
+          Collection<TaskSpec> tasksInStep = step
+              .getPodInstanceRequirement()
+              .get().getPodInstance().getPod().getTasks();
+          for (TaskSpec taskSpec : tasksInStep) {
+            //filter tasks in pod belonging to this step
+            if (step.getName().contains(taskSpec.getName())) {
```
timcharper: I'm feeling like this using Java8 streams would be very helpful here; this conditional here could be a filter and this body could be a map function; taskArray could be built from the resulting stream.

sdk/scheduler/src/main/java/com/mesosphere/sdk/debug/TaskStatusesTracker.java
```diff
@@ -0,0 +1,98 @@
...
+              String taskInstanceName = TaskSpec.getInstanceName(
+                  step.getPodInstanceRequirement().get().getPodInstance(),
+                  taskSpec.getName()
+              );
+              JSONObject taskStatus = new JSONObject();
+              Optional<Protos.TaskStatus> status = stateStore.fetchStatus(taskInstanceName);
+              taskStatus.put("taskName", taskInstanceName);
+              if (status.isPresent()) {
```
timcharper: use a lambda?

```
status.ifPresent(s -> {
  taskStatus.put("taskId", s.getTaskId());
  ...
})
```

PR #2857, opened 2018-12-28
[DCOS-46514] Adding a regex validator

soumasish:
# What changes were proposed in this pull request?
Added a regex validator to check stub universe name for the service instance consisted of a series of words separated by slashes. Each word must be at least 1 alphanumeric character and may only contain digits (`0-9`), dashes (`-`), dots (`.`), and lowercase letters (`a-z`). The word may not begin or end with a dash",
# How were the changes tested?
Manually by creating a HDFS package on a CCM cluster.

### DIFF
--- a/frameworks/hdfs/tests/config.py
+++ b/frameworks/hdfs/tests/config.py
@@ -22,7 +22,7 @@ TEST_CONTENT_SMALL = "This is some test data"
 DEFAULT_HDFS_TIMEOUT = 5 * 60
 HDFS_POD_TYPES = {"journal", "name", "data"}
 KEYTAB = "hdfs.keytab"
-HADOOP_VERSION = "hadoop-2.6.0-cdh5.16.1"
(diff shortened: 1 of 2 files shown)


frameworks/hdfs/tests/config.py
```diff
@@ -22,7 +22,7 @@
 DEFAULT_HDFS_TIMEOUT = 5 * 60
 HDFS_POD_TYPES = {"journal", "name", "data"}
 KEYTAB = "hdfs.keytab"
-HADOOP_VERSION = "hadoop-2.6.0-cdh5.9.1"
+HADOOP_VERSION = "hadoop-2.6.0-cdh5.16.1"
```
akirillov: Doesn't look like this change belongs here

soumasish: Pushed accidentally, reverted this change to it related branch.

PR #2866, opened 2018-12-31
COPS-4320: Configurable timeouts for readiness checks.

kaiwalyajoshi:
Make readiness-check interval, timeout and delay configurable for the following:
- Cassandra
- HDFS (current defaults maintained for journal-node, name-node and data-node)
- Elasticsearch
- Hello-World

Default readiness interval is set to 60 and timeout to 20 to match Marathon default values. Values still need to be fine-tuned via discussion through this PR.

### DIFF
--- a/frameworks/cassandra/universe/config.json
+++ b/frameworks/cassandra/universe/config.json
@@ -108,7 +108,28 @@
               }
             }
           }
-        }
+        },
+        "readiness_check" : {
+          "description" : "Readiness check settings",
+          "type": "object",
+          "properties": {
+            "interval": {
+              "type": "integer",
+              "description": "The period in seconds to wait after the last readiness check has completed to start the next check.",
+              "default": 60
+            },
+            "delay": {
+              "type": "integer",
+              "description": "An amount of time in seconds to wait before starting the readiness check attempts.",
+              "default": 0
+            },
+            "timeout": {
+              "type": "integer",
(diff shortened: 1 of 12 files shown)


frameworks/cassandra/universe/config.json
```diff
@@ -108,7 +108,28 @@
               }
             }
           }
-        }
+        },
+        "readiness_check" : {
+          "description" : "Readiness check settings",
```
porridge: The descriptions should probably contain a warning that messing with these values can lead to spurious failures of the tasks in the framework.

frameworks/cassandra/universe/config.json
```diff
@@ -108,7 +108,28 @@
...
+          "type": "object",
+          "properties": {
+            "interval": {
+              "type": "integer",
+              "description": "The period in seconds to wait after the last readiness check has completed to start the next check.",
+              "default": 60
```
porridge: All these values must default to what they were before this PR.

kaiwalyajoshi:
@vishnu2kmohan any objections to @porridge's approach above? We'll leave the current values as default, but they can be overriden as needed by the customer.

kaiwalyajoshi:
@akirillov we can indeed set minimums in config.json. I figured we'd set sane minimums *after* we had conducted our benchmarks.

PR #2869, opened 2019-01-05
[DCOS-46650] Added a regex name validator for Cassandra to check that stub universe names follow a pattern.

soumasish:
## What this PR aims to do?
1. Part of  Release Cassandra-2.5.0-3.11.3.
2. Adds a regex name validator to Cassandra to check that it follows a specific pattern.
## How were the features tested?
The regex name validator was tested manually by creating a CCM cluster and trying an incorrect name.

### DIFF
--- a/frameworks/cassandra/universe/config.json
+++ b/frameworks/cassandra/universe/config.json
@@ -6,8 +6,9 @@
       "description": "DC/OS Apache Cassandra service configuration properties",
       "properties": {
         "name": {
-          "description": "The name of the Cassandra service instance.",
+          "description": "Unique name for the service instance consisting of a series of words separated by slashes. Each word must be at least 1 alphanumeric character and may only contain digits (`0-9`), dashes (`-`), dots (`.`), and lowercase letters (`a-z`). The word may not begin or end with a dash",
           "type": "string",
+          "pattern": "^(\\/?((\\.\\.)|(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9]))?($|\\/))+$",
           "default": "cassandra"
         },
         "user": {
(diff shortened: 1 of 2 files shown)


frameworks/cassandra/universe/config.json
```diff
@@ -6,8 +6,9 @@
       "description": "DC/OS Apache Cassandra service configuration properties",
       "properties": {
         "name": {
-          "description": "The name of the Cassandra service instance.",
+          "description": "Unique name for the service instance consisting of a series of words separated by slashes. Each word must be at least 1 alphanumeric character and may only contain digits (`0-9`), dashes (`-`), dots (`.`), and lowercase letters (`a-z`). The word may not begin or end with a dash",
           "type": "string",
+          "pattern": "^(\\/?((\\.\\.)|(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9]))?($|\\/))+$",
```
porridge: A few questions:
- do we really need to escape forward slashes?
- do we need to escape a dash in a character class if it's the last character? All regex engines I've seen so far do not require this.

soumasish: Thanks @porridge for the review. This regex has been tested manually during HDFS release, by creating a cluster and giving it an incorrect name. But I think it makes sense to have a unit test for this. 
As for the engine, locally I've tested it in Python and using some free tools which uses a Javascript regex engine.
Finally, I'll rename the PR and make it pertinent to the specific issue we're addressing here.

porridge:
Hm, I noticed that this PR was closed, with the unit test still failing, and without any additional comment. Instead another PR, https://github.com/mesosphere/dcos-commons/pull/2924 which does *not* have the unit test, was merged. Was this intended?

I also noticed today that the unit test here is in Python and tests a copy (?) of the regex in the YAML file, which makes it rather useless.

PR #2877, opened 2019-01-09
[DCOS-46788] Prepare diagnostic script release.

porridge:

PR #2878, opened 2019-01-09
[DIAGNOSTICS] Fix condition where scheduler task can't be found.

mpereira:
This happens due to a name mismatch. The DCOS_SERVICE_NAME label in SDK services and the service name provided through the CLI (i.e. `dcos $package --name=$service_name ...`) may or may not be prefixed by a slash, so comparing the two as-is sometimes doesn't work.

This change makes sure to get the scheduler task with and without a slash prefix in the name.

### DIFF
--- a/tools/diagnostics/full_bundle.py
+++ b/tools/diagnostics/full_bundle.py
@@ -15,6 +15,9 @@ import config
 log = logging.getLogger(__name__)
 
 
+DCOS_SERVICES_JSON_FILE_NAME = "dcos_services.json"
+
+
 @config.retry
 def get_dcos_services() -> (bool, str):
     rc, stdout, stderr = sdk_cmd.run_cli(
@@ -118,7 +121,7 @@ class FullBundle(Bundle):
 
         all_services = json.loads(all_services_or_error)
 
-        self.write_file("dcos_services.json", all_services, serialize_to_json=True)
+        self.write_file(DCOS_SERVICES_JSON_FILE_NAME, all_services, serialize_to_json=True)
 
         # An SDK service might have multiple DC/OS service entries. We expect that at most one is
         # "active".
@@ -152,8 +155,9 @@ class FullBundle(Bundle):
         ]
         if len(scheduler_tasks) == 0:
             log.warn(
-                "Could not find scheduler tasks for '%s'. Please check the 'dcos_services.json' file.",
+                "Could not find scheduler tasks for '%s' under the Marathon service ('\"name\": \"marathon\"') 'tasks' key in '%s'.",
                 self.service_name,
+                DCOS_SERVICES_JSON_FILE_NAME,
             )
 
         ServiceBundle(
--- a/tools/diagnostics/full_bundle.py
+++ b/tools/diagnostics/full_bundle.py
@@ -30,7 +30,7 @@ def get_dcos_services() -> (bool, str):
         return (True, stdout)
 
 
-def is_service_named(sdk_service_name: str, dcos_service_name: dict) -> bool:
+def is_service_named(sdk_service_name: str, dcos_service_name: str) -> bool:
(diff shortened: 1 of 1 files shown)


tools/diagnostics/full_bundle.py
```diff
@@ -60,7 +60,7 @@ def is_service_scheduler_task(package_name: str, service_name: str, task: dict)
     dcos_service_name = next(
         iter([l.get("value") for l in labels if l.get("key") == "DCOS_SERVICE_NAME"]), ""
     )
-    return dcos_package_name == package_name and dcos_service_name == service_name
+    return dcos_package_name == package_name and is_service_named(service_name, dcos_service_name)
```
mpereira: This is the important bit.

tools/diagnostics/full_bundle.py
```diff
@@ -146,6 +150,11 @@ def create(self) -> (int, "FullBundle"):
             for t in active_marathon_service.get("tasks", [])
             if is_service_scheduler_task(self.package_name, self.service_name, t)
         ]
+        if len(scheduler_tasks) == 0:
+            log.warn(
+                "Could not find scheduler tasks for '%s'. Please check the 'dcos_services.json' file.",
```
porridge: I see a few issues here:
1. please make `dcos_services.json` a constant and use it both here and on line 121. Such reference will ensure that the warning will not rot if someone changes the filename.
2. the warning is not very precise, we should explain that we're looking among active tasks of marathon service with id such-and-such
3. it's not terribly clear what the end-user is supposed to look for in that file? Look at other marathon services? unreachable or completed tasks too?
4. the json file tends to be large and very hard to view without a viewer which can collapse subtrees, even if someone knows what structure to expect. Perhaps we can do this search for them? Although this is probably outside of scope of this PR.

tools/diagnostics/full_bundle.py
```diff
@@ -146,6 +150,11 @@ def create(self) -> (int, "FullBundle"):
             for t in active_marathon_service.get("tasks", [])
             if is_service_scheduler_task(self.package_name, self.service_name, t)
         ]
+        if len(scheduler_tasks) == 0:
```
porridge: `if not scheduler_tasks:` would be more Pythonic...

tools/diagnostics/full_bundle.py
```diff
@@ -30,26 +30,30 @@ def get_dcos_services() -> (bool, str):
         return (True, stdout)
 
 
-def is_service_named(service_name: str, service: dict) -> bool:
+def is_service_named(sdk_service_name: str, dcos_service_name: str) -> bool:
```
porridge: Have you perhaps investigated why the leading slash is missing in some cases?
Perhaps stripping leading slash from both arguments will make it harder to misuse the function?
Since both arguments are now strings, it could then be called `service_names_match`?

tools/diagnostics/full_bundle.py
mpereira: I agree with `4`, but I'll leave it out of the scope of this PR.

I think I addressed the other points here: 5bd6e17e976286677a13aad18b6f56a26dc55895

tools/diagnostics/full_bundle.py
mpereira: Aye 82deb678bb7c08d97013021d636e2b5571ea20ae

tools/diagnostics/full_bundle.py
mpereira: Good suggestion 1b334c89b58b85e1eb8e39c276dce1de2eee7ec4

porridge:
Could you please describe these changes in https://github.com/mesosphere/dcos-commons/pull/2877 once you merge?

mpereira:
Will do!

PR #2879, opened 2019-01-09
Bump SDK snapshot version post release of v0.55.0

kaiwalyajoshi:


build.gradle
```diff
@@ -43,7 +43,7 @@ allprojects {
     // 2. In the GitHub UI, add a Release entry against the tag created in step 1. List any changes in release notes.
     // 3. Create a PR which bumps this from '0.10.0-SNAPSHOT' to '0.11.0-SNAPSHOT' in master to start the 0.11.0 track.
     // --
-    version = '0.55.0-SNAPSHOT'
+    version = '0.56.0-SNAPSHOT'
```
takirala: I think in the past this is what we have done:

- Keep making patch releases such as `v0.55.1` , `v0.55.2` from master even if master is `v0.55.0-SNAPSHOT`.
- Bump master only when we need to make a `v0.56.0` release. This delay until requirement also helps us release `v0.54.x` from `v0.55.0-SNAPSHOT` or `v0.54.(x-1)` easily _if_ needed.

But i don't have a strong preference for anything.

kaiwalyajoshi: I'm perfectly fine with following the convention you mentioned @takirala.
Any objections @kvish or @vishnu2kmohan? Will close this PR otherwise.

vishnu2kmohan: I'll defer to @takirala and @kvish here since they have had the most experience with how the SDK has been versioned.

But I do admit that I too was a bit surprised that `master` wasn't `0.56.0-SNAPSHOT` - isn't that the typical convention for most software projects?

If ya'll get together and decide whatever is the most sane for everyone I'll go with that :)

kaiwalyajoshi:
Closing PR in favour of the conventions mentioned by Tarun above. Will bump version when we release v0.56.0

PR #2882, opened 2019-01-15
Increase Elastic xpack_upgrade_matrix test timeout.

mpereira:
This test has failed a few times because the timeout for its runtime is exactly 20 minutes and it takes on average 20 minutes to complete (sometimes a bit more, sometimes a bit less).

screen shot 2019-01-15 at 14 15 57


mpereira:
Thanks @Fabs!

PR #2886, opened 2019-01-16
[DIAGNOSTICS] Don't raise exception when there's a missing agent.

mpereira:
Also:
- Show helpful message in the logs.
- Add TODO to code.

### DIFF
--- a/tools/diagnostics/agent.py
+++ b/tools/diagnostics/agent.py
@@ -28,6 +28,9 @@ def debug_agent_files(agent_id: str) -> List[str]:
         # Retry.
         raise Exception(response)
 
+    if response.status_code == 404:
+        return {}
+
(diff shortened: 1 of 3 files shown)


tools/diagnostics/agent.py
```diff
@@ -28,6 +28,9 @@ def debug_agent_files(agent_id: str) -> List[str]:
         # Retry.
         raise Exception(response)
 
+    if response.status_code == 404:
+        return {}
+
```
mpereira: This is the important bit.

edgarlanting: Looks good to me @mpereira

edgarlanting:
Thanks @mpereira looks good to me.

PR #2888, opened 2019-01-17
[DCOS-47365] Allow passing range constraints when requesting a dynamic port

kvish:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -262,10 +268,25 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
             .get(Constants.PORTS_RESOURCE_TYPE);
     Optional<Integer> dynamicPort = Optional.empty();
     if (availablePorts != null) {
-      dynamicPort = availablePorts.getRanges().getRangeList().stream()
-          .flatMap(r -> IntStream.rangeClosed((int) r.getBegin(), (int) r.getEnd()).boxed())
-          .filter(p -> !consumedPorts.contains(p))
-          .findFirst();
+      if (!spec.getRanges().isEmpty()) {
```
takirala: (nit)
instead of writing
```
if (!condition) {
  ...
} else {
  ...
}
```
it is more readable to write
```
if (condition) {
  ...
} else {
  ...
}
``` 
but this is a super nit :)

sdk/scheduler/src/test/resources/ranges.yml
```diff
@@ -0,0 +1,24 @@
...
+        cmd: "some command"
+        ports:
+          name1:
+            port: 0
+            env-key: key1
+            ranges:
+              - begin: 1
+                end: 21
+              - begin: 2000
+                end: 5050
```
takirala: Does this mean both port `2000` and `5050` are allowed as well or is this a left closed, right open interval ? Is this documented somewhere (or are there any plans to update docs) ?

sdk/scheduler/src/test/resources/ranges.yml
```diff
@@ -0,0 +1,24 @@
...
+          name2:
+            port: 0
+            ranges:
+              - begin:
```
takirala: Assuming the following is incorrect:
```
...
...
  - begin:
    end: 21
  - begin: 
    end: 42
  - begin: 44
    end: 
```
where is this validated ? Is it possible to add a test to test that this is an invalid configuration?

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/RangeSpec.java
```diff
@@ -0,0 +1,41 @@
+package com.mesosphere.sdk.specification;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+
+/**
+ * Representation of a port range, consisting of a begin and an end.
```
takirala: May be update the javadoc here to clarify on the nature of (open vs closed on left and right) interval ?!

sdk/scheduler/src/test/resources/ranges.yml
kvish: this defines an open interval (left or right) this means give me any port less than 21 or any port less than 42 or any port greater than 44, this yml is validated here validPortRangesTest()

sdk/scheduler/src/test/resources/ranges.yml
kvish: 2000 and 5050 are allowed as well, leaving begin or end empty defines an open interval

sdk/scheduler/src/test/resources/ranges.yml
takirala: Oh, I am torn here. Although it makes sense when you explained it, i am inclined to make the check more stringent so that the above would be a invalid config. The user can rephrase this as 
```
  - begin:
    end: 42
  - begin: 44
    end: 
```
(which would be exactly the same thing - right?!)
I felt it may be confusing to keep empty values for `begin: ` or `end: ` when they are not the first (or last) occurrences in `ranges: ` but I don't have a strong preference for either.

takirala: Also, perhaps add a config validator here - https://github.com/mesosphere/dcos-commons/tree/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate (for example the user cannot decrease the port range with service update/upgrade option)?!

kvish: ya that would be the same thing, I just copy pasted the yml from a dummy sample I had which is where the overlap comes from, ya I agree it can be confusing, but @vishnu2kmohan may have thoughts

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -232,7 +235,10 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
   }
 
   private static Optional<Integer> selectDynamicPort(
-      MesosResourcePool mesosResourcePool, PodInfoBuilder podInfoBuilder, String preReservedRole)
+      MesosResourcePool mesosResourcePool,
+      PodInfoBuilder podInfoBuilder,
+      String preReservedRole,
+      PortSpec spec)
   {
```
kensipe: based on previous PR reviews I understand that dcos-commons has it's own unique style of bracket alignment... I'll state again how much I dislike it.   It should all be right aligned or match... not both.   Too many cases to remember.   To be consistent with other orchestration projects it is all right aligned.   not a blocker.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -262,10 +268,25 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
...
     if (availablePorts != null) {
-      dynamicPort = availablePorts.getRanges().getRangeList().stream()
-          .flatMap(r -> IntStream.rangeClosed((int) r.getBegin(), (int) r.getEnd()).boxed())
-          .filter(p -> !consumedPorts.contains(p))
-          .findFirst();
+      if (!spec.getRanges().isEmpty()) {
+        List<Integer> constrainedPorts = new ArrayList<>();
+        for (RangeSpec range : spec.getRanges()) {
+          constrainedPorts.addAll(
+              IntStream.range(range.getBegin(), range.getEnd())
```
kaiwalyajoshi: Looking at the reply to this comment from @takirala 
> 2000 and 5050 are allowed as well, leaving begin or end empty defines an open interval. also validated in validPortRangesTest()

Shouldn't we have a `IntStream.rangeClosed` which is inclusive of the `range.getEnd()` whereas `IntStream.range` is exclusive of `range.getEnd()`

I agree with @takirala's comment about making the docs clear about inclusive vs exclusive ranges.

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
```diff
@@ -262,10 +268,25 @@ protected void setProtos(PodInfoBuilder podInfoBuilder, Protos.Resource resource
...
+                  .boxed()
+                  .collect(Collectors.toList()));
+        }
+        dynamicPort = availablePorts.getRanges().getRangeList().stream()
+            .flatMap(r -> IntStream.rangeClosed((int) r.getBegin(), (int) r.getEnd()).boxed())
```
kensipe: I'm not a fan of single letter variables (with the exception of i and the like for indexes)
I assume r == range and p == port
I love terse... but not at the loss of readability

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/RangeSpec.java
kensipe: please state if inclusive in the docs

kaiwalyajoshi:
Minor update, might want to change the JIRA to refer to https://jira.mesosphere.com/browse/DCOS-47365 instead which is the SDK feature for this.

sdk/scheduler/src/test/resources/ranges.yml
kvish: I dont think we should restrict users from updating the port range constraints ill follow up this PR with corresponding changes to the yml reference and developer guide

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
kvish: oh ya good catch i should change this

sdk/scheduler/src/test/resources/ranges.yml
kvish: closing this for now, we can make changes based on feedback from the yml changes but for now I think its good to be as accepting as possible

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
kvish: ya this is something enforced by our linter, open to change? (maybe when we move to the USI)

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java
kvish: resolving since this is a convention already followed in dcos-commons

kvish:
Resolved all comments and adding to the default HW svc spec for "free" integration testing since tests will already verify that the tasks come up healthy. Will merge pending CI

PR #2890, opened 2019-01-18
[DCOS-47365][DOCS] Adding documentation to yml reference on port range constraints

kvish:


docs/pages/yaml-reference.md
```diff
@@ -362,6 +362,29 @@ This documentation effectively reflects the Java object tree under [RawServiceSp
 
         This may be used to define an environment variable used to advertise this port within the task. This is most useful when a random dynamic port is being used, as it allows the task to know what port was allocated for it.
 
+      * `ranges`
+
+        When requesting a dynamic port, a list of ranges may be passed to to constrain the value of the dynamic port. The range is inclusive of `begin` and `end` when they are set. If `begin` or `end` is not set the range constraint is implicit and unbounded on that side. Ranges may be defined as follows:
```
kaiwalyajoshi: The docs here aren't consistent with the code right now, the code uses an exclusive range whereas the docs mention an inclusive range. 

Code section here: https://github.com/mesosphere/dcos-commons/blob/bd6927fd012bdeb9cd7d769759036522c60b62c1/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PortEvaluationStage.java#L280

`IntStream.range` on that line should be `IntStream.rangeClosed` to be in-sync with the docs here.

kaiwalyajoshi: I've created PR #2891 reconcile the above. Approving this PR.

PR #2912, opened 2019-01-25
[DCOS-42593][WIP] Batch taskInfo writes to protect against bulk updates

kvish:


kvish:
Curator protects us against the size of an individual transaction but not the number of transactions in a given commit

sdk/scheduler/src/main/java/com/mesosphere/sdk/state/StateStore.java
```diff
@@ -201,19 +206,29 @@ public Persister getPersister() {
...
-    }
-    try {
-      persister.setMany(taskBytesMap);
-    } catch (PersisterException e) {
-      throw new StateStoreException(e, String.format("Failed to store %d TaskInfos", tasks.size()));
+    List<Protos.TaskInfo> taskInfos = tasks.stream().collect(Collectors.toList());
+    //Only taskInfos are bulk written and require batching, this should only occur at startup when a
+    //new configuration is issued, before the scheduler is initialized and doing work,
+    // so the loss of atomicity from batching should not be noticeable.
+    for (List<Protos.TaskInfo> batchedTaskInfos: Lists.partition(taskInfos, MAX_TRANSACTION_SIZE)) {
```
takirala: This would still fail if `sizeOf(singeTaskInfo) * 100 > 1 MB` right?
Instead of partitioning by taskInfos, we have to calculate the size of the actual transaction that would be made with ZK. I don't think the approach is foolproof for k8 load testing case.

kvish:
upgrading curator will be non trivial, lot of methods are using are deprecated and will require a refactor

PR #2915, opened 2019-01-29
DCOS-47532: Update SDK Diagnostic bundle to add new debuggability endpoints.

kaiwalyajoshi:
- Adds new debug endpoints.
- Uses `sdk_cmd.service_request` instead of `sdk_cmd.marathon_task_exec` which didn't work on the endpoints.

### DIFF
--- a/tools/diagnostics/README.md
+++ b/tools/diagnostics/README.md
@@ -11,6 +11,9 @@ Script that gathers SDK-related service diagnostics artifacts.
 - stdout and stderr log files for all pod tasks and their executors (even killed
   or finished ones) and scheduler tasks
 - `https://$cluster/$service/v1/debug/offers` output
+- `https://$cluster/$service/v2/debug/offers` output
+- `https://$cluster/$service/v1/debug/plans` output
+- `https://$cluster/$service/v1/debug/taskStatuses` output
 - `https://$cluster/$service/v1/configurations/$configuration_id` output for all
   `$configuration_id`s
 - Output of base-tech specific diagnostics commands (e.g.: Cassandra's `nodetool
--- a/tools/diagnostics/service_bundle.py
+++ b/tools/diagnostics/service_bundle.py
@@ -7,7 +7,6 @@ from typing import List
 
 import sdk_cmd
 import sdk_diag
-import sdk_hosts
 
 from bundle import Bundle
 import agent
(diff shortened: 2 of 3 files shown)


tools/diagnostics/service_bundle.py
```diff
@@ -149,62 +148,78 @@ def download_log_files(self):
 
     @config.retry
     def create_offers_file(self):
-        scheduler_vip = sdk_hosts.scheduler_vip_host(self.service_name, "api")
-        scheduler = self.scheduler_tasks[0]
+        response = sdk_cmd.service_request("GET", self.service_name, "/v1/debug/offers",
```
takirala: (nit) wdyt of adding a deprecated annotation here since we are adding a new method now ?!

tools/diagnostics/service_bundle.py
```diff
@@ -149,62 +148,78 @@ def download_log_files(self):
...
+    def create_taskstatuses_file(self):
+        response = sdk_cmd.service_request("GET", self.service_name, "/v1/debug/taskStatuses",
+                                           raise_on_error=False)
+        if not response.ok:
+            log.error(
+                "Could not get scheduler task-statuses\nstatus_code: '%s'\nstderr: '%s'",
+                response.status_code, response.text
+            )
+        else:
+            self.write_file("service_v1_debug_taskstatuses.json", response.text)
```
takirala: (nit) how about naming this method and the filename identical to the endpoint as `service_v1_debug_taskStatuses` ?!

tools/diagnostics/service_bundle.py
```diff
@@ -227,6 +242,9 @@ def create(self):
         self.create_pod_status_file()
         self.create_plans_status_files()
         self.create_offers_file()
+        self.create_v2_offers_file()
```
takirala: 🎊

tools/diagnostics/README.md
```diff
@@ -11,6 +11,9 @@ Script that gathers SDK-related service diagnostics artifacts.
 - stdout and stderr log files for all pod tasks and their executors (even killed
   or finished ones) and scheduler tasks
 - `https://$cluster/$service/v1/debug/offers` output
```
takirala: Does it make sense to add a note here that this endpoint is deprecated ?!

tools/diagnostics/service_bundle.py
```diff
@@ -17,7 +16,7 @@
 
 
 class ServiceBundle(Bundle):
-    DOWNLOAD_FILES_WITH_PATTERNS = ["^stdout(\.\d+)?$", "^stderr(\.\d+)?$"]
+    DOWNLOAD_FILES_WITH_PATTERNS = ["^stdout(\\.\\d+)?$", "^stderr(\\.\\d+)?$"]
```
takirala: was the previous regex invalid / not working as expected ?

PR #2918, opened 2019-01-31
[DIAGNOSTICS][DCOS-477820] Add initial support for kubernetes diagnostic bundles

kaiwalyajoshi:
In this PR we refactor the code to allow the introduction of a kubernetes_bundle

Diagnostics for the following commands have been added in this patch:
- `manager describe`
- `manager pod status --json`
- `manager plan list` and associated `manager plan status` for each plans from the list.
- `cluster list`
- `cluster debug state properties`
- `cluster debug endpoints`
- `cluster debug pod status --json`

Commands which deal with iterating through properties of the cluster such as
`dcos kubernetes cluster debug config list` aren't supported right now as not all commands in the `dcos kubernetes` cli sub-command group support a full JSON output as of this moment (particularly ones such as `cluster list`).

Lacking a JSON output makes it error-prone to parse the output. Work will be required on the kubernetes cli to emit full json versions before we can add diagnostic support for them here in a future release.

### DIFF
--- a/tools/diagnostics/base_tech_bundle/__init__.py
+++ b/tools/diagnostics/base_tech_bundle/__init__.py
@@ -1,12 +1,89 @@
 from service_bundle import ServiceBundle
 
+import json
+import logging
+import config
+import sdk_cmd
 
+log = logging.getLogger(__name__)
 class BaseTechBundle(ServiceBundle):
+
+    def __init__(self, package_name, service_name, scheduler_tasks, service, output_directory):
+        self.package_name = package_name
+        self.service_name = service_name
+        self.scheduler_tasks = scheduler_tasks
+        self.service = service
+        self.framework_id = service.get("id")
+        self.output_directory = output_directory
+        self.install_cli()
+
+    @config.retry
+    def install_cli(self):
+        sdk_cmd.run_cli(
+            "package install {} --cli --yes".format(self.package_name),
+            print_output=False,
+            check=True,
+        )
+
+    @config.retry
+    def create_configuration_file(self):
+        rc, stdout, stderr = sdk_cmd.svc_cli(
+            self.package_name, self.service_name, "describe", print_output=False
+        )
+
+        if rc != 0 or stderr:
+            log.error(
+                "Could not get service configuration\nstdout: '%s'\nstderr: '%s'", stdout, stderr
+            )
+        else:
+            self.write_file("service_configuration.json", stdout)
+
+    @config.retry
+    def create_pod_status_file(self):
+        rc, stdout, stderr = sdk_cmd.svc_cli(
+            self.package_name, self.service_name, "pod status --json", print_output=False
+        )
+
+        if rc != 0 or stderr:
+            log.error("Could not get pod status\nstdout: '%s'\nstderr: '%s'", stdout, stderr)
+        else:
+            self.write_file("service_pod_status.json", stdout)
+
+    @config.retry
+    def create_plan_status_file(self, plan):
+        rc, stdout, stderr = sdk_cmd.svc_cli(
+            self.package_name,
+            self.service_name,
+            "plan status {} --json".format(plan),
+            print_output=False,
+        )
+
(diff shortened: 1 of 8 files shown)


tools/diagnostics/base_tech_bundle/__init__.py
```diff
@@ -1,18 +1,98 @@
...
+        else:
+            self.write_file("service_configuration.json", stdout)
+
+    @config.retry
+    def create_pod_status_file(self):
+        rc, stdout, stderr = sdk_cmd.svc_cli(
+            self.package_name, self.service_name, "pod status --json", print_output=False
+        )
+
+        if rc != 0 or stderr:
```
takirala: i _think_ there were some cases in past where we can have `rc==0` && `stderr` being non empty. e.g.: CLI prints (or at least used to) prints the warnings (such as version X of CLI is not compatible with version Y of dc/os cluster) to stderr.
May be we can print the `stderr` and write `stdout` to json file if `stdout` is non empty ?!

tools/diagnostics/base_tech_bundle/__init__.py
```diff
@@ -1,18 +1,98 @@
...
+        sdk_cmd.run_cli(
+            "package install {} --cli --yes".format(self.package_name),
+            print_output=False,
+            check=True,
+        )
+
+    @config.retry
+    def create_configuration_file(self):
+        rc, stdout, stderr = sdk_cmd.svc_cli(
+            self.package_name, self.service_name, "describe", print_output=False
```
takirala: (nit) This would be a subset of `dcos marathon app show ....`, is the app definition already captured somewhere (or may be do you think it would be useful to capture the app definition) ?

tools/diagnostics/base_tech_bundle/kubernetes_bundle.py
```diff
@@ -0,0 +1,126 @@
...
+            self.package_name, self.service_name, "cluster debug pod status --json", print_output=False
+        )
+
+        if rc != 0 or stderr:
+            logging.error("Could not get cluster pod status\nstdout: '%s'\nstderr: '%s'", stdout, stderr)
+        else:
+            self.write_file("cluster_debug_pod_status.json", stdout)
+
+    @config.retry
+    def task_exec(self, task_id, cmd):
```
takirala: (nit) we don't have to override this here right ?!

tools/diagnostics/full_bundle.py
```diff
@@ -168,26 +168,19 @@ def create(self) -> (int, "FullBundle"):
             self.output_directory,
         ).create()
 
-        if base_tech.is_package_supported(self.package_name):
-            BaseTechBundle = base_tech.get_bundle_class(self.package_name)
+        log.info("Completed creating DC/OS and service-level diagnostics.")
```
takirala: > Completed creating DC/OS

we are creating a dc/os bundle as well?

tools/diagnostics/base_tech_bundle/__init__.py
takirala: OR we can "try" loading stdout as json and if it succeeds, we can write it to the file.

mpereira:
Ping @kaiwalyajoshi re: the above.

tools/diagnostics/full_bundle.py
```diff
@@ -168,26 +168,19 @@ def create(self) -> (int, "FullBundle"):
...
+        BaseTechBundle(
+            self.package_name,
+            self.service_name,
+            scheduler_tasks,
+            active_service,
+            self.output_directory,
+        ).create()
 
-        log.info("\nCreated %s", os.path.abspath(self.output_directory))
+        log.info("\nCreated base-tech bundle at %s", os.path.abspath(self.output_directory))
```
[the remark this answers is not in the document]
kaiwalyajoshi: @mpereira sure, I've merged the changes above but haven't made a release yet for Tools. Was waiting for input from more folks. Happy to sync up on the fixes.

PR #2919, opened 2019-02-04
[DIAGNOSTICS][DCOS-47783]: Initial support for Edge-LB Diagnostic Bundle

kaiwalyajoshi:
In this PR we add initial support for Edge-LB Pools with the following edgelb related commands:
- `dcos edgelb version`
- `dcos edgelb ping`
- `dcos edgelb list --json`

Additionally for each pool in the list above, the following edgelb commands are issued:
- `dcos edgelb status <pool-name> --json`
- `dcos edgelb endpoints <pool-name> --json`
- `dcos edgelb lb-config <pool-name>`
- `dcos edgelb template show <pool-name>`

Executor logs for *only the specified edge-lb pool* are collected ie:
`./tools/diagnostics/create_service_diagnostics_bundle.sh --package-name=edgelb-pool --service-name=dcos-edgelb/pools/test-http-pool --yes`
will collect the executor logs only for `dcos-edgelb/pools/test-http-pool`. Other pools executor logs will not be present in this collection.

Executor logs for other pools are collected via additional invocations with the pool supplied via `--service-name`.

### DIFF
--- a/tools/diagnostics/base_tech_bundle/__init__.py
+++ b/tools/diagnostics/base_tech_bundle/__init__.py
@@ -118,6 +118,7 @@ from .elastic_bundle import ElasticBundle  # noqa: E402
 from .hdfs_bundle import HdfsBundle  # noqa: E402
 from .kafka_bundle import KafkaBundle  # noqa: E402
 from .kubernetes_bundle import KubernetesBundle   # noqa: E402
+from .edgelb_bundle import EdgeLBBundle   # noqa: E402
 
 
 BASE_TECH_BUNDLE = {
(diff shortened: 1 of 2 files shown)


kaiwalyajoshi:
@takirala thanks for the review
1. If the commands don't work, the CLI errors out but it doesn't stop the entire bundle from being collected.
2. Yes, will add more documentation to the README for bundle specific commands. I wanted to get an initial version out for the teams to try out as the details might change depending on their feedback.

mainred:
@kaiwalyajoshi  @takirala ,  Thanks so much for adding Edge-LB diagnose bundle.
Edge-LB is not a framework, so we cannot get a service related to edge-lb, and this PR may finally not work as expected? correct me if I am wrong. Or it's just as you said, it is initial support, so we may need further work to make it work?

kaiwalyajoshi:
@mainred 
This PR adds initial support for EdgeLB, I was hoping the EdgeLB team would take this initial work and build ontop of it such that we have useful data to diagnose failures.

Feel free to issue another PR to correct what was done here for EdgeLB or to add additional diagnostics.

PR #2923, opened 2019-02-07
[DCOS-47829] Use dcos node decommission only in DC/OS 1.11 and above

takirala:
According to [DCOS-18981](https://jira.mesosphere.com/browse/DCOS-18981) and [1.11 Release Notes](https://docs.mesosphere.com/1.11/release-notes/1.11.0-rc1/#platform) decommission is relatively new and thus we should run tests 1.11 and above only.

### DIFF
--- a/frameworks/helloworld/tests/test_zzzrecovery.py
+++ b/frameworks/helloworld/tests/test_zzzrecovery.py
@@ -347,6 +347,7 @@ def test_config_update_while_partitioned():
 # WARNING: KEEP THIS TEST AT THE END
 # @@@@@@@
 @pytest.mark.sanity
+@pytest.mark.dcos_min_version("1.11")
 def test_auto_replace_on_decommission():
     candidate_tasks = sdk_tasks.get_tasks_avoiding_scheduler(
         config.SERVICE_NAME, re.compile("^(hello|world)-[0-9]+-server$")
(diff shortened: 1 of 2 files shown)


kvish:
this lgtm! thanks for fixing my test @takirala 

As a side thought, since we are only running the decommission test in versions greater 1.11 and above perhaps we want to enable the partition aware cap for only 1.11 and above to be consistent:
https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/dcos/Capabilities.java#L102

however, a user in versions less than 1.11 can still use the mesos MARK_AGENT_GONE operator call so maybe we should continue to support auto replacement in 1.10? no strong thoughts but let me know what you guys think.

takirala:
> however, a user in versions less than 1.11 can still use the mesos MARK_AGENT_GONE operator call so maybe we should continue to support auto replacement in 1.10

I don't think this is possible. https://issues.apache.org/jira/browse/MESOS-7443 is mesos 1.5.0 (dc/os 1.11) and above only. we can keep the code as-is. we have support for a case that would never happen, i guess that's fine. also, i think it's better to not _decrease_(in a way) the capability of SDK in a newer version.

kvish:
sure makes sense to me, approved

takirala:
@kaiwalyajoshi @kvish merging this change to get this tested in today's nightlies. but please review and i can address any comments in a later PR.

PR #2927, opened 2019-02-15
[DCOS-48462][Cassandra] Honor config JMX_PORT

samvantran:
## What changes were proposed in this PR? 

Resolves [DCOS-48462](https://jira.mesosphere.com/browse/DCOS-48462)

Apache Cassandra 3.11+ adds in new logic and classes for parsing the `JMX_PORT` from `conf/cassandra-env.sh` which is hardcoded to `7199`. This commit replaces the value in the file with the user-specified framework config.

Added an integration tests for checking logs do output the correct jmx port

### DIFF
--- a/frameworks/cassandra/tests/test_sanity.py
+++ b/frameworks/cassandra/tests/test_sanity.py
@@ -1,12 +1,15 @@
 import pytest
 import logging
 
+import sdk_cmd
 import sdk_hosts
 import sdk_install
 import sdk_jobs
 import sdk_metrics
 import sdk_networks
 import sdk_plan
+import sdk_service
+import sdk_tasks
 import sdk_upgrade
 
 from tests import config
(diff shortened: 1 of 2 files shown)


frameworks/cassandra/tests/test_sanity.py
```diff
@@ -96,3 +99,25 @@ def expected_metrics_exist(emitted_metrics):
...
+        new_config,
+        config.DEFAULT_TASK_COUNT,
+    )
+
+    sdk_plan.wait_for_completed_deployment(config.get_foldered_service_name())
+
+    tasks = sdk_tasks.get_service_tasks(config.get_foldered_service_name(), "node")
+
+    for task in tasks:
+        _, stdout, _ = sdk_cmd.run_cli("task log --lines=1000 {}".format(task.id))
```
akirillov: It would be better to check if this port is actually used by the process instead of relying on the log output.

samvantran: good call. Made the change in 515e77a

PR #2928, opened 2019-02-16
[DCOS-48617] Dont evaluate existing reservations against new portSpecs

kvish:


kvish:
merging this PR, the elastic tests have been experiencing flakiness and the same tests have failed against master and other PRs, merging this since it is a blocker ticket and will follow up with a ticket to fix elastic CI

PR #2932, opened 2019-02-19
[DCOS-47352][COPS-4389] Don't escape characters for Elastic custom YAML.

mpereira:
As per the manual/spec[1] on which the Java library we use bases its
implementation, one needs to use triple braces for variables that
should not have escaped characters when expanded. Without this, for
example, if one has quotes in the YAML content, they will get expanded
to their HTML entity: `&#34;`.

[1]: http://mustache.github.io/mustache.5.html

### DIFF
--- a/frameworks/elastic/src/main/dist/elasticsearch.yml
+++ b/frameworks/elastic/src/main/dist/elasticsearch.yml
@@ -283,4 +283,4 @@ cluster.routing.allocation.balance.threshold: {{CLUSTER_ROUTING_ALLOCATION_BALAN
 indices.requests.cache.size: {{INDICES_REQUESTS_CACHE_SIZE}}
 {{/INDICES_REQUESTS_CACHE_SIZE}}
 
-{{CUSTOM_YAML_BLOCK}}
+{{{CUSTOM_YAML_BLOCK}}}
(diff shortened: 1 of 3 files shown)


mpereira:
@samvantran they look like flakes. I'm re-triggering the build to see if they pass, I think the Elastic CI has been a bit spotty the last few weeks.

mpereira:
@samvantran looks like it might be a bug in our test tooling.

The first failure is a test that:
1. Updates the service configuration to include a plugin
2. Makes sure all tasks are restarted (due to service configuration change) and base-tech settings reflect service configuration (plugin present)
3. Updates the service configuration to include *no* plugins
4. Makes sure all tasks are restarted (due to service configuration change) and base-tech settings reflect service configuration (no plugins)

Looking at the logs, it looks like steps 1, 2, and 3 are successful, but 4 fails because apparently the test tooling (or some flake inducing edge condition) failed to identify that the service configuration changed:
```
[2019-02-19 20:45:10,376|sdk_upgrade-_wait_for_deployment(225)|INFO]: No config change detected. Tasks should not be restarted
```
Resulting in it trying to verify that tasks were _not_ restarted, which is the opposite of what it should be doing.

So I'm inclined to create a specific JIRA to investigate this flake, and merge this PR in the meantime.

mpereira:
@samvantran created https://jira.mesosphere.com/browse/DCOS-48772

Merging this for now and will look into the flakes asap.

mpereira:
Opened #2934.

PR #2934, opened 2019-02-20
[DCOS-48772] Wait for scheduler Marathon task to finish deployments.

mpereira:
This is required because the following line in this function gets the
scheduler configuration state based on a previous configuration
update. We need to make sure the scheduler finishes its deployment
first to get the new configuration state.

### DIFF
--- a/testing/sdk_upgrade.py
+++ b/testing/sdk_upgrade.py
@@ -219,6 +219,10 @@ def _update_service_with_cli(
 
 
 def _wait_for_deployment(package_name, service_name, initial_config, task_ids, timeout_seconds):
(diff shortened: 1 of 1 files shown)


mpereira:
This PR should fix this scenario detailed in another PR: https://github.com/mesosphere/dcos-commons/pull/2932#issuecomment-465520927

mpereira:
@mpereira it looks like this PR indeed fixed the flakes related to service updates sometimes waiting for redeployed tasks incorrectly. The HDFS build failed with timeout failures waiting for all service tasks to come up, which is unrelated to the former issue. I'm gonna merge this one for now, as it should help with the CI health.

PR #2935, opened 2019-02-20
[DCOS-48617] Allow reorder of existing ports

kvish:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/TaskResourceMapper.java
```diff
@@ -162,15 +162,14 @@
         if (!priorPort.isPresent()) {
           //this is a new portSpec and will never match a previously reserved taskResource
           continue;
+        } else if (RangeUtils.isInAny(ranges.getRangeList(), priorPort.get())) {
```
takirala: Do you think it would be useful to add a test case with a yaml file in helloworld covering this scenario and the one you fixed in earlier in #2928 ?

PR #2936, opened 2019-02-21
[DCOS-48059] Handle duplicate StoreTaskInfoRecommendations generated during step evaluation

takirala:


sdk/scheduler/src/main/java/com/mesosphere/sdk/state/PersistentLaunchRecorder.java
```diff
@@ -69,18 +72,44 @@ public PersistentLaunchRecorder(
...
+     *   - Step B would generate Launch+Store for task Y and only Store for task X.
+     *     During this step B assumes X is not launched right away and thus uses "" (empty string) for its task id.
+     *  We do need to do adhere to above behavior because of DCOS-42539.
+     *
+     * Because of above constraint, we do the following:
+     * 1. Filter given {@link OfferRecommendation} by only looking for {@link StoreTaskInfoRecommendation}
+     * 2. Persist all {@link StoreTaskInfoRecommendation}s that have empty task ID.
+     * 3. After above, persist the rest of {@link StoreTaskInfoRecommendation} that have non empty task ID.
+     */
+    Iterator<StoreTaskInfoRecommendation> storeTaskRecommendations = offerRecommendations
```
mainred: Sorry, Tarun, I want to fully understand your implementation and did not get your change.
You replaced with for loop with an iterator, and everything else seems to be unchanged in essence?

mainred: And the task id will be overwritten(?) to "" only when these two steps run simultaneously, right?

takirala: @mainred https://github.com/mesosphere/dcos-commons/pull/2936/files#diff-efe7be5120dc7a98f90cbf8beb3bb703R102 is the only thing i have changed. 

Let me clarify this with the example structure at https://github.com/mesosphere/dcos-edge-lb/blob/0502fad32dd5ddb7c6f997aa7d2f469d580ce640/framework/edgelb-pool/src/main/dist/svc.yml#L68-L85. Assume `deploy` and `reload` are running. 

Previously:
Deploy plan would generate following recommendations : 
 - 1 recommendation for launching `server` task
 - 1 recommendation for storing the `server` task in zookeeper
 - 1 recommendation for storing the `sidercar` task in zookeeper
reload plan would generate following recommendations : 
 - 1 recommendation for launching `sidecar` task
 - 1 recommendation for storing the `sidecar` task in zookeeper
 - 1 recommendation for storing the `server` task in zookeeper

Note that deploy plan above generates a `store` recommendation for `sidecar` task and it will have empty taskId. This collides with the `store` recommendation generated by `reload` plan that would have a non-empty taskId. The reason for generating these duplicate recommendations are discussed in DCOS-42539. While it would have been more ideal to fix this duplicate recommendation generation, it requires a re-architecture of how plans are evaluated.

What I have done is to order these recommendations such that the empty taskIds are written first followed by non-empty taskIds.

mainred: ❤️ , Got it. So that the empty taskid can always be overwritten by it's non-empy taskid if it has one.

mainred:
@alenkacz @kvish @kaiwalyajoshi , could you PTAL. The CI is green.

takirala:
@kaiwalyajoshi @kvish i am merging this to run this through some nightlies before we decide to cut a release, please review and i can address any comments in a later PR.

PR #2937, opened 2019-02-21
[COPS-4469] Shutdown the scheduler when service user changes in a ServiceSpec update

kaiwalyajoshi:
Previously the SDK would detect a change of `user` in the `ServiceSpec`, emit a warning and *not* propagate and store these updates into ZK. The scheduler would however continue to keep running with the updated user and this can cause issues with downstream components such as volumes which may have permissions set for the previous user.

This change explictly disallows changing of users on a previously deployed service. The scheduler issues a *shutdown* with a prescriptive message of how to resolve this issue.

Operators should note the scheduler crash-looping with the following message in the scheduler logs when a service user change is applied:

    'INVALID CONFIGURATION UPDATE. Cannot change user of deployed service from 'userA' to 'userB'.
    Revert to previous user 'userA' to proceed. Current set of configuration updates will NOT be applied.'

The violating updates to the `ServiceSpec` are again *not* propagated to ZK maintaining existing semantics.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ConfigValidationError.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ConfigValidationError.java
@@ -16,16 +16,20 @@ public final class ConfigValidationError {
 
   private final String message;
 
+  private final boolean isFatal;
+
   private ConfigValidationError(
       String configField,
       String oldConfigValue,
       String newConfigValue,
-      String message)
+      String message,
+      boolean isFatal)
   {
     this.configField = configField;
     this.oldConfigValue = oldConfigValue;
     this.newConfigValue = newConfigValue;
     this.message = message;
+    this.isFatal = isFatal;
   }
 
   /**
@@ -36,7 +40,18 @@ public final class ConfigValidationError {
       String configField, String configValue, String message)
   {
     // Set oldValue to null
-    return new ConfigValidationError(configField, null, configValue, message);
+    return new ConfigValidationError(configField, null, configValue, message, false);
+  }
+
+  /**
+   * Returns a new validation error which indicates that a configuration field has an invalid
+   * value. This is equivalent to a transition error, except with no prior value.
+   */
+  public static ConfigValidationError valueError(
+      String configField, String configValue, String message, boolean isFatal)
+  {
+    // Set oldValue to null
+    return new ConfigValidationError(configField, null, configValue, message, isFatal);
   }
 
   /**
@@ -46,7 +61,18 @@ public final class ConfigValidationError {
   public static ConfigValidationError transitionError(
       String configField, String oldConfigValue, String newConfigValue, String message)
   {
-    return new ConfigValidationError(configField, oldConfigValue, newConfigValue, message);
+    return new ConfigValidationError(configField, oldConfigValue, newConfigValue, message, false);
+  }
+
+  /**
+   * Returns a new validation error which indicates that a configuration field has an invalid
+   * transition from its previous value to the current value.
+   */
+  public static ConfigValidationError transitionError(
+      String configField, String oldConfigValue, String newConfigValue, String message,
+      boolean isFatal)
+  {
+    return new ConfigValidationError(configField, oldConfigValue, newConfigValue, message, isFatal);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


sdk/scheduler/src/test/java/com/mesosphere/sdk/config/ConfigurationUpdaterTest.java
```diff
@@ -261,12 +266,21 @@ public void testUserSetAtServiceLevelButNotPodLevel() throws ConfigStoreExceptio
...
         // Note: the new service spec sets pod users with a non-root user
-        ConfigurationUpdater.UpdateResult result =
-                configurationUpdater.updateConfiguration(ORIGINAL_SERVICE_SPECIFICATION_WITH_USER);
-        Assert.assertEquals(TARGET_ID, result.getTargetId());
-        // since the 2 pods don't set the user their user defaults to "root" which conflicts with the user set as noted above
-        Assert.assertEquals(2, result.getErrors().size());
+        boolean caughtError = false;
+        boolean correctErrorMessage = false;
+        ConfigurationUpdater.UpdateResult result = null;
+        try {
```
takirala: (nit) i like the style we use here https://github.com/mesosphere/dcos-commons/blob/60e9caf214eed5bf0c61bea26afe05744ed93ae4/sdk/scheduler/src/test/java/com/mesosphere/sdk/curator/CuratorUtilsTest.java#L115-L118
use a fail() at the end of try and assert inside the catch, feels more readable. same pattern in other two places in this file.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/UserCannotChange.java
```diff
@@ -20,8 +20,22 @@
  * change as "nobody" will not have access to "root"-owned files in persistent volumes.
  */
 public class UserCannotChange implements ConfigValidator<ServiceSpec> {
```
takirala: iiuc, previously we were validating the user change and throwing an error? (trying to understand what was the point of the error if we are using the new value for user anyway).

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ConfigValidationError.java
```diff
@@ -36,7 +40,18 @@ public static ConfigValidationError valueError(
       String configField, String configValue, String message)
   {
     // Set oldValue to null
-    return new ConfigValidationError(configField, null, configValue, message);
+    return new ConfigValidationError(configField, null, configValue, message, false);
```
takirala:
```suggestion
    return valueError(configField, configValue, message, false);
```
and also remove the `// Set oldValue to null` comment

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ConfigValidationError.java
```diff
@@ -46,7 +61,18 @@ public static ConfigValidationError valueError(
   public static ConfigValidationError transitionError(
       String configField, String oldConfigValue, String newConfigValue, String message)
   {
-    return new ConfigValidationError(configField, oldConfigValue, newConfigValue, message);
+    return new ConfigValidationError(configField, oldConfigValue, newConfigValue, message, false);
```
takirala:
```suggestion
    return transitionError(configField, oldConfigValue, newConfigValue, message, false);
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/UserCannotChange.java
kaiwalyajoshi: We threw and error (which gets emitted as a log.warn) but the service continued on with the new values that *weren't* persisted to ZK. This might have been an edge case that was missed up till now.

kaiwalyajoshi:
> Did you try verify fix (as in trying to update a running scheduler with new user and then scheduler crashes and then the user can re-update with correct user at which point scheduler should be healthy again) ?

@takirala 
Yes this has been verified, see [this JIRA comment](https://jira.mesosphere.com/browse/COPS-4469?focusedCommentId=220102&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-220102).

sdk/scheduler/src/test/java/com/mesosphere/sdk/config/ConfigurationUpdaterTest.java
```diff
@@ -304,9 +314,37 @@ public void testPodsUsersUnsetInPreviousServiceSpecButSetToRootInNewServiceSpec(
...
+                mockStateStore,
+                mockConfigStore,
+                DefaultServiceSpec.getComparatorInstance(),
+                DefaultConfigValidators.getValidators(SchedulerConfigTestUtils.getTestSchedulerConfig()),
+                Optional.empty());
+        when(mockConfigStore.getTargetConfig()).thenReturn(TARGET_ID);
+
+        when(mockConfigStore.fetch(TARGET_ID)).thenReturn(ORIGINAL_SERVICE_SPECIFICATION_WITH_USER);
+
+        //ConfigurationUpdater.UpdateResult result = null;
```
takirala: this unused code can be removed ?! (Same in other places)

kaiwalyajoshi: whoops! fixed now, thanks.

PR #2942, opened 2019-02-26
Change long decline to 1 hour from 14 days

alenkacz:


kvish:
nit can you update the comments and docs to reflect this change:

https://github.com/mesosphere/dcos-commons/blob/master/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/WorkSetTracker.java#L41

https://github.com/mesosphere/dcos-commons/blob/master/docs/pages/developer-guide.md#default (line 1089)

alenkacz:
@kvish nice catch!

PR #2943, opened 2019-02-26
[DCOS-49103][SDK Utils] Nightly CI fixes, read SECURITY env var as fallback.

kaiwalyajoshi:
Read environment variable if call to GET dcos-metadata/bootstrap-config.json fails.

### DIFF
--- a/testing/sdk_utils.py
+++ b/testing/sdk_utils.py
@@ -139,9 +139,19 @@ def is_strict_mode():
 
 
 def get_security_mode() -> DCOS_SECURITY:
-    r = get_metadata().json()
-    mode = r['security']
-    return DCOS_SECURITY[mode]
+    try:
+        r = get_metadata().json()
+        mode = r['security']
+        return DCOS_SECURITY[mode]
+    except Exception:
+        # Read environment variable only if get_metadata call fails
+        # Defaults to disabled.
+        mode = os.environ.get("SECURITY", "disabled")
+        return {
+            'disabled': DCOS_SECURITY.disabled,
+            'permissive': DCOS_SECURITY.permissive,
+            'strict': DCOS_SECURITY.strict
+        }[mode.lower]
 
 
 def get_metadata():


testing/sdk_utils.py
```diff
@@ -139,9 +139,19 @@ def is_strict_mode():
...
 def get_security_mode() -> DCOS_SECURITY:
-    r = get_metadata().json()
-    mode = r['security']
-    return DCOS_SECURITY[mode]
+    try:
+        r = get_metadata().json()
+        mode = r['security']
+        return DCOS_SECURITY[mode]
+    except Exception:
+        # Read environment variable only if get_metadata call fails
```
takirala: why would it fail? Is it based on DC/OS version? If so, we can do an if-else like
```
if version < 1.11:
  // read env
else:
  // make http call
```
wdyt ?

kaiwalyajoshi: This is actually the core question, I'm not sure why retrieving `<cluster>/dcos-metadata/bootstrap-config.json` failed in the batch of these [Nightly CI Runs](https://teamcity.mesosphere.io/viewLog.html?buildId=1632123&tab=buildResultsDiv&buildTypeId=Frameworks_Orchestration_DcosCommons_Nightly_HelloWorld_Open_DcOs111Open). Is bootstrap-config.json something special that's not always available? What's the canonical way of determining a cluster security mode?

My goal was to "ask the cluster" what its security settings were, and if that didn't work rely on a env-var set somewhere, hence the current call structure.

takirala: Looks like that endpoint is EE only?
Found it in ee repo : https://github.com/mesosphere/dcos-enterprise/blob/master/packages/adminrouter/extra/src/docs/api/nginx.master.yaml#L317
but OSS doesn't have this endpoint : https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/docs/api/nginx.agent.yaml#L29

kaiwalyajoshi:
Merging this PR to resolve nightly CI failures on DC/OS Open.

PR #2945, opened 2019-03-01
[DCOS-49318] Move declaration to fix UnboundLocalError

porridge:


porridge:
FWIW, I checked manually that the cassandra/elastic/hdfs tests did pass. There is an unrelated problem with propagating their result to GH.

PR #2946, opened 2019-03-01
[DCOS-49331] Write data to zk only when needed

takirala:
When SDK is writing data to zk, in some cases it is creating a `KeeperException.NodeExistsException` and this may lead to large exhibitor log files in unstable (agent disconnected etc..) clusters with multiple SDK schedulers running. This PR addresses this by problem checking for the existing zk node data before creating/updating it.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
@@ -13,7 +13,6 @@ import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.api.ACLProvider;
 import org.apache.curator.framework.api.transaction.CuratorOp;
-import org.apache.curator.framework.api.transaction.CuratorTransactionFinal;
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/curator/CuratorPersister.java
```diff
@@ -182,18 +179,21 @@ private static String getInfo(byte[] bytes) {
...
-    LOGGER.debug("Setting {} => {}", path, getInfo(bytes));
+    LOGGER.debug("Setting {} => {}", path, getInfo(newData));
     try {
-      try {
-        client.create().creatingParentsIfNeeded().forPath(path, bytes);
-      } catch (KeeperException.NodeExistsException e) {
-        client.setData().forPath(path, bytes);
+      if (client.checkExists().forPath(path) != null) {
+        byte[] oldData = client.getData().forPath(path);
+        if (oldData != newData) {
```
kvish: maybe !Array.equals instead?

takirala: 🤦‍♂️ of course, thanks!

PR #2950, opened 2019-03-04
[TOOLS] Upgrade dcos-launch

takirala:
Tools infra recommended to upgrade dcos-launch and use us-west-2 to get rid of the recent spike in ROLLBACK_IN_PROGRESS errors.


frozen_requirements.txt
```diff
@@ -14,110 +14,107 @@
...
 oauth2client==3.0.0
 oauthlib==2.1.0
 pager==3.3
 paramiko==2.4.2
+pathlib2==2.3.3
 pefile==2018.8.8
-pkginfo==1.4.2
-pluggy==0.8.0
-pre-commit==1.12.0
+pkginfo==1.2.1
```
samvantran: you meant to downgrade this?

takirala: yeah it's the result of the `pip freeze` as documented at https://github.com/mesosphere/dcos-commons/blob/master/frozen_requirements.txt#L3-L13

takirala:
@samvantran i meant to say us-west-2, this is the info i got from tools infra : 

```
us-east-1 was a temporary measure while we were putting in a fix for dcos-launch
us-west-2 with m5 seems like a good idea, but you might still hit rollback in progress
us-west-2 with up to date dcos-launch should have the lowest incidence of rollback in progress
```

PR #2952, opened 2019-03-05
Remove tests which are not run on CI

adamtheturtle:
This removes test files which are not run on CI.
Passing CI on https://github.com/mesosphere/dcos-commons/pull/2949 (a PR which adds `assert False`) shows that at least one of these test files is not run on CI.

This will allow us to move on with adding mypy checks to this repository - https://jira.mesosphere.com/browse/DCOS-48816.

See https://jira.mesosphere.com/browse/DCOS-49574 for re-adding these tests.


adamtheturtle:
In the interest of speed I am merging this without all CI checks having run. I feel confident in this because this is in the `tools` directory and we are confident that these tests do not run.

PR #2953, opened 2019-03-05
Remove unused raw_data parameter from Package

adamtheturtle:
While adding mypy checks to this repository - https://jira.mesosphere.com/browse/DCOS-48816, we noticed that this parameter / attribute is unused.

That is - nothing is ever supplied to the parameter, and the private attribute is never used.


adamtheturtle:
In the interest of speed I am merging this without all CI checks having run. I feel confident in this because this is in the `tools` directory and my searching shows that this parameter is unused.

PR #2954, opened 2019-03-05
Use the `Version` type when creating a Package

adamtheturtle:
While adding mypy checks to this repository ([DCOS-48816](https://jira.mesosphere.com/browse/DCOS-48816)), we found that there was a type error.

In particular, `universe.Package`'s "version" parameter is expected in the class to be a `Version`.

However, multiple callers of the class were giving `version` as a `str`.

A `Version` includes a `release_version` (`int`) and a `package_version` (`str`).
In this PR, we choose for all existing callers which ignored `release_version` to set that to 0.
I am happy to take other suggestions!
From what I can tell, it just so happens that none of these callers ever call use any functionality of `Package` which rely on the `version` being set correctly.

Tests for `tools` are not touched because they are removed in https://github.com/mesosphere/dcos-commons/pull/2952.

### DIFF
--- a/tools/publish_aws.py
+++ b/tools/publish_aws.py
@@ -18,6 +18,7 @@ import sys
 import time
 
 import universe
+from universe.package import Version
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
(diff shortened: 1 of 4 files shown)


adamtheturtle:
I don't really understand what we are waiting for here in terms of CI.
I guess there is some queue.
However, I am willing to take that risk.

PR #2964, opened 2019-03-08
[DCOS-49667] Fix test_rack to test against the region it was deployed in

farhan5900:
Instead of checking against single region "us-west", added checks against all the valid regions.

### DIFF
--- a/frameworks/cassandra/tests/test_racks.py
+++ b/frameworks/cassandra/tests/test_racks.py
@@ -43,7 +43,7 @@ def test_rack():
     log.info("node: {}".format(node))
 
     assert node.get_rack() != "rack1"
-    assert sdk_utils.get_cluster_region() in node.get_rack()
+    assert sdk_utils.get_cluster_zones()[node.get_address()] == node.get_rack()
 
 
 @pytest.mark.sanity
--- a/frameworks/cassandra/tests/test_racks.py
+++ b/frameworks/cassandra/tests/test_racks.py
@@ -43,7 +43,7 @@ def test_rack():
     log.info("node: {}".format(node))
 
     assert node.get_rack() != "rack1"
-    assert "us-west" in node.get_rack()
+    assert any(rack in node.get_rack() for rack in config.AWS_REGIONS)
 
 
 @pytest.mark.sanity
--- a/frameworks/cassandra/tests/test_racks.py
+++ b/frameworks/cassandra/tests/test_racks.py
@@ -43,7 +43,7 @@ def test_rack():
     log.info("node: {}".format(node))
 
     assert node.get_rack() != "rack1"
-    assert any(rack in node.get_rack() for rack in config.AWS_REGIONS)
+    assert sdk_utils.get_cluster_region() in node.get_rack()
 
 
 @pytest.mark.sanity
--- a/testing/sdk_utils.py
+++ b/testing/sdk_utils.py
@@ -161,7 +161,7 @@ def get_metadata():
                                    retry=False)
 
 
-def get_cluster_region():
+def get_cluster_zones():
     rc, stdout, _ = sdk_cmd.run_cli("node --json")
 
     if rc != 0:
@@ -169,8 +169,12 @@ def get_cluster_region():
         return None
 
     r = json.loads(stdout)
+    ip_zn = {}
 
-    return r[1]['domain']['fault_domain']['region']['name']
+    for ag in r[1:]:
+        ip_zn[ag['hostname']] = ag['domain']['fault_domain']['zone']['name']
+
+    return ip_zn
 
 
 """Annotation which may be used to mark test suites or test cases as EE-only.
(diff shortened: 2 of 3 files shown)


frameworks/cassandra/tests/test_racks.py
```diff
@@ -43,7 +43,7 @@ def test_rack():
     log.info("node: {}".format(node))
 
     assert node.get_rack() != "rack1"
-    assert "us-west" in node.get_rack()
+    assert any(rack in node.get_rack() for rack in config.AWS_REGIONS)
```
samvantran: This checks if the node is on a rack in _any_ region, not specifically if the node is on a rack in the _same_ region as the cluster. I was thinking more along the lines of determining what region the cluster is in eg `us-east-1` and then verifying that the rack is indeed in `us-east` as well.

Look through the python `sdk_` utils in `testing/` or `tools/` directories. There might be something that can help you with that. Or maybe create your own helper method to determine cluster region!

farhan5900: I have written `get_cluster_region` helper method in `sdk_utils` to get the region of the cluster.

testing/sdk_utils.py
```diff
@@ -165,6 +166,22 @@ def get_metadata() -> requests.Response:
                                    retry=False)
 
 
+def get_cluster_zones():
+    rc, stdout, _ = sdk_cmd.run_cli("node --json")
+
+    if rc != 0:
+        log.warning("return code for command 'node --json' is non-zero: %s", rc)
```
mpereira: `sdk_cmd.run_cli` ultimately already logs out something to that effect, so we can skip on doing this: https://github.com/mesosphere/dcos-commons/blob/4ef375d8c95272041d553c6549f7ed88c21889c1/testing/sdk_cmd.py#L216-L217

testing/sdk_utils.py
```diff
@@ -165,6 +166,22 @@ def get_metadata() -> requests.Response:
                                    retry=False)
 
 
+def get_cluster_zones():
```
mpereira: We should probably add some retrying to this function. Example: https://github.com/mesosphere/dcos-commons/blob/4ef375d8c95272041d553c6549f7ed88c21889c1/testing/sdk_cmd.py#L276-L278

testing/sdk_utils.py
```diff
@@ -165,6 +166,22 @@ def get_metadata() -> requests.Response:
...
+    rc, stdout, _ = sdk_cmd.run_cli("node --json")
+
+    if rc != 0:
+        log.warning("return code for command 'node --json' is non-zero: %s", rc)
+        return None
+
+    r = json.loads(stdout)
```
mpereira: Attempting to improve variable name to be more descriptive:
```suggestion
    nodes = json.loads(stdout)
```

testing/sdk_utils.py
```diff
@@ -165,6 +166,22 @@ def get_metadata() -> requests.Response:
...
+    ip_zn = {}
```
mpereira: Attempting to improve variable name to be more descriptive:
```suggestion
    zones_by_ip = {}
```
or
```suggestion
    ips_to_zone = {}
```

testing/sdk_utils.py
```diff
@@ -165,6 +166,22 @@ def get_metadata() -> requests.Response:
...
+
+    for ag in r[1:]:
```
mpereira: Attempting to improve variable name to be more descriptive:
```suggestion
    for agent in nodes[1:]:
```

testing/sdk_utils.py
farhan5900: Removed this check and logging both.

testing/sdk_utils.py
farhan5900: Added retrying.

testing/sdk_utils.py
farhan5900: Thanks for the suggestion. Changed the variable name as suggested.

testing/sdk_utils.py
```diff
@@ -166,20 +166,17 @@ def get_metadata() -> requests.Response:
                                    retry=False)
 
 
+@retrying.retry(stop_max_attempt_number=3, wait_fixed=1000, retry_on_result=lambda result: not result)
 def get_cluster_zones():
-    rc, stdout, _ = sdk_cmd.run_cli("node --json")
+    _, stdout, _ = sdk_cmd.run_cli("node --json")
 
-    if rc != 0:
```
mpereira: We should keep
```python
if rc != 0:
    return None
```
here so that the method retries when the `dcos node` command fails. Just not the warning.

testing/sdk_utils.py
mpereira: Removing the logging was good, but the check is valid. We should keep it.

testing/sdk_utils.py
mpereira: Great!

testing/sdk_utils.py
mpereira: 👍

mpereira:
@porridge let's merge? The two HW failures are due to TC issues it seems.

PR #2978, opened 2019-03-12
Add multiple custom plans to plan.yml

takirala:
In `test_parallel_plans.py` we use plans from sidecar.yml which has multiple pods with all tasks sharing a single resource-set, this sometimes leads to inconsistencies (offer starvation) when all the tasks are run in parallel and the test times out before all the offers are received back to make progress. This PR adds new plans to `plan.yml` that can be used more reliably to make sure that all the plans progress simultaneously without waiting for each other.


takirala:
Merging this before a green CI in order to get this in to the nightly CI. I have tested this fix locally and it should pass. We can always revert the fix in case this breaks sth.

PR #2979, opened 2019-03-13
Add strict mypy type hints to all Python files

adamtheturtle:
This change adds mypy checking to the Python files in the `dcos-commons` codebase.
This has found and will find bugs in the code - some of those bugs have been fixed in other PRs.
The main purpose of this change is to make SDK bumps for frameworks easier.

## CI

The pre-commit setup is really not suited to `mypy`.
There are `mypy` pre-commit repos but they do not work well - they fail when there are duplicate file names given to `mypy` directly (as we have e.g. `auth.py`).

My goal was to add CI with this branch.
However, as I am soon to leave the DS team I want to ship the progress here and push on with bumping Confluent Zookeeper.
I will maek sure that JIRA reflects this.

For now I have been running:

```
pip install mypy
mypy frameworks testing tools *.py
```

Morally, I do not see an issue with merging this without CI, because there were already a bunch of type hints in the codebase.

### Setup

* Add mypy checking to a framework (if in a different repository)
* Add type hints to a framework (all frameworks in this repository are type hinted by this PR)

### On SDK bump

* Run `mypy` against the framework to tell us if we are calling functions which no longer exist, or if which have changed signatures


adamtheturtle:
I see tests are failing and I will fix this!

frameworks/helloworld/tests/test_tls.py
```diff
@@ -292,28 +298,7 @@ def _export_cert_from_task_keystore(task_name, keystore_path, alias, password=KE
...
-    Returns:
-        A string that can be used as `dcos exec` argument.
-    """
-    return _java_command(
-        "keytool -list -keystore {keystore_path} "
-        "-noprompt {args}".format(keystore_path=keystore_path, args=args)
-    )
-
-
-def _keystore_export_command(keystore_path, cert_alias, args=None):
```
wavesoft: Why did  you remove this comment block? :thinking:

testing/sdk_auth.py
```diff
@@ -38,24 +38,27 @@
...
-        tasks = json.loads(raw_tasks)
-        for task in tasks:
-            if task["name"] == task_name:
-                return task
-
-    raise RuntimeError(
-        "Expecting marathon KDC task but no such task found. Running tasks: {tasks}".format(
-            tasks=raw_tasks
+    @retrying.retry(stop_max_attempt_number=3, wait_fixed=2000)
+    def _get_kdc_task(task_name: str) -> dict:
```
wavesoft: Can we change this name into something like `_get_kdc_task_retried` to avoid name collision?

testing/sdk_cmd.py
```diff
@@ -176,22 +184,26 @@ def svc_cli(
     return rc, stdout, stderr
 
 
-def _get_json_output(cmd, print_output=True, check=False):
-    _, stdout, stderr = run_cli(cmd, print_output, check=check)
+# def _get_json_output(cmd: str, print_output: bool = True, check: bool = False) -> Any:
```
wavesoft: Why not remove this instead of commenting-out?

testing/sdk_networks.py
```diff
@@ -23,7 +23,7 @@
 @retrying.retry(wait_fixed=1000, stop_max_delay=5 * 1000, retry_on_result=lambda res: not res)
 def _wait_for_endpoint_info(
     package_name: str, service_name: str, endpoint_name: str, json: bool
-) -> typing.Union[typing.Dict, str]:
+) -> Any:
```
wavesoft: Why so generic? Is there really nothing we could restrict this to? :thinking:

testing/sdk_networks.py
```diff
@@ -34,12 +34,14 @@ def _wait_for_endpoint_info(
...
+def get_endpoint_names(package_name: str, service_name: str) -> List[str]:
     """Returns a list of endpoint names for the specified service."""
-    return _wait_for_endpoint_info(package_name, service_name, None, json=True)
+    result = _wait_for_endpoint_info(package_name, service_name, None, json=True)
+    assert isinstance(result, list)
+    return result
 
 
-def get_endpoint(package_name: str, service_name: str, endpoint_name: str) -> typing.Dict:
+def get_endpoint(package_name: str, service_name: str, endpoint_name: str) -> Any:
```
wavesoft: Same question like before here, can we restrict this `Any` even further?

testing/sdk_tasks.py
```diff
@@ -96,24 +101,34 @@ def parse(task_entry, agentid_to_hostname):
...
+        executor_id: str,
+        framework_id: str,
+        agent_id: str,
+        resources: Dict[str, int],
+    ) -> None:
+        self.name: str = name
+        self.host: str = host
+        self.state: str = state  # 'TASK_RUNNING', 'TASK_KILLED', ...
+        self.is_completed: bool = state in COMPLETED_TASK_STATES
         self.id = task_id
```
wavesoft: Do we need to do type hinting also on these variables? Because you are already doing this with `self.name` and the others.

adamtheturtle:
> That's an amazing effort @adamtheturtle 

Oops! Should also have explicitly called out @mhrabovcin who worked on a bunch of this!

testing/sdk_cmd.py
adamtheturtle: Done

frameworks/cassandra/tests/nodetool.py
```diff
@@ -1,15 +1,17 @@
+from typing import List, Tuple
+
 import sdk_cmd
 
 
-def cmd(service_name, task_name, command):
+def cmd(service_name: str, task_name: str, command: str) -> Tuple[int, str, str]:
```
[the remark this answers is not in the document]
adamtheturtle: Why?

conftest.py
```diff
@@ -82,6 +86,7 @@ def pytest_runtest_makereport(item: pytest.Item, call):  # _pytest.runner.CallIn
     # Execute all other hooks to obtain the report object.
     outcome = yield
 
+    assert isinstance(outcome, pluggy.callers._Result)
```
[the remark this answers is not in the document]
adamtheturtle: Yes, else we get an error that `None has no attribute...`. This is a `pytest` quirk I believe.

testing/sdk_tasks.py
adamtheturtle: So far I'm only hinting what `mypy` says needs to be hinted.

frameworks/elastic/tests/config.py
```diff
@@ -76,61 +77,73 @@
...
     retry_on_result=lambda res: not res,
 )
-def check_kibana_adminrouter_integration(path):
+def check_kibana_adminrouter_integration(path: str) -> bool:
     curl_cmd = 'curl -L -I -k -H "Authorization: token={}" -s {}/{}'.format(
         sdk_utils.dcos_token(), sdk_utils.dcos_url().rstrip("/"), path.lstrip("/")
     )
     rc, stdout, _ = sdk_cmd.master_ssh(curl_cmd)
-    return rc == 0 and stdout and "HTTP/1.1 200" in stdout
+    return bool(rc == 0 and stdout and "HTTP/1.1 200" in stdout)
```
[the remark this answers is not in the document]
adamtheturtle: There are three conditions here - and the `stdout` one is a `str`:

```
~/D/mypy-experiment (e2e) ~> mypy example.py
example.py:2: error: Incompatible return value type (got "Union[bool, str]", expected "bool")
~/D/mypy-experiment (e2e) ~> cat example.py 
def example() -> bool:
    return 'a' == 'a' and 'c' and 'b' in 'o'

print(example())
```

frameworks/elastic/tests/config.py
```diff
@@ -143,44 +156,53 @@ def check_kibana_plugin_installed(plugin_name, service_name=SERVICE_NAME):
...
     result = _get_hosts_with_plugin(service_name, plugin_name)
     return result is not None and result == []
 
 
-def _get_hosts_with_plugin(service_name, plugin_name):
+def _get_hosts_with_plugin(service_name: str, plugin_name: str) -> List[str]:
     output = _curl_query(service_name, "GET", "_cat/plugins", return_json=False)
     if output is None:
-        return None
+        return []
```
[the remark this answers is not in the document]
adamtheturtle: I have reverted this change.

PR #2980, opened 2019-03-13
[SEE DESCRIPTION] Add mypy type hints to the 'testing' directory

adamtheturtle:
This is a subset of #2979.

This is split out for easier review.
Perhaps we will not merge this - and only merge  #2979.
Or perhaps we will merge this and then make #2979 smaller.


porridge:
`code-checks` failed - is this expected?

testing/sdk_auth.py
```diff
@@ -38,24 +38,27 @@
...
-    if raw_tasks:
-        tasks = json.loads(raw_tasks)
-        for task in tasks:
-            if task["name"] == task_name:
-                return task
-
-    raise RuntimeError(
-        "Expecting marathon KDC task but no such task found. Running tasks: {tasks}".format(
-            tasks=raw_tasks
+    @retrying.retry(stop_max_attempt_number=3, wait_fixed=2000)
```
porridge: Why move this, is this a mypy limitation?

testing/sdk_cmd.py
```diff
@@ -176,22 +184,26 @@ def svc_cli(
     return rc, stdout, stderr
 
 
-def _get_json_output(cmd, print_output=True, check=False):
-    _, stdout, stderr = run_cli(cmd, print_output, check=check)
+# def _get_json_output(cmd: str, print_output: bool = True, check: bool = False) -> Any:
```
porridge: Please remove this rather than comment out.

testing/sdk_upgrade.py
```diff
@@ -176,14 +179,25 @@ def update_or_upgrade_or_downgrade(
         )
     else:
         _update_service_with_cli(package_name, service_name, to_package_version, additional_options)
-    return not wait_for_deployment or _wait_for_deployment(
-        package_name, service_name, initial_config, task_ids, timeout_seconds
-    )
+
+    if not wait_for_deployment:
```
porridge: This seems backwards.

testing/security/transport_encryption.py
```diff
@@ -95,7 +102,7 @@ def fetch_dcos_ca_bundle_contents() -> str:
         log.error("Error fetching DC/OS CA bundle")
         raise Exception("Errot fetching DC/OS CA bundle")
 
-    return cert
+    return cert.decode('utf-8')
```
porridge: This seems to change the semantics of the function. Should we fix the type annotation instead?

porridge:
I'm also wondering, if there is anything in the checks that will remind/require people in the future that newly added code should also be annotated?

adamtheturtle:
@porridge As per the description on the other PR - there will be CI - that is something I'm working towards.

testing/sdk_cmd.py
adamtheturtle: Done - good spot.

porridge:
I understand there will be CI, but it's not obvious to me what it will ensure? Just the type-integrity of the codebase, or does it also require *everything* to be annotated?

adamtheturtle:
> code-checks failed - is this expected?

No and I do not yet understand why. On the other, bigger PR, this isn't an issue and I haven't yet reproduced this locally.

testing/sdk_auth.py
adamtheturtle: We want to say that `_get_kdc_task` returns a `Dict`.
However, in the original, `_get_kdc_task` is decorated with a third party decorator - `@retrying.retry`.
That decorator can arbitrarily change the

testing/sdk_upgrade.py
adamtheturtle: Changing now!

adamtheturtle:
@porridge To that end I added the `mypy.ini` to this branch, so that what I've been working with as a configuration (i.e. "require everything to be hinted) is clear.

testing/security/transport_encryption.py
adamtheturtle: Addressed - not sure what this was about!

testing/sdk_auth.py
```diff
@@ -366,8 +376,8 @@ def __encode_secret(self, keytab_path: str) -> List[str]:
                     keytab = f.read()
 
                 base64_encoding = base64.b64encode(keytab).decode("utf-8")
-                with open(base64_encoded_keytab_path, "w") as f:
-                    f.write(base64_encoding)
+                with open(base64_encoded_keytab_path, "wb") as f:
```
adamtheturtle: Without this we get:

```
testing/sdk_auth.py:379: error: Incompatible types in assignment (expression has type "TextIO", variable has type "BinaryIO")
testing/sdk_auth.py:380: error: No overload variant of "write" of "BinaryIO" matches argument type "str"
testing/sdk_auth.py:380: note: Possible overload variants:
testing/sdk_auth.py:380: note:     def write(self, s: bytearray) -> int
testing/sdk_auth.py:380: note:     def write(self, s: bytes) -> int
```

adamtheturtle:
AFAIK we have exactly one actual code change (not hints).

This is at https://github.com/mesosphere/dcos-commons/pull/2980/files#diff-61b2ec0ab12111d3a3ed40fba2428b2cL370. I think that this should not cause anything to break, but some tests fail. I am re-running in case these are flakes.

mypy.ini
```diff
@@ -0,0 +1,15 @@
...
+disallow_untyped_defs = True
+follow_imports = 'normal'
+ignore_missing_imports = True
+no_implicit_optional = True
+strict_optional = True
+warn_no_return = True
+warn_redundant_casts = True
+warn_return_any = True
+warn_unused_configs = True
+warn_unused_ignores = True
```
adamtheturtle: I recognise that this should have a newline but I would like to put fixing this off until a follow-up.

adamtheturtle:
There are some minor comments on these files from @wavesoft on #2979. I would like to handle those in a follow-up.

From Slack:

```
Adam: Hey @here - I’m trying to push forward the work I started in the last weeks. In particular https://github.com/mesosphere/dcos-commons/pull/2980. There is one `Required` builder failing (`infinity/helloworld` ). In particular, the cluster does not start. I have re-run this multiple times. This builder has passed once in the last 10 master runs - https://teamcity.mesosphere.io/viewType.html?buildTypeId=Frameworks_Orchestration_DcosCommons_PrJobs_HelloWorldPr&branch_Frameworks_Orchestration_DcosCommons_PrJobs=%3Cdefault%3E&tab=buildTypeStatusDiv.

Marcin: I'd ignore it if it's a big pain and failures are unrelated.
```

Thank you @porridge - this is ready for another round of review and I hope merge!

testing/sdk_auth.py
```diff
@@ -38,24 +38,27 @@
...
-        tasks = json.loads(raw_tasks)
-        for task in tasks:
-            if task["name"] == task_name:
-                return task
-
-    raise RuntimeError(
-        "Expecting marathon KDC task but no such task found. Running tasks: {tasks}".format(
-            tasks=raw_tasks
+    @retrying.retry(stop_max_attempt_number=3, wait_fixed=2000)
+    def _get_kdc_task(task_name: str) -> dict:
```
porridge: Please rename this so we don't have two functions called exactly the same.

testing/sdk_auth.py
porridge: Hm, `decode` returns a `str`, so passing `base64_encoding` to `open(..., "w").write()` should just work, no?

I think all you need to do here is use a variable name different than `f`, and not change to binary mode. In theory this should be all ASCII, but safe is better than sorry.

adamtheturtle:
@porridge - I addressed your comments - please can I merge?

It looks as though a bunch of PRs are failing with the same errors as `infinity/helloworld` and `infinity/helloworld/strict` are failing here:

https://teamcity.mesosphere.io/viewType.html?buildTypeId=Frameworks_Orchestration_DcosCommons_PrJobs_HelloWorldPr&tab=buildTypeHistoryList&branch_Frameworks_Orchestration_DcosCommons_PrJobs=__all_branches__

PR #2988, opened 2019-03-20
Add mypy type hints to the 'tools' directory

adamtheturtle:
This is a subset of #2979.
Please see the description there for more information.

I split it out as requested to allow for easier reviews.

### DIFF
--- a/tools/airgap_linter.py
+++ b/tools/airgap_linter.py
@@ -11,9 +11,10 @@
 import re
 import sys
 import os
+from typing import List, Match
 
 
-def readlines_if_text_file(filename):
+def readlines_if_text_file(filename: str) -> List[str]:
     try:
         with open(filename, "r", encoding="utf8") as fh:
             return fh.readlines()
@@ -23,7 +24,7 @@ def readlines_if_text_file(filename):
         return []
 
 
-def extract_uris(file_name):
+def extract_uris(file_name: str) -> List[str]:
     lines = readlines_if_text_file(file_name)
 
     matcher = re.compile(r".*https?:\/\/([^\?\s]*)", re.IGNORECASE)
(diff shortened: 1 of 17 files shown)


adamtheturtle:
It looks as though a bunch of PRs are failing with the same errors as `infinity/helloworld` and `infinity/helloworld/strict` are failing here:

https://teamcity.mesosphere.io/viewType.html?buildTypeId=Frameworks_Orchestration_DcosCommons_PrJobs_HelloWorldPr&tab=buildTypeHistoryList&branch_Frameworks_Orchestration_DcosCommons_PrJobs=__all_branches__

Therefore I would like to ignore these errors.

adamtheturtle:
Thank you @akirillov for the review and approval.

> A quick question: why in e.g. package.py class names are quoted like 'Version' and 'Package'?

When we are using a class defined in the same module as a type hint, we would get the following error when importing the file:

```
~/D/w/m/r/dcos-commons (dcos-commons) ~> python -c 'from tools.universe import package'
  File "<string>", line 1, in <module>
  File "/Users/adam/Documents/work/mesosphere/repositories/dcos-commons/tools/universe/__init__.py", line 2, in <module>
    from .package import Package
  File "/Users/adam/Documents/work/mesosphere/repositories/dcos-commons/tools/universe/package.py", line 6, in <module>
    @functools.total_ordering
  File "/Users/adam/Documents/work/mesosphere/repositories/dcos-commons/tools/universe/package.py", line 9, in Package
    def from_json(json: Dict[str, Any]) -> Package:
NameError: name 'Package' is not defined
```

`mypy` allows us to use strings for class names to avoid this issue.
I will merge this and I will soon submit the next set of type hints.

porridge:
Unfortunately some of these type hints are incompatible with Python 3.5 or lower, and we happen to use that version in the Jenkins jobs. We need these jobs to be functional and upgrading Python there does not look like will be a straightforward change, so we might have to temporarily revert this PR.

PR #2989, opened 2019-03-20
[TOOLS] Source required file for setting MESOS_WORK_DIR in clusters created with onprem amis 

takirala:
verified this via https://teamcity.mesosphere.io/viewLog.html?buildId=1715989&buildTypeId=Frameworks_Orchestration_DcosCommons_PrJobs_TestOnpremHelloWorldPr&tab=buildResultsDiv


takirala:
This is a tooling only change to support onprem provision of clusters in HW CI. I _think_ the CI will detect no changes in src code / test code and thus skip to run the integration tests - thus merging this without green CI, will revert the change if anything breaks.

PR #2990, opened 2019-03-21
Add mypy types to the elastic framework

adamtheturtle:
This is a subset of #2979.

@wavesoft already looked at the changes there.
However, as that PR was huge, I was fairly asked to split it up.

A key question from @wavesoft was why we wrap a few things in `bool()`.
The answer is that sometimes `mypy` is not sure that something will evaluate to a boolean.

For example, if we take the following pseudocode:

```
@retry.retry(...)
def my_function() -> bool:
    return True

my_var = my_function()
```

We can fairly interpret as humans that `my_var` is likely a boolean.
However, the `retry` function has no type hints and could in theory change the return value of the function.

### DIFF
--- a/frameworks/elastic/tests/conftest.py
+++ b/frameworks/elastic/tests/conftest.py
@@ -1,8 +1,10 @@
+from typing import Iterable
+
 import pytest
 import sdk_security
 from tests import config
 
 
 @pytest.fixture(scope="session")
-def configure_security(configure_universe):
+def configure_security(configure_universe: None) -> Iterable:
     yield from sdk_security.security_session(config.SERVICE_NAME)
--- a/frameworks/elastic/tests/conftest.py
+++ b/frameworks/elastic/tests/conftest.py
@@ -1,4 +1,4 @@
-from typing import Iterable
+from typing import Iterator
 
 import pytest
 import sdk_security
@@ -6,5 +6,5 @@ from tests import config
 
 
 @pytest.fixture(scope="session")
-def configure_security(configure_universe: None) -> Iterable:
+def configure_security(configure_universe: None) -> Iterator[None]:
     yield from sdk_security.security_session(config.SERVICE_NAME)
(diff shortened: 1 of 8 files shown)


frameworks/elastic/tests/conftest.py
```diff
@@ -1,8 +1,10 @@
+from typing import Iterable
+
 import pytest
 import sdk_security
 from tests import config
 
 
 @pytest.fixture(scope="session")
-def configure_security(configure_universe):
+def configure_security(configure_universe: None) -> Iterable:
```
akirillov: shouldn't `Iterable` have a type parameter?

frameworks/elastic/tests/test_tls.py
```diff
@@ -34,7 +35,7 @@ def service_account(configure_security):
 
 
 @pytest.fixture(scope="module")
-def elastic_service(service_account):
+def elastic_service(service_account: Dict[str, Any]) -> Iterable[Dict[str, Any]]:
```
akirillov: and only here `Iterable` is type-parameterized

adamtheturtle:
Thanks both for the review.

@akirillov - I have added more precise types for `Iterator`s.
@mpereira - `assert`s are actually used by `mypy`. For example - if I have:

```python
from typing import Union

def my_function(my_parameter: Union[str, bool]) -> Union[str, bool]:
    return my_parameter

result = my_function('hello')
print(result + 'a')
```

`mypy` will show:

```
example.py:7: error: Unsupported operand types for + ("bool" and "str")
example.py:7: note: Left operand is of type "Union[str, bool]"
```

However, as humans we are sure that it will be a `str`.
We can use `assert` to tell `mypy` this:

```
result = my_function('hello')
assert isinstance(result, str)
print(result + 'a')  # no mypy error!
```

I will make a judgement call and merge as the `code-check` CI is finished as the changes I made really should not affect other things.

PR #2993, opened 2019-03-25
[DCOS-50694] Revert "Add mypy type hints to the 'tools' directory"

porridge:
Reverts mesosphere/dcos-commons#2988 due to https://jira.mesosphere.com/browse/DCOS-50694


porridge:
I think this looks healthy enough to consider it safe to merge. Pressing the red button.

PR #3003, opened 2019-04-05
[DCOS-51307] Propagate options in sdk_upgrade when to_options not specified.

kaiwalyajoshi:
Propagate `from_options` as `to_options` in the upgrade scenario when `to_options` are not explicitly defined to maintain compatibility.
Explicitly propagate hello-world install-options for upgrade scenario in test_decomission tests.


kaiwalyajoshi:
@takirala I've added code to propagate `from_options` to `to_options` and changed the title. PTAL.

PR #3018, opened 2019-04-23
[DCOS-51901] Support user defined functions configuration

viivek46:
[DCOS-51901] - Support user_defined_functions configuration

## What changes were proposed in this pull request?

Resolves [COPS-3394] - Support user_defined_functions to Cassandra configuration options (https://jira.mesosphere.com/browse/COPS-3394) 

Added an integration tests to verify the changes.
in integration test created simple UDF and using it to verify it.

## Release Notes

With this change user would be able to configure enable_user_defined_functions and enable_scripted_user_defined_functions.
By default its set to false but user can make it true if they want to use UFD in cassandra.
This options are added under Cassandra tab in DOCS.


frameworks/cassandra/tests/test_udf.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        test_jobs = config.get_udf_jobs()
+        # destroy/reinstall any prior leftover jobs, so that they don't touch the newly installed service:
+        for job in test_jobs:
+            sdk_jobs.install_job(job)
+
+        service_options = {
+            "service": {
+                "name": config.SERVICE_NAME,
+                "security": {"authentication": {"enabled": True}, "authorization": {"enabled": True}},
```
samvantran: I'm surprised this works... your other PR where you add these configs hasn't been merged yet

frameworks/cassandra/tests/test_udf.py
```diff
@@ -0,0 +1,52 @@
...
+            },
+            "cassandra": {
+                "enable_user_defined_functions": True,
+                "enable_scripted_user_defined_functions": True,
+            }
+        }
+
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
```
samvantran: Can we reuse an existing Cassandra cluster from a different test? We're only testing UDFs (a simple one) - it seems a little cumbersome to un+install a fresh C* cluster just to test this feature. From TC log, it looks like this test alone adds an additional 3.5 minutes to the already very long CI job.

frameworks/cassandra/tests/config.py
```diff
@@ -208,6 +201,44 @@ def get_write_data_job(
...
+
+
+def get_verify_udf_data_job(
+    node_address: str = DEFAULT_NODE_ADDRESS,
+    node_port: str = DEFAULT_NODE_PORT,
+    dcos_ca_bundle: Optional[str] = None,
+) -> Dict[str, Any]:
+    cmds = [
+        "{} | grep 200".format(
+            _cqlsh("SELECT maxOf(val1,val2) FROM testspace1.testtable1;", node_address, node_port)
```
samvantran: ha, looks like the function names are case-insensitive since you created the function as `maxof()`

frameworks/cassandra/tests/test_udf.py
viivek46: correct, its just do not apply options which are not available rather then failing. but removed it as well.

frameworks/cassandra/tests/test_udf.py
viivek46: added UDF test to test_sanity.py

frameworks/cassandra/tests/config.py
viivek46: ye abut updated it as well.

frameworks/cassandra/tests/test_sanity.py
```diff
@@ -125,3 +125,33 @@ def test_custom_jmx_port() -> None:
...
+        for job in test_jobs:
+            sdk_jobs.install_job(job)
+
+        new_config = {
+            "cassandra": {
+                "enable_user_defined_functions": True,
+                "enable_scripted_user_defined_functions": True,
+            }
+        }
+        sdk_service.update_configuration(
```
samvantran: Sorry I wasn't clear but by changing the test to update configs you're still effectively relaunching all the tasks in the service but I reconsidered my request. Because we don't install UDFs by default we _should_ follow this pattern and update configs and test the feature like you are doing.

samvantran:
Looks like you might have to fix up some of the python code to make the linters happy

PR #3019, opened 2019-04-24
[DCOS-52245] Preparatory change for wrapping dcos-launch

porridge:
Introduce a wrapper script around `dcos-launch` as the first stage of encapsulating it and its dependencies. Actual logic will be added once the callers defined in TeamCity configs are converted to use it.
More details about the design in the ticket.


porridge:
FWIW, the failures seem to be unrelated flakes.

porridge:
> LGTM. Would like to find out more about how we're going to use the venvs here in the upcoming design.

@kaiwalyajoshi are you looking for more detail than what's described in https://jira.mesosphere.com/browse/DCOS-52245? Happy to chat.

PR #3021, opened 2019-04-25
[DCOS-52239] Reinstate the init script

porridge:
This will allow the updating instructions in older copies of UPDATING.md
to still work.


porridge:
Docker image built fine, merging.

PR #3022, opened 2019-04-25
[DCOS-52239] Replace the spec section with a quickstart one

porridge:
Remove the short section on the service spec - there is a very similar but
better one in the developer guide already, and it's not really appropriate
to attack a complete newcomer with such details.

Instead, re-add the quick-start section that should be more useful to
someone who wants to start using the SDK. In there, point them at the
dcosdev tool and the helloworld example for the lack of a better HOWTO
guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,58 +13,15 @@
 __DC/OS SDK__ is a collection of tools, libraries, and documentation for easy integration of technologies such as Kafka, Cassandra, HDFS, Spark, and TensorFlow with DC/OS.
 
 ---
-### Understanding the Hello World Service Specification
+### Quick Start
 
-The service specification declaratively defines the `helloworld` service:
+To create a new service based on the SDK:
+1. read the [developer guide](https://mesosphere.github.io/dcos-commons/developer-guide/) and the [FAQ](https://mesosphere.github.io/dcos-commons/faq/)
+1. either install and use the [dcosdev tool](https://github.com/mesosphere/dcosdev) or follow the example of [hello-world framework](frameworks/helloworld)
 
-```yaml
-name: "helloworld"
-pods:
-  helloworld:
-    count: {{COUNT}}
-    tasks:
-      server:
-        goal: RUNNING
-        cmd: "echo 'Hello World!' >> helloworld-container-volume/output && sleep 10"
-        cpus: {{SERVER_CPU}}
-        memory: 32
-        volume:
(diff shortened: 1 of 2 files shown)


porridge:
I'll be away for a week starting tomorrow, but it would be great if someone could merge this in time for `0.56.1` release if this LGTY.

README.md
```diff
@@ -13,58 +13,15 @@
 __DC/OS SDK__ is a collection of tools, libraries, and documentation for easy integration of technologies such as Kafka, Cassandra, HDFS, Spark, and TensorFlow with DC/OS.
 
 ---
-### Understanding the Hello World Service Specification
+### Quick Start
 
-The service specification declaratively defines the `helloworld` service:
+To create a new service based on the SDK:
+1. read the [developer guide](https://mesosphere.github.io/dcos-commons/developer-guide/) and the [FAQ](https://mesosphere.github.io/dcos-commons/faq/)
+1. either install and use the [dcosdev tool](https://github.com/mesosphere/dcosdev) or follow the example of [hello-world framework](frameworks/helloworld)
```
kaiwalyajoshi: I think you meant a numbered list here.

```suggestion
2. either install and use the [dcosdev tool](https://github.com/mesosphere/dcosdev) or follow the example of [hello-world framework](frameworks/helloworld)
```

README.md
```diff
@@ -13,58 +13,15 @@
...
-
-```yaml
-volume:
-  path: "helloworld-container-volume"
-  type: ROOT
-  size: 64
-```
+To update an existing framework to use the latest version of the SDK:
+1. review the [SDK changelog](changelog.md), and
+1. refer to the current [update instructions](tools/distribution/UPDATING.md)
```
kaiwalyajoshi: I think you meant a numbered list here.
```suggestion
2. refer to the current [update instructions](tools/distribution/UPDATING.md)
```

README.md
porridge: I did mean a numbered list, but using all ones is a common pattern in markdown, lets you take the renderer take care about keeping the numbers right.
See [the MD guide](https://www.markdownguide.org/basic-syntax/#lists) and that [it actually renderes correctly](https://github.com/mesosphere/dcos-commons/tree/DCOS-52239/fix#quick-start)

README.md
porridge: ditto

PR #3024, opened 2019-04-29
[DCOS-32730]Added rlimit_memlock parameter to the elastic configuration

rishabh96b:
In order to make `bootstrap.memory_lock: true` working in Elastic, we need _rlimit_memlock_ to be set to `infinity`.


takirala:
Elastic is no longer in monorepo

PR #3026, opened 2019-04-30
[DCOS-52831] Added Grafana dashboard from Cassandra and Readme

alembiewski:
This PR contains Grafana dashboard for Cassandra package as well as readme file describing installation steps

### DIFF
--- a/frameworks/cassandra/dashboards/DC_OS_Cassandra_Dashboard.json
+++ b/frameworks/cassandra/dashboards/DC_OS_Cassandra_Dashboard.json
@@ -1,0 +1,4159 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "description": "Dashboard for Cassandra Package",
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": 28,
+  "iteration": 1556615593345,
+  "links": [],
+  "panels": [
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 6,
+      "panels": [],
+      "title": "Errors",
+      "type": "row"
+    },
+    {
+      "cacheTimeout": null,
+      "colorBackground": false,
+      "colorValue": true,
+      "colors": [
+        "#299c46",
+        "rgba(237, 129, 40, 0.89)",
+        "#d44a3a"
+      ],
+      "datasource": "prometheus",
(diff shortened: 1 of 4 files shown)


samvantran:
@zmalik tagging you on this to make sure that dashboards are aligned with yours and that we try to maintain a consistent UX across FWs.

alembiewski:
@samvantran, thanks for the feedback! Yes, the problem indeed in metric format: for 1.12.3, metrics get into Prometheus in all-lowercase format, but for 1.13 - in camel case. I remember the same issue with Kafka metrics as well (@zmalik, could you please confirm?)
Prometheus queries is case-sensitive, therefore I guess we need to create a separate dashboard for 1.12, or unify metric names (if it's possible to do so). 
Also, could you please check the dashboard on 1.13?

zmalik:
@alembiewski @samvantran 
metrics for DC/OS >= 1.13 are camelCase and for DC/OS version <=1.12 are lowercase

To tackle this we would be supporting two equal dashboards with different queries(lowercase/camelCase) present in two branches of grafana-dashboards:

https://github.com/dcos/grafana-dashboards/tree/1.13.x
https://github.com/dcos/grafana-dashboards/tree/1.12.x

@alembiewski to get `service_name` you are using a metric like `org_apache_cassandra_metrics_Client_connectedNativeClients` and probably that won't work if @samvantran is using 1.12.x DC/OS cluster. 

To get it more similar with Kafka and data-services related dashboards lets sync. @shubhanilBag is already working on getting `Mesos` level singlestats/graphs that I guess we all can use, as the first row for data-services frameworks.

samvantran:
Thanks @alembiewski for adding a 2nd graph for version 1.12 but the issue of reporting still exists. When I import the graph, I see a lot of zero values but if I go in and remove some of the variable interpolation, I get non-zero values. 

For example: the  `Live SSTable Count` graph:
if I add another query for: `org_apache_cassandra_metrics_table_livesstablecount_all` I see these values reported 

Screen Shot 2019-05-06 at 3 26 53 PM

Please take a 2nd look at these graphs and check whether reported zeros are true. I think you may have to adjust your variables. Also, please sync with @zmalik and @shubhanilBag so we can make the UX a little more consistent across FWs.

alembiewski:
@samvantran, the query `org_apache_cassandra_metrics_table_livesstablecount_all` returns a sum of all `LiveSSTableCount` across all tables and keyspaces, for all cassandra node, which is likely non-zero value. I will add this metric to the dashboard as well. In addition to that, we are having a possibility to filter this metric by keyspace/table. Zero values, that you see on your graph, is valid for selected keyspace/table. You can try to select another table or switch keyspace and you will see, that graph is changing.

image

Issue with `service_name` was caused by metric name, and has been fixed.

samvantran:
Thanks @alembiewski you're right the graph was specific to the table. I didn't realize that was the default view - sorry about that. But I think adding an overall view of the cluster first is the right step. As operators first thing to look as is the whole state of the cluster: "how do _all_ the Cassandra clusters look? All the keyspaces? All the tables?". Once an anomaly is detected only then do we drill down deeper into specific services/tasks/tables.

PR #3032, opened 2019-05-03
[DCOS-53276] Parameterize remaining values in cassandra.yaml

viivek46:

Resolves [DCOS-53276](https://jira.mesosphere.com/browse/DCOS-53276)

There were some remaining hardcoded values in cassandra.yaml, parameterizing all the hardcoded values.

TC CI tests


frameworks/cassandra/src/main/dist/cassandra.yaml
```diff
@@ -143,6 +143,18 @@ otc_coalescing_strategy: {{CASSANDRA_OTC_COALESCING_STRATEGY}}
 credentials_validity_in_ms: {{CREDENTIALS_VALIDITY_IN_MS}}
 credentials_update_interval_in_ms: {{CREDENTIALS_UPDATE_INTERVAL_IN_MS}}
 permissions_cache_max_entries: {{PERMISSIONS_CACHE_MAX_ENTRIES}}
+cdc_enabled: {{CASSANDRA_CDC_ENABLED}}
+prepared_statements_cache_size_mb: {{CASSANDRA_PREPARED_STATEMENTS_CACHE_SIZE_MB}}
+thrift_prepared_statements_cache_size_mb: {{CASSANDRA_THRIFT_PREPARED_STATEMENTS_CACHE_SIZE_MB}}
+column_index_cache_size_in_kb: {{CASSANDRA_COLUMN_INDEX_CACHE_SIZE_IN_KB}}
+slow_query_log_timeout_in_ms: {{CASSANDRA_SLOW_QUERY_LOG_TIMEOUT_IN_MS}}
+back_pressure_enabled: {{CASSANDRA_BACK_PRESSURE_ENABLED}}
```
samvantran: Should we template this ? If backpressure is disabled (ie not enabled) then we do not need to include the 
```
back_pressure_strategy:
  ...
```
configs right?

This also seems like it opens doors for customers to customize (like authn custom yaml block) _but_ this is outside the scope of this PR so we can leave that for another day if needed

frameworks/cassandra/universe/config.json
```diff
@@ -730,6 +735,116 @@
...
+          "type": "boolean",
+          "description": "Enables or disables taking a snapshot before each compaction. A snapshot is useful to back up data when there is a data format change.",
+          "default": false
+        },
+        "cross_node_timeout": {
+          "type": "boolean",
+          "description": "operation timeout information exchange between nodes (to accurately measure request timeouts).",
+          "default": false
+        },
+        "role_manager": {
```
samvantran: This config is not actually used or passed through to marathon.mustache but the `security.authentication.role_manager` is.  You can delete this one

frameworks/cassandra/universe/config.json
```diff
@@ -730,6 +735,116 @@
...
+        },
+        "counter_cache_size_in_mb": {
+          "type": "string",
+          "description": "When no value is set, Cassandra uses the smaller of minimum of 2.5% of Heap or 50MB.",
+          "default": ""
+        },
+        "commitlog_sync": {
+          "type": "string",
+          "description": "The method that Cassandra uses to acknowledge writes in milliseconds",
+          "default": "periodic"
```
samvantran: is there a finite set of possible configs? I would suggest using enums in this case

frameworks/cassandra/universe/config.json
```diff
@@ -730,6 +735,116 @@
...
+        },
+        "role_manager": {
+          "type": "string",
+          "description": "Part of the Authentication & Authorization backend that implements IRoleManager to maintain grants and memberships between roles.",
+          "default": "CassandraRoleManager"
+        },
+        "commit_failure_policy": {
+          "type": "string",
+          "description": "Policy for commit disk failures.",
+          "default": "stop"
```
samvantran: Please use enums here if there is a finite set

frameworks/cassandra/universe/config.json
```diff
@@ -730,6 +735,116 @@
...
+        },
+        "index_summary_capacity_in_mb": {
+          "type": "string",
+          "description": "Fixed memory pool size in MB for SSTable index summaries.",
+          "default": ""
+        },
+        "rpc_server_type": {
+          "type": "string",
+          "description": "Cassandra provides three options for the RPC server. sync and hsha performance is about the same, but hsha uses less memory.",
+          "default": "sync"
```
samvantran: please use enums since there are only 3 options

frameworks/cassandra/universe/config.json
viivek46: nice catch @samvantran , i will remove it.

frameworks/cassandra/universe/config.json
viivek46: Cassandra provides two out-of-the-box options for the RPC Server : sync, hsha.
Alternatively,  can provide your own RPC server by providing the fully-qualified class name
of an o.a.c.t.TServerFactory that can create an instance of it.
@samvantran should I use enum with 2 values sync, hsha or should I keep it as it is.

frameworks/cassandra/universe/config.json
viivek46: done.

frameworks/cassandra/universe/config.json
viivek46: done

frameworks/cassandra/universe/config.json
samvantran: If a user can provide their own server then you can leave as is. This goes for any config that allows for custom values. It's only if there is a finite set do we use enums.

PR #3039, opened 2019-05-09
Build and push docker image on release tags too.

porridge:
Needs tag discovery to be enabled to be active:
https://jenkins.io/blog/2018/05/16/pipelines-with-git-tags/

### DIFF
--- a/Jenkinsfile.dcos-commons
+++ b/Jenkinsfile.dcos-commons
@@ -15,6 +15,17 @@ pipeline {
  
     stages {
         stage('docker login') {
+            when {
+                anyOf {
+                    branch '*';
+                    // Do not re-build old tags, which were pushed by dockerhub autobuilder
+                    tag pattern "^0\\.5[6-9]\\..*", comparator: "REGEXP";
+                    tag pattern "^0\\.[6-9][0-9]\\..*", comparator: "REGEXP";
+                    tag pattern "^0\\.[1-9][0-9][0-9]\\..*", comparator: "REGEXP";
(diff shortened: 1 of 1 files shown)


Jenkinsfile.dcos-commons
```diff
@@ -15,19 +15,40 @@ pipeline {
  
     stages {
         stage('docker login') {
+            when {
+                anyOf {
+                    branch '*';
+                    // Do not re-build old tags, which were pushed by dockerhub autobuilder
+                    tag pattern "^0\\.5[6-9]\\..*", comparator: "REGEXP";
```
takirala: What do you think if we change the regex to sth like `^[0-9]{1,2}.[0-9]{1,2}.[0-9]{0,3}$` and do this:
- Try to do a docker pull of the image we want to push.
- If the pull fails, we can continue to push the image.
- If the push succeeds, the image is already present and we SHOULD NOT override the image, this also protects us from overriding an existing image (we can use the jenkins job under `release-tools` if we ever need to override anyway). 

This would also make the regex expr simpler/less redundant. I don't have a hard opinion on this though.

porridge: > If the pull fails, we can continue to push the image.

I'm afraid that if the pull fails for a reason other than it not existing (like a network flake or timeout) we will misunderstand and potentially overwrite it :-/

porridge:
Hm, looks like our jenkins does not like the syntax, perhaps it's too old :-(

```
WorkflowScript: 22: expecting '}', found ',' @ line 22, column 51.
   tag pattern "^0\\.5[6-9]\\..*", comparat
```
Reverting.

PR #3040, opened 2019-05-09
Update SDK to run on Open-JDK 8

kaiwalyajoshi:
This is an intermediate step in the migration towards OpenJDK 11.

This PR only replaces Oracle JRE 8 with OpenJDK 8.

### DIFF
--- a/tools/universe/package_builder.py
+++ b/tools/universe/package_builder.py
@@ -16,7 +16,7 @@ import urllib.request
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
-_jre_url = "https://downloads.mesosphere.com/java/openjdk-8u212b03-hotspot-linux-x64.tar.gz"
+_jre_url = "https://downloads.mesosphere.com/java/openjdk-jre-8u212b03-hotspot-linux-x64.tar.gz"
 _libmesos_bundle_url = (
     "https://downloads.mesosphere.com/libmesos-bundle/libmesos-bundle-1.12.0.tar.gz"
 )
--- a/tools/universe/package_builder.py
+++ b/tools/universe/package_builder.py
@@ -16,7 +16,7 @@ import urllib.request
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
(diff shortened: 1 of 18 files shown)


tools/universe/package_builder.py
```diff
@@ -16,7 +16,7 @@
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
-_jre_url = "https://downloads.mesosphere.com/java/server-jre-8u192-linux-x64.tar.gz"
+_jre_url = "https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u212-b03/OpenJDK8U-jdk_x64_linux_hotspot_8u212b03.tar.gz"
```
VladimirKravtsov: Hi dear @kaiwalyajoshi, @samvantran,
I just noticed that previously **JRE** was used here and in other places, but now it's pointing to **JDK**, although I guess make sense switch it to **JRE**.  You can find it here https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u212-b03/OpenJDK8U-jre_x64_linux_hotspot_8u212b03.tar.gz

kaiwalyajoshi: We're using `https://downloads.mesosphere.com/java/openjdk-8u212b03-hotspot-linux-x64.tar.gz` to maintain consistency with the other named packages.

The two images are exactly the same, just named differently.

tools/universe/package_builder.py
```diff
@@ -16,7 +16,7 @@
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
 
-_jre_url = "https://downloads.mesosphere.com/java/server-jre-8u192-linux-x64.tar.gz"
+_jre_url = "https://downloads.mesosphere.com/java/openjdk-8u212b03-hotspot-linux-x64.tar.gz"
```
takirala: Wouldn't jre's at https://adoptopenjdk.net/archive.html?variant=openjdk8&jvmVariant=hotspot be sufficient ?

kaiwalyajoshi: Good catch! Updated to fetch JRE here.

tools/universe/package_builder.py
VladimirKravtsov: After editing your the message, LGTM:-)

kaiwalyajoshi:
Replaced by this PR: https://github.com/mesosphere/dcos-commons/pull/3060

PR #3043, opened 2019-05-10
[DCOS-53489] cassandra test_auth failing on DC/OS Open

viivek46:

Resolves [DCOS-53489](https://jira.mesosphere.com/browse/DCOS-53489)

Recently added integrations test test_auth.py is failing in open DCOS , this change will fix it.

TC CI tests and manually tested over Open DCOS

### DIFF
--- a/frameworks/cassandra/tests/test_auth.py
+++ b/frameworks/cassandra/tests/test_auth.py
@@ -6,7 +6,6 @@ import sdk_jobs
 import sdk_cmd
 import sdk_install
 import sdk_tasks
-import sdk_utils
 import os
 from tests import config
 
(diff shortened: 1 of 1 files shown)


frameworks/cassandra/tests/test_auth.py
```diff
@@ -6,11 +6,18 @@
 import sdk_cmd
 import sdk_install
 import sdk_tasks
+import os
 from tests import config
 
 log = logging.getLogger(__name__)
 
 
+no_auth_test_for_open_dcos = pytest.mark.skipif(
```
samvantran: This looks near identical to the `sdk_utils.dcos_ee_only` method [here](https://github.com/mesosphere/dcos-commons/blob/DCOS-53489-cassandra-test-auth-failing-on-DC/OS-Open/testing/sdk_utils.py#L198). No need to reinvent, can you import and use the method instead?

viivek46: Correct @samvantran , 
this is what I did previously,import sdk_utils, and marked @sdk_utils.dcos_ee_only  # at top of test.
but it was not working, so I had to use this way.

PR #3063, opened 2019-05-22
[HDFS] Ignore CLI version warnings in stderr

samvantran:
Tests were failing due to warnings like:
```
The attached cluster is running DC/OS 1.14 but this CLI only supports DC/OS 1.12.
```
that were sneaking into the `stderr` causing tests to incorrectly think that failures existed.

This PR ignores those warnings and allows the tests to proceed.


samvantran:
Merging now that HDFS tests are all green

PR #3064, opened 2019-05-24
[DCOS-54393] wait for marathon app deployment during service update correctly

takirala:
Fixes the flaky `test_sidecar.test_envvar_accross_restarts` in helloworld

### DIFF
--- a/frameworks/helloworld/universe/package.json
+++ b/frameworks/helloworld/universe/package.json
@@ -2,7 +2,7 @@
   "packagingVersion": "4.0",
   "upgradesFrom": ["*"],
   "downgradesTo": ["*"],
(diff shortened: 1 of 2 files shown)


frameworks/helloworld/universe/package.json
```diff
@@ -2,7 +2,7 @@
   "packagingVersion": "4.0",
   "upgradesFrom": ["*"],
   "downgradesTo": ["*"],
-  "minDcosReleaseVersion": "1.10",
+  "minDcosReleaseVersion": "1.13",
```
takirala: This is not strictly mandatory but it still makes sense to have this just to ensure we don't deploy `seccomp.yaml` in older clusters.

PR #3069, opened 2019-05-28
[DCOS-53561] Add support for accurate Scheduler Plan status in the UI with SDK 0.56.0+

vibhujain:

Resolves [DCOS-53561](https://jira.mesosphere.com/browse/DCOS-53561)
Added code for a check to show accurate scheduler plan status in UI with SDK 0.56.0 or later. Also, The default values are parameterized so that they can be configured at the time of deployment.

TC CI tests


vibhujain:
@samvantran I have rebased my PR with master so that the `mesosphere/dcos-commons-base` test could pass that was failing due to OPENJDK issue, but seems like tests are still pending (when going in test details, we can see that they have successfully passed). There may be some issue on the Jenkins side to reflect back the successful results to Github. As @porridge already pinged me that these tests can be ignored, can I go ahead and merge the PR?

samvantran:
Yup you've had multiple approvals. Merge when ready

PR #3071, opened 2019-05-29
Bump version to 0.56.1-SNAPSHOT post release

porridge:
I was under impression that doing this is a part of the release checklist?
Please merge this if it LGTY

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,7 +31,7 @@ allprojects {
     // 2. In the GitHub UI, add a Release entry against the tag created in step 1. List any changes in release notes.
     // 3. Create a PR which bumps this from '0.10.0-SNAPSHOT' to '0.11.0-SNAPSHOT' in master to start the 0.11.0 track.
     // --
-    version = '0.56.0-SNAPSHOT'
+    version = '0.56.1-SNAPSHOT'
 
     sourceCompatibility = '1.8'
     targetCompatibility = '1.8'


build.gradle
```diff
@@ -31,7 +31,7 @@ allprojects {
     // 2. In the GitHub UI, add a Release entry against the tag created in step 1. List any changes in release notes.
     // 3. Create a PR which bumps this from '0.10.0-SNAPSHOT' to '0.11.0-SNAPSHOT' in master to start the 0.11.0 track.
     // --
-    version = '0.56.0-SNAPSHOT'
+    version = '0.56.1-SNAPSHOT'
```
kaiwalyajoshi: Should be the following if we're following the Wiki for [SDK Releases](https://wiki.mesosphere.com/display/ENG/%5BSDK%5D+Releasing+the+SDK).

```suggestion
    version = '0.57.0-SNAPSHOT'
```

porridge: Hm, that would currently break our CI, since' [were expecting](https://github.com/mesosphere/data-services-ci/blob/a85ff086c4665ca22b909e5ad24a833691ffc380/.teamcity/DataServices/buildTypes/ClusterConfig.kt#L160) a `mesosphere/dcos-commons:x.y.z` image to exist (where `x.y.z` is taken by ignoring an optional `-SNAPSHOT`). :-/

porridge:
I will merge this for the time being, since it is no worse than the current situation, and in the background will think on how we can make this work with the wiki-prescribed version string.

PR #3077, opened 2019-06-03
DCOS-52237 Remove dcos-commons/tools/vagrant

VladimirKravtsov:


porridge:
Adding folks from the SDK team to double-check nobody uses it.

PR #3095, opened 2019-06-10
[DCOS-51992] Auditing and changing Non-HTTPS resources & references to HTTPS

vibhujain:

Resolves [DCOS-51992](https://jira.mesosphere.com/browse/DCOS-51992)
Audited the files to remove Non-HTTPS resources and references by replacing them with HTTPS ones.

The links were manually checked by browsing their https equivalent and replacing them in place.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,4 +38,4 @@ This site used to also host per-service user documentation, but that was moved i
 
 # Notes
 
-The page uses the [Dropdown](http://code.iamkate.com/javascript/touch-friendly-drop-down-menus/) Javascript library by Kate Morley for rendering the menus. The library is licensed [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/legalcode).
+The page uses the [Dropdown](https://code.iamkate.com/javascript/touch-friendly-drop-down-menus/) Javascript library by Kate Morley for rendering the menus. The library is licensed [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/legalcode).
--- a/docs/generate.sh
+++ b/docs/generate.sh
@@ -21,7 +21,7 @@ OUTPUT_DIR=dcos-commons-gh-pages/dcos-commons
 SWAGGER_CODEGEN_VERSION=2.2.2
 SWAGGER_OUTPUT_DIR=reference/swagger-api
 SWAGGER_JAR=swagger-codegen-cli-${SWAGGER_CODEGEN_VERSION}.jar
-SWAGGER_URL=http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/${SWAGGER_CODEGEN_VERSION}/${SWAGGER_JAR}
+SWAGGER_URL=https://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/${SWAGGER_CODEGEN_VERSION}/${SWAGGER_JAR}
 
 # Default value, override with "HTTP_PORT" envvar:
 DEFAULT_HTTP_PORT=8888
--- a/docs/pages/developer-guide.md
+++ b/docs/pages/developer-guide.md
@@ -19,7 +19,7 @@ The four major components are Apache Mesos, Marathon, Universe, and Apache ZooKe
 
 DC/OS is modeled on an operating system with Mesos as its kernel. Mesos provides an abstraction to enable consumption of the resources a datacenter provides.  In a typical case, these resources are CPU, memory, disk space, and ports. Tasks are launched in the datacenter and consume particular subsets of resources. The programs that can receive resource offers and launch tasks that consume are called **frameworks**. The component of a framework that receives offers and launches tasks is called the **scheduler**.
 
-Mesos determines which frameworks should be offered resources at any given time. It sends update events regarding the status of those tasks. These events include *staging, running, failed*,* *etc.  To learn more about Mesos, consult "[Apache Mesos](http://mesos.apache.org/)".
+Mesos determines which frameworks should be offered resources at any given time. It sends update events regarding the status of those tasks. These events include *staging, running, failed*,* *etc.  To learn more about Mesos, consult "[Apache Mesos](https://mesos.apache.org/)".
 
 ## Marathon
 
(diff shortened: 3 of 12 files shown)


vibhujain:
@samvantran This is a WIP PR as I still need to check some sub-directories just to be sure of not missing any such non-HTTPS references. Still, can you please have a look at the current changes if they look neat to you?

Also, I need to confirm if I need to correct references in `govendor` CLI directory in case of almost all FWs and `gradle` directory in case of `dcos-commons` in particular as they contains lots of third-party packages references in their guide,code,README,etc.

tools/airgap_linter.py
```diff
@@ -178,7 +180,7 @@ def main(argv):
 
     if invalid:
         print(
-            "Airgap check FAILED. This framework will NOT work in an airgap. Fix the detected issues."
+            "Airgap check FAILED. The framework's working code directory contains references to non-https URL resources and/or there are http URL references made inside marathon/config JSON files. Please consider replacing these resources to their HTTPS equivalent. In case of internal Mesos DNS URLs, please mention http links as URI under resources.json. This framework will NOT work in an airgap. Fix the detected issues."
```
samvantran: 👍

docs/README.md
```diff
@@ -38,4 +38,4 @@ This site used to also host per-service user documentation, but that was moved i
 
 # Notes
 
-The page uses the [Dropdown](http://code.iamkate.com/javascript/touch-friendly-drop-down-menus/) Javascript library by Kate Morley for rendering the menus. The library is licensed [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/legalcode).
+The page uses the [Dropdown](https://code.iamkate.com/javascript/touch-friendly-drop-down-menus/) Javascript library by Kate Morley for rendering the menus. The library is licensed [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/legalcode).
```
kaiwalyajoshi: You get an invalid certificate warning on this URL with `https`

docs/generate.sh
```diff
@@ -21,7 +21,7 @@ OUTPUT_DIR=dcos-commons-gh-pages/dcos-commons
 SWAGGER_CODEGEN_VERSION=2.2.2
 SWAGGER_OUTPUT_DIR=reference/swagger-api
 SWAGGER_JAR=swagger-codegen-cli-${SWAGGER_CODEGEN_VERSION}.jar
-SWAGGER_URL=http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/${SWAGGER_CODEGEN_VERSION}/${SWAGGER_JAR}
+SWAGGER_URL=https://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/${SWAGGER_CODEGEN_VERSION}/${SWAGGER_JAR}
```
kaiwalyajoshi: Why did we change from `central` to `repo1` here?

docs/pages/developer-guide.md
```diff
@@ -96,7 +96,7 @@ pods:
 
             * **cmd**: The command to run to start a task. Here, the task will print "hello world" to stdout and sleep for 1000 seconds. Because its goal state is `RUNNING`, it will be started again upon exit.
 
-            * **cpus**: This entry defines how many CPUs will be allocated to the task’s container.  For discussion of how resources are isolated and allocate [see the Mesos documentation here](http://mesos.apache.org/documentation/latest/containerizer/).
+            * **cpus**: This entry defines how many CPUs will be allocated to the task’s container.  For discussion of how resources are isolated and allocate [see the Mesos documentation here](https://mesos.apache.org/documentation/latest/containerizers/).
```
kaiwalyajoshi: :+1:

tools/airgap_linter.py
kaiwalyajoshi: @takirala can you review this?

docs/generate.sh
vibhujain: @kaiwalyajoshi This is because when we reach to [this link](https://central.maven.org/maven2/io/swagger/swagger-codegen-cli/) while using https, it shows us a message to use https://repo1.maven.org. The contents of the pages/directory is same at the replaced host.
Attaching screenshot for the same:
Screenshot from 2019-06-11 14-36-27

tools/airgap_linter.py
```diff
@@ -26,7 +26,7 @@ def readlines_if_text_file(filename):
 def extract_uris(file_name):
     lines = readlines_if_text_file(file_name)
 
-    matcher = re.compile(r".*https?:\/\/([^\?\s]*)", re.IGNORECASE)
+    matcher = re.compile(r".*http?:\/\/([^\?\s]*)", re.IGNORECASE)
```
takirala: Just fyi, this may break downstream packages (edgelb, dss, k8 etc who use our tooling), but I can't think of a reason that breakage is not easily fixable.

tools/airgap_linter.py
takirala: > In case of internal Mesos DNS URLs, please mention http links as URI under resources.json

what do you mean? This will not work. All the uris in resource.json needs to be "downloadable" without a running DC/OS cluster.

tools/airgap_linter.py
vibhujain: I agree to this, but this was the purpose of the whole ticket to remove all non-HTTPS assets. I hope that this will us tackle those resources well.

tools/airgap_linter.py
vibhujain: @takirala I may have interpreted it in a different way, but what can we do here to make it sensible? Should we remove this altogether or suggest user to "mention external resources URLs in resources.json"

vibhujain:
After checking again for sub-directories, I have found no more such non-https references. @samvantran Let's make the decision over changes requested and then I hope that we can go ahead with merging the changes.

tools/airgap_linter.py
takirala: Wdyt of following rephrasing ?
```suggestion
            "Airgap check FAILED. The framework package definition contains references to non-https URL resources and/or there are invalid URL references in the package definition. Please consider replacing these resources to their HTTPS equivalent. In case of DC/OS internal URLs, please move them to service yaml definitions (which will be populated with FQN on the fly). This framework will NOT work in an air-gapped clusters without fixing the detected issues."
```

docs/generate.sh
vibhujain: @kaiwalyajoshi marking this as resolved if it looks fine by your side.

tools/airgap_linter.py
vibhujain: @takirala I agree with your changes and they make sense now! Thank you. I'm going ahead and committing these changes.

PR #3096, opened 2019-06-10
[DCOS-54802] Add fixes for proper RLIMIT_INFINITY support.

kaiwalyajoshi:
Fix previous bugs with setting `RLIMIT_INFINITY` which was incorrectly set.

JIRA: DCOS-54802

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
@@ -673,7 +673,13 @@ public class PodInfoBuilder {
           .setType(rLimit.getEnum());
 
       // RLimit itself validates that both or neither of these are present.
-      if (soft.isPresent() && hard.isPresent()) {
+      if ((soft.isPresent() && hard.isPresent()) && (soft.get() != RLimitSpec.RLIMIT_INFINITY &&
+          hard.get() != RLimitSpec.RLIMIT_INFINITY))
+      {
+        // If RLIMIT_INFINITY is desired, the RLimitInfo Protobuf exists but both
+        // the hard and soft values are unset.
+        //http://mesos.apache.org/api/latest/java/org/apache/mesos/Protos.RLimitInfo.RLimit.Builder.html#setHard-long-
+
         rLimitsBuilder.setSoft(soft.get()).setHard(hard.get());
       }
       rLimitInfoBuilder.addRlimits(rLimitsBuilder);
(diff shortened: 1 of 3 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/RLimitSpec.java
```diff
@@ -79,6 +85,16 @@ private void validate() throws InvalidRLimitException {
     if (soft != null && soft > hard) {
       throw new InvalidRLimitException("soft rlimit must be less than or equal to the hard rlimit");
     }
+
+    if ((soft != null && soft < RLIMIT_INFINITY) || (hard != null && hard < RLIMIT_INFINITY)) {
+      throw new InvalidRLimitException("soft and hard rlimits must be positive with the exception"
+          + " of " + RLIMIT_INFINITY + " which represents unlimited.");
+    }
+
+    if ((soft != null && soft == RLIMIT_INFINITY) ^ (hard != null && hard == RLIMIT_INFINITY)) {
```
mpereira: Didn't know about `^` as an XOR, nice!

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -673,7 +673,13 @@ private static String getConfigTemplateDownloadPath(ConfigFileSpec config) {
           .setType(rLimit.getEnum());
 
       // RLimit itself validates that both or neither of these are present.
-      if (soft.isPresent() && hard.isPresent()) {
+      if ((soft.isPresent() && hard.isPresent()) && (soft.get() != RLimitSpec.RLIMIT_INFINITY &&
+          hard.get() != RLimitSpec.RLIMIT_INFINITY))
```
mpereira: Does it make sense to add a unit test that exercises this condition as well?

mpereira: Also, should we log something in the case where only one of `soft` or `hard` are set to `RLIMIT_INFINITY`?

kaiwalyajoshi: We currently throw an exception in the event that only one is exclusively set: 

https://github.com/mesosphere/dcos-commons/blob/709a827ee07c0aaf01cfc54c3615196ff40aaf30/sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/RLimitSpec.java#L95

kaiwalyajoshi: There's this existing test case which excercises setting normal RLimits:
https://github.com/mesosphere/dcos-commons/blob/709a827ee07c0aaf01cfc54c3615196ff40aaf30/sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/RLimitSpecTest.java#L8

PR #2884, opened 2019-01-15
[DCOS-40703] Automate pod replacement when a node is decomissioned

kvish:

PR #3100, opened 2019-06-11
Add PARTITION_AWARE task-statues.

kaiwalyajoshi:
Add support for handling of addtional task-statuses enabled by the PARTITION_AWARE capability added in DCOS-40703 (#2884)

JIRA: DCOS-53490

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
@@ -181,6 +181,13 @@ public class DeploymentStep extends AbstractStep {
       case TASK_KILLED:
       case TASK_KILLING:
       case TASK_LOST:
+      //Tasks statuses relating to PARTITION_AWARE capability.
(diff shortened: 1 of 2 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/DeploymentStep.java
```diff
@@ -181,6 +181,13 @@ public synchronized void update(Protos.TaskStatus status) {
       case TASK_KILLED:
       case TASK_KILLING:
       case TASK_LOST:
+      //Tasks statuses relating to PARTITION_AWARE capability.
+      //Currently these Task statuses are handled in the same
+      //way as TASK_LOST.
+      case TASK_GONE:
+      case TASK_DROPPED:
+      case TASK_UNREACHABLE:
+      case TASK_GONE_BY_OPERATOR:
```
kvish: Out of scope for this PR, but TASK_GONE_BY_OPERATOR, and other future states that automate pod replace should probably have some new status outside of pending to distinguish when a task was explicitly drained.

PR #3102, opened 2019-06-12
SDK Quota Support

kaiwalyajoshi:
This feature requires support for `enforceRole` on Marathon groups found in Marathon v1.9.73 and Mesos v1.9.0 available starting DC/OS 2.0.
By default Marathon does not set `enforceRole=true` on group creation, and existing semantics are maintained.

#### Deploy new service in a group with quota enabled
Hello-World is used in the example below but this is applicable to any SDK based service.
To create a service named `/dev/hello-world` in group `dev` with quota consumed from role `dev`
1. Create a group with enforceRole:
```
cat >create-group.json<<EOF
{
    "id":"/dev",
    "enforceRole":true
}
EOF
```
2. Create Marathon group:
```
dcos marathon group add create-group.json
```
3. Populate the service options:
```
cat >hello-world-dev-options.json<<EOF
{
    "service":{
        "name":"/dev/hello-world"
    }
}
EOF
```
4. Install service.
```
dcos package install hello-world --options=namespace-options-foo-enforce-role.json --yes
```
5. Ensure SDK scheduler and pods have been launched under the `dev` role via Mesos UI.

#### Migrate an existing deployed service to use quota support
To upgrade an existing service to a new version of the SDK with quota support, use the following procedure.
We will use Hello-World again pre-installed in group `foo` in the example below but this is applicable to any SDK based service.

1. Create a file with the current service-name and the following additional options:
```
cat >hello-world-foo-options.json<<EOF
{
    "service":{
        "name":"/foo/hello-world",
        "role": "foo",
        "enable_role_migration": true
    }
}
EOF
```
- `role`: Specifies the quota enforced role we're migrating towards, which is `foo` in this example.
- `enable_role_migration`: Notifies the scheduler that its pods will be migrated between legacy and quota enforced roles. The scheduler subscribes with *both roles* when this flag is set.

2. Update the scheduler to use the quota enforced role.
```
dcos hello-world --name="/foo/hello-world" update start --options=hello-world-foo-options.json.json
```
3. At this point the scheduler will be upgraded and will use quota from the `foo` role. The deployed pods will be unaffected and will use their previous roles.
4. Issue `pod replace` commands to migrate all the pods in the service to the quota enforced role.
```
dcos hello-world --name="/foo/hello-world" pod replace hello-0
```
The `hello-0` pod will be migrated to consume quota from `foo`
5. Create a file with the current service-name and the following options to signal the end of the migration:
```
cat >hello-world-foo-disable-migration.json<<EOF
{
    "service":{
        "name":"/foo/hello-world",
        "role": "foo",
        "enable_role_migration": false
    }
}
EOF
```
6. Update the scheduler to stop subscribing to the legacy role.
```
dcos hello-world --name="/foo/hello-world" update start --options=hello-world-foo-disable-migration.json.json
```
At this point, the scheduler and all the previous running pods have been migrated to the quota enforced role.

#### Strict Mode Clusters

For strict mode clusters, additional role permissions are required and must be setup before deploying the service.

1. New service in a group with `enforceRole=true`
Example:
-  New service with name `/dev/hello-world` will need permissions to the `dev` role
```
dcos security org users grant <service-account> dcos:mesos:master:reservation:role:dev create
```
2. Migrating an existing service to a quota enforced role.
Example:
-  Existing service with name `/foo/hello-world` will need permissions to *both* the `foo` and `foo__hello-world-role` roles
```
dcos security org users grant <service-account> dcos:mesos:master:reservation:role:foo create
dcos security org users grant <service-account> dcos:mesos:master:reservation:role:foo__hello-world-role create
```
#### Pod Pre-Reserved Roles
For pods which specify pre-reserved roles (eg `slave_public`), the scheduler will issue a hierarhical role depending on the value of `role`. 

Example:
- Pod Pre-Reserved Role: `slave_public` and `role=slave_public`. These permissions are required:
```
dcos security org users grant <service-account> dcos:mesos:master:reservation:role:slave_public/dev__hello-world-role create
```
- Pod Pre-Reserved Role: `slave_public` and `role=dev`. These permissions are required:
```
dcos security org users grant <service-account> dcos:mesos:master:reservation:role:slave_public/dev create
```
When performing migration between legacy to enforced group roles via `enable_role_migration`, both permissions above will be required.

#### Downgrading to and older non-quota aware version of the scheduler
This section details the procedures to downgrade from a quota enforced role to a shipped non-quota enforced release.
The process is the same as [migrating an existing service to the quoted role](#migrate-an-existing-deployed-service-to-use-quota-support)
The key difference is that `role` should be `slave_public` to indicate migration towards the legacy roles.
```
cat >hello-world-foo-downgrade.json<<EOF
{
    "service":{
        "name":"/foo/hello-world",
        "role": "slave_public",
        "enable_role_migration": true
    }
}
EOF
```
The remaining scheduler update and `pod-replace` operations must be issued to move the scheduler and pods into the legacy roles.
Once all the pods have been migrated, the scervice can be downgraded to an earlier release which isn't quota aware.

JIRA: DCOS-54278


testing/sdk_upgrade.py
```diff
@@ -215,8 +215,14 @@ def _wait_for_deployment(
         log.info("No config change detected. Tasks should not be restarted")
         sdk_tasks.check_tasks_not_updated(service_name, "", task_ids)
     else:
-        log.info("Checking that all tasks have restarted")
-        sdk_tasks.check_tasks_updated(service_name, "", task_ids)
+        deployment_is_complete = sdk_plan.get_deployment_plan(service_name)["status"] == "COMPLETE"
+        recovery_is_complete = sdk_plan.get_recovery_plan(service_name)["status"] == "COMPLETE"
+
```
kvish: +1 also seeing this

testing/sdk_upgrade.py
```diff
@@ -215,8 +215,14 @@ def _wait_for_deployment(
...
+        # A difference between configs might be trivial that the scheduler doesn't consider the deployment
```
kvish: this may still be an issue. if we have tasks hello-0 world-0 and world-1 and we get an update that only affects the hello pod, only hello-0 will be relaunched, while this code would expect world-0 and world-1 to be relaunched as well leading to a timeout

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/DefaultConfigValidators.java
```diff
@@ -31,4 +31,17 @@ private DefaultConfigValidators() {
...
         new ServiceNameCannotBreakDNS());
   }
+
+  public static Collection<ConfigValidator<ServiceSpec>> getRoleValidators(boolean hasRoleChanged,
+      boolean hasCompletedDeployment)
+  {
+    //Conditionally include addtional validators here for role related changes.
+    if (hasRoleChanged && hasCompletedDeployment) {
+      return Collections.emptyList();
+    } else if (hasRoleChanged && !hasCompletedDeployment) {
```
takirala:
```suggestion
    } else if (hasRoleChanged) {
      return Collections.singletonList(new ServiceRoleCannotChangeOnIncompleteDeployment());
    } else {
      return Collections.singletonList(new TaskVolumesCannotChange());
    }
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/DefaultConfigValidators.java
```diff
@@ -31,4 +31,17 @@ private DefaultConfigValidators() {
         new RegionCannotChange(),
         new ServiceNameCannotBreakDNS());
   }
+
+  public static Collection<ConfigValidator<ServiceSpec>> getRoleValidators(boolean hasRoleChanged,
+      boolean hasCompletedDeployment)
+  {
+    //Conditionally include addtional validators here for role related changes.
```
takirala:
```suggestion
    // Conditionally include additional validators here for role related changes.
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ServiceRoleCannotChangeOnIncompleteDeployment.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Configuration validator which halts the schdeduler due to a role change on an incoming
+ * {@link com.mesosphere.sdk.specification.ServiceSpec} where the previous deployment wasn't complete.
+ * <p>
+ * This is to prevent the scheduler from continuing from previous incomplete deployments on role changes.
```
takirala:
```suggestion
 * This is to prevent the scheduler from continuing previous incomplete deployments on role changes.
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkConfig.java
```diff
@@ -132,6 +144,20 @@ public static FrameworkConfig fromEnvStore(EnvStore envStore) {
...
   }
 
+  /**
+   * Returns the configured Mesos role to use for running the service.
+   * Priority is given to the role specified by MESOS_ALLOCATION_ROLE environment variable set by Mesos.
+   * If no such environment-variable is found, or MARATHON_APP_ENFORCE_GROUP_ROLE is false we revert to the legacy
+   * version which uses the service name. Leading slashes are removed and intermediate slashes are escaped.
+   */
+  private static String getServiceRole(String frameworkName, Optional<String> serviceNamespace) {
+    if (serviceNamespace.isPresent()) {
```
takirala: (nit) simplify :
```java
    return serviceNamespace.map(SchedulerUtils::withEscapedSlashes).orElseGet(() -> getServiceRole(frameworkName));
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkConfig.java
```diff
@@ -96,10 +97,11 @@ public static FrameworkConfig fromRawServiceSpec(RawServiceSpec rawServiceSpec)
    * Creates a {@link FrameworkConfig} instance based off the contents of the provided
    * {@link ServiceSpec}.
    */
-  public static FrameworkConfig fromServiceSpec(ServiceSpec serviceSpec) {
+  public static FrameworkConfig fromServiceSpec(ServiceSpec serviceSpec, Optional<String> serviceNamespace) {
+    String serviceRole = getServiceRole(serviceSpec.getName(), serviceNamespace);
     return new FrameworkConfig(
         serviceSpec.getName(),
-        serviceSpec.getRole(),
+        serviceRole,
```
takirala: i think `serviceRole` needs to be used in few lines below (L108) as well.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkConfig.java
```diff
@@ -117,9 +119,19 @@ public static FrameworkConfig fromServiceSpec(ServiceSpec serviceSpec) {
...
     String frameworkName = envStore.getRequired("FRAMEWORK_NAME");
+
+    // We can only use MESOS_ALLOCATION_ROLE if MARATHON_APP_ENFORCE_GROUP_ROLE is true..
+    Optional<String> serviceNamespace;
+    boolean enforceRole = envStore.getOptionalBoolean("MARATHON_APP_ENFORCE_GROUP_ROLE", false);
+    if (enforceRole) {
+      serviceNamespace = Optional.ofNullable(envStore.getOptional("MESOS_ALLOCATION_ROLE", null));
+    } else {
+      serviceNamespace = Optional.empty();
+    }
```
takirala: (nit) simplify:
```java
    Optional<String> serviceNamespace = envStore.getOptionalBoolean("MARATHON_APP_ENFORCE_GROUP_ROLE", false) ?
      Optional.ofNullable(envStore.getOptional("MESOS_ALLOCATION_ROLE", null)) : Optional.empty();
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkConfig.java
```diff
@@ -281,4 +308,13 @@ public String getWebUrl() {
   public String toString() {
     return ReflectionToStringBuilder.toString(this);
   }
+
+  /**
+   * Returns the non-namespaced role used in earlier versions.
+   * Can be compared with {@code getRole} to determine which role
```
takirala:
```suggestion
   * Can be compared with {@link #getRole()} to determine which role
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkConfig.java
takirala: Let's merge this method with the ` private static String getServiceRole(String frameworkName)` (https://github.com/mesosphere/dcos-commons/pull/3102/files#diff-c06ef9fa9ce86e5f8888d227fa91d0cfR177) . Even the javadoc needs to be merged/rephrased. Both are `private static` so it should be straight forward.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -277,6 +277,16 @@
    */
   private static final AtomicBoolean PRINTED_BUILD_INFO = new AtomicBoolean(false);
 
+  /**
+   * When defined, this is the role to use when subscribing to Mesos and for recovering or adding footprint.
+   */
+  private static final String MESOS_ALLOCATION_ROLE = "MESOS_ALLOCATION_ROLE";
+
+  /**
+   * When set to true, the MESOS_ALLOCATION_ROLE value should be used to subscribe and recover or add footprint.
```
takirala:
```suggestion
   * When set to true, the {@link #MESOS_ALLOCATION_ROLE} value should be used to subscribe and recover or add footprint.
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -386,6 +396,15 @@ public String getDcosSpace() {
     return Optional.ofNullable(envStore.getOptional(SERVICE_REGION_ENV, null));
   }
 
+  public Optional<String> getServiceNamespace() {
```
takirala: This is duplicated in https://github.com/mesosphere/dcos-commons/pull/3102/files#diff-c06ef9fa9ce86e5f8888d227fa91d0cfR124

kaiwalyajoshi: Yes, this is due to `FrameworkConfig` requring access to `EnvStore` to pull all the env-vars, and passing in a `EnvStore` wasn't possible at all initialization paths in the scheduler and across the services.

`SchedulerConfig` was easily available however and had easy access to `EnvStore`, as such the context for this is split across these two classes. So its not duplication,even though it looks that way.

The responsibility is split and there is a coupling required here, but so far I haven't found a way to get this done. Happy to hear ways of fixing this as I'm not a fan of this approach either, but havnen't found a better low-impact alternative.

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/DefaultConfigValidators.java
```diff
@@ -31,4 +31,17 @@ private DefaultConfigValidators() {
...
   }
+
+  public static Collection<ConfigValidator<ServiceSpec>> getRoleValidators(boolean hasRoleChanged,
+      boolean hasCompletedDeployment)
+  {
+    // Conditionally include additional validators here for role related changes.
+    if (hasRoleChanged && hasCompletedDeployment) {
+      return Collections.emptyList();
+    } else if (hasRoleChanged) {
+      return Arrays.asList(new ServiceRoleCannotChangeOnIncompleteDeployment());
```
takirala:
```suggestion
      return Collections.singletonList(new ServiceRoleCannotChangeOnIncompleteDeployment());
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/DefaultConfigValidators.java
```diff
@@ -31,4 +31,17 @@ private DefaultConfigValidators() {
...
+    } else {
+      return Arrays.asList(new TaskVolumesCannotChange());
```
takirala:
```suggestion
      return Collections.singletonList(new TaskVolumesCannotChange());
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/ServiceRoleCannotChangeOnIncompleteDeployment.java
```diff
@@ -0,0 +1,42 @@
...
+import com.mesosphere.sdk.specification.ServiceSpec;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Configuration validator which halts the schdeduler due to a role change on an incoming
```
takirala:
```suggestion
 * Configuration validator which halts the scheduler due to a role change on an incoming
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkConfig.java
```diff
@@ -281,4 +280,26 @@ public String getWebUrl() {
...
+   */
+  public String getNonNamespacedRole() {
+    return SchedulerUtils.withEscapedSlashes(frameworkName) + DEFAULT_ROLE_SUFFIX;
+  }
+
+  /**
+   * Returns the namespaced role derived from the framework name.
+   */
+  public Optional<String> getNamespacedRole() {
+    String[] marathonGroups = getNonNamespacedRole().split("__");
```
takirala: it would be nice to get `__` somehow from `SchedulerUtils` 
One way is to make `SLASH_REPLACEMENT` in that class public and use it here, or create a new `public static` method. anyway this is a nit.

sdk/scheduler/src/main/java/com/mesosphere/sdk/framework/FrameworkRunner.java
```diff
@@ -235,4 +236,21 @@ private void runSkeletonScheduler() {
...
+  private Set<String> getResourceRoles() {
+    // Collection of all the roles we might subscribe with.
+    HashSet<String> resourceRoles = new HashSet<String>();
+    // Always Add the main role we're going to subscribe with.
+    resourceRoles.add(frameworkConfig.getRole());
+    // Add all pre-reserved roles if they exist.
+    if (!frameworkConfig.getPreReservedRoles().isEmpty()) {
+      resourceRoles.addAll(frameworkConfig.getPreReservedRoles());
+    }
+    // Add both preferred and legacy roles, if we're in migration mode.
```
takirala: Thanks for all the explanatory comments!

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -277,6 +277,29 @@
...
+
+  /**
+   * Default role that the scheduler gets created under via Marathon without any other options.
+   */
+  private static final String DEFAULT_SCHEDULER_ROLE = "slave_public";
+
+  /*
+   * Determines if the scheduler should subscribe with both legacy role and new quota roles.
+   * Setting this to true is harmless and might incur a slight startup penalty as Mesos offers
+   * with this role get declined if the pod is looking for role set by FrameworkConfig.getRole()
```
takirala: change to javadoc and use `{@link com.mesosphere.sdk.framework.FrameworkConfig#getRole()}`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -277,6 +277,29 @@
    */
   private static final AtomicBoolean PRINTED_BUILD_INFO = new AtomicBoolean(false);
 
+  /**
+   * When defined, this is the role to use when subscribing to Mesos and for recovering or adding footprint.
+   */
+  private static final String MESOS_ALLOCATION_ROLE_ENV = "MESOS_ALLOCATION_ROLE";
+
+  /**
+   * When set to true, the {@link #MESOS_ALLOCATION_ROLE} value should be used to subscribe and recover or
```
takirala:
```suggestion
   * When set to true, the {@link #MESOS_ALLOCATION_ROLE_ENV} value should be used to subscribe and recover or
```

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
```diff
@@ -386,6 +409,34 @@ public String getDcosSpace() {
...
+      // service name is `dev/foo-service` and is under the group `dev`.
+      // Marathon validation resets the role to the default of `slave_public`.
+      // We cannot launch all pods under `slave_public` as it is bad form.
+      if (preferredServiceRole != null && !DEFAULT_SCHEDULER_ROLE.equals(preferredServiceRole)) {
+        // Here enforceRole is not set and we have valid preferred service role, we can use this.
+        return Optional.of(preferredServiceRole);
+      } else {
+        // Here enforceRole is not set and we have an invalid preferred service role we can't use.
+        // Revert to legacy semantics.
+        return Optional.empty();
```
takirala: Should we revert to legacy semantics or should we fail loudly ?! 

I think of it as a reserved keyword in mustache template and thus if a user mentions `role` field (may be with some other intentions) then it should fail. User should not specify invalid roles

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/NamedVIPSpec.java
```diff
@@ -75,6 +75,23 @@ public static Builder newBuilder() {
     return new Builder();
   }
 
+  public static Builder newBuilder(NamedVIPSpec copy) {
```
takirala: I think we can get rid of this if this is not used anywhere.

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/PortSpec.java
```diff
@@ -68,6 +68,20 @@ public static Builder newBuilder() {
     return new Builder();
   }
 
+  public static Builder newBuilder(PortSpec copy) {
```
takirala: I think we can get rid of this if this is not used anywhere.

sdk/scheduler/src/test/java/com/mesosphere/sdk/framework/FrameworkRunnerTest.java
```diff
@@ -145,14 +145,247 @@ public void testExhaustiveFrameworkInfo() {
...
+    /*
+     *  Matrix for verification of the following variables.
+     *  {Pre-reserved-role, Service-role, Enforce-role, Migration-Mode}
+     *  Pre-reserved-role: Does the service have static reservations via pre-reserved-role
+     *  Service-role: Is the service using a the new quota-role (T) or legacy-role (F)
+     *  Enforce-role: Is the service in a group where the role is enforced.
+     *  Migration-Mode: Is the service set to migrate quotas. 
+     *   
+     *  The following combinations are invalid:
+     *  XFTX - Here pre-reserved-role and migration-mode are irrelevant. You cannot have
```
takirala: Neat!

sdk/scheduler/src/test/java/com/mesosphere/sdk/framework/FrameworkRunnerTest.java
```diff
@@ -145,14 +145,247 @@ public void testExhaustiveFrameworkInfo() {
...
+    
     private static Protos.FrameworkInfo.Capability getCapability(Protos.FrameworkInfo.Capability.Type type) {
         return Protos.FrameworkInfo.Capability.newBuilder().setType(type).build();
     }
 
     @SuppressWarnings("deprecation")
     private static void checkRole(Optional<String> expectedRole, Protos.FrameworkInfo info) {
         if (expectedRole.isPresent()) {
-            Assert.assertEquals(info.getRole(), expectedRole.get());
+            Assert.assertEquals(expectedRole.get(), info.getRole());
```
takirala: 👍 for cleanup

testing/sdk_security.py
```diff
@@ -145,10 +145,7 @@ def get_default_permissions(
 
 
 def grant_permissions(
-    linux_user: str,
-    role_name: str,
-    service_account_name: str,
-    permissions: List[Dict[str, str]],
+    linux_user: str, role_name: str, service_account_name: str, permissions: List[Dict[str, str]]
```
takirala: i mean some linters DO complain if there is no trailing comma - but no strong feelings :)

testing/sdk_security.py
```diff
@@ -247,15 +242,34 @@ def delete_secret(secret: str) -> None:
     sdk_cmd.run_cli("security secrets delete {}".format(secret))
 
 
-def _get_service_role(service_name: str) -> List[str]:
+def _get_service_role(service_name: str):
```
takirala: why not retain the return type ?

testing/sdk_utils.py
```diff
@@ -243,3 +242,61 @@ def merge_dictionaries(dict1: Dict, dict2: Optional[Dict]) -> Dict:
...
+    return current_service_roles
+
+
+def filter_role_from_config(unfiltered_config: Optional[Dict[str, Any]]) -> None:
+    if unfiltered_config is None:
+        return
+
+    # Remove any role related fields as role changes
+    # are allowed to happen across config targets
+    # to accomodate for quota migration.
```
takirala:
```suggestion
    # to accommodate for quota migration.
```

testing/sdk_utils.py
```diff
@@ -243,3 +242,61 @@ def merge_dictionaries(dict1: Dict, dict2: Optional[Dict]) -> Dict:
...
+    # Find our service.
+    for service in mesos_state["frameworks"]:
+        if service["name"] == service_name:
+            service_state = service
+
+    # Create a map of tasks to roles.
+    if service_state:
+
+        if "roles" in service_state:
+            # MUTI_ROLE
```
takirala:
```suggestion
            # MULTI_ROLE
```

testing/sdk_utils.py
```diff
@@ -243,3 +242,61 @@ def merge_dictionaries(dict1: Dict, dict2: Optional[Dict]) -> Dict:
...
+            current_service_roles["framework-roles"] = service_state["roles"]
```
takirala: So if this is a multi role, what would be the result of `service_state["role"]` - is it a non existent key OR defaulted to an empty list ?

testing/sdk_utils.py
```diff
@@ -243,3 +242,61 @@ def merge_dictionaries(dict1: Dict, dict2: Optional[Dict]) -> Dict:
...
+def filter_role_from_config(unfiltered_config: Optional[Dict[str, Any]]) -> None:
+    if unfiltered_config is None:
+        return
+
+    # Remove any role related fields as role changes
+    # are allowed to happen across config targets
+    # to accomodate for quota migration.
+
+    # Inspired by:
+    # https://stackoverflow.com/a/50444005/10840685
```
takirala: ah, TIL! 

i remembered the scala collection api's `flatten` method which can be used to accomplish this in a much simpler way.

tools/universe/package_builder.py
```diff
@@ -18,7 +18,7 @@
 
 _jre_url = "https://downloads.mesosphere.com/java/openjdk-jre-8u212b03-hotspot-linux-x64.tar.gz"
 _libmesos_bundle_url = (
-    "https://downloads.mesosphere.com/libmesos-bundle/libmesos-bundle-1.12.0.tar.gz"
+    "https://downloads.mesosphere.com/libmesos-bundle/libmesos-bundle-1.14-alpha.tar.gz"
```
takirala: 🚀

frameworks/helloworld/tests/test_quota_deployment.py
```diff
@@ -0,0 +1,340 @@
...
+log = logging.getLogger(__name__)
+
+ENFORCED_ROLE = "quota"
+SERVICE_NAME = "/{}/hello-world".format(ENFORCED_ROLE)
+LEGACY_ROLE = "{}-role".format(SERVICE_NAME.strip("/").replace("/", "__"))
+
+RECOVERY_TIMEOUT_SECONDS = 20 * 60
+
+# This test does a fresh install of a new service and verifies that role
+# creation was correctly done.
```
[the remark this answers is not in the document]
kaiwalyajoshi: nope, this was intentional.

frameworks/helloworld/tests/test_quota_upgrade.py
```diff
@@ -0,0 +1,247 @@
...
+@pytest.mark.dependency(depends=["test_disable_legacy_role_post_update"])
+def test_more_pods_disable_legacy_role_post_update():
+    # Ensure we can scale out more still with legacy role disabled.
+
+    # Add new pods to service which should be launched with the new role.
+    marathon_config = sdk_marathon.get_config(SERVICE_NAME)
+
+    # Add an extra pod to each.
+    marathon_config["env"]["HELLO_COUNT"] = "3"
+    marathon_config["env"]["WORLD_COUNT"] = "4"
```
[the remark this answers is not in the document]
kaiwalyajoshi: Both `hello` and `world` pods have the placement constraint `"[[\"hostname\", \"UNIQUE\"]]"`, in our SI we spin-up a cluster with 5 agents, so this fits into the normal test configuration.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/SchedulerConfig.java
kaiwalyajoshi: It's Marathon that does the validation for us here, essentially either the specified role is is correct or we revert to `slave_public`. So for now I don't think there's a need to fail loudly. If at some point in the future, `slave_public` is deprecated (or something along those lines), we'll fail loudly.

testing/sdk_security.py
kaiwalyajoshi: I think `black` was ok with this, the PR code checks will fail anyways if its not conformant.

testing/sdk_utils.py
kaiwalyajoshi: `service_state["role"] = "*"`

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/NamedVIPSpec.java
kaiwalyajoshi: I wanted to leave them in for completeness sake :man_shrugging:

sdk/scheduler/src/main/java/com/mesosphere/sdk/specification/PortSpec.java
kaiwalyajoshi: Same as above, wanted to leave them in for compleness sake.

frozen_requirements.txt
```diff
@@ -98,6 +98,7 @@ PyJWT==1.7.1
 pylint==2.3.1
 PyNaCl==1.3.0
 pytest==4.1.0
+pytest-dependency==0.4.0
```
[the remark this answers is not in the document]
kaiwalyajoshi: Don't think so, pytest harness gets all of its deps from `frozen_requirements.txt`

PR #3105, opened 2019-06-13
Add config validator for Seccomp

kvish:


sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/PodSpecsCannotUseUnsupportedFeatures.java
```diff
@@ -162,6 +165,17 @@ private static boolean podRequestsProfileMountVolumes(PodSpec podSpec) {
             "This DC/OS cluster does not support profile mount volumes"
         ));
       }
+
+      if (!supportsSeccomp && ((podSpec.getSeccompUnconfined() != null
+              && podSpec.getSeccompUnconfined()) ||
```
takirala: Do we really need `&& podSpec.getSeccompUnconfined())` ?

won't this suffice :
```java
if (!supportsSeccomp && (podSpec.getSeccompUnconfined() != null ||
              podSpec.getSeccompProfileName().isPresent()))
      {
```

kvish: there is the case that seccomp unconfined can be null or the case that we set it to false based on the HDFS issue, this covers both cases

PR #3111, opened 2019-06-19
[HOTFIX][TOOLS] dcos-launch is supposed to be duplicated there.

mpereira:
The first parameter to `/venvs.sh` is the name of the virtual environment, and the rest are the command itself to be run in that virtual environment.

### DIFF
--- a/tools/ci/launch_cluster.sh
+++ b/tools/ci/launch_cluster.sh
@@ -17,7 +17,7 @@ while [ x"${LAUNCH_SUCCESS}" == x"False" ]; do
     else
         set -e
     fi
-    /venvs/wrap.sh dcos-launch wait --info-path=${CLUSTER_INFO_FILE} 2>&1 | tee dcos-launch-wait-output.stdout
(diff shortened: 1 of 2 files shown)


tools/ci/launch_cluster.sh
```diff
@@ -17,7 +17,7 @@ while [ x"${LAUNCH_SUCCESS}" == x"False" ]; do
     else
         set -e
     fi
-    /venvs/wrap.sh dcos-launch wait --info-path=${CLUSTER_INFO_FILE} 2>&1 | tee dcos-launch-wait-output.stdout
+    /venvs/wrap.sh dcos-launch dcos-launch wait --info-path=${CLUSTER_INFO_FILE} 2>&1 | tee dcos-launch-wait-output.stdout
```
wavesoft: Please add a comment here, explaining that the first parameter is the environment name and *NOT* the command :)

mpereira: https://github.com/mesosphere/dcos-commons/pull/3111/commits/63d62c9116066e5a3b18512fd2b211befd32a9a8

mpereira:
Merging without checks because integration test checks are broken without this.

PR #3114, opened 2019-06-21
[DCOS-51362] Prevent pytest process from crashing during diagnostics fetching.

mpereira:


conftest.py
```diff
@@ -82,10 +83,28 @@ def pytest_runtest_makereport(item: pytest.Item, call):  # _pytest.runner.CallIn
...
+            # runs, instead of when the actual test runs. That manifests as an "INTERNALERROR>
+            # Failed: Timeout" that crashes the test process. That is undesirable because then we
+            # get no diagnostics artifacts which would allow us to investigate why the test timed
+            # out in the first place.
+            #
+            # Here, we retry the diagnostics-fetching method in case of pytest exceptions.
+            #
+            # Check DCOS-51362 for more context.
+            stop_max_attempt_number=3,
+            retry_on_exception=lambda e: isinstance(e, pytest.fail.Exception),
```
phrynchanka: Maybe it's better to add retrying on any exception? (f.e. BaseException)

mpereira: The reasoning of only retrying on `pytest.fail.Exception` is that in the case of different exception-raising-scenarios than the one described in the comment, we'd want to know what they are so that we could potentially actually fix them.

PR #3115, opened 2019-06-24
[CLEANUP] Delete unnecessary files

takirala:
I believe this was added mistakenly from #3098


RezviyBelorus:
Hm, I was sure I revert it and remove from commit

PR #3120, opened 2019-07-01
[DCOS-54275] Add backoff delay for failed/error tasks

takirala:
### Implementation details:
* Introduces a `BackOff` class that is responsible for storing all the delays for any applicable tasks.
* This is a singleton class that is instantiated to an `ExponentialBackoff` or a `DisabledBackoff` class based on the environment configuration. Sensible defaults are assumed but these are changeable by tweaking the environment. Singleton also makes it easier to access from anywhere (and we DO need to access this class from various places).
* `Deploy` (All yaml plans and `Recovery` plan extend this plan) consult the above class to 1) store delays, 2) advance delays and to 3) clear delays.
* `Recovery` plan consults the above class to calculate what is ready for recovery.
* When the plan is delayed (meaning all its phases are delayed - meaning all steps in each phase are delayed) then the scheduler assumes `IDLE` (suppress). 
* All the state is stored in memory - restarting scheduler process will clear all delays. Upgrades and downgrades should continue to work as is. 
* Similar to above, restarting a pod instance would clear its associated task delays.

### Alternatives considered

* `LaunchConstrainer` in recovery plan is somewhat similar and I guess we can remove that code. I tried extending that initially but it does not work in cases where the first ever launch was unsuccessful and we want to use `DELAYED` status for deploy plan.

### Testing
* Added an integration test that verified `deploy` and `recovery` plans can cause backoff and suppress the framework accordingly.
* Added unit tests for `DeploymentStep`, `TaskUtils` and `PlanUtils`.


sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/backoff/ExponentialBackOff.java
```diff
@@ -0,0 +1,101 @@
...
+    } else {
+      logger.info("Current delay for {} is {}", taskName, delay.getCurrentDelay().toString());
+      delay.advanceDelay();
+      logger.info("Advancing existing delay for {} to {}",
+              taskName, delay.getCurrentDelay().toString());
+    }
+    delays.put(taskName, delay);
+  }
+
+  public boolean clearDelay(Protos.TaskID taskId) {
```
kaiwalyajoshi: Can we invoke this from [PodQueries::restartPod](https://github.com/mesosphere/dcos-commons/blob/60fa2d80d75e63f2d934d66a915bb5ae4042196b/sdk/scheduler/src/main/java/com/mesosphere/sdk/http/queries/PodQueries.java#L280)? We can explicitly clear the backoff delay when the pods are restarted, which gives a nice way to reset backoff.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/AbstractStep.java
```diff
@@ -97,6 +97,8 @@ public boolean isInterrupted() {
   @Override
   public void restart() {
     logger.warn("Restarting step: '{} [{}]'", getName(), getId());
+    //TODO@kjoshi: Reset backoff status here when issued by command line.
+    //Implementation detail: Set backoff back to zero.
```
kaiwalyajoshi: Another option for where we can issue `Delay::clearDelay`

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/backoff/ExponentialBackOff.java
```diff
@@ -0,0 +1,101 @@
...
+    this.maxLaunchDelay = Duration.ofSeconds(maxLaunchDelaySeconds);
+    delays = new ConcurrentHashMap<>();
+  }
+
+  public static ExponentialBackOff getInstance() {
+    if (instance == null) {
+      synchronized (lock) {
+        if (instance == null) {
+          // SUPPRESS CHECKSTYLE MagicNumberCheck
+          instance = new ExponentialBackOff(1.5, 60, 300);
```
kaiwalyajoshi: We'll likely need to parametrize this down the line, this is fine for investigation.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/AbstractScheduler.java
```diff
@@ -135,25 +136,26 @@ public void registered(boolean reRegistered) {
   public ClientStatusResponse getClientStatus() {
     // Get the current work, and save it for the following offers() call.
     candidateSteps.clear();
+    //TODO@kjoshi: This is where our upcoming footprint is computed.
```
kaiwalyajoshi: We can purge all of these now.

frameworks/helloworld/universe/config.json
```diff
@@ -144,6 +144,11 @@
           "enum": [0, 1],
           "default": 1
         },
+        "enable_backoff": {
```
kaiwalyajoshi: I think we should extend this from a simple boolean flag into a subsection under `service` i.e
```
"task_failure_backoff": {
  "disabled" : true
  "backoff-factor": {...}
  "max-launch-delay": {...}
  "initial-backoff": {...}
}
```
In addition, we should switch the flag from `enabled` to `disabled` such that for now it will be opt-in but after the next n+2 releases, we can flip the `disabled` flag to default to false.

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/backoff/BackOff.java
```diff
@@ -0,0 +1,74 @@
...
+   *    {@link #FRAMEWORK_INITIAL_BACKOFF} and/or {@link #FRAMEWORK_MAX_LAUNCH_DELAY}.
+   *    If more than zero params are initialized, defaults are used for the uninitialized params.
+   *  * Absence of {@link #ENABLE_BACKOFF} and other environment params would be considered
+   *    as hint to disable backoff
+   */
+  public static BackOff getInstance() {
+    if (instance == null) {
+      synchronized (lock) {
+        EnvStore envStore = EnvStore.fromEnv();
+        if (envStore.isPresent(ENABLE_BACKOFF)
```
kaiwalyajoshi: Likely need to change this to be in sync with the options change mentioned earlier. (i.e use `disable` flag instead of `enable`).

sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/backoff/BackOff.java
```diff
@@ -0,0 +1,74 @@
...
+                || envStore.isPresent(FRAMEWORK_BACKOFF_FACTOR)
```
kaiwalyajoshi: I don't think we need to check for these three env vars, the latter call uses `envStore.getOptional*` which means its enough to rely on the enable/disable flag.

takirala:
Thanks for the review @kaiwalyajoshi and @kvish

PR #3122, opened 2019-07-04
[DCOS-56081] Make "service configuration JSON"-standardizing script more flexible.

mpereira:
This allows for the script to also standardize project specific sections. Right
now it only standardizes the order of the "service" section.

For example, I'm using it to standardize the order of the "elasticsearch"
section for the Elastic service. By providing a configuration file that looks
like the following:

`dcos_commons.json`:
```
{
  "standardize_config_json": {
    "sections": {
      "elasticsearch": {
        "head": [
          "custom_elasticsearch_yml",
          "health_user",
          "health_user_password",
          "plugins",
          "plugin_http_proxy_host",
          "plugin_http_proxy_port",
          "plugin_https_proxy_host",
          "plugin_https_proxy_port",
          "xpack_http_proxy_host",
          "xpack_http_proxy_port",
          "xpack_graph_enabled",
          "xpack_ml_enabled",
          "xpack_monitoring_collection_enabled",
          "xpack_monitoring_enabled",
          "xpack_security_audit_enabled",
          "xpack_security_enabled",
          "xpack_sql_enabled",
          "xpack_watcher_enabled"
        ],
        "tail": ["http_content_type_required", "xpack_enabled"]
      }
    }
  }
}
```

and calling the standardization script like:

```
./dcos-commons/tools/standardize_config_json.py frameworks/elastic/universe/config.json dcos_commons.json
```

the script will make sure that:
1. the first elements in the "elasticsearch" section are the ones in the "head" section
2. the last elements in the "elasticsearch" section are the ones in the "tail" section
3. the remaining elements (in the middle) will be sorted alphabetically

The "head" section is usually reserved for important settings, and the "tail"
section for deprecated settings.

### DIFF
--- a/tools/standardize_config_json.py
+++ b/tools/standardize_config_json.py
@@ -94,9 +94,13 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
     o = json.dumps(original, indent=2)
     c = json.dumps(new, indent=2)
 
-    diff = difflib.unified_diff(o.split("\n"), c.split("\n"))
+    diff = list(difflib.unified_diff(o.split("\n"), c.split("\n")))
 
-    LOG.info("\n".join(diff))
+    if len(diff) == 0:
+        LOG.info("No changes")
+    else:
+        LOG.info("Applied diff:")
+        LOG.info("\n".join(diff))
 
 
 def process(service_config_json_path: str, sdk_tools_config: dict):
--- a/tools/standardize_config_json.py
+++ b/tools/standardize_config_json.py
@@ -3,7 +3,7 @@
 usage = """Standardizes the ordering of sections in config.json files.
 
 Usage:
-$ ./standardize_config_json.py $path_to_config_json"""
+$ ./standardize_config_json.py $path_to_config_json $path_to_dcos_commons_json"""
 
 import collections
 import difflib
--- a/tools/standardize_config_json.py
+++ b/tools/standardize_config_json.py
@@ -123,10 +123,12 @@ def process(service_config_json_path: str, sdk_tools_config: dict):
 
     write_json_file(service_config_json_path, contents)
 
+    return 0
+
 
 def main(argv):
     parser = argparse.ArgumentParser(
-        description="Standardizes the ordering of sections in SDK service config.json files."
+        description="Standardizes the ordering of sections in SDK service JSON configuration files."
     )
 
     parser.add_argument(
@@ -134,7 +136,7 @@ def main(argv):
         type=str,
         required=True,
         default=None,
-        help="Path to the service configuration JSON file to be standardized",
+        help="Path to the SDK service configuration file to be standardized (e.g. config.json)",
     )
 
     parser.add_argument(
@@ -142,7 +144,7 @@ def main(argv):
         type=str,
         required=False,
         default=None,
-        help="Path to the SDK Tools configuration file",
+        help="Path to the SDK Tools configuration file (e.g. sdk-tools.json)",
     )
 
     args = parser.parse_args()
@@ -151,20 +153,26 @@ def main(argv):
     sdk_tools_config_path = args.sdk_tools_config
 
     if not os.path.isfile(service_config_json_path):
-        LOG.info("'%s' is not a file, was expecting a 'config.json' file", service_config_json_path)
+        LOG.info(
+            "'%s' is not a file, was expecting an SDK service configuration file",
+            service_config_json_path,
+        )
+        return 1
 
     if sdk_tools_config_path:
         if not os.path.isfile(sdk_tools_config_path):
             LOG.info(
-                "'%s' is not a file, was expecting a 'sdk-tools.json' file", sdk_tools_config_path
+                "'%s' is not a file, was expecting an SDK Tools configuration file",
+                sdk_tools_config_path,
             )
+            return 1
 
-        LOG.info("Parsing SDK tooling configuration from '%s'", sdk_tools_config_path)
+        LOG.info("Parsing SDK Tools configuration from '%s'", sdk_tools_config_path)
         sdk_tools_config = json.loads(read_file(sdk_tools_config_path))
     else:
         sdk_tools_config: dict = {}
 
-    process(service_config_json_path, sdk_tools_config.get("standardize_config_json", {}))
+    return process(service_config_json_path, sdk_tools_config.get("standardize_config_json", {}))
 
 
 if __name__ == "__main__":
--- a/tools/standardize_config_json.py
+++ b/tools/standardize_config_json.py
@@ -1,16 +1,16 @@
 #!/usr/bin/env python3
 
-usage = """Standardizes the ordering of sections in config.json files.
-
-Usage:
-$ ./standardize_config_json.py $path_to_config_json $path_to_dcos_commons_json"""
-
 import collections
 import difflib
 import json
 import logging
 import os.path
 import sys
+
+usage = """Standardizes the ordering of sections in config.json files.
+
+Usage:
+$ ./standardize_config_json.py $path_to_config_json $path_to_dcos_commons_json"""
 
 
 logging.basicConfig(
--- a/tools/standardize_config_json.py
+++ b/tools/standardize_config_json.py
@@ -103,14 +103,15 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
     LOG.info("\n".join(diff))
 
 
-def process(config_json_path: str, configuration: list):
+def process(config_json_path: str, configuration: dict):
     contents = read_json_file(config_json_path)
     original = read_json_file(config_json_path)
 
     reordered = reorder_service(contents["properties"]["service"]["properties"])
     contents["properties"]["service"]["properties"] = reordered
 
-    for section_name, head_and_tail in configuration["sections"].items():
+    sections = configuration.get("sections", {})
+    for section_name, head_and_tail in sections.items():
         head = head_and_tail["head"]
         tail = head_and_tail["tail"]
         properties = contents["properties"][section_name]["properties"]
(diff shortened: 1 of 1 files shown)


mpereira:
Head:
image

Tail:
image

tools/standardize_config_json.py
```diff
@@ -104,24 +103,57 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
+    write_json_file(config_json_path, contents)
 
 
 if __name__ == "__main__":
-    files = [
-        "frameworks/cassandra/universe/config.json",
-        "frameworks/hdfs/universe/config.json",
-        "frameworks/helloworld/universe/config.json",
-    ]
+    args = sys.argv[1:]
```
elezar: Do we not want to check the length of args before doing this? Also, since argparse is part of the standard library these days we could use that instead to avoid checks on arguments and specific handling of the help arguments and specific usage docs.

tools/standardize_config_json.py
```diff
@@ -104,24 +103,57 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
+        reordered = reorder(properties, head, tail, reorder_property)
+        contents["properties"][section_name]["properties"] = reordered
+
     print_diff(original, contents)
 
-    write_json_file(filename, contents)
+    write_json_file(config_json_path, contents)
 
 
 if __name__ == "__main__":
```
elezar: Nit: do we want to call a 'main' function here instead as this may make things easier to test or script against.

tools/standardize_config_json.py
```diff
@@ -104,24 +103,57 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
+        exit(0)
+
+    config_json_path = args[0]
+
+    if not os.path.isfile(config_json_path):
+        LOG.info("'%s' is not a file, was expecting a config.json file", config_json_path)
+
+    configuration_path = None
+    configuration: dict = {}
+    LOG.info("config_json_path: %s", config_json_path)
```
elezar: Nit: log this after it is actually read.

tools/standardize_config_json.py
```diff
@@ -104,24 +103,57 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
+
+    if len(args) == 2:
+        configuration_path = args[1]
+        LOG.info("configuration_path: %s", configuration_path)
+
+    if configuration_path:
+        if not os.path.isfile(configuration_path):
```
elezar: The two parameters are handled in similar ways. What about a "try_load" function? Also, should the following not log an error and terminate?

tools/standardize_config_json.py
```diff
@@ -104,24 +103,57 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
 
+    sections = configuration.get("sections", {})
+    for section_name, head_and_tail in sections.items():
+        head = head_and_tail["head"]
+        tail = head_and_tail["tail"]
+        properties = contents["properties"][section_name]["properties"]
+        reordered = reorder(properties, head, tail, reorder_property)
+        contents["properties"][section_name]["properties"] = reordered
+
     print_diff(original, contents)
```
elezar: Should this be optional?

tools/standardize_config_json.py
```diff
@@ -104,24 +103,57 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
+    original = read_json_file(config_json_path)
 
     reordered = reorder_service(contents["properties"]["service"]["properties"])
     contents["properties"]["service"]["properties"] = reordered
 
+    sections = configuration.get("sections", {})
+    for section_name, head_and_tail in sections.items():
+        head = head_and_tail["head"]
+        tail = head_and_tail["tail"]
+        properties = contents["properties"][section_name]["properties"]
```
elezar: This will fail if the config file gets out of sync and the file does not contain the specified section. Should we check and log an error?

tools/standardize_config_json.py
```diff
@@ -104,24 +103,57 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
+    contents = read_json_file(config_json_path)
+    original = read_json_file(config_json_path)
 
     reordered = reorder_service(contents["properties"]["service"]["properties"])
     contents["properties"]["service"]["properties"] = reordered
 
+    sections = configuration.get("sections", {})
+    for section_name, head_and_tail in sections.items():
+        head = head_and_tail["head"]
+        tail = head_and_tail["tail"]
```
elezar: Use get here and above instead.

tools/standardize_config_json.py
```diff
@@ -1,17 +1,16 @@
...
 import json
 import logging
+import os.path
 import sys
-import difflib
+
+usage = """Standardizes the ordering of sections in config.json files.
+
+Usage:
+$ ./standardize_config_json.py $path_to_config_json $path_to_dcos_commons_json"""
```
porridge: `...dcos_commons_json` does not sound like a good name, because it does not convey the intent. `...formatting_policy_json` perhaps?

mpereira: My idea with this file was having configuration for any dcos-commons tool (i.e. a "dcos-commons tooling" configuration file). If you notice the example configuration, there's a key specific for this script's configuration (`"standardize_config_json"`). If more scripts need to be configured in a way that doesn't lend itself well to command line arguments (like it's the case here) they could have their own sections.

Maybe we could rephrase the variable name in the usage output (and in the script itself) so that it's intent is clearer. Do you have suggestions?

porridge: Oh, I see. Perhaps we should have a config file per tool, then? This seems to be the pattern for most UNIX-like utilities, rather than having "the config file" with sections per tool...

mpereira: We could. What I had in mind was less UNIX-like and more [`pyproject.toml`](https://black.readthedocs.io/en/stable/pyproject_toml.html)-like, so as to avoid requiring services that use dcos-commons tooling as a _unit of dependency_ to have to have in their source control several configuration files.

porridge: :thinking: that makes sense if was not for the naming... `pyproject` is much more convenient than `dcos-commons-tools-consuming-project` :-/

porridge: `sdktools.json`?

mpereira: Sounds good to me 👍

tools/standardize_config_json.py
mpereira: The script should not terminate when not given a "SDK Tools" configuration file. In that case it standardizes just the `service` section.

tools/standardize_config_json.py
mpereira: Fixed in 072eb00997b0f6d39c184bcd8501dc19f34db4e9.

tools/standardize_config_json.py
mpereira: Fixed in 072eb00997b0f6d39c184bcd8501dc19f34db4e9.

tools/standardize_config_json.py
mpereira: Fixed in 072eb00997b0f6d39c184bcd8501dc19f34db4e9.

tools/standardize_config_json.py
mpereira: With 2fa054c69fc5da12528e5d47308819d78a56e74b this function will still be called unconditionally, but it will output `no changes` in case the diff was empty.

tools/standardize_config_json.py
mpereira: Fixed in a065501b359f31ed7741b29565283a6f234cea4e.

tools/standardize_config_json.py
mpereira: With 2fa054c69fc5da12528e5d47308819d78a56e74b this function will still be called unconditionally, but it will output `no changes` in case the diff was empty.

tools/standardize_config_json.py
mpereira: Fixed in a065501b359f31ed7741b29565283a6f234cea4e.

tools/standardize_config_json.py
mpereira: Fixed in 072eb00997b0f6d39c184bcd8501dc19f34db4e9 and https://github.com/mesosphere/dcos-elastic-service/pull/21/commits/bfc8a75c3178176a132e67b3ea3f8679ea8d0e99.

mpereira:
Thanks for the thoughtful feedback @elezar and @porridge. I think I've addressed everything. If not, please let me know.

tools/standardize_config_json.py
```diff
@@ -99,29 +94,78 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
+
+    service_config_json_path = args.service_config_json
+    sdk_tools_config_path = args.sdk_tools_config
+
+    if not os.path.isfile(service_config_json_path):
+        LOG.info("'%s' is not a file, was expecting a 'config.json' file", service_config_json_path)
+
+    if sdk_tools_config_path:
+        if not os.path.isfile(sdk_tools_config_path):
+            LOG.info(
```
porridge: If this is the case..

tools/standardize_config_json.py
```diff
@@ -99,29 +94,78 @@ def print_diff(original: collections.OrderedDict, new: collections.OrderedDict):
...
+                "'%s' is not a file, was expecting a 'sdk-tools.json' file", sdk_tools_config_path
+            )
+
+        LOG.info("Parsing SDK tooling configuration from '%s'", sdk_tools_config_path)
+        sdk_tools_config = json.loads(read_file(sdk_tools_config_path))
```
porridge: .. then this is going to crash right here, no?

mpereira: Of course! deb7333e0f92c50103625f85493b0f5a5d4cc48f

tools/standardize_config_json.py
mpereira: Also: deb7333e0f92c50103625f85493b0f5a5d4cc48f

PR #3123, opened 2019-07-08
[DCOS-55542] Shared Memory Support SDK

kvish:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -322,6 +322,15 @@ public void setExecutorVolume(VolumeSpec volumeSpec) {
 
     taskInfoBuilder.setContainer(getContainerInfo(podInstance.getPod(), true, true));
 
+    if (taskSpec.getSharedMemory().isPresent()) {
```
kaiwalyajoshi: Question about improper/invalid ServiceSpec's

If we have `podSpec.getSharedMemory().isPresent() == false` and we set 
`taskInfoBuilder.getContainerBuilder().getLinuxInfoBuilder().setIpcMode(taskSpec.getSharedMemory().get()).build()`
due to bad ServiceSpec, will Mesos reject this?

I think it might be worthwhile to extend and add validation for SHM in [DefaultConfigValidators](https://github.com/mesosphere/dcos-commons/blob/0f532c3c76f7826c028248f9985731423ace207d/sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/DefaultConfigValidators.java#L17) to filter out these types of cases.

kvish: added a new config validator for taskSpec validation

frameworks/helloworld/src/main/dist/shm.yml
```diff
@@ -0,0 +1,22 @@
...
+  principal: {{FRAMEWORK_PRINCIPAL}}
+  user: {{FRAMEWORK_USER}}
+pods:
+  hello:
+    count: {{HELLO_COUNT}}
+    ipc-mode: PRIVATE
+    shm-size: 128
+    tasks:
+      server:
+        shm-size: 256
```
Gilbert88: Probably we should validate this field at framework level - If the `ipc-mode` is SHARE_PARENT, uses are not allowed to specify the `shm-size` field, because we should be explicit otherwise users may think the 256 mb shm may be used somewhere.

Based on the [API](https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L3322~L3329), no matter this field is being ignored or errored out, I would suggest frameworks always add this validation, so that users know their config error before launching the task.

Sorry for the previous example, we have already changed the unit tests correspondingly. Mesos will return a failure if shm-size is specified in SHARE_PARENT mode. Please update here. :)

frameworks/helloworld/src/main/dist/svc.yml
```diff
@@ -6,12 +6,16 @@ pods:
...
+    ipc-mode: PRIVATE
+    shm-size: 128
     rlimits:
       RLIMIT_NOFILE:
         soft: {{HELLO_RLIMIT_NOFILE_SOFT}}
         hard: {{HELLO_RLIMIT_NOFILE_HARD}}
     tasks:
       server:
+        ipc-mode: SHARE_PARENT
+        shm-size: 256
```
Gilbert88: ditto.

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultServiceSpecTest.java
```diff
@@ -182,6 +182,46 @@ public void validSeccompInfoAndProfile() throws Exception {
...
+        Assert.assertEquals(Protos.LinuxInfo.IpcMode.PRIVATE, spec.getSharedMemory().get());
+        Assert.assertTrue(spec.getSharedMemorySize().get().equals(1024));
+
+        TaskSpec spec1 = serviceSpec.getPods().get(0).getTasks().get(0);
+        Assert.assertEquals(Protos.LinuxInfo.IpcMode.PRIVATE, spec1.getSharedMemory().get());
+        Assert.assertTrue(spec1.getSharedMemorySize().get().equals(256));
+
+        TaskSpec spec2 = serviceSpec.getPods().get(0).getTasks().get(1);
+        Assert.assertEquals(Protos.LinuxInfo.IpcMode.SHARE_PARENT, spec2.getSharedMemory().get());
+        Assert.assertEquals(Optional.empty(), spec2.getSharedMemorySize());
```
Gilbert88: ditto.

sdk/scheduler/src/test/java/com/mesosphere/sdk/specification/DefaultServiceSpecTest.java
```diff
@@ -182,6 +182,46 @@ public void validSeccompInfoAndProfile() throws Exception {
...
+
+        TaskSpec spec3 = serviceSpec.getPods().get(0).getTasks().get(2);
+        Assert.assertEquals(Protos.LinuxInfo.IpcMode.SHARE_PARENT, spec3.getSharedMemory().get());
+        Assert.assertEquals(Optional.empty(), spec3.getSharedMemorySize());
```
Gilbert88: ditto.

frameworks/helloworld/src/main/dist/shm.yml
```diff
@@ -0,0 +1,22 @@
+name: {{FRAMEWORK_NAME}}
+scheduler:
+  principal: {{FRAMEWORK_PRINCIPAL}}
+  user: {{FRAMEWORK_USER}}
+pods:
+  hello:
+    count: {{HELLO_COUNT}}
+    ipc-mode: PRIVATE
+    shm-size: 128
```
Gilbert88: another thing I would like to double check here is that - when the ipc-mode is PRIVATE, the shm-size field could still be optional.

The shm-size take precedence as follow:
1. users specified shm-size
2. If users do not specify any, Mesos takes [cluster default shm-size](https://github.com/qianzhangxa/dcos/blob/552c13891a0d63c4dacc18a0c0cba4828d62b2b7/gen/dcos-config.yaml#L1394)
3. If there is no default cluster shm size provided, Mesos takes the kernel default behavior - half size of the host memory.

so tl;dr, please double check when ipc-mode is PRIVATE, shm-size is optional.

Thank you!

PR #3125, opened 2019-07-09
[WIP] SI Tests for Quota migration.

kaiwalyajoshi:
- Update scheduler to new role.
- Update pods to new role via `pod replace`, restart scheduler in between each replaced pod to ensure mixed-mode roles are applicable.
- Add additional pods post scheduler update, ensure that new pods are launched under the new role.


kaiwalyajoshi:
Wrong target branch! Closing PR :(

PR #3128, opened 2019-07-15
[DCOS-55083] Pass more params to setBuildStatus to fix status checks

porridge:
Zrzut ekranu z 2019-07-17 18-15-52
This time seems to work.


porridge:
@samvantran and @takirala PTAL.
FWIW, I think that red cross is from one of the earlier runs.

PR #3139, opened 2019-07-25
[DCOS-56415] Emit Delayed status code

kvish:


sdk/scheduler/src/main/java/com/mesosphere/sdk/http/endpoints/HealthResource.java
```diff
@@ -340,10 +344,12 @@ private ServiceStatusEvaluationStage evaluatePendingOrStartingStatusCode(
 
     // We're biasing pessimistically here, pick cases that are halting the
     // deployment from becoming complete.
-    if (pendingSteps > 0 || preparedSteps > 0 || delayedSteps > 0) {
+    if (pendingSteps > 0 || preparedSteps > 0) {
```
takirala: So `if (delayedSteps > 0) ` needs to be the first check here right?

kvish: resolved

PR #3140, opened 2019-07-26
Including keytab checksum on KDC API Response

wavesoft:
This PR adds a `checksum` property on the list and check API calls that is unique for the contents of the keytab stored in the secret.

This checksum can be used to detect unwanted changes in the keytab. (eg. when KDC is restarted and the keytab is re-issued).

### DIFF
--- a/testing/sdk_auth.py
+++ b/testing/sdk_auth.py
@@ -287,7 +287,9 @@ class KerberosEnvironment:
         parsed = res.json()
         print(parsed)
 
-        return parsed.get("data", [])
+        principals = parsed.get("principals", {})
+        return principals.get("list", [])
+
 
     def add_principals(self, principals: List[str]) -> None:
         """
--- a/testing/sdk_auth.py
+++ b/testing/sdk_auth.py
@@ -290,7 +290,6 @@ class KerberosEnvironment:
         principals = parsed.get("principals", {})
         return principals.get("list", [])
 
-
     def add_principals(self, principals: List[str]) -> None:
         """
         Adds a list of principals to the KDC. A principal is defined as a concatenation of 3 parts
--- a/tools-internal/kdc-api-server/server/server.go
+++ b/tools-internal/kdc-api-server/server/server.go
@@ -64,13 +64,14 @@ func (p KPrincipalList) Swap(i, j int) {
 	p[i], p[j] = p[j], p[i]
 }
 
-/**
- * Sortable list of encryption keys
- */
 type KTKey struct {
 	Type         int32
 	HashContents string
 }
+
+/**
+ * Sortable list of encryption keys
+ */
 type KTKeyList []KTKey
 
 func (p KTKeyList) Len() int {
@@ -83,9 +84,6 @@ func (p KTKeyList) Swap(i, j int) {
 	p[i], p[j] = p[j], p[i]
 }
 
-/**
- * createKDCAPIServer creates a configured instance of a server interface
- */
 func createKDCAPIServer(kadmin *KAdminClient, port string, host string) *KDCAPIServer {
 	inst := &KDCAPIServer{
 		kadmin:   kadmin,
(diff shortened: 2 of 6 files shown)


tools-internal/kdc-api-server/server/server.go
```diff
@@ -27,18 +31,61 @@ type KDCRequestListPrincipals struct {
...
+}
+func (p KPrincipalList) Less(i, j int) bool {
+	return p[i].Full() < p[j].Full()
+}
+func (p KPrincipalList) Swap(i, j int) {
+	p[i], p[j] = p[j], p[i]
+}
+
+/**
+ * Sortable list of encryption keys
```
porridge: move comment down, this struct is not the list

tools-internal/kdc-api-server/server/server.go
```diff
@@ -27,18 +31,61 @@ type KDCRequestListPrincipals struct {
...
+}
+func (p KTKeyList) Less(i, j int) bool {
+	return p[i].Type < p[j].Type
+}
+func (p KTKeyList) Swap(i, j int) {
+	p[i], p[j] = p[j], p[i]
+}
+
+/**
+ * createKDCAPIServer creates a configured instance of a server interface
```
porridge: I'd remove this comment, this is a non-exported function, and the name explains the purpose well enough...

tools-internal/kdc-api-server/server/server.go
```diff
@@ -342,9 +459,19 @@ func (s *KDCAPIServer) handleListPrincipals(rw http.ResponseWriter, req *http.Re
 			}
 		}
 		list = newList
+
+		// Calculate checksum of found principals
```
porridge: Please remove this comment, the code is self-explanatory.

tools-internal/kdc-api-server/server/server.go
```diff
@@ -342,9 +459,19 @@ func (s *KDCAPIServer) handleListPrincipals(rw http.ResponseWriter, req *http.Re
...
+		csum, err := getPrincipalsChecksum(ktBytes, list)
+		if err != nil {
+			s.replyReject(rw, req, err.Error())
+			return
+		}
+		resp.Checksum = csum
 	}
 
-	s.replySuccess(rw, req, list)
+	// Prepare list response
```
porridge: Please remove this comment, the code is self-explanatory.

tools-internal/kdc-api-server/server/server.go
```diff
@@ -563,12 +690,18 @@ func (s *KDCAPIServer) handleCheckPrincipals(rw http.ResponseWriter, req *http.R
...
-				false, fmt.Sprintf("Principal '%s' does not exist in keytab", principal.Full()),
+				false, fmt.Sprintf("Principal '%s' does not exist in keytab", principal.Full()), "",
 			})
 			return
 		}
 	}
 
-	// We are done
-	s.replySuccess(rw, req, &KDCCheckStatus{true, ""})
+	// Calculate the principals checksum from the keytab contents
```
porridge: Please remove this comment, the code is self-explanatory.

testing/sdk_auth.py
```diff
@@ -287,7 +287,8 @@ def list_principals(self, filter: str = "*") -> List[str]:
         parsed = res.json()
         print(parsed)
 
-        return parsed.get("data", [])
+        principals = parsed.get("principals", {})
```
porridge: It seems you're changing the API in an incompatible way, but this function is not used anywhere yet, right?

wavesoft: You are correct. The HTTP API has changed slightly in order to include the `checksum` in the response, in addition to the list of `principals`. 

This API is already in use by the KDC terraform provider and already up-to-date : https://github.com/mesosphere/data-services-terraform/blob/master/providers/terraform-provider-kdc/client/api.go#L31

In the scope of `dcos-commons` and the respective integration tests, this change has no impact.

PR #3144, opened 2019-08-01
Allow configuration of the host volume mode

kvish:
mode was previously defaulted to RW, now a user can pass RO to override this


sreis:
I can't review it but I've tested this branch with the Kubernetes framework and it worked as expected. When can we have an SDK release with this PR?

kvish:
I can merge this today, and have a release mid next week

PR #3146, opened 2019-08-07
[DCOS-54125] Add configs to enable secure JMX for Cassandra

viivek46:

Resolves [DCOS-54125](https://jira.mesosphere.com/browse/DCOS-54125) 
Add configs to enable secure JMX for Cassandra


frameworks/cassandra/universe/config.json
```diff
@@ -217,6 +217,138 @@
...
+                    "key_store_password_file": {
+                      "description": "Secret name that contains key store password file.",
+                      "type": "string",
+                      "default": "",
+                      "media": {
+                        "type": "application/x-secret+string"
+                      }
+                    },
+                    "add_trust_store": {
+                      "description": "By default user specified trust store is disabled and user provided certs are added to the default trust store.",
```
samvantran: If trust store is empty string by default, can't we assume that this means users want them disabled and we can remove this boolean checkbox? And if users specify a trust store secret path, then it means this is "enabled"?

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -76,9 +110,61 @@ pods:
...
+                sed -i '/  #JVM_OPTS="$JVM_OPTS -Djavax.net.ssl.keyStorePassword=*/s/^  #/  /g;s/<keystore-password>/'$SECURE_JMX_KEY_STORE_KEY'/g' apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra-env.sh
+                sed -i '/  #JVM_OPTS="$JVM_OPTS -Djavax.net.ssl.trustStore=\/path\/to\/truststore*/s/^  #/  /g;s/\/path\/to\/truststore/\$TRUST_STORE_PATH/g' apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra-env.sh
+                sed -i '/  #JVM_OPTS="$JVM_OPTS -Djavax.net.ssl.trustStorePassword*/s/^  #/  /g;s/<truststore-password>/'$SECURE_JMX_TRUST_STORE_KEY'/g' apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra-env.sh
+                sed -i '/JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.password.file=*/s/\/etc\/cassandra\/jmxremote.password/jmx_properties\/key_file/g' apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra-env.sh
+                sed -i '/#JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.access.file*/s/^#//g;s/\/etc\/cassandra\/jmxremote.access/jmx\/access_file/g' apache-cassandra-{{CASSANDRA_VERSION}}/conf/cassandra-env.sh
```
samvantran: wow, a lot of settings needed to change.

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -146,17 +262,23 @@ pods:
         {{/TASKCFG_ALL_SECURITY_TRANSPORT_ENCRYPTION_ENABLED}}
       snapshot:
         goal: ONCE
-        cmd: >
-                export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/) ;
+        cmd:  |
+                export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/) 
+                {{#TASKCFG_ALL_SECURE_JMX_ENABLED}}
+                export HOME=$MESOS_SANDBOX
```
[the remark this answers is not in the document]
viivek46: yes @samvantran when we want to run nodetool with ssl properties, it look for property file in home directory, and home directory was not set hence was failing, so i have to set home directory.

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,208 @@
...
+        "deleteme",
+        "-keypass",
+        "deleteme",
+        "-storetype",
+        "jks",
+    ]
+
+    subprocess.check_output(create_keystore_cmd)
+
+    create_keystore_cmd = [
```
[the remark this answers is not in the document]
viivek46: updated variable name.

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,208 @@
...
+            trust_store_password = "deleteme"
+        else:
+            trust_store = "$JAVA_HOME/lib/security/cacerts"
+            trust_store_password = "changeit"
+
+        cmd = (
+            "export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/) && "
+            "$JAVA_HOME/bin/java "
+            "-Duser.home=$MESOS_SANDBOX "
+            "-Djdk.tls.client.protocols=TLSv1.2 -Djavax.net.ssl.trustStore={trust_store} "
```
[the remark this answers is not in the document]
viivek46: updated

frameworks/cassandra/universe/config.json
```diff
@@ -217,6 +217,138 @@
             }
           }
         },
+        "jmx_port": {
```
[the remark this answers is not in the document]
viivek46: moved inside ```jmx```

frameworks/cassandra/universe/config.json
viivek46: Yes @samvantran you are correct but it is looking good in UI. you can try if you think its not how it should be then i will remove it.

viivek46:
@zmalik not sure why com.sun.management.jmxremote.ssl.config.file is not working, i had to directly put config values in cassandra-env.sh, i will remove that part from jmx-ssl-properties.sh.

frameworks/cassandra/universe/config.json
samvantran: actually you're right. I like it in the UI too - it hides configs by default. Okay keep it

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -108,11 +206,29 @@ pods:
...
+              ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool --ssl -u $user -pwf $MESOS_SANDBOX/jmx_properties/key_file repair -pr -p ${JMX_PORT} -- $CASSANDRA_KEYSPACE
+              {{/TASKCFG_ALL_SECURE_JMX_ENABLED}}
+              {{^TASKCFG_ALL_SECURE_JMX_ENABLED}}
+              ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool repair -pr -p ${JMX_PORT} -- $CASSANDRA_KEYSPACE
+              {{/TASKCFG_ALL_SECURE_JMX_ENABLED}}
         resource-set: sidecar-resources
       cleanup:
         goal: ONCE
-        cmd: export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/) && ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool cleanup -p ${JMX_PORT} -- $CASSANDRA_KEYSPACE
+        cmd: |
```
zmalik: do we have any tests for these commands? these are the only places where we are enabling secure JMX options and if we break them in the future would be hard to detect.  
Can we add a test with secure JMX options enabled and run a cleanup for example?

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -108,11 +206,29 @@ pods:
         {{/TASKCFG_ALL_SECURITY_TRANSPORT_ENCRYPTION_ENABLED}}
       repair:
         goal: ONCE
-        cmd: export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/) && ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool repair -pr -p ${JMX_PORT} -- $CASSANDRA_KEYSPACE
+        cmd: |
+              export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/)
+              {{#TASKCFG_ALL_SECURE_JMX_ENABLED}}
+              export HOME=$MESOS_SANDBOX
+              user=$(sed -nr '0,/re/s/.*(^)(.+) readwrite.*/\2/p' $MESOS_SANDBOX/jmx/access_file)
```
zmalik: works for 
```
controlRole readwrite
monitorRole readonly
```
but not for 
```
monitorRole readonly
controlRole readwrite
```
and both are valid access file content

viivek46: Thanks @zmalik , I will update it.

frameworks/cassandra/src/main/dist/svc.yml
viivek46: Yes @zmalik , i had discussion with Sam . i am currently working on to add test where jmx code is added like in task repair/cleanup/snapshot/cleanup-snapshot.

frameworks/cassandra/src/main/dist/svc.yml
viivek46: updated sed command.

frameworks/cassandra/universe/config.json
```diff
@@ -391,11 +530,6 @@
           "description": "The name of the cluster managed by the Service",
           "default": "cassandra"
         },
-        "jmx_port": {
```
zmalik: that would be a breaking change right? for customers using non-secure JMX? 
@samvantran is that intended for secure JMX in Cassandra?

frameworks/cassandra/src/main/dist/svc.yml
viivek46: -Dcom.sun.management.jmxremote.ssl=true 6 times. yea that was working, as i was not replacing it, i was just adding line below it.
but now i have moved it under ```jmx-ssl-properties.sh```

frameworks/cassandra/universe/config.json
samvantran: Good catch! Yes, we definitely don't want to introduce breaking changes (unless we tell customers and give them a transition period where we deprecate the config). Let's revert this to what it used to be... Actually, our upgrade test should have told us that the upgrade failed due to breaking configs: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/cassandra/tests/test_sanity.py#L32-L37. Did that happen?

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,208 @@
...
+
+def install_jmx_configured_cassandra(self_signed_trust_store: bool = True):
+    foldered_name = config.get_foldered_service_name()
+    sdk_install.uninstall(config.PACKAGE_NAME, foldered_name)
+    install_jmx_secrets()
+    service_options = {
+        "service": {
+            "name": foldered_name,
+            "jmx": {
+                "enabled": True,
```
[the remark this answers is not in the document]
viivek46: @samvantran i am not very sure about it , we can take advice form other folks as well.

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,261 @@
...
+
+    delete_secret(secret=secret_path)
+    sdk_cmd.run_cli(
+        'security secrets create --value="{account}" "{secret}"'.format(
+            account=secret_value, secret=secret_path
+        )
+    )
+
+
+def delete_secret(secret: str) -> None:
```
samvantran: Why have this method at all? why not use thje `sdk_security` cmd  like you do [here](https://github.com/mesosphere/dcos-commons/pull/3146/files#diff-43a0548ff83f66aac8579c084a3972e5R138-R141)

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,261 @@
...
+    full_cmd = "bash -c '{}'".format(cmd)
+    rc, stdout, stderr = sdk_cmd.run_cli(
+        "task exec {} {}".format(task_id, full_cmd), print_output=True
+    )
+    assert rc == 0, "Error downloading jmxterm {}".format(jmx_term_url)
+
+
+def create_secret(secret_value: str, secret_path: str) -> None:
+
+    delete_secret(secret=secret_path)
```
samvantran: Replace with the sdk_security cmd as this extra method doesn't seem to do anything different

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,261 @@
...
+    cmd = "/opt/mesosphere/bin/curl {} --output jmxterm-1.0.1-uber.jar".format(jmx_term_url)
+    full_cmd = "bash -c '{}'".format(cmd)
+    rc, stdout, stderr = sdk_cmd.run_cli(
+        "task exec {} {}".format(task_id, full_cmd), print_output=True
+    )
+    assert rc == 0, "Error downloading jmxterm {}".format(jmx_term_url)
+
+
+def create_secret(secret_value: str, secret_path: str) -> None:
+
```
samvantran:
```suggestion
```

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,261 @@
...
+        if self_signed_trust_store:
+            trust_store = "$MESOS_SANDBOX/jmx/trust_store"
+            trust_store_password = "deleteme"
+        else:
+            trust_store = "$JAVA_HOME/lib/security/cacerts"
+            trust_store_password = "changeit"
+
+        if not authentication:
+            test_sanity.test_repair_cleanup_plans_complete()
+            test_backup_and_restore.test_backup_and_restore_to_s3()
```
samvantran: These two tests seem out of place. The entire file thus far been building up deploying jmx cassandra and setting up beans/domains commands with uberterm and then all of a sudden you're testing backup/restore. Even the following code goes back to setting up the test for jmx commands. 

Instead, can you add a separate test method to test these operations with jmx-cassandra to verify they still work? Let's not mix too many things in the same test.

frameworks/cassandra/tests/test_sanity.py
```diff
@@ -109,7 +108,7 @@ def expected_metrics_exist(emitted_metrics: List[str]) -> bool:
 def test_custom_jmx_port() -> None:
     expected_open_port = ":7200 (LISTEN)"
 
-    new_config = {"cassandra": {"jmx_port": 7200}}
+    new_config = {"service": {"jmx": {"jmx_port": 7200}}}
```
samvantran: if you reverted changes back, you'll have to do it here too

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,261 @@
...
+        generate_jmx_command_files(task_id=node_task_id_0)
+
+        if self_signed_trust_store:
+            trust_store = "$MESOS_SANDBOX/jmx/trust_store"
+            trust_store_password = "deleteme"
+        else:
+            trust_store = "$JAVA_HOME/lib/security/cacerts"
+            trust_store_password = "changeit"
+
+        if not authentication:
```
samvantran: How comes these tests only work when authentication is off?

frameworks/cassandra/tests/test_jmx.py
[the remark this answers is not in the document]
samvantran: it's fine we can leave as is since it's all about jmx

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -313,7 +534,35 @@ pods:
                     cqlsh -u cassandra -p cassandra {{{CASSANDRA_CQLSH_SSL_FLAGS}}} -e "alter keyspace system_traces      WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 3};" "node-0-server.${FRAMEWORK_HOST}" {{TASKCFG_ALL_CASSANDRA_NATIVE_TRANSPORT_PORT}} &&
                     cqlsh -u cassandra -p cassandra {{{CASSANDRA_CQLSH_SSL_FLAGS}}} -e "alter keyspace system_auth        WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 3};" "node-0-server.${FRAMEWORK_HOST}" {{TASKCFG_ALL_CASSANDRA_NATIVE_TRANSPORT_PORT}} &&
                     cqlsh -u cassandra -p cassandra {{{CASSANDRA_CQLSH_SSL_FLAGS}}} -e "alter keyspace system_distributed WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 3};" "node-0-server.${FRAMEWORK_HOST}" {{TASKCFG_ALL_CASSANDRA_NATIVE_TRANSPORT_PORT}} &&
-                    nodetool repair system_auth &&
+                    {{#TASKCFG_ALL_SECURE_JMX_ENABLED}}
+                    ./bootstrap --resolve=false
```
samvantran: you're already calling this a [few lines up](https://github.com/mesosphere/dcos-commons/pull/3146/files#diff-fe0c20da1fba698643bc101afa99c204R529)

Also does all of this code get executed? The beginning of this has `cmd: >` and every line has either a `;` or `&&` while this new section does not. I wonder if this code path is getting tested?

frameworks/cassandra/src/main/dist/jmx-ssl-properties.sh
```diff
@@ -0,0 +1,46 @@
...
+sed -i '/  JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.rmi.port*/s/$JMX_PORT/"'$SECURE_JMX_RMI_PORT'"/g' apache-cassandra-$CASSANDRA_VERSION/conf/cassandra-env.sh 
+#sed -i 's/# JVM_OPTS="$JVM_OPTS -Djava.rmi.server.hostname=<public name>"/JVM_OPTS="$JVM_OPTS -Djava.rmi.server.hostname='$MESOS_CONTAINER_IP'"/g' apache-cassandra-$CASSANDRA_VERSION/conf/cassandra-env.sh
+sed -i '/JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.password.file=*/s/\/etc\/cassandra\/jmxremote.password/jmx_properties\/key_file/g' apache-cassandra-$CASSANDRA_VERSION/conf/cassandra-env.sh
+
+NODETOOL_SSL_PROPERTIES_FILE=$MESOS_SANDBOX/apache-cassandra-$CASSANDRA_VERSION/conf/cassandra-env.sh
+
+tee -a $NODETOOL_SSL_PROPERTIES_FILE <<EOF
+
+JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.ssl=true"
+JVM_OPTS="\$JVM_OPTS -Djavax.net.ssl.enabled.protocols=TLSv1.2"
```
samvantran: just for neatness can you move this `javax.net` next to its other configs of the same name?

frameworks/cassandra/src/main/dist/jmx-ssl-properties.sh
```diff
@@ -0,0 +1,46 @@
...
+-Djavax.net.ssl.keyStorePassword=$SECURE_JMX_KEY_STORE_KEY 
+-Djavax.net.ssl.trustStore=$TRUST_STORE_PATH
+-Djavax.net.ssl.trustStorePassword=$SECURE_JMX_TRUST_STORE_KEY 
+-Djavax.rmi.ssl.client.enabledProtocols=TLSv1.2
+
+EOF
+
+sed -i 's/    LOCAL_JMX=yes*/    LOCAL_JMX=no/' apache-cassandra-$CASSANDRA_VERSION/conf/cassandra-env.sh
+sed -i '/  JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.rmi.port*/s/$JMX_PORT/"'$SECURE_JMX_RMI_PORT'"/g' apache-cassandra-$CASSANDRA_VERSION/conf/cassandra-env.sh 
+#sed -i 's/# JVM_OPTS="$JVM_OPTS -Djava.rmi.server.hostname=<public name>"/JVM_OPTS="$JVM_OPTS -Djava.rmi.server.hostname='$MESOS_CONTAINER_IP'"/g' apache-cassandra-$CASSANDRA_VERSION/conf/cassandra-env.sh
```
samvantran: If this isnt needed then delete

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -313,7 +534,35 @@ pods:
...
+                    export SECURE_JMX_KEY_STORE_KEY=`cat jmx/key_store_pass`
+                    export KEY_STORE_PATH=jmx/key_store
+                    if [ -f "jmx/trust_store" ]; then
+                        # load user provided trust store
+                        export TRUST_STORE_PATH=jmx/trust_store
+                        export SECURE_JMX_TRUST_STORE_KEY=`cat jmx/trust_store_pass`
+                    else
+                        # load user provided keystore certs to default truststore
+                        export TRUST_STORE_PATH=$JAVA_HOME/lib/security/cacerts
+                        export SECURE_JMX_TRUST_STORE_KEY="changeit"
```
samvantran: `changeit`? Is that like java default? Why are we hardcoding the trust store key ourselves?

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -313,7 +534,35 @@ pods:
...
+                        ${JAVA_HOME}/bin/keytool -noprompt -importkeystore -srckeystore $KEY_STORE_PATH -srcstorepass $SECURE_JMX_KEY_STORE_KEY -destkeystore $TRUST_STORE_PATH -deststorepass $SECURE_JMX_TRUST_STORE_KEY
+                    fi
+
+                    mkdir jmx_properties
+                    mkdir .cassandra
+                    (umask u=rw,g=,o= && cp jmx/key_file jmx_properties/key_file)
+                    chmod +x .ssl/jmx-ssl-properties.sh
+                    .ssl/jmx-ssl-properties.sh
+                    export HOME=$MESOS_SANDBOX &&
```
samvantran: You already exported home [above](https://github.com/mesosphere/dcos-commons/pull/3146/files#diff-fe0c20da1fba698643bc101afa99c204R540)

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -179,11 +366,45 @@ pods:
...
+                    export SECURE_JMX_TRUST_STORE_KEY="changeit"
+                    ${JAVA_HOME}/bin/keytool -noprompt -importkeystore -srckeystore $KEY_STORE_PATH -srcstorepass $SECURE_JMX_KEY_STORE_KEY -destkeystore $TRUST_STORE_PATH -deststorepass $SECURE_JMX_TRUST_STORE_KEY
+                fi
+
+                mkdir jmx_properties
+                mkdir .cassandra
+                (umask u=rw,g=,o= && cp jmx/key_file jmx_properties/key_file)
+                chmod +x .ssl/jmx-ssl-properties.sh
+                .ssl/jmx-ssl-properties.sh
+                export HOME=$MESOS_SANDBOX
```
samvantran: HOME already exported to same path

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -179,11 +366,45 @@ pods:
...
+                export SECURE_JMX_KEY_STORE_KEY=`cat jmx/key_store_pass`
+                export KEY_STORE_PATH=jmx/key_store
+                if [ -f "jmx/trust_store" ]; then
+                    # load user provided trust store
+                    export TRUST_STORE_PATH=jmx/trust_store
+                    export SECURE_JMX_TRUST_STORE_KEY=`cat jmx/trust_store_pass`
+                else
+                    # load user provided keystore certs to default truststore
+                    export TRUST_STORE_PATH=$JAVA_HOME/lib/security/cacerts
+                    export SECURE_JMX_TRUST_STORE_KEY="changeit"
```
samvantran: same question regarding hardcoded `changeit`

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -146,19 +300,52 @@ pods:
...
+                    ${JAVA_HOME}/bin/keytool -noprompt -importkeystore -srckeystore $KEY_STORE_PATH -srcstorepass $SECURE_JMX_KEY_STORE_KEY -destkeystore $TRUST_STORE_PATH -deststorepass $SECURE_JMX_TRUST_STORE_KEY
+                fi
 
+                mkdir jmx_properties
+                mkdir .cassandra
+                (umask u=rw,g=,o= && cp jmx/key_file jmx_properties/key_file)
+                chmod +x .ssl/jmx-ssl-properties.sh
+                .ssl/jmx-ssl-properties.sh
+                export HOME=$MESOS_SANDBOX
```
samvantran: exported already

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -108,12 +190,84 @@ pods:
...
+                  export SECURE_JMX_TRUST_STORE_KEY="changeit"
+                  ${JAVA_HOME}/bin/keytool -noprompt -importkeystore -srckeystore $KEY_STORE_PATH -srcstorepass $SECURE_JMX_KEY_STORE_KEY -destkeystore $TRUST_STORE_PATH -deststorepass $SECURE_JMX_TRUST_STORE_KEY
+              fi
+
+              mkdir jmx_properties
+              mkdir .cassandra
+              (umask u=rw,g=,o= && cp jmx/key_file jmx_properties/key_file)
+              chmod +x .ssl/jmx-ssl-properties.sh
+              .ssl/jmx-ssl-properties.sh
+              export HOME=$MESOS_SANDBOX
```
samvantran: same

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -108,12 +190,84 @@ pods:
...
```
samvantran: again

frameworks/cassandra/src/main/dist/svc.yml
viivek46: correct i will remove one.

frameworks/cassandra/src/main/dist/svc.yml
viivek46: correct i will remove one.

frameworks/cassandra/src/main/dist/svc.yml
viivek46: The default password for this keystore is changeit.

frameworks/cassandra/tests/test_jmx.py
viivek46: because test_backup_and_restore_to_s3  does not work with auth, as for auth we have separate test which have different dependencies. 
With auth we just need to check JMX is working or not, as in task ```init_system_keyspaces``` under auth  some changes were done as part of JMX. 
and with auth back/restore - repair/cleanup test are not required as JMX changes are already evaluating in non-auth back/restore - repair/cleanup.

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,261 @@
...
+    sdk_install.install(
+        config.PACKAGE_NAME,
+        foldered_name,
+        config.DEFAULT_TASK_COUNT,
+        additional_options=service_options,
+    )
+
+
+def install_jmx_secrets():
+    test_run = random_string()
```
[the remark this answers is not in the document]
viivek46: updated as per suggestion.

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,261 @@
...
+    sdk_security.delete_secret(KEY_STORE)
+    sdk_security.delete_secret(KEY_STORE_PASS)
+    sdk_security.delete_secret(ACCESS_FILE)
+    sdk_security.delete_secret(PASSWORD_FILE)
+
+
+@pytest.mark.sanity
+@sdk_utils.dcos_ee_only
+@pytest.mark.parametrize("self_signed_trust_store", [True, False])
+@pytest.mark.parametrize("authentication", [True, False])
```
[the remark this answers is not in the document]
viivek46: updated as per suggestion.

frameworks/cassandra/src/main/dist/svc.yml
viivek46: updated as per suggestion.

frameworks/cassandra/src/main/dist/svc.yml
viivek46: updated as per suggestion.

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -146,19 +298,51 @@ pods:
...
-                ./apache-cassandra-{{CASSANDRA_VERSION}}/bin/nodetool snapshot -p ${JMX_PORT} -t "$SNAPSHOT_NAME" -- $(eval "echo $CASSANDRA_KEYSPACES")
+        cmd:  |
+                export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/) 
+                {{#TASKCFG_ALL_SECURE_JMX_ENABLED}}
+                ./bootstrap --resolve=false
+                cd $MESOS_SANDBOX
+                export HOME=$MESOS_SANDBOX
+                export SECURE_JMX_KEY_STORE_KEY=`cat jmx/key_store_pass`
+                export KEY_STORE_PATH=jmx/key_store
+                if [ -f "jmx/trust_store" ]; then
```
zmalik: we are using a lot of repeated code in commands, for example, this piece of code block 
```
export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/)
{{#TASKCFG_ALL_SECURE_JMX_ENABLED}}
./bootstrap --resolve=false
cd $MESOS_SANDBOX
export HOME=$MESOS_SANDBOX
export SECURE_JMX_KEY_STORE_KEY=`cat jmx/key_store_pass`
export KEY_STORE_PATH=jmx/key_store
if [ -f "jmx/trust_store" ]; then
    # load user provided trust store
    export TRUST_STORE_PATH=jmx/trust_store
    export SECURE_JMX_TRUST_STORE_KEY=`cat jmx/trust_store_pass`
else
    # load user provided keystore certs to default truststore
    export TRUST_STORE_PATH=$JAVA_HOME/lib/security/cacerts
    export SECURE_JMX_TRUST_STORE_KEY="changeit"
    ${JAVA_HOME}/bin/keytool -noprompt -importkeystore -srckeystore $KEY_STORE_PATH -srcstorepass $SECURE_JMX_KEY_STORE_KEY -destkeystore $TRUST_STORE_PATH -deststorepass $SECURE_JMX_TRUST_STORE_KEY
fi
mkdir jmx_properties
mkdir .cassandra
(umask u=rw,g=,o= && cp jmx/key_file jmx_properties/key_file)
chmod +x .ssl/jmx-ssl-properties.sh
.ssl/jmx-ssl-properties.sh
user=$(sed -nr '0,/readwrite/s/.*(^)(.+) readwrite.*/\2/p' $MESOS_SANDBOX/jmx/access_file)
```
is used 6 times, for the sake of maintainability I think we should move the repeated code to a script and call the script from everywhere its needed.

frameworks/cassandra/src/main/dist/jmx-ssl-properties.sh
```diff
@@ -0,0 +1,45 @@
...
+
+JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.ssl=true"
+JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.local.only=false"
+JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.registry.ssl=true"
+JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote=true"
+JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.host=$MESOS_CONTAINER_IP"
+JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.ssl.need.client.auth=true"
+JVM_OPTS="\$JVM_OPTS -Djavax.net.ssl.enabled.protocols=TLSv1.2"
+JVM_OPTS="\$JVM_OPTS -Djavax.net.ssl.keyStore=$KEY_STORE_PATH"
+JVM_OPTS="\$JVM_OPTS -Djavax.net.ssl.keyStorePassword=$SECURE_JMX_KEY_STORE_KEY"
```
zmalik: as you are not suppressing the output as you do in above redirection to `>/dev/null` 
here it will print the `SECURE_JMX_KEY_STORE_KEY` and `SECURE_JMX_TRUST_STORE_KEY` in logs. I think we don't want that.

frameworks/cassandra/src/main/dist/svc.yml
samvantran: +1

frameworks/cassandra/tests/test_jmx.py
```diff
@@ -0,0 +1,304 @@
...
+def uninstall_jmx_secrets():
+    sdk_security.delete_secret(KEY_STORE)
+    sdk_security.delete_secret(KEY_STORE_PASS)
+    sdk_security.delete_secret(ACCESS_FILE)
+    sdk_security.delete_secret(PASSWORD_FILE)
+
+
+@pytest.mark.sanity
+@sdk_utils.dcos_ee_only
+def secure_jmx_configuration(self_signed_trust_store, authentication):
```
samvantran: Can you rename this to something that's closer to its intent? Like
```
check_secure_jmx_output()
```
or 
```
verify_jmx_cmd_succeeded()
```

viivek46: changed to ```check_secure_jmx_output```

frameworks/cassandra/src/main/dist/svc.yml
viivek46: added repeated code to jmx-ssl-setup.sh.

PR #3158, opened 2019-08-20
[DCOS-57609] Rename the file to make autobuilder work again

porridge:
https://jenkins.mesosphere.com/service/jenkins/job/dcos-commons-base%20docker%20image/indexing/console
has already been pointed at the new name.


porridge:
FTR, the failure is expected at this stage. #3159 fixes that, but is orthogonal to this change.

PR #3163, opened 2019-08-21
Revert "Add config validator for Seccomp"

kvish:
Reverts mesosphere/dcos-commons#3105

Revert config validation, seccomp is a noop if on older dcos versions


sdk/scheduler/src/main/java/com/mesosphere/sdk/config/validate/PodSpecsCannotUseUnsupportedFeatures.java
```diff
@@ -165,17 +162,6 @@ private static boolean podRequestsProfileMountVolumes(PodSpec podSpec) {
             "This DC/OS cluster does not support profile mount volumes"
         ));
       }
-
```
takirala: Instead of reverting this code, how about:

1. Retain the if condition but delete the code that adds to `errors`
2. print a loud warning that the framework behavior may be undefined due to discarded fields in yaml.
3. create a ticket to change the log message back to va;odation error.

PR #3175, opened 2019-09-02
[DCOS-57593][HDFS] Add base tech cli commands

alexeygorobets:
Added new command `hdfs` that allows to run HDFS CLI commands via DC/OS CLI
```
$ dcos hdfs hdfs dfs -mkdir /dir
$ dcos hdfs hdfs dfs -ls /
Found 1 items
drwxr-xr-x   - nobody supergroup          0 2019-09-02 10:11 /dir
```
List of HDFS CLI commands: https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/HDFSCommands.html

- added new command to hdfs framework - frameworks/hdfs/cli/main.go, remaining arguments after `hdfs` command will be considered as subcommands and flags for HDFS CLI
- added building CLI tool as part HDFS framework build
- `test_cli.py` test creates a directory in HDFS via command line commands and verifys it

### DIFF
--- a/frameworks/hdfs/cli/main.go
+++ b/frameworks/hdfs/cli/main.go
@@ -5,6 +5,7 @@ import (
 	"github.com/mesosphere/dcos-commons/cli"
 	"github.com/mesosphere/dcos-commons/cli/client"
 	"gopkg.in/alecthomas/kingpin.v3-unstable"
+	"os"
 	"strings"
 )
 
@@ -67,7 +68,11 @@ func (args *HdfsCliArgs) runHdfs(a *kingpin.Application, e *kingpin.ParseElement
 	hdfsCommand := fmt.Sprintf("export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/); $HDFS_VERSION/bin/hdfs %s", args.String())
 	fullCommand := append(dcosExecCommand, hdfsCommand)
 
-	value, _ := client.RunCLICommand(fullCommand...)
+	value, error := client.RunCLICommand(fullCommand...)
+	if error != nil {
+		client.PrintMessage(error.Error())
+		os.Exit(1)
+	}
 
 	client.PrintMessage(value)
 	return nil
--- a/frameworks/hdfs/cli/main.go
+++ b/frameworks/hdfs/cli/main.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"fmt"
+	"github.com/mesosphere/dcos-commons/cli"
+	"github.com/mesosphere/dcos-commons/cli/client"
+	"gopkg.in/alecthomas/kingpin.v3-unstable"
+	"strings"
+)
+
+func main() {
+	app := cli.New()
+	arguments := cli.GetArguments()
+
+	cli.HandleDefaultSections(app)
+
+	hdfsCliCommandIndex := HandleHdfsSection(app, arguments)
+
+	if hdfsCliCommandIndex != -1 {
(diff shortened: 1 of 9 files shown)


frameworks/hdfs/cli/main.go
```diff
@@ -0,0 +1,74 @@
...
+type HdfsCliArgs struct {
+	arguments []string
+}
+
+func (args *HdfsCliArgs) String() string {
+	return strings.Join(args.arguments, " ")
+}
+
+func (args *HdfsCliArgs) runHdfs(a *kingpin.Application, e *kingpin.ParseElement, c *kingpin.ParseContext) error {
+	dcosExecCommand := []string{"task", "exec", "name-0-node", "bash", "-c"}
```
samvantran: I think this is fine for now but eventually we'd probably want to make it possible to target specific hdfs nodes. For example, depending on which namenode you run the command `hdfs haadmin -getServiceState` on, you'd get `active|standby`. 

Thinking it over, is it easy to add a flag for targeting a certain node and default to `name-0-node` if no flag provided?

frameworks/hdfs/tests/test_cli.py
```diff
@@ -0,0 +1,40 @@
...
+    finally:
+        sdk_install.uninstall(config.PACKAGE_NAME, config.SERVICE_NAME)
+
+
+@pytest.mark.sanity
+def test_cli():
+    sdk_cmd.run_cli("package install {} --yes --cli".format(config.PACKAGE_NAME))
+    dirpath = "/test"
+    sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, "hdfs dfs -mkdir {}".format(dirpath))
+    _, stdout, _ = sdk_cmd.svc_cli(config.PACKAGE_NAME, config.SERVICE_NAME, "hdfs dfs -ls /")
```
samvantran: I'd say let's add more tests to verify other commands work as expected but given team's current priorities, I think this is fine for now

alexeygorobets: I will think about corner cases that can be tested, thanks!

frameworks/hdfs/cli/main.go
alexeygorobets: yes, thanks for the suggestion! will do that

frameworks/hdfs/cli/main.go
```diff
@@ -0,0 +1,74 @@
...
+	hdfsCommand := fmt.Sprintf("export JAVA_HOME=$(ls -d $MESOS_SANDBOX/jdk*/); $HDFS_VERSION/bin/hdfs %s", args.String())
+	fullCommand := append(dcosExecCommand, hdfsCommand)
+
+	value, _ := client.RunCLICommand(fullCommand...)
```
zmalik: you can use the `error` here to return it if it's not `nil`
The method is returning an error but that value isn't used anywhere. So just print the error also here in case it's not nil

frameworks/hdfs/cli/main.go
```diff
@@ -0,0 +1,74 @@
+package main
```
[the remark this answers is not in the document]
alexeygorobets: thank you! forgot about this awesome tool

frameworks/hdfs/cli/main.go
alexeygorobets: done!

samvantran:
Just found out one thing we missed was adding a message in the `dcos hdfs -h` about this new feature!

PR #3193, opened 2019-09-24
[DCOS-29593] Use introspection of stub universe JSON to determine package name

mediapills:
- Added  package name loader from the stub universe URL

### DIFF
--- a/tools/release_builder.py
+++ b/tools/release_builder.py
@@ -13,6 +13,7 @@ import sys
 import tempfile
 import universe
 import urllib.request
+import urllib.error
 
 log = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
@@ -44,7 +45,21 @@ class UniverseReleaseBuilder(object):
                 )
 
             package_name = name_match.group(1)
-            log.info("Got package name %s from stub universe URL")
+
+            try:
+                log.info("Parse stub universe")
+                with urllib.request.urlopen(stub_universe_url) as response:
+                    data = json.loads(response.read())
+                    for package in data["packages"]:
+                        if "name" in package:
+                            package_name = package["name"]
+                            break
+            except urllib.error.HTTPError:
+                raise Exception("Could not open URL: {}".format(stub_universe_url))
+            except json.decoder.JSONDecodeError:
+                raise Exception("Wrong file content format: {}".format(stub_universe_url))
+
+            log.info("Got package name '{}' from stub universe URL".format(package_name))
 
         return package_name
 
--- a/tools/release_builder.py
+++ b/tools/release_builder.py
@@ -47,17 +47,26 @@ class UniverseReleaseBuilder(object):
             package_name = name_match.group(1)
 
             try:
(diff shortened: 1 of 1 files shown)


tools/release_builder.py
```diff
@@ -44,7 +45,21 @@ def get_package_name(stub_universe_url: str) -> str:
                 )
 
             package_name = name_match.group(1)
-            log.info("Got package name %s from stub universe URL")
+
+            try:
+                log.info("Parse stub universe")
```
porridge:
```suggestion
                log.info("Parsing stub universe...")
```

mediapills:
> This seems to include changes from another PR. Please make them independent.

Done 

> It's not clear to me where and how many times `get_package_name` is called. We might have to cache the stub `json` file to not fetch it every time.

It is called only once so I left it as is.

> Also, for the time being I think we should attempt to parse the file, but in case of any problem, we should fall back to the previous (filename-based) logic with a warning. Once we establish this works for all frameworks, we can phase out the old logic.

I have added a check in a case we have more than one different names `log.warn` will notify us about this.
> And probably most importantly, it's not clear to me what should happen in case there are multiple packages in the `json` file. Perhaps this is the case that's not working for DSE and the reason that repo has custom modifications around that area? Can you please investigate?

tools/release_builder.py
```diff
@@ -44,7 +45,30 @@ def get_package_name(stub_universe_url: str) -> str:
...
+                                log.warning(
+                                    "Two different package names have been found '{}' and '{}'".format(
+                                        universe_name, package["name"]
+                                    )
+                                )
+                            else:
+                                universe_name = package["name"]
+                if universe_name is not None:
+                    package_name = universe_name
+            except urllib.error.HTTPError:
```
porridge: Unfortunately from my experience `urllib` is not well behaved w.r.t. exceptions, especially when TLS is involved - we cannot assume that this is the only exception type that will be raised. Please use `except Exception` to handle this gracefully.

tools/release_builder.py
```diff
@@ -44,7 +45,30 @@ def get_package_name(stub_universe_url: str) -> str:
...
+                log.warning("Could not open URL: {}".format(stub_universe_url))
```
porridge: Please do not hide the details provided in the original exception. This makes troubleshooting very hard - we'd have no idea whether the problem was transient, permanent, server-side, client-side, bad proxy or what.

tools/release_builder.py
```diff
@@ -44,7 +45,30 @@ def get_package_name(stub_universe_url: str) -> str:
...
+            try:
+                universe_name = None
+                log.info("Parsing stub universe...")
+                with urllib.request.urlopen(stub_universe_url) as response:
+                    data = json.loads(response.read())
+                    for package in data["packages"]:
+                        if "name" in package:
+                            if universe_name is not None and universe_name != package["name"]:
+                                log.warning(
+                                    "Two different package names have been found '{}' and '{}'".format(
```
porridge: s/Two/More than one/

mediapills:
Could you please check the changes?

PR #3207, opened 2019-11-12
Make uploading compatiable with other S3 storage.

harryge00:
Currently, the `uploading` only works with AWS S3. But we may use other cloud providers like Azure, GCP, Aliyun. So we can use env var `S3_ENDPOINT_URL`. If this var is not defined, skip it. Otherwise, append the `--endpoint-url`,  uploading to other cloud providers through the `url`


harryge00:
Similar to https://github.com/mesosphere/dcos-commons/pull/2475/files#diff-fe0c20da1fba698643bc101afa99c204R139

PR #3208, opened 2019-11-19
[DCOS-28911] : Disabling short circuit local reads

rathohit311356:
## What changes were proposed in this pull request?
Disabled\turned off short-circuit local reads 

Resolves [HDFS client shows DomainSocket warning when reading files](https://jira.mesosphere.com/browse/DCOS-28911)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ allprojects {
     // 2. In the GitHub UI, add a Release entry against the tag created in step 1. List any changes in release notes.
     // 3. Create a PR which bumps this from '0.10.0-SNAPSHOT' to '0.11.0-SNAPSHOT' in master to start the 0.11.0 track.
     // --
-    version = '0.58.0-SNAPSHOT'
+    version = '0.57.0'
 
     sourceCompatibility = '1.8'
     targetCompatibility = '1.8'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -33,7 +33,7 @@ allprojects {
     // 2. In the GitHub UI, add a Release entry against the tag created in step 1. List any changes in release notes.
     // 3. Create a PR which bumps this from '0.10.0-SNAPSHOT' to '0.11.0-SNAPSHOT' in master to start the 0.11.0 track.
     // --
-    version = '0.58.0-SNAPSHOT'
+    version = '0.57.0'
```
samvantran: Head branch is pointing to master so you shouldn't be changing this value

frameworks/hdfs/universe/config.json
```diff
@@ -1490,7 +1490,7 @@
         "client_read_shortcircuit": {
           "type": "boolean",
           "description": "This configuration parameter turns on short-circuit local reads.",
-          "default": true
+          "default": false
```
samvantran: We should provide an explanation why we're changing it to false as this may result in degraded performance. Original purpose was to fix the issue in our tests - by changing this we're actually affecting the standard deployment of any HDFS service

minyk: @samvantran If any process would use HDFS's short circuit read, that process should access `dn_socket` in the HDFS datanode's  mesos sandbox directory. I don't think this socket can be accessed other mesos tasks like spark executors. It is already inaccessible, so it's safe to turn this off. Am I something wrong?

PR #3211, opened 2019-12-04
[DCOS-61615] Bump Cassandra version to 3.11.5

farhan5900:
Resolves [DCOS-61615: Bump Cassandra version to 3.11.5](https://jira.mesosphere.com/browse/DCOS-61615)

Currently Cassandra Framework is running with version `3.11.4`. Here we are addressing upgrade to use version `3.11.5`.


frameworks/cassandra/universe/resource.json
```diff
@@ -7,7 +7,7 @@
       "cassandra-openssl-tar-gz": "https://downloads.mesosphere.com/cassandra/assets/openssl-1.0.2n-libs.tar.gz",
       "libmesos-bundle-tar-gz": "{{libmesos-bundle-url}}",
       "scheduler-zip": "{{artifact-dir}}/cassandra-scheduler.zip",
-      "cassandra-tar-gz": "https://downloads.mesosphere.com/cassandra/assets/apache-cassandra-3.11.4-bin-dcos.tar.gz",
+      "cassandra-tar-gz": "https://infinity-artifacts.s3-us-west-2.amazonaws.com/autodelete7d/cassandra/apache-cassandra-3.11.5-bin-dcos.tar.gz",
```
kaiwalyajoshi: This needs to be a permanent asset and not from `autodelete7d` folder.

frameworks/cassandra/src/main/dist/cassandra.yaml
```diff
@@ -167,3 +167,12 @@ back_pressure_strategy:
...
 {{/AUTHENTICATION_CUSTOM_YAML_BLOCK_BASE64}}
+{{#CASSANDRA_REPAIR_SESSION_MAX_TREE_DEPTH}}
+repair_session_max_tree_depth: {{CASSANDRA_REPAIR_SESSION_MAX_TREE_DEPTH}}
+{{/CASSANDRA_REPAIR_SESSION_MAX_TREE_DEPTH}}
+
+#########################
+# EXPERIMENTAL FEATURES #
+#########################
+enable_sasi_indexes: {{CASSANDRA_ENABLE_SASI_INDEXES}}
+enable_materialized_views: {{CASSANDRA_ENABLE_MATERIALIZED_VIEWS}}
```
viivek46:
```suggestion
enable_materialized_views: {{CASSANDRA_ENABLE_MATERIALIZED_VIEWS}}

```

frameworks/cassandra/universe/config.json
```diff
@@ -1108,6 +1108,21 @@
           "type": "string",
           "description": "The flow speed to apply rate limiting: FAST - rate limited to the speed of the fastest replica. SLOW - rate limit to the speed of the slowest replica.",
           "default": "FAST"
+        },
+        "repair_session_max_tree_depth": {
+          "type": "string",
```
samvantran: Based on description, I think this should be an integer?

farhan5900: Yes, you are right. But the field is optional and if we make it integer we have to provide some value, that's why kept it string.

rishabh96b: Yes @farhan5900 , we should provide the default value. The default value for `repair_session_max_tree_depth` is 18 and minimum value is 10.

farhan5900: Yes the value of `repair_session_max_tree_depth` is 18 but by default it is not enabled.

samvantran: If the base-tech default is 18, then we should set it to that. By changing the type to integer, you can now take advantage of adding the conditional `minimum:10` like Rishabh suggested. Now you no longer need to do the mustache if check [here](https://github.com/mesosphere/dcos-commons/pull/3211/files#diff-2c8bf3256c231656da3676fe93e6ccd0R170)

viivek46: I think we should follow what is by default in Apache Cassandra and here by default [this property](https://github.com/apache/cassandra/blob/b697af87f8e1b20d22948390d516dba1fbb9eee7/conf/cassandra.yaml#L512) is disabled, hence we should keep it like this.

samvantran: Guys, if you don't set it here Cassandra will [set it to that value anyway](https://github.com/jolynch/cassandra/blob/12c177db6313dc04d383df49f470e0d875e1ae57/src/java/org/apache/cassandra/config/Config.java#L126).

By choosing to use strings you give the user the option to put something invalid eg "apple" and then deployment will crash on them later. It will also fail deployment if users put a value less than 10 even as a string.

By choosing an integer, you can validate before deployment which makes for a better UX.

Btw, there are _many_ commented cassandra configs that we add defaults for. Here are 3 examples or probably 50-100:
- [commitlog_total_space_in_mb](https://github.com/mesosphere/dcos-commons/blob/2015e6e7545ba4aa5a38937a4452edd7dd36a124/frameworks/cassandra/src/main/dist/cassandra.yaml#L32)
- [memtable_flush_writers ](https://github.com/mesosphere/dcos-commons/blob/2015e6e7545ba4aa5a38937a4452edd7dd36a124/frameworks/cassandra/src/main/dist/cassandra.yaml#L126)
- [memtable_cleanup_threshold ](https://github.com/mesosphere/dcos-commons/blob/2015e6e7545ba4aa5a38937a4452edd7dd36a124/frameworks/cassandra/src/main/dist/cassandra.yaml#L125)

farhan5900: Thanks @samvantran, for clarification. Changed it to `integer` type.

PR #3215, opened 2020-01-10
[DCOS-61858] Quota Framework Uninstall Fixes

kaiwalyajoshi:
**Background**

This series of patches fixes the uninstall bug found in DCOS-61858 where frameworks under the same group & quota'ed role where sharing a service-principal.  Upon uninstallation, the framework being uninstalled `DESTROYED` and `UNRESERVED` resources belonging to other frameworks in the same role.

This bug occurs on DC/OS 2.0 and later clusters where the service-principal is shared.

**Resolution**

This series of patches applies the `framework_id` label to all resources created by the framework. 
The framework scheduler now filters out any offer with resources that have a mismatched `framework_id` label.

The bulk of this PR is propagating the `framework_id` from where the scheduler subscribes to Mesos, to where OfferEvaluation and the resources are built.

This patch-set also adds a debug reservations endpoint (DCOS-44705) for easier debugging of which reservations are owned by the current set of running tasks.

### DIFF
--- a/frameworks/helloworld/src/main/dist/svc.yml
+++ b/frameworks/helloworld/src/main/dist/svc.yml
@@ -21,7 +21,6 @@ pods:
           type: ROOT
           size: {{HELLO_DISK}}
         env:
-          PARSE_EMPTY_KEY_AS_BLANK_TEST:
           SLEEP_DURATION: {{SLEEP_DURATION}}
         health-check:
           cmd: stat hello-container-path/output
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
@@ -154,7 +154,10 @@ public final class ResourceUtils {
    * <li>Static against "pre-reserved-role" (we can reserve against it)</li>
    * <li>Dynamic against "pre-reserved-role" (DOESN'T belong to us at all! Likely created by Marathon)</li></ul>
    *
-   * <p>This function should return {@code false} for all reservations of the last type.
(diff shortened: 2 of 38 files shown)


frameworks/helloworld/src/main/dist/svc.yml
```diff
@@ -21,7 +21,6 @@ pods:
           type: ROOT
           size: {{HELLO_DISK}}
         env:
-          PARSE_EMPTY_KEY_AS_BLANK_TEST:
```
takirala: This was intentionally introduced iirc. it was added as a unit test for a ui parsing bug fix. any reason to remove this?

sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/ResourceUtils.java
```diff
@@ -138,14 +161,27 @@ public static boolean hasResourceId(Protos.Resource resource) {
    *                 {@link #getReservationRoles(org.apache.mesos.Protos.FrameworkInfo)}
    * @return whether this resource should be processed by our framework. if false then this resource should be ignored
    */
-  public static boolean isProcessable(Protos.Resource resource, Collection<String> ourRoles) {
+  public static boolean isProcessable(
```
takirala: maybe update javadoc with the newly added constraint(s)?

frameworks/helloworld/src/main/dist/svc.yml
kaiwalyajoshi: It prevents the quota upgrade/downgrade tests as the service-spec diverges for the hello pod (which doesn't have this ENV var). I'd rather introduce this as a seperate scenario we can test for.

PR #3216, opened 2020-01-15
Fixed publish http server launch

r2dedios:
Bug fixed when the python program "publish_http.py" try to add the repository to a DC/OS cluster with the dcos-cli.


r2dedios:
Thanks @kaiwalyajoshi !!

PR #3222, opened 2020-01-22
Create KDC database on CMD step

alexeygorobets:
To allow mount volume to KDC database folder we have to create KDC database after launching image. So I've moved this step from image building to CMD step.


alexeygorobets:
Closed PR as no longer needed

PR #3224, opened 2020-01-24
Feature/improvment build package

r2dedios:
I add some changes to make more comfortable the 'tools/build_package.sh' script output.

1. The `set -x` option is disabled except if the `-v` flag is added as an argument.
2. The publisher script output has been redirected to a log file in `/tmp` so that the standard output does not bother the terminal in which it is working

### DIFF
--- a/tools/build_package.sh
+++ b/tools/build_package.sh
@@ -42,6 +42,7 @@ custom_artifacts=
 while getopts 'va:' opt; do
     case $opt in
         v) echo "Verbose mode enabled"
+           VERBOSE=true
            set -x
            ;;
         a) custom_artifacts="$custom_artifacts $OPTARG"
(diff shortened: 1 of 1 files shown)


tools/build_package.sh
```diff
@@ -112,7 +122,12 @@ esac
...
 PACKAGE_VERSION=${1:-"stub-universe"}
 
+
+# Launch Publisher script if is defined
 if [ -n "$PUBLISH_SCRIPT" ]; then
     # All the scripts use the same argument format:
-    $PUBLISH_SCRIPT "${FRAMEWORK_NAME}" "${PACKAGE_VERSION}" "${UNIVERSE_DIR}" ${custom_artifacts}
+    publisher_log_file="/tmp/$(basename ${PUBLISH_SCRIPT}).log"
+    echo "Logs of publisher script in $publisher_log_file"
+    $PUBLISH_SCRIPT "${FRAMEWORK_NAME}" "${PACKAGE_VERSION}" "${UNIVERSE_DIR}" ${custom_artifacts} &> $publisher_log_file
```
kaiwalyajoshi: @r2dedios could we add a `tee` to both `stdout` and the `$publisher_log_file` here when verbose mode is set?

During development, we execute the build script hundreds of times a day, so having the `STUB_UNIVERSE_URL` and the build instructions emitted to the output really streamlines the development.

r2dedios: I add a new commit to add that request

r2dedios:
> @r2dedios thanks for the patch, I'm happy to merge these changes if you could make a few adjustments to the build output with `verbose` mode.

Change request attended

kaiwalyajoshi:
@r2dedios many thanks! LGTM.

r2dedios:
Thanks Kaiwalya, sorry for spelling

PR #3232, opened 2020-02-03
Added new publisher for azure blob storage

r2dedios:
Added a python script to publish the universes from Azure blob storage


r2dedios:
Of course! Count me in.

PR #3252, opened 2020-05-08
Add support for local maven repositories in the docker images.

kaiwalyajoshi:
Add support for `maven-cache` as an option to the docker image which can point to the local maven repository on the host. This is useful for cases where a snapshot build of the SDK is being tested but not fully published out to Maven repositores.

This can significantly help in trying out fixes in SDK for the various frameworks or trying out upcoming versions via artifacts from a local Maven repository.


kaiwalyajoshi:
Merging this as rebase cleared out the old approvals, but the original patch didn't change.

PR #3259, opened 2020-06-29
HDFS & Cassandra CI Fixes

kaiwalyajoshi:
- Fixes issues in `test_sanity.test_custom_jmx_port` where the `lsof` command is missing.
- Skip `test_backup_and_restore_to_s3_compatible_storage` till [D2IQ-69805](https://jira.d2iq.com/browse/D2IQ-69805) is resolved.
- Grants `dcos_marathon` permissions to run Apps with `root` user, this is specifically needed to launch `hdfs-client` which is used in many tests for HDFS.

### DIFF
--- a/frameworks/cassandra/tests/test_backup_and_restore.py
+++ b/frameworks/cassandra/tests/test_backup_and_restore.py
@@ -114,6 +114,7 @@ def test_backup_and_restore_to_s3() -> None:
     )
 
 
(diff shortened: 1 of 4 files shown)


kaiwalyajoshi:
Yeah, I'll attack those in a seperate PR. These should clear up the nightly integration tests now.

PR #3260, opened 2020-06-30
[COPS-5547] REMOTE_SEED Fixes.

pierrebeitz:
customers might not have `REMOTE_SEEDS`. in that case the comma in the end tells the config parser that the list end with a NULL, which in turn is converted to `127.0.0.1`, which then results in the debug log constantly spewing errors about the gossip protocol not being able to communicate with localhost.

i'd love to learn how to properly test this change. it's currently untested and just a best guess to illustrate what i think needs to happen in order to fix COPS-5547


kaiwalyajoshi:
Cassandra and HDFS CI failures were fixed in #3259 which was merged after this PR was created.

PR #3265, opened 2020-07-27
Add external volumes to container info

alexeygorobets:


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/PodInfoBuilder.java
```diff
@@ -19,39 +20,15 @@
...
-import com.mesosphere.sdk.specification.NetworkSpec;
-import com.mesosphere.sdk.specification.PodInstance;
-import com.mesosphere.sdk.specification.PodSpec;
-import com.mesosphere.sdk.specification.PortSpec;
-import com.mesosphere.sdk.specification.RLimitSpec;
-import com.mesosphere.sdk.specification.ReadinessCheckSpec;
-import com.mesosphere.sdk.specification.SecretSpec;
-import com.mesosphere.sdk.specification.TaskSpec;
-import com.mesosphere.sdk.specification.VolumeSpec;
+import com.mesosphere.sdk.specification.*;
```
kaiwalyajoshi, line 23: Again, we avoid `*` imports.

frameworks/cassandra/src/main/dist/svc.yml
```diff
@@ -47,14 +47,16 @@ pods:
...
     {{/TASKCFG_ALL_SECURE_JMX_ENABLED}}
-    volume:
-      path: container-path
-      type: {{CASSANDRA_DISK_TYPE}}
-      size: {{CASSANDRA_DISK_MB}}
-      {{#CASSANDRA_VOLUME_PROFILE}}
-      profiles:
-        - {{CASSANDRA_VOLUME_PROFILE}}
-      {{/CASSANDRA_VOLUME_PROFILE}}
+    external-volumes:
```
kaiwalyajoshi, line 50: We should expose `driver-options` and `volume-name` to the user in the next patch series.

frameworks/cassandra/universe/config.json
```diff
@@ -468,6 +468,32 @@
...
+              "type": "boolean",
+              "description": "If true, external profile will be used.",
+              "default": false
+            },
+            "portworx_volume_options": {
+              "type": "string",
+              "default": "size=10",
+              "description": "Volume options."
+            },
+            "use_service_path": {
```
kaiwalyajoshi, line 485: I think we can eliminate this option and make things cleaner.

if `volume-name` == "": retreive service path.
else use `volume_name`

In the code, we should be able to return an `Optional<String>` and check if the Optional is empty.

frameworks/cassandra/universe/config.json
```diff
@@ -468,6 +468,32 @@
...
+              "type": "boolean",
+              "description": "If true, use service path for volume name.",
+              "default": true
+            },
+            "volume_name": {
```
kaiwalyajoshi, line 490: See comment above, we should leave the default for this is as `""`

PR #3231, opened 2020-01-31
Support for Mesos resource-limits

timcharper:
JIRA Issues: D2IQ-62959, D2IQ-62961

PR #3321, opened 2020-11-03
Add --insecure flag to dcos-login. Enable external volumes by default. Add --disable-external-volumes flag.

kaiwalyajoshi:
- Add `--insecure` flag to dcos-login which prevents authentication with dev clusters.
- Add `--disable-external-volumes` flag.
- Enable external-volumes by default.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -76,7 +76,7 @@ def configure_universe(tmpdir_factory):
 
 @pytest.fixture(scope="session", autouse=True)
 def configure_external_volumes():
-    if is_env_var_set("EXTERNAL_VOLUMES_ENABLED", default=""):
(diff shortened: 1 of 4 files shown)


kaiwalyajoshi:
Manually cancelled a few builds as docker image has to be updated out-of-band.

kaiwalyajoshi:
Looks like some patches from Vertical Bursting (#3231) changed the requirements for Cassandra, this might be what's causing the tests to fail. Reverted those patches and we're only going to test Cassandra here as HDFS seems unaffected.

kaiwalyajoshi:
Manually stopped HDFS builds as no changes affect them, previous builds were successful.

PR #3323, opened 2020-11-25
Add container-path for pods with multiple external-volumes.

kaiwalyajoshi:
Add `container-path` to the external volume when a pod has multiple external-volumes defined.
This is only used when the user does not explicitly set a external-volume name.

### DIFF
--- a/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/GenericDockerVolumeProvider.java
+++ b/sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/GenericDockerVolumeProvider.java
@@ -17,6 +17,7 @@ public class GenericDockerVolumeProvider implements ExternalVolumeProvider {
   public GenericDockerVolumeProvider(
       String serviceName,
       Optional<String> providedVolumeName,
+      Optional<String> containerPath,
       String podType,
       int podIndex,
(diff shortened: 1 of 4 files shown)


sdk/scheduler/src/main/java/com/mesosphere/sdk/offer/evaluate/GenericDockerVolumeProvider.java
```diff
@@ -27,6 +28,9 @@ public GenericDockerVolumeProvider(
       volumeNameUnescaped = providedVolumeName.get();
     } else {
       volumeNameUnescaped = providedVolumeName.filter(name -> !name.isEmpty()).orElse(serviceName) + '_' + podType;
+      if (containerPath.isPresent()) {
+        volumeNameUnescaped = volumeNameUnescaped + '-' + containerPath.get();
```
farhan5900: I think we should use underscore (`_`) here instead of hyphen (`-`). Otherwise it will not work for NetApp.
```suggestion
        volumeNameUnescaped = volumeNameUnescaped + '_' + containerPath.get();
```

farhan5900 (changes requested):
Thanks, @kaiwalyajoshi for the changes. Please address the comment.

farhan5900 (approved):

rishabh96b (approved):

